PR #33, opened 2023-11-06
Update README.md

yebai:


README.md
```diff
@@ -6,12 +6,20 @@
...
+Some notable features that we try to improve on over ReverseDiff and Zygote / ChainRules are 
+
+- optimise performance, usability and robustness while balancing generalizability
+- guaranteed safety against dynamic control flows when using cached and compiled tapes
+- compatibility with more array types, e.g. GPU arrays
+- builtin support for mutation (writing to arrays, modifying fields of `mutable struct`s, etc), which is a shared limitation of `ReverseDiff.jl` and `Zygote.jl`, and has been the elephant-in-the-room of reverse-mode AD in Julia for years
+
+`Taped.jl` inherits many features and characteristics from `ReverseDiff.jl` and aims to be a drop-in replacement. However, compared to `ReverseDiff.jl`'s operator-overloading approach to tracing, we adopt an IR-based tracking based on [Umlaut](https://github.com/dfdx/Umlaut.jl). This IR-based tracing mechanism allows us to avoid the need to rewrite functions with generic type signatures (similar to [`Cassette.jl`](https://github.com/JuliaLabs/Cassette.jl)). 
+
```
willtebbutt, line 22: I'm not in favour of this re-phrasing.

The reason that I only went with mutation is that mutation is missing in _both_ ReverseDiff and Zygote. This is not true of the other points -- the dynamic control flow point is only relevant to ReverseDiff, as is the compatibility with more array types point (Zygote is fine with whatever array you like).

Additionally, my feeling is that the "optimise performance, usability..." point is too vague to be helpful. If 10 different people read it, you'll get 10 different impressions of what it means. My feeling is that the goals at the top of the readme better capture what this sentence states.

I would prefer it if you removed this proposed change.

willtebbutt (changes requested):

README.md
willtebbutt: I will include a careful comparison of Taped vs ReverseDiff / Zygote / ChainRules / Enzyme in the docs, so I can add a sentence here along the lines of `An in-depth comparison can be found in the [docs](link to dicussion)` when they're live.

yebai: Feel free to modify/rephrase the points. We want a balanced message here. I showed this to a colleague who needed clarification on why we highlight mutation so prominently. Let's add other considerations so it doesn't mislead people.

willtebbutt: I don't know what you mean by a "balanced message" -- could you please be more specific?

Does the colleague you mention have much experience with AD in Julia? I ask, because the readme is not currently intended for a general audience -- it's aimed at people who have some experience in the development of AD systems in Julia, and are likely to already have an appreciation of the problems that lack of mutation support causes. I'm targetting this group of people beacuse they are the ones that we are planning to share this with in the immediate future. If your colleague falls into this group, then I'll happily add further clarification / more context to expain why this is a central issue.

willtebbutt: I'm also happy to add a note at the top of the readme that clarifies who the (current) intended audience of the package is.

yebai: yes, he is fairly familiar with AD. 

I have similar confusion when reading about the paragraph. By balanced message, it should be clear we want a good tradeoff between performance, usability and generalizability (e.g. mutation support). The current message highlights generalizability, specifically mutation support, but needs more details on considerations.

willtebbutt: Okay. I'll find a way to justify the point better, but I might do it in a separate PR.

Re balanced message, do you feel that the goals stated at the top of the readme (where the stated goals are performance, correctness, support for a wide range of language features) are inadequate in addressing your points?

PR #68, opened 2024-02-01
Dynamic Dispatch Perf in Primal

willtebbutt:
Improves the performance of dynamic dispatch for primal interpretation


willtebbutt:
Also redundant in light of #90 , so closing.

PR #69, opened 2024-02-02
Simplify PhiNodes

willtebbutt:
Turns out we don't need to maintain a stack for each PhiNode. This is important from a code-complexity perspective, and for the performance of hot loops.


willtebbutt:
This definitely won't be merged, but I'll derived some additional unit tests from the insights I've gained from understanding why this definitely won't work.

PR #80, opened 2024-02-12
Cache Rules in AbstractInterpreter

willtebbutt:
If a rule has already been derived, don't re-derived it, just look it up from a cache of rules.


willtebbutt:
There's quite a lot of state associated to the object returned by `build_rrule!!`, so the test failures mean that something isn't composing properly there at the minute. 🤞 it's something simple.

PR #82, opened 2024-02-18
Eager Instruction Creation

willtebbutt:
At present, the instructions are constructed when-needed. This is helpful for avoiding things which are a little awkward to construct, but has a run-time hit associated to checking whether each instruction is / is-not assigned each time it is encountered.

This PR will attempt to modify this. It necessarily taxes the robustness of our system further, because it requires that _any_ instruction we encounter can be constructed. The basic tests pass locally for the interpreted function as-is, so I'm opening this PR to see whether the integration tests also pass. Presumably they won't.


willtebbutt:
Completely redundant -- closing.

PR #83, opened 2024-02-20
Routinely comparing performance between frameworks

willtebbutt:
Starts to compare performance between frameworks. Adds a benchmarking run which compares between Taped, Enzyme, ReverseDiff, and Zygote. Problems are chosen such that most frameworks are able to differentiate most of the examples.

Currently results are just printed to the output of the `comparison` CI run.

Fixes: https://github.com/withbayes/Taped.jl/issues/3


.github/workflows/CI.yml
```diff
@@ -60,7 +78,30 @@ jobs:
...
+          PERF_GROUP: 'comparison'
+          GKSwstype: '100'
         shell: bash
+      - uses: actions/upload-artifact@v4
+        with:
+          name: benchmarking-results
+          path: bench_results/
+      - name: Convert graph to Base64
+        run: |
+          base64 benchmarking_results.png > benchmarking_results_base64.txt
```
yebai: That's a clever trick -- I hope it can work!

willtebbutt: Alas, it does not. I'm currently playing around with unicode plots to see if I can't just produce a comment with a unicode plot. I'm pretty sure that chatgpt was lying to me when it said that the above is a thing that works (there are other discussions on the internet that suggest it doesn't work)

yebai: Producing a comment with the benchmark result table is probably good enough! If the user is interested, he can navigate to artefacts to download the PNG file.

willtebbutt: That's a good point. Probably easier to produce the table than a graph!

PR #84, opened 2024-02-21
Fix Recursion and add test

willtebbutt:
Because I wasn't properly testing recursion, it was badly broken. The simplest fix by far is to drop caching of rules.

This is a little bit sad, but only a little bit -- rule caching wasn't adding a lot, and was more of a nice-to-have than something that's really important. Certainly, I can't imagine a situation where users are likely to see any runtime difference in performance due to caching.

On the other hand, removing caching removes complexity, and means that we have to worry about less state related stuff.

### DIFF
--- a/src/interpreter/interpreted_function.jl
+++ b/src/interpreter/interpreted_function.jl
@@ -476,9 +476,6 @@ explosion of types. Moreover, type-stability it maintained.
 """
 function InterpretedFunction(ctx::C, sig::Type{<:Tuple}, interp) where {C}
 
-    # If we've already constructed this interpreted function, just return it.
-    sig in keys(interp.in_f_cache) && return interp.in_f_cache[sig]
-
     # Grab code associated to this function.
     output = Base.code_ircode_by_type(sig; interp)
     if isempty(output)
@@ -522,7 +519,6 @@ function InterpretedFunction(ctx::C, sig::Type{<:Tuple}, interp) where {C}
     make_phi_instructions!(in_f)
 
     # Cache this InterpretedFunction so that we don't have tobuild it again.
-    interp.in_f_cache[sig] = in_f
     return in_f
 end
 
--- a/src/interpreter/reverse_mode_ad.jl
+++ b/src/interpreter/reverse_mode_ad.jl
@@ -444,9 +444,6 @@ end
 
 function build_rrule!!(in_f::InterpretedFunction{sig}) where {sig}
 
-    # If we've already constructed this interpreted function, just return it.
-    sig in keys(in_f.interp.in_f_rrule_cache) && return in_f.interp.in_f_rrule_cache[sig]
-
     return_slot = SlotRef{codual_type(eltype(in_f.return_slot))}()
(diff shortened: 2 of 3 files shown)


willtebbutt:
Regular control flow is fine (gotoifnot nodes and phi nodes), it's just recursion that's the problem. It's easiest to build an intuition for this via a concrete example, so consider the following function:
```julia
function foo(x)
    a = 5x
    return x > 1 ? a + foo(x / 2) : a
end
```
When this is executed by an `InterpretedFunction`, before hitting the `foo(x / 2)` call, a value is assigned to the slot associated to `a`. Now, when `foo(x / 2)` runs, this same slot is written to again, but with a value of half whatever was written to it previously. When we exit `foo(x / 2)`, the value associated to `a` does not change, and this is now the wrong value in the context of `foo(x)`. This then gets added to the output of `foo(x / 2)`, resulting in the wrong answer.

So, more abstractly, when doing recursion we wind up corrupting the state associated to an `InterpretedFunction` (or the rrule version of this) with the current implementation. The implication of this is that, in general, you _need_ to ensure that each call to an `InterpretedFunction` gets its own state that no other on-going call is able to modify.

Now, I could spend some time trying to figure out how to do this, but it will take a decent chunk of time and add some substantial complexity to the code. So the question is: would it be worth it?

In the case of static dispatch, there really isn't any run-time benefit to caching interpreted functions (notice how all bar one of the performance tests continue to pass), so it doesn't look to me like it's worth it there. Additionally, there's not obviously much in the way of a memory benefit. I'll grant you that there are potentially some small memory benefits to sharing some instructions / `IRCode` between different instances of an `InterpretedFunction`, but it's definitely a second-order concern.

The only place that this might have an appreciable run-time impact is in dynamic dispatch for reverse-mode, but that can be addressed via a much more limited caching mechanism (which is definitely safe), and which is specific to the rrule for `DelayedInterpretedFunction`. Specifically, the `DelayedInterpretedFunction` (the thing which implements primal dynamic dispatch) maintains its own local cache of `InterpretedFunction`s to make doing dynamic dispatch performant -- if you call a `DelayedInterpretedFunction` with the same argument types repeatedly, it doesn't build a new `InterpretedFunction` each time -- see [here](https://github.com/withbayes/Taped.jl/blob/0376c004b470e25aac7ba1649c537a0d32a97c6f/src/interpreter/interpreted_function.jl#L598). I don't currently have the same mechanism in place for its rrule (see [here](https://github.com/withbayes/Taped.jl/blob/0376c004b470e25aac7ba1649c537a0d32a97c6f/src/interpreter/reverse_mode_ad.jl#L319)), so dynamic dispatch + reverse-mode may have taken a bit of a hit in this change. That being said, the performance is already bad, and I need to overhaul it anyway, so I'd rather take the hit in this PR, and revisit the implementation at a later point in time.

willtebbutt:
>I hit the merge button too soon. I suggest we add this explanation (or a ref to this PR) to the recursion test case so we don't forget.

Good idea

PR #112, opened 2024-04-02
Deepcopy output rather than copy

willtebbutt:
@gdalle could you see if this branch sorts your problem out? If so, I'll merge it, as I don't think it does any harm on my end.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Tapir"
 uuid = "07d77754-e150-4737-8c94-cd238a1fb45b"
 authors = ["Will Tebbutt and contributors"]
-version = "0.1.2"
+version = "0.1.3"
 
 [deps]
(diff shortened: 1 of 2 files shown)


gdalle:
It probably slows things down and makes them type unstable. Is it really a deep copy that you need? Why even copy?

gdalle:
Related:

- https://github.com/JuliaLang/julia/issues/31995
- https://github.com/JuliaLang/julia/issues/45143

gdalle:
Probably worth closing for now? Maybe in favor of a custom `_copy` that handles the `nothing` case as suggested in https://github.com/gdalle/DifferentiationInterface.jl/pull/126#issuecomment-2033173761?
I'm usually not a fan of these manual patches, but I think it's justified by the relevance of DifferentiationInterface for testing Tapir

PR #122, opened 2024-04-29
Remove JET deps

yebai:
Fix https://github.com/withbayes/Tapir.jl/issues/117


willtebbutt:
As-is, this PR won't work, because JET is used in `src/test_utils.jl`. We'll need to factor it out into an extension or something -- I'm not 100% sure what the right way to do this is really.

willtebbutt:
My point is it's unclear whether an extension offers the functionality we need, so it needs some thought -- extensions aren't _super_ flexible.

PR #123, opened 2024-04-29
LogDensityProblemsAD + fix CI bug

willtebbutt:
- [x] make TemporalGPs integration testing actually run during CI (I forgot to do this in the last PR)
- [x] add extension to interface with LogDensityProblemsAD

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Tapir"
 uuid = "07d77754-e150-4737-8c94-cd238a1fb45b"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.2.0"
+version = "0.2.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"


test/runtests.jl
```diff
@@ -41,6 +41,7 @@ include("front_matter.jl")
         end
         include("chain_rules_macro.jl")
     elseif test_group == "integration_testing/misc"
+        include(joinpath("integration_test", "logdensityproblemsad_interop.jl"))
```
yebai:
```suggestion
        include(joinpath("integration_testing", "logdensityproblemsad_interop.jl"))
```

willtebbutt:
@yebai please don't merge -- need to add ADTypes interop still.

willtebbutt:
Yup, I just need to make sure that interop between it and LogDensityProblemsAD is done correctly

willtebbutt:
Okay. @yebai I can't support ADTypes yet because it's held back by Turing -- the current version restricts it to 0.2. @torfjeld mentioned to me that this will be bumped at some point in the near future, so I'll revisit when that happens.

devmotion:
Should this be moved to LogDensityProblemsAD? Generally, there's no clear rule regarding which package defines an extension but since all other AD backends are extensions in LogDensityProblemsAD that would seem a bit more consistent.

willtebbutt:
I think for now it's more straightforward for me to have this code here so that I have direct control over it, and can trivially fix it when I make breaking changes to Tapir, but once things have stabilised a bit I agree that it would probably make sense to move it over to LogDensityProblemsAD

Issue #135, opened 2024-05-01
Add a "getting started" early in the README


gdalle:
Right now you have to scroll all the way down the README to figure out how to use Tapir. Maybe you could add a small section where you point to DifferentiationInterface and give a syntax example with `AutoTapir()`?

https://github.com/JuliaRegistries/General/pull/105953 will be merged soon with Tapir v0.2 compat

willtebbutt:
This is a good idea.

willtebbutt:
Resolved by #145

PR #145, opened 2024-05-13
Getting started guide in the readme

willtebbutt:
Will address #135. @gdalle your "Modern Julia Workflows" blog is excellent -- it pointed me towards the `TestReadme.jl` package, which I _love_ and am going to be making use of here!


willtebbutt:
Sadly `TestReadme` doesn't seem to work in my use-case, and I tried fairly hard. I don't want to let it block this, so I'm going to not use it for now :(

willtebbutt:
There's not really anything to review here, so I'll merge when CI passes (I don't see how it could fail though).

PR #152, opened 2024-05-14
Makes apply_iterate work

willtebbutt:


willtebbutt:
Ohh okay -- reading it back, it's not clear on this point. I'll ammend it.

PR #162, opened 2024-05-16
WIP: Disable more unsafe casting

willtebbutt:
I discovered more ways that casting can cause problems in seemingly innocuous code. In particular, if you ask Tapir to differentiate
```julia
function h(p)
    a = [1,2,3,4,5]
    buf_view = view(a,3:4)
    buf_view[1] = p
    return a[3]*2.0
end
```
at `p = 2.0`, the answer is currently zero. I was amazed that this works at all until I realised that Julia will convert `p` to an `Int` when it tries to write it to `buf_view` if it's the case that `p` happens to be integer valued. If this happens, gradients get dropped and the wrong answer is given.

As part of this PR, I went back over the intrinsics involved in casting to check if there are any more, and I don't believe there are any more which risk causing problems -- hopefully that will prove to be the case.

edit: I've converted this to a WIP because it's going to take a little bit of time to figure out what's going on in all of the cases where `fptosi` and `fptoui` are used in actually innocuous code that doesn't result in dropped gradient info. For the most part, it's just going to be declaring things non-differentiable, but there might be a couple of tricky cases.

This may also motivate a more general approach to this in which we have a macro / trait which can be applied to methods of functions which asserts that it's fine to "drop" gradients for all code inside the method, as we're confident that it's not doing it in a way which risks giving the wrong answer, but that they should otherwise be differentiated as usual. Writing the macro / trait / whatever winds up being a convenient approach to this ought really to be straightforward.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Tapir"
 uuid = "07d77754-e150-4737-8c94-cd238a1fb45b"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.2.15"
+version = "0.2.16"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/src/rrules/builtins.jl
+++ b/src/rrules/builtins.jl
@@ -88,12 +88,12 @@ end
 # atomic_pointerswap
 
 @intrinsic bitcast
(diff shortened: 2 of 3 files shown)


willtebbutt:
I'm closing this for now because it's not high on my priority list. I will attempt to address this in the future.

PR #169, opened 2024-05-28
WIP: Basic Lux Integration

willtebbutt:
@yebai this PR adds some basic Lux integration testing. There are some test failures at the minute that I'll patch up -- they'll just be intrinsics + rules etc I imagine.

@avik-pal I've just copied over the Enzyme tests that you pointed us towards, and am converting everything to `Float64` (because that's what our testing infrastructure is best equipped to handle at the minute). If you could let me know if anything looks obviously wrong in how I'm making use of the tests, that would be great!

Update: the current failures are all rule-related. I think we basically just need to work through NNlib.jl and make sure that everything there works.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -31,6 +31,7 @@ jobs:
           - 'integration_testing/array'
           - 'integration_testing/turing'
           - 'integration_testing/temporalgps'
+          - 'integration_testing/lux'
           - 'interface'
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


avik-pal:
Yeah mostly the needed rules would be same as the ones Enzyme needed https://github.com/FluxML/NNlib.jl/blob/master/ext/NNlibEnzymeCoreExt/NNlibEnzymeCoreExt.jl

willtebbutt:
I'm closing this in favour of #254 , which addresses all of this.

PR #175, opened 2024-05-29
WIP: Docs Round 1

willtebbutt:
This is the first round of documentation. The plan is to explain Tapir.jl's rule abstraction, and the tangent type system. The audience is intended to be those who are potentially interested in developin Tapir, rather than just using it.

This is _not_ ready for review yet.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -1,3 +1,144 @@
 # Tapir.jl
 
 Documentation for Tapir.jl is on it's way!
+
+Note (29/05/2024): I (Will) am currently actively working on the documentation.
+It will be merged in chunks over the next month or so as good first drafts of sections are completed.
+Please don't be alarmed that not all of it is here!
+
+## The Rule Abstraction
+
+Tapir.jl makes use of a rule system which is at first glance similar to the `rrule` function offered by ChainRules.jl.
+However, owing to Tapir.jl's support for mutation (e.g. differentiating through functions which write to arrays) and high degree of precision around the types used to represent (co)-tangent-like data, the number of situations in which the two are identical are actually rather small.
+
+Nevertheless, we begin this explanation with an example which should be familiar to anyone who has used ChainRules.jl and seen its rrule.
+Once this example has been explained, we move into new territory.
+
+### Functions of Scalars: from ChainRules.rrule to Tapir.rrule!!
+
+Consider the simple Julia function
+```julia
+mul(a::Float64, b::Float64) = a * b
+```
+
+A `ChainRules.rrule` for this might look something like
+```julia
+function ChainRules.rrule(::typeof(mul), a::Float64, b::Float64)
+    mul_pullback(dc::Float64) = NoTangent(), dc * b, dc * a
+    return a * b, mul_pullback
+end
+```
+
+The corresponding `Tapir.rrule!!` would be something like
+```julia
+function Tapir.rrule!!(::CoDual{typeof(mul)}, a::CoDual{Float64}, b::CoDual{Float64})
+    _a = primal(a)
+    _b = primal(b)
+    mul_pullback!!(dc::Float64) = NoRData(), dc * _b, dc * _a
+    return CoDual(_a * _b, NoFData()), mul_pullback!
+end
+```
+
+The core differences between the `rrule` and `rrule!!` are:
+1. each argument is a `CoDual`, which contains the primal and one other piece of data (more on this later),
+1. we must extract the primal values from `a` and `b` using the `primal` function in order to access them,
+1. `NoTangent()` is replaced by `NoRData()`, and
+1. we must return another `CoDual`, rather than just the primal value (more on this later).
+
+The point of this example is to highlight that `Tapir.rrule!!`s look a lot like `ChainRules.rrule`s in some situations, so some of your existing knowledge should transfer over.
+
+### Functions of Vectors
+
+We now turn to the obvious question: why do `Tapir.rrule!!`s differ from `ChainRules.rrule`s?
+The short answer is that Tapir.jl requires that each unique primal memory address associated to differentiable data be associated to a unique tangent (a.k.a. shadow) memory address.
+(See [Why Unique Memory Address](@ref) to understand why this is necessary.)
+
+To see how this is achieved, consider the function
+```julia
+function set_1!(x::Vector{Float64}, y::Float64)
+    x[1] = y
+    return x
+end
+```
+A valid `Tapir.rrule!!` for this function given below.
+There are a lot of concepts introduced here, so you'll need to hop back and forth between this and the text below which explains everything.
+```julia
+function Tapir.rrule!!(
+    ::CoDual{typeof(set_1!)}, x::CoDual{Vector{Float64}}, y::CoDual{Float64}
+)
+    # Extract the primal and "fdata" from x.
+    px = primal(x)
+    dx = tangent(x)
+
+    # Store the current values.
+    px_1_old = px[1]
+    dx_1_old = dx[1]
+
+    # Set x_p[1] to `y` and zero-out x_f[1].
+    px[1] = primal(y)
+    dx[1] = 0.0
+
+    function set_1_pullback!!(::NoRData)
+
+        # The (co)tangent to `y` is just the value in the first position of x_f.
+        dy = dx
+
+        # We _must_ undo any changes which occur on the forwards-pass, both to the primal
+        # and the fdata (the forwards-component of the tangent).
+        px[1] = px_1_old
+        dx[1] = dx_1_old
+
+        # There's nothing to propagate backwards for `f` because it's non-differentiable.
+        # It has "no reverse data", hence `NoRData`.
+        df = NoRData()
+
+        # There's nothing to propagate backwards for `x`, because its tangent is entirely
+        # represented by `dx` on the forwards-pass, hence `NoRData`.
+        dx = NoRData()
+
+        return df, dx, dy
+    end
+
+    # Just return x (the CoDual) -- this propagates forwards the correct unique tangent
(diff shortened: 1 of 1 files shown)


docs/src/algorithmic_differentiation.md
```diff
@@ -0,0 +1,451 @@
...
+Consequently, it will be much easier to avoid these kinds of "flattening" operations wherever possible.
+In order to do so, we make use of a generalised notion of the derivative.
+
+_**Functions Between More General Spaces**_
+
+In order to avoid the difficulties described above, we consider we consider functions ``f : \mathcal{X} \to \mathcal{Y}``, where ``\mathcal{X}`` and ``\mathcal{Y}`` are _finite_ dimensional real Hilbert spaces (read: finite-dimensional vector space with an inner product, and real-valued scalars).
+This definition includes functions to / from ``\RR``, ``\RR^D``, but also real-valued matrices.
+Furthermore, we shall see later how we can model all sorts of structured representations of data directly as such spaces.
+
+For such spaces, the derivative of ``f`` at ``x \in \mathcal{X}`` is the linear operator (read: linear function) ``D f [x] : \mathcal{X} \to \mathcal{Y}`` satisfying
```
mhauru, line 47: It took me a while to parse this correctly. I think two things threw me off a bit:
1) It wasn't obvious to me that I can consider dx to be an element of `\mathcal{X}`, considering it's infinitesimal and gradients I think are elements of the dual space.
2) There are multiple layers of functions here, where `D` takes the function `f` as an argument, `D f` takes the point `x` as an argument, and `D f [x]` takes the infinitesimal d x as an argument.

Because of the above two, for a moment I thought the `\mathcal{X} \to \mathcal{Y}` part was referring to the `[x]` argument of `D f`, and this got me confused.

mhauru: This became much clearer once I got to the definition of Frechet derivative.

docs/src/algorithmic_differentiation.md
```diff
@@ -0,0 +1,451 @@
...
+Forwards-mode AD achieves this by breaking down ``f`` into the composition ``f = f_N \circ \dots \circ f_1``, # where each ``f_n`` is a simple function whose derivative (function) ``D f_n [x_n]`` we know for any given ``x_n``. By the chain rule, we have that
+```math
+D f [x] (\dot{x}) = D f_N [x_N] \circ \dots \circ D f_1 [x_1] (\dot{x})
+```
+which suggests the following algorithm:
+1. let ``x_1 = x``, ``\dot{x}_1 = \dot{x}``, and ``n = 1``
+2. let ``\dot{x}_{n+1} = D f_n [x_n] (\dot{x}_n)``
+3. let ``x_{n+1} = f(x_n)``
+4. let ``n = n + 1``
+5. if ``n = N+1`` then return `\dot{x}_{N+1}`, otherwise go to 2.
```
mhauru:
```suggestion
5. if ``n = N+1`` then return ``\dot{x}_{N+1}``, otherwise go to 2.
```

docs/src/algorithmic_differentiation.md
```diff
@@ -0,0 +1,451 @@
...
+
+_**Step 3: Compute Adjoint of Derivative**_
+
+The argument to the adjoint of the derivative must be a 2-tuple whose elements are drawn from ``\{\RR^N \times \RR \}``.
+Denote such a tuple as ``(\bar{y}_1, \bar{y}_2)``.
+Plugging this into an inner product with the derivative and rearranging yields
+```math
+\begin{align}
+    \langle (\bar{y}_1, \bar{y}_2), D \phi_{\text{f!}} [x] (\dot{x}) \rangle &= \langle (\bar{y}_1, \bar{y}_2), (2 x \odot \dot{x}, 2 \sum_{n=1}^N x_n \dot{x}_n) \rangle \nonumber \\
+        &= \langle (2 x \odot \bar{y}_1, 2 \bar{y}_2 x), (\text{d} x, \text{d} x) \rangle \nonumber \\
```
mhauru: I find the middle step here confusing. Would it be clearer to use the definition of the inner product of tuples to spell this out as a sum over inner products?

Also, there's been a switch here from `\dot{x}` to `\text{d} x`.

mhauru (commented):
I read through the Intro and Algorithmic Differentiation pages, and took a peek at the Mathematical Interpretation of Functions one as well. Overall, this is very clearly written, and helped me a lot. I have essentially no previous experience of AD implementations and how they work.

I left a few comments. Related to the first comment, I think my largest confusion was with the infinitesimals/differentials `\text{d} x`. The text also at a couple of points switched between that and `\dot{x}`. They seem to be considered as elements of the same set as `x` itself, which I found confusing given they are infinitesimal and more like tangents. I'm not sure what the best thing to do here is, but given that the later section about implementation seems to have to introduce the notion of tangent types anyway, I wonder if it would be useful to do so here in mathsland already. If not that, then a note about how the reader should think of `\text{d} x`/`\dot{x}` would be good.

torfjelde:
Had a skim through it all and it looks very nice:)

I don't really have any concrete feedback beyond what @mhauru has already given. I was also a bit confused by the switch between $d x$ and $\dot{x}$, in part because it eludes to the fact that you should actually define the gradient as the time-derivative for a particular curve at that given point but this is (AFAIK) not mentioned anywhere. Buuut those who know this can probably can probably handle this minor thing, while those who don't know this won't ofc have this moment of "so _is_ this the time derivative or is it something else?" 🙃

But all in all, I definitively think the docs achieve what it sets out to do! Many of the things you've mentioned to me about Tapir.jl before now makes more sense and I understand some of the design choices made much better:)

willtebbutt:
Thanks for the feedback @torfjelde and @mhauru . I'll process your comments at some point today.

@torfjelde could you expand on what you mean by `I was also a bit confused by the switch between and
, in part because it eludes to the fact that you should actually define the gradient as the time-derivative for a particular curve at that given point but this is (AFAIK) not mentioned anywhere.` Are you saying that I _ought_ to be defining the gradient as a time derivative of something, or that this "dot" notation hints at time derivatives?

docs/src/algorithmic_differentiation.md
willtebbutt: This sounds like a good idea. I'll spell it out in more detail.

> Also, there's been a switch here from \dot{x} to \text{d} x.

Good catch. This is definitely a typo.

torfjelde:
> Are you saying that I ought to be defining the gradient as a time derivative of something, or that this "dot" notation hints at time derivatives?

Sorry, that was not very clear. I personally would have liked to see explicit definition of the gradient as a time-derivative, but thinking about it it's probably too much for this document 🤷

willtebbutt:
> Sorry, that was not very clear. I personally would have liked to see explicit definition of the gradient as a time-derivative, but thinking about it it's probably too much for this document 🤷

Okay, so I think I'm missing something basic here -- it's unclear to me why we need to talk about time derivatives of curves here. Is there something limiting about defining it in terms of the derivative? i.e. per [this definition on wikipedia](https://en.wikipedia.org/wiki/Gradient#Definition)

torfjelde:
Haha aight, so I really don't think this is useful for these docs 😅 

But basically, AFAIK the "cotangent" terminology comes from defining tangents on non-Euclidean spaces. And the way you do that is by considering smooth curves $\gamma: \mathbb{R} \to \mathcal{M}$ which satisfies the property $\gamma(0) = p \in \mathcal{M}$ for some point $p$ of interest, where $\mathcal{M}$ is your non-Euclidean manifold, and inspecting what happens to a given function $f: \mathbb{M} \to \mathbb{R}^d$, i.e. $\gamma \circ f: \mathbb{R} \to \mathbb{R}^d$. This is just a "standard" function, so you can consider derivative. Applying the chain rule, we get
$$\frac{d}{dt} (\gamma \circ f)(t) = \sum_i \dot{\gamma}(t)^i \frac{\partial f}{\partial x^i} \bigg|_{x = \gamma(t)}$$
Since $f$ is just an arbitrary function and the above is a linear operator applied to $f$, we see that we can identify the curve $\gamma(t)$ with a particular linear operator
$$\gamma \mapsto \dot{\gamma}^i \frac{\partial}{\partial x^i}$$

Sometimes all of this is written as the curve being $x(t)$ instead of $\gamma(t)$, and so you get $\dot{x}$ in the above, which typically denotes the _time_ derivative! That's why I thought you were using the notation $\dot{x}$ 😅 

But yeah, as I said, this is really not that relevant for this tutorial 🙃 It was just that I initially thought you were effectively eluding to this by using the $\dot{x}$ notation.

willtebbutt:
Ohhhhh sure, fine, yes. Everything here is going to be restricted to things which are isomorphic to Euclidean space, so there's really no need to discuss differential geometry, or to make use of definitions of things which work in those kinds of spaces.

This actually isn't what I'm alluding to with the dot-notation -- it's just notation that gets used in the AD literature to represent "directions" / "sensitivities" in forwards mode.

docs/src/algorithmic_differentiation.md
willtebbutt: Okay. I've clarified the fact that `D f [x]` is a function, and that I denote its application to an argument as `D f [x] (\dot{x})` sooner in the text.

docs/src/safe_mode.md
```diff
@@ -0,0 +1,35 @@
+# Safe Mode
```
yebai, line 1: Maybe rename this to `debugg_mode.md` to avoid renaming it again soon?

willtebbutt: I don't think this is a good idea -- it's currently called safe mode everywhere else. I think we should keep the naming consistent within a version, rather than have two quite different things in different places.

PR #176, opened 2024-05-30
add more compute in sum/_sum

yebai:


willtebbutt:
Why do you want to do this?

willtebbutt:
I agree that we include a few worst-case examples, but I really like including them to demonstrate that we're being up-front about them. We also include things which are handled pretty well though (i.e. everything other than `sum` / `_sum` / `kron_sum` / `kron_view_sum`), which I also think is reasonable.

I would definitely be in favour of giving more context for them though (i.e. making it clear to readers that we're intentionally showing you some of the current worst-cases).

Issue #184, opened 2024-06-17
Type-unstable `struct` arguments


yebai:
Discovered and discussed [here](https://github.com/TuringLang/AdvancedVI.jl/issues/63#issuecomment-2167594525).

willtebbutt:
This is going to be trivial to solve once I've modified the rrule!! interface to provide the arguments passed to the rrule also to the forwards-pass of a rule to the reverse-pass (this is a performance optimisation). Since this is a fairly high-priority problem, I'm going to bump this API change up the list of things which are important, and try to get it done this week.

PR #186, opened 2024-06-19
Abstract Fields in Arguments

willtebbutt:
Addresses #184


willtebbutt:
Hope fully the latest commits fix the issue with the Turing.jl integration testing. This should not be merged yet though, because I need to add a unit test which captures the behaviour that was causing the Turing.jl integration test to fail.

Issue #168, opened 2024-05-24
Rename safe_mode to debug_mode


yebai:
It’s not obvious what safe mode is, and what limitations it induces. Let’s document that. 

EDIT: also, given the significant performance penalty of using safe mode, does it make sense to enable it by default?

willtebbutt:
Is addressed via #175 and https://github.com/SciML/ADTypes.jl/pull/68 . Will close this when they are merged.

willtebbutt:
This is going to stay open so that we don't forget to rename to debug mode at some point. I've let @gdalle know that we would like to do this at some point, so when a breaking change in ADTypes next happens, we'll bundle it in with that.

willtebbutt:
There don't appear to be any breaking released -- `ADTypes` is still in version 1.x.x.

willtebbutt:
No, 1.5 -> 1.6 is a minor (feature) release. 1.5 -> 2.0 is a major (breaking) release. See e.g. https://semver.org/ for details.

willtebbutt:
> This is not breaking in major ways, so I won't suggest waiting for ADTypes 2.0.

This is something that you'll have to discuss with @gdalle -- I do not have control over this, and I don't view this as being sufficiently important to really push him to make a breaking release sooner rather than later.

> It also only affects Tapir's constructor AutoTapir, so if needed, we can support both constructors AutoTapir(debug_mode=false) and AutoTapir(safe_mode=false) but throw a depreciation warning for the latter.

We could potentially just add a constructor, but we can't change the name of the field from `safe_mode` to `debug_mode` (as this is breaking). My preference is to change it all, or change none of it, rather than have two things for users to look out for.

willtebbutt:
This is definitely also true in Julia -- there have been no breaking releases since 1.0 in the public interface (or, at least, nothing intentional). What kinds of changes are you thinking about in 1.7?

willtebbutt:
> I'm not saying these are major, but it does seem that minor breaking changes do happen from time to time, for good reasons.

Fair point.

gdalle:
I would rather add a new keyword argument and deprecate the old one. Even though it's minor, such a change would break things e.g. for people using Tapir through DifferentiationInterface

willtebbutt:
I've opened a PR here: https://github.com/SciML/ADTypes.jl/pull/75

[...]

PR #189, opened 2024-07-01
Improve safe mode messaging

willtebbutt:
Addresses #168 . As discussed in https://github.com/SciML/ADTypes.jl/pull/68, `safe_mode` is part of the public interface of `ADTypes` as it is a documented field of the `AutoTapir` type. This means that we cannot change it without making a breaking change to `ADTypes`. Since we can't do this, it seems preferable to simply clarify the nature of safe mode / the `safety_on` kwarg, then to rename it and have multiple names floating around.

The purpose of this PR, therefore, is to improve the documentation around safe_mode / safety_on, to ensure that users are aware of what it does.


willtebbutt:
Closed in favour of improving #175

PR #201, opened 2024-07-26
Debugging + MWE Docs

willtebbutt:
@sunxd3 hopefully these docs will make it clear how to conveniently go about running code to isolate your errors. Could you let me know if they're clear to you? If so, I'll merge.

### DIFF
--- a/src/chain_rules_macro.jl
+++ b/src/chain_rules_macro.jl
@@ -1,7 +1,7 @@
 _to_rdata(::ChainRulesCore.NoTangent) = NoRData()
 _to_rdata(dx::Float64) = dx
 
-"""
+@doc"""
     @from_rrule ctx sig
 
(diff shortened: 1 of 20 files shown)


src/chain_rules_macro.jl
```diff
@@ -1,7 +1,7 @@
 _to_rdata(::ChainRulesCore.NoTangent) = NoRData()
 _to_rdata(dx::Float64) = dx
 
-"""
+@doc"""
```
sunxd3, line 4: What difference does this make?

willtebbutt: (Unless I'm mistaken) it just lets you write `[name of thing to reference](@ref)`, which disappears when you look at the docstring at the REPL, but turns into an appropriate cross-reference link in the docs.

sunxd3: Interesting, thanks!

Issue #202, opened 2024-07-26
Errors when compute the gradient of `logdensity` of a JuliaBUGS model


sunxd3:
This is not quite a MWE, but still maybe useful. 

* Using `https://github.com/TuringLang/JuliaBUGS.jl.git#sunxd/resolve_tapir_issue`, which contains an efforts to resolve the issue with recursive type definition
```julia
using Tapir, JuliaBUGS, Random, LogDensityProblems

(;model_def, data, inits) = JuliaBUGS.BUGSExamples.rats
model = compile(model_def, data, inits)

Tapir.TestUtils.test_rule(
    Random.default_rng(),
    LogDensityProblems.logdensity,
    model,
    rand(65); # num of dimension of model params is 65
    perf_flag=:none,
    is_primitive=false,
    interp=Tapir.TapirInterpreter(),
)
```

produces, 
<details>
<summary>error (and stack trace)</summary>

```julia
MethodError(Tapir.rrule!!, (Tapir.CoDual{typeof(Core._compute_sparams), NoFData}(Core._compute_sparams, NoFData()), Tapir.CoDual{Method, NoFData}(getsym(vn::AbstractPPL.VarName{sym}) where sym @ AbstractPPL ~/.julia/packages/AbstractPPL/kb4q5/src/varname.jl:103, NoFData()), Tapir.CoDual{typeof(AbstractPPL.getsym), NoFData}(AbstractPPL.getsym, NoFData()), Tapir.CoDual{AbstractPPL.VarName{:mu, Accessors.IndexLens{Tuple{Int64, Int64}}}, NoFData}(mu[7, 4], NoFData())), 0x0000000000007b8f)

[...]

PR #203, opened 2024-07-26
Fix Assorted Problems

willtebbutt:
Resolves a couple of small problems needed to make #202 run.


willtebbutt:
@sundx3 could you confirm that, if you check out to this branch, your example runs?

willtebbutt:
Great stuff. Yeah, I have the same thing. I suspect that there's something going on with some non-determinism in the `Dict`s or something which means that certain values can't be compared properly. The rest of the tests seem to pass without issue, so I'm not too worried.

willtebbutt:
I'll merge this in and tag a release when CI passes.

Issue #204, opened 2024-07-29
Circular References


willtebbutt:
Properly writing this up is motivated by ongoing problems with circular references, highlighted in a PR linked to by https://github.com/compintell/Tapir.jl/issues/197 . These circular references appear in the testing infrastructure for Turing.jl -- while they could in principle be removed, it's inconvenient to do so, and Tapir.jl ought to be able to handle them.

# The Problem

You can construct circular references via seemingly straightforward types as follows:
```julia
julia> mutable struct Foo
           x
           y::Float64
       end

julia> foo = Foo(nothing, 5.0)
Foo(nothing, 5.0)

julia> foo.x = foo
Foo(Foo(#= circular reference @-1 =#), 5.0)
```
Tapir.jl can handle these if they appear inside functions that are being differentiated. For example,
```julia
julia> function f(x::Float64)
           foo = Foo(nothing, x)
           @noinline foo.x = foo # no-inline to avoid optimising everything away
           return foo.y
       end

f (generic function with 1 method)

julia> using Tapir

julia> rule = Tapir.build_rrule(f, 5.0);

julia> Tapir.value_and_gradient!!(rule, f, 5.0)
(5.0, (NoTangent(), 1.0))
```
However, the same is not true if they occur as an argument, or a value returned from, a function being differentiated:
```julia
julia> foo = Foo(nothing, 5.0)
Foo(nothing, 5.0)

julia> foo.x = foo
Foo(Foo(#= circular reference @-1 =#), 5.0)

julia> g(foo::Foo) = foo.y
g (generic function with 1 method)

julia> rule = Tapir.build_rrule(g, foo);

julia> Tapir.value_and_gradient!!(rule, g, foo)
ERROR: Found a StackOverFlow error when trying to wrap inputs. This often means that Tapir.jl has encountered a self-referential type. Tapir.jl is not presently able to handle self-referential types, so if you are indeed using a self-referential type somewhere, you will need to refactor to avoid it if you wish to use Tapir.jl.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] __create_coduals(args::Tuple{typeof(g), Foo})
   @ Tapir ~/.julia/packages/Tapir/mkYSZ/src/interface.jl:134
 [3] value_and_gradient!!(::Tapir.DerivedRule{…}, ::Function, ::Foo)
   @ Tapir ~/.julia/packages/Tapir/mkYSZ/src/interface.jl:126
 [4] top-level scope
[...]

PR #210, opened 2024-07-31
Add a `zero_tangent_internal` that uses `stackdict` to avoid circular reference induced infinite recursion 

sunxd3:
Address #204

The error we are to address is, e.g.
```julia
mutable struct Foo
    x
    y::Float64
end

foo = Foo(nothing, 1.0)
foo.x = foo # what is the `zero_tangent` of this object?
```

I think it is reasonable to set the tangent type as `{NoTangent(), Float64}`. This is quite subtle as `foo.x === foo` but  `zero_tangent(foo).fields.x !== zero_tangent(foo)`. But I think a recursive base case is required, I am not sure how this may break things.

Following the discussion and suggestions, I added a `stackdict` to track the `zero_tangent`s that has been encountered. This should give the benefits of sharing the same `zero_tangent` when two fields are egal.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Tapir"
 uuid = "07d77754-e150-4737-8c94-cd238a1fb45b"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.2.34"
+version = "0.2.35"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"


willtebbutt:
A few misc points to aid offline discussion:

#### Arrays

`Array`s can contain circular references. e.g.
```julia
julia> x = Vector{Any}(undef, 1)
1-element Vector{Any}:
 #undef

julia> x[1] = x
1-element Vector{Any}:
 1-element Vector{Any}:#= circular reference @-1 =#
```
so their method of `zero_tangent_internal` will require the `IdDict`.

I think the same is probably true of `SimpleVector`s.

#### Tuples / NamedTuples

`Tuple`s (also `NamedTuple`s) aren't always free of alias / circular-references. Consider, for example,
```julia
julia> x = randn(2)
2-element Vector{Float64}:
 -1.4593203693602337
  1.8727501637988224

julia> y = (x, x)
([-1.4593203693602337, 1.8727501637988224], [-1.4593203693602337, 1.8727501637988224])
```

willtebbutt:
> WIth the Foo example above, the result gradient of foo should have type looks like MutableTangent{NT{typeof(foo), Float64}}

This isn't _quite_ correct. The typeof the gradient of `foo` should be `tangent_type(Foo)`, which is `Tangent{@NamedTuple{x, y::Float64}}`. However, the _value_ of the `x` field of this tangent will be something concrete (e.g. `tangent_type(Foo)`)

src/tangents.jl
```diff
@@ -436,16 +440,89 @@ end
...
+            if isassigned(x, i)
+                stackdict[x][i] = zero_tangent_internal(x[i], stackdict)
+            end
+        end
+        return stackdict[x]
+    end
+end
+function zero_tangent_internal(x::P, stackdict::IdDict) where {P<:Union{Tuple, NamedTuple}}
+    if haskey(stackdict, x)
+        return stackdict[x]
```
sunxd3: this will make `zero_tangent_internal` type-unstable

willtebbutt: Could you try adding a type assertion here? i.e.
```suggestion
        return stackdict[x]::fdata_type(tangent_type(P))
```
ought to work nicely.

src/tangents.jl
```diff
@@ -410,18 +413,19 @@ It is an error for the zero element of the tangent space of `x` to be represente
...
-@inline function zero_tangent(x::SimpleVector)
-    return map!(n -> zero_tangent(x[n]), Vector{Any}(undef, length(x)), eachindex(x))
-end
-@inline function zero_tangent(x::Array{P, N}) where {P, N}
-    return _map_if_assigned!(zero_tangent, Array{tangent_type(P), N}(undef, size(x)...), x)
+function zero_tangent(x::P) where {P}
+    if isbitstype(P)
+        return zero_tangent_internal(x)
+    end
+    return zero_tangent_internal(x, IdDict())
```
willtebbutt: This could be simplified to
```suggestion
    return isbitstype(P) ? zero_tangent_internal(x) : zero_tangent_internal(x, IdDict())
```

src/tangents.jl
```diff
@@ -436,16 +440,89 @@ end
...
+    if tangent_type(P) == NoTangent
+        return NoTangent()
+    end 
+    stackdict[x] = tuple_map(Base.Fix2(zero_tangent_internal, stackdict), x)
+    return stackdict[x]
+end
+function zero_tangent_internal(x::P, stackdict::IdDict) where {P}
+    if haskey(stackdict, x)
+        return stackdict[x]
+    end
```
willtebbutt: Should this maybe live inside the `tangent_type(P) <: MutableTangent` if statement? It'll be correct either way, but where it currently is, I imagine it'll damage the performance if `tangent_type(P) == NoTangent`, or is a `struct`.

src/tangents.jl
```diff
@@ -436,16 +440,89 @@ end
...
+                end
+            else
+                return (name => zero_tangent_internal(getfield(x, n), stackdict))
+            end
+        end
+
+        stackdict[x].fields = NamedTuple(zt)
+        return stackdict[x]
+    else
+        zt = ntuple(fieldcount(P)) do n
```
willtebbutt: Should regular immutable `struct`s be put into the `IdDict`? I don't _think_ that they can ever alias one another / contain circular references, but would be happy to be shown wrong on this point!

src/tangents.jl
sunxd3: The clause with condition `!( tangent_type(P) <: MutableTangent)` is for `struct` that are not bitstype. One example I encountered here is `SubArray`.

for instance,
```julia
struct A
    x
    y
end

x = [1.0, 2.0, 3.0]
it = Tapir.zero_tangent(A(view(x, 1:2), view(x, 1:2))

# returns
Tangent{@NamedTuple{x, y}}(@NamedTuple{x, y}((Tangent{@NamedTuple{parent::Vector{Float64}, indices::NoTangent, offset1::NoTangent, stride1::NoTangent}}((parent = [0.0, 0.0, 0.0], indices = NoTangent(), offset1 = NoTangent(), stride1 = NoTangent())), Tangent{@NamedTuple{parent::Vector{Float64}, indices::NoTangent, offset1::NoTangent, stride1::NoTangent}}((parent = [0.0, 0.0, 0.0], indices = NoTangent(), offset1 = NoTangent(), stride1 = NoTangent())))))

# and
it.fields.x === it.fields.y
```

without saving to stackdict,

returns
```julia
(([0.0, 0.0, 0.0], NoTangent(), NoTangent(), NoTangent()), ([0.0, 0.0, 0.0], NoTangent(), NoTangent(), NoTangent()))
```

This is not necessary for resolving the error, but will have the same tangent for the same object, as you indicated in the github issue description

src/tangents.jl
sunxd3: similar to the above response

willtebbutt:
It looks like the CI for the inter-AD benchmarks is failing because it doesn't have permission to comment on this PR, as your PR is coming from a fork of this repo. I'm going to make a PR from my own fork which turns this benchmark suite off in this situation.

willtebbutt:
inter-AD benchmarks are now (correctly) being skipped.

src/tangents.jl
```diff
@@ -55,6 +55,13 @@ Base.:(==)(x::Tangent, y::Tangent) = x.fields == y.fields
 
 mutable struct MutableTangent{Tfields<:NamedTuple}
     fields::Tfields
+    MutableTangent{Tfields}(fields::Tfields) where {Tfields} = new{Tfields}(fields)
```
willtebbutt: This method doesn't appear to be used. Could we remove it?

src/tangents.jl
```diff
@@ -55,6 +55,13 @@ Base.:(==)(x::Tangent, y::Tangent) = x.fields == y.fields
...
+    function MutableTangent{Tfields}(fields::NamedTuple{names}) where {Tfields, names}
+        @assert names == fieldnames(Tfields) "field names mismatch"
+        new{Tfields}(fields)
+    end
```
willtebbutt: Would you consider using diagonal dispatch here, rather than error checking? Something like
```julia
    MutableTangent{Tfields}(fields::Tfields) where {Tfields} = new{Tfields}(fields)
```

src/tangents.jl
```diff
@@ -436,16 +441,88 @@ end
...
+        return stackdict[x]::tangent_type(typeof(x))
+    end
+    if isbitstype(P)
+        zt = _map_if_assigned!(Base.Fix2(zero_tangent_internal, stackdict), Array{tangent_type(P), N}(undef, size(x)...), x)
+        stackdict[x] = zt
+        return zt
+    else
+        zt = Array{tangent_type(P), N}(undef, size(x)...)
+        stackdict[x] = zt
+        for i in eachindex(x)
```
willtebbutt: Could `_map_if_assigned!` be used here, rather than writing out a for loop?

src/tangents.jl
```diff
@@ -436,16 +441,88 @@ end
...
+            end
+        end
+        return zt
+    end
+end
+function zero_tangent_internal(x::P, stackdict::IdDict) where {P<:Union{Tuple, NamedTuple}}
+    if tangent_type(P) == NoTangent
+        return NoTangent()
+    end
+    if haskey(stackdict, x)
```
willtebbutt: I think we can probably get away without doing this here. Since `Tuple`s and `NamedTuple`s are immutable, they don't have an address, so the same reasoning applies to them as does to `struct`s.

So I think you can avoid the check, and you don't ever need to put them in the `stackdict`, so the line just before the end of the function can be dropped as well.

willtebbutt (commented):
This is great. I've left direct comments on several details.

Some high-level thoughts:
- do we really need all four inner constructors for `MutableTangent`? It definitely looks like one of them isn't actually used, but I wonder whether others could go? Plainly, we need the partial-initialisation one, but I'm not clear on why we need the others any more.
- would it make sense to update the docstring for `zero_tangent` to tell users who wish to implement `zero_tangent` for their type that they should look at `zero_tangent_internal`?
- why are we calling the `IdDict` variable `stackdict`? In particular, where does the `stack` part of the name come from?

src/tangents.jl
sunxd3: This is motivated by a test error, e.g.
```julia
julia> MutableTangent{@NamedTuple{a, b::Float64}}((;a = 1.0, b = 2.0))
ERROR: MethodError: no method matching MutableTangent{@NamedTuple{a, b::Float64}}(::@NamedTuple{a::Float64, b::Float64})

Closest candidates are:
  MutableTangent{Tfields}(::Tfields) where Tfields
   @ Main REPL[1]:3
  MutableTangent{Tfields}() where Tfields
   @ Main REPL[1]:4

Stacktrace:
 [1] top-level scope
   @ REPL[6]:1
```
where `@NamedTuple{a, b::Float64}` can be a `tengent_type` return. 


Technically, this should include
```julia
MutableTangent{Tfields}(fields::Tfields) where {Tfields} = new{Tfields}(fields)
```
but I left it there to catch special case.

src/tangents.jl
sunxd3: agreed, I put it there before to catch aliasing, but as we established, a better way to handle aliasing of struct is taking care of the field that actually require tangents

src/tangents.jl
willtebbutt: Ahh right, yes, I see why this is needed -- I'm very glad we have a decent number of test cases now!

Perhaps this would work:
```suggestion
    function MutableTangent{Tfields}(fields::NamedTuple{names}) where {names, Tfields<:NamedTuple{names}}
        return new{Tfields}(fields)
    end
```

src/tangents.jl
```diff
@@ -436,16 +441,75 @@ end
...
+    end
+    if isbitstype(P)
+        zt = _map_if_assigned!(Base.Fix2(zero_tangent_internal, stackdict), Array{tangent_type(P), N}(undef, size(x)...), x)
+        stackdict[x] = zt
+        return zt
+    else
+        zt = Array{tangent_type(P), N}(undef, size(x)...)
+        stackdict[x] = zt
+        return _map_if_assigned!(Base.Fix2(zero_tangent_internal, stackdict), zt, x)::Array{tangent_type(P), N}
+    end
```
willtebbutt: Another thing: do we actually need these two separate paths here? If I'm not mistaken, the second case should cover everything that we need, and be just as performant. I might be wrong though...

willtebbutt:
Huh, I see. I'm happy to follow `deepcopy`'s conventions. Actually, on that note, could you leave a comment somewhere in the source saying that this design follows that of `deepcopy`? It's just helpful context for anyone else who might see this who is already familiar with `deepcopy`.

src/tangents.jl
sunxd3: yeah, I think so

src/tangents.jl
sunxd3: that works, thanks for the suggestion

src/tangents.jl
```diff
@@ -408,20 +411,24 @@ end
 Returns the unique zero element of the tangent space of `x`.
 It is an error for the zero element of the tangent space of `x` to be represented by
 anything other than that which this function returns.
+
+Internally, `zero_tangent` calls `zero_tangent_internal`, which handles different types of inputs differently.
+`zero_tangent_internal` has two variants:
+1. For `isbitstype` types, `zero_tangent_internal` takes one argument. 
+2. Otherwise, `zero_tangent_internal` takes another argument which is an `IdDict`, which solves the issue of circular references
+and also allows more efficient handling of aliased objects.
```
willtebbutt:
```suggestion
2. Otherwise, `zero_tangent_internal` takes another argument which is an `IdDict`, which
handles both circular references and aliasing correctly.
```

willtebbutt (approved):
I'm basically happy with this. Just one comment.

Could you please bump the patch version so that we can make a release?

willtebbutt:
Thanks for sorting this. I'll merge when CI passes.

willtebbutt:
Actually, you have merge right's I believe, so feel free to merge when CI passes.

willtebbutt:
Apologies, but I've made the patch version in this PR out of date -- you'll need to bump it again!

PR #219, opened 2024-08-08
symm rrule implementation

willtebbutt:
Adds a missing rule for `BLAS.symm!`. I believe that this is the last missing Level 3 BLAS rule that we were missing!


willtebbutt:
codecov failure is due to additional error handling code which I can't easily test. Merging.

PR #220, opened 2024-08-09
Reading doc and some discussion

sunxd3:

### DIFF
--- a/docs/src/mathematical_interpretation.md
+++ b/docs/src/mathematical_interpretation.md
@@ -123,7 +123,7 @@ Crucially, observe that we distinguish between the state of the arguments before
 
 For our example, the exact form of ``f`` is
 ```math
-f((x, y, z, s)) = ((x, y, x \odot y, \text{Ref}(2 x \odot y)), (2 x \odot y, \sum_{d=1}^D x \odot y))
+f((x, y, z, s)) = ((x, y, x \odot y, 2 x \odot y), (2 x \odot y, \sum_{d=1}^D x \odot y))
 ```
 Observe that ``f`` behaves a little like a transition operator, in the that the first element of the tuple returned is the updated state of the arguments.
 


docs/src/algorithmic_differentiation.md
```diff
@@ -118,7 +115,7 @@ We then generalise.
 _**Reverse-Mode AD: what does it do in Euclidean space?**_
 
 In this setting, the goal of reverse-mode AD is the following: given a function ``f : \RR^P \to \RR^Q`` which is differentiable at ``x \in \RR^P`` with Jacobian ``J[x]`` at ``x``, compute ``J[x]^\top \bar{y}`` for any ``\bar{y} \in \RR^Q``.
-This is useful because we can obtain the gradient from this when ``Q = 1`` by letting ``\bar{y} = 1``.
+A special case is: obtaining the gradient (of scalar output) correcponds to when ``Q = 1``, by letting ``\bar{y} = 1``.
```
willtebbutt: Could we stick with the original wording here?

docs/src/algorithmic_differentiation.md
```diff
@@ -259,11 +256,9 @@ g -> (g, (y[2] * g, y[1] * g))
 
 As before, we work through in detail.
 
-
-
 _**Step 1: Differentiable Mathematical Model**_
 
-There are a couple of aspects of `f` which require thought:
+There are a couple of aspects of `f` which require thoughts:
```
willtebbutt:
```suggestion
There are a couple of aspects of `f` which require thought:
```
I'm pretty sure the original wording is correct (grammatically speaking)

willtebbutt (changes requested):
Basically happy with your proposals -- only a couple of things I disagree with.

willtebbutt:
Regarding your questions:

>  should s be included in this?

Almost certainly. We should fix this. I think I re-worked the example, but must have forgotten to update this bit.

>  these are not quite obvious to me

Could you expand one which bit is not clear?

>  any update to this?

Not yet, although I really need to push it forward. I think this is also how JAX thinks about these kinds of objects, so I might take a closer look at what they do.

>  maybe mention this is for demonstration, in the sense that Tapir will generate this automatically? (if I understand right)

Great point. Would you mind adding something to this effect?

docs/src/mathematical_interpretation.md
```diff
@@ -123,7 +123,7 @@ Crucially, observe that we distinguish between the state of the arguments before
 
 For our example, the exact form of ``f`` is
 ```math
-f((x, y, z)) = ((x, y, x \odot y), (2 x \odot y, \sum_{d=1}^D x \odot y))
+f((x, y, z, s)) = ((x, y, x \odot y, \text{Ref}(2 x \odot y)), (2 x \odot y, \sum_{d=1}^D x \odot y))
```
sunxd3: not sure this is standard (or even correct)

docs/src/mathematical_interpretation.md
```diff
@@ -173,7 +173,8 @@ Consider the usual inner product to derive the adjoint:
 ```math
 \begin{align}
     \langle \bar{y}, D f [x] (\dot{x}) \rangle &= \langle (\bar{y}_1, \bar{y}_2), (\dot{x}, D \varphi [x](\dot{x})) \rangle \nonumber \\
-        &= \langle \bar{y}_1, \dot{x} \rangle + \langle D \varphi [x]^\ast (\bar{y}_2), \dot{x} \rangle \nonumber \\
+        &= \langle \bar{y}_1, \dot{x} \rangle + \langle \bar{y}_2, D \varphi [x](\dot{x}) \rangle \nonumber \\
+        &= \langle \bar{y}_1, \dot{x} \rangle + \langle D \varphi [x]^\ast (\bar{y}_2), \dot{x} \rangle \nonumber \quad \text{(by definition of the adjoint)} \\
```
sunxd3, line 177: I think I was confused because I missed the `*` for adjoint at first read, so I added an extra step, hope this make sense

docs/src/mathematical_interpretation.md
```diff
@@ -340,6 +341,7 @@ where ``\mathbf{1}`` is the vector of length ``N`` in which each element is equa
 (Observe that this agrees with the result we derived earlier for functions which don't mutate their arguments).
 
 Now that we know what the adjoint is, we'll write down the `rrule!!`, and then explain what is going on in terms of the adjoint.
+Tapir.jl will generate `rrule!!`s of similar effect (mostly) automatically, but it is good to see a manual implementation for understanding.
```
sunxd3: also made an addition here (just for visibility)

docs/src/mathematical_interpretation.md
willtebbutt: This is great -- much clearer now

docs/src/mathematical_interpretation.md
willtebbutt: I like this. I think I'd like to re-word very slightly to:
```suggestion
This hand-written implementation is to aid your understanding -- Tapir.jl should be relied upon to generate this code automatically in practice.
```

docs/src/mathematical_interpretation.md
willtebbutt: Nearly -- you can drop the `Ref` (we've not given `Ref` a mathematical interpreteration):
```julia
f((x, y, z, s)) = ((x, y, x \odot y, 2 x \odot y), (2 x \odot y, \sum_{d=1}^D x \odot y))
```

willtebbutt:
Thanks for the updated items -- think we're nearly there!

docs/src/mathematical_interpretation.md
sunxd3:
```suggestion
f((x, y, z, s)) = ((x, y, x \odot y, 2 x \odot y), (2 x \odot y, \sum_{d=1}^D x \odot y))
```

willtebbutt (approved):
This is a really helpful PR, thanks!

Issue #209, opened 2024-07-31
More Informative Error Message Needed: Encounter `AssertionError: _typeof(tangent(out)) == fdata_type(T)` for mutable struct with duplicated fields that are both view of an array


sunxd3:
```julia
mutable struct Bar
    x
    y
end

x = [1.0, 2.0, 3.0]
bar = Bar(view(x, 1:2), view(x, 1:2))
f(bar::Bar) = bar.x .* bar.y
rule = Tapir.build_rrule(f, bar)
v_and_g = Tapir.value_and_gradient!!(rule, f, bar)
```

error and stack

```julia
julia> v_and_g = Tapir.value_and_gradient!!(rule, f, bar)
ERROR: AssertionError: _typeof(tangent(out)) == fdata_type(T)
Stacktrace:
 [1] __value_and_pullback!!(::Tapir.DerivedRule{MistyClosures.MistyClosure{Core.OpaqueClosure{Tuple{Tapir.CoDual{typeof(f), NoFData}, Tapir.CoDual{Bar, MutableTangent{…}}}, P}} where P<:Tapir.CoDual, MistyClosures.MistyClosure{Core.OpaqueClosure{Tuple{Any}, Tuple{NoRData, NoRData}}}, Val{false}, Val{2}}, ::Float64, ::Tapir.CoDual{typeof(f), NoTangent}, ::Tapir.CoDual{Bar, MutableTangent{@NamedTuple{x, y}}})
   @ Tapir ~/Tapir.jl/src/interface.jl:15
 [2] __value_and_gradient!!(::Tapir.DerivedRule{MistyClosures.MistyClosure{Core.OpaqueClosure{Tuple{Tapir.CoDual{typeof(f), NoFData}, Tapir.CoDual{Bar, MutableTangent{@NamedTuple{x, y}}}}, P}} where P<:Tapir.CoDual, MistyClosures.MistyClosure{Core.OpaqueClosure{Tuple{Any}, Tuple{NoRData, NoRData}}}, Val{false}, Val{2}}, ::Tapir.CoDual{typeof(f), NoTangent}, ::Tapir.CoDual{Bar, MutableTangent{@NamedTuple{x, y}}})
   @ Tapir ~/Tapir.jl/src/interface.jl:66
[...]

PR #227, opened 2024-08-13
Some very basic Float32 support

willtebbutt:
We still can't properly test `Float32`, so do not treat this PR as saying that Tapir.jl is now safe to use with `Float32`. Rather, the purpose is to unblock `Float32` usage in some situations that I'm pretty sure that it's safe to do so.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -49,7 +49,7 @@ jobs:
         with:
           files: lcov.info
           token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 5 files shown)


willtebbutt:
Do not merge -- I need to check that the error I've introduced throws when someone tries to ask for the gradient of a thing that returns a non-scalar.

willtebbutt:
@sunxd3 I believe that this PR also addresses #209 -- could you let me know if what I've done makes sense to you?

Issue #226, opened 2024-08-12
Need to match the new `zero_tangent` implementation for `randn_tangent`


sunxd3:
#210 introduce `zero_tangent_internal`s to fix circular reference issue. 

We need to update `randn_tangent` function for properly testing.

PR #228, opened 2024-08-16
Add `randn_tangent` to match `zero_tangent` fix for circular references

sunxd3:
partially address #226 

~~This allows us to properly test the newly introduced `zero_tangent` fix for circular references.~~

update `randn_tangent` and `has_equal_data` to handle circular ref

### DIFF
--- a/test/test_utils.jl
+++ b/test/test_utils.jl
@@ -22,6 +22,8 @@
         @test has_equal_data(Diagonal(ones(5)), Diagonal(ones(5)))
         @test has_equal_data("hello", "hello")
         @test !has_equal_data("hello", "goodbye")
+        @test has_equal_data(TypeUnstableMutableStruct(4.0, 5), TypeUnstableMutableStruct(4.0, 5))
+        @test !has_equal_data(TypeUnstableMutableStruct(4.0, 5), TypeUnstableMutableStruct(4.0, 6))
         @test has_equal_data(TypeUnstableStruct(4.0, 5), TypeUnstableStruct(4.0, 5))
         @test !has_equal_data(TypeUnstableStruct(0.0), TypeUnstableStruct(4.0))
         @test has_equal_data(make_circular_reference_struct(), make_circular_reference_struct())


willtebbutt:
This assertion should just checks a circular reference / aliased array satisfies the usual "primal aliases primal => tangent aliases tangent" relationship. i.e. it should only break if something has gone wrong with this relationship.

Could you provide more info about the error? I wonder whether there's something going on that we haven't considered yet.

willtebbutt:
The `Ptr{Nothing}` here makes sense -- whenever you use `pointer_from_objref` this is what you'll get back (regardless the type of the object in question I believe).

What is weird is the `KeyError: key Ptr{Nothing} @0x0000000170f70430 not found` -- is that happening when you hit the assertion?

willtebbutt (commented):
This is looking really good, but I would request that the tests be tidied up slightly before we merge this. Specifically, you've got a selection of excellent functions for generating awkward test cases (`make_circular_reference_array`, `make_indirect_circular_reference_array`, and the ones for mutable structs), but they're only used in the `has_equal_data` tests at the minute -- a different collection of test cases are used for checking circular referencing is handled properly in `randn_tangent` and `zero_tangent`.

Would it be possible to e.g. move these functions to `Tapir.TestResources`, and make use of them for all tests? I appreciate that eventually we'll roll them all in to `tangent_test_cases`, but we might as well avoid duplicating logic now as well.

Does this seem reasonable, or is there a reason that we ought not to do this in this particular instance?

willtebbutt:
Great. These will just be type stability things. I'm happy to help fix them with you if you're free tomorrow, or to leave you to hack away at them if you prefer.

src/tangents.jl
```diff
@@ -474,45 +448,36 @@ end
...
-@inline function zero_tangent_internal(x::P, stackdict::IdDict) where {P<:Union{Tuple, NamedTuple}}
-    return tangent_type(P) == NoTangent ? NoTangent() : tuple_map(Base.Fix2(zero_tangent_internal, stackdict), x)
-end
-function zero_tangent_internal(x::P, stackdict::IdDict) where {P}
-
+function zero_tangent_internal(x::P, stackdict) where {P}
     tangent_type(P) == NoTangent && return NoTangent()
 
     if tangent_type(P) <: MutableTangent
+        stackdict isa IdDict || error("...")
```
willtebbutt: Would you mind fleshing out the error message here. This should probably be `throw(ArgumentError("informative message"))`, rather than just a plain error.

src/tangents.jl
```diff
@@ -474,45 +448,36 @@ end
...
-        else
-            return tangent_type(P)(backing_type(P)(zero_tangent_struct_field(x, stackdict)))
-        end
+        return tangent_type(P)(zero_tangent_struct_field(x, stackdict))
     end
 end
 
-@inline function zero_tangent_struct_field(x::P, stackdict::IdDict) where {P}
-    return ntuple(fieldcount(P)) do n
+@generated function zero_tangent_struct_field(x::P, stackdict::Union{IdDict, Nothing}) where {P}
```
willtebbutt: You've gone for `Union{IdDict, Nothing}` here as the restriction on the type of `stackdict`, whereas in other places `Any` has been used. Can we use `Any` here as well, or is there a good reason not to that I'm missing?

src/tangents.jl
```diff
@@ -529,47 +494,56 @@ details -- this docstring is intentionally non-specific in order to avoid becomi
...
+        stackdict[x] = tangent_type(P)()
+        stackdict[x].fields = randn_tangent_struct_field(rng, x, stackdict)
+        return stackdict[x]::tangent_type(P)
+    else
+        return tangent_type(P)(randn_tangent_struct_field(rng, x, stackdict))
+    end
+end
 
-    # Assume `P` is a generic struct type, and derive the tangent recursively.
+@generated function randn_tangent_struct_field(rng::AbstractRNG, x::P, stackdict::Union{IdDict, Nothing}) where {P}
```
willtebbutt: Same thing here re type constraints.

test/tangents.jl
```diff
@@ -163,51 +163,46 @@
...
-        @test zt.fields.b === zt
+        @test zt.fields.b.tangent === zt
+        rt = Tapir.randn_tangent(Xoshiro(123456), foo)
+        @test rt.fields.b.tangent === rt
     end
 
     @testset "struct with non-concrete fields" begin
         bar = Tapir.TestResources.TypeUnstableStruct(5.0, 1.0)
-        @test Tapir.zero_tangent(bar) == Tangent{@NamedTuple{a::Float64, b}}(@NamedTuple{a::Float64, b}((0.0, 0.0)))
+        @test Tapir.zero_tangent(bar) == Tangent{@NamedTuple{a::PossiblyUninitTangent{Float64}, b::PossiblyUninitTangent{Any}}}((a=PossiblyUninitTangent{Float64}(0.0), b=PossiblyUninitTangent{Any}(0.0)))
```
willtebbutt: Formatting, maybe consider something like
```suggestion
@test ==(
    Tapir.zero_tangent(bar),
    Tangent{@NamedTuple{a::PossiblyUninitTangent{Float64}, b::PossiblyUninitTangent{Any}}}(
        (a=PossiblyUninitTangent{Float64}(0.0), b=PossiblyUninitTangent{Any}(0.0))
    )
)
```
I really need to set up formatting CI...

willtebbutt (commented):
Just a few minor points.

willtebbutt:
Also, is there any chance of getting the coverage up a little bit? Doesn't have to be at 100%, I'm just trying to avoid loweing the coverage of the package in any given PR.

src/tangents.jl
sunxd3: no specific reason, that's was my carelessness

willtebbutt:
This is looking good to me now. @sunxd3 is this done, as far as you're concerned?

willtebbutt (approved):
Yes -- please merge + tag etc! 🍾

PR #229, opened 2024-08-16
Various fixes and extensions

willtebbutt:
In the course of working through the various issues associated to Tapir + Turing integration, I've found a bunch more problems.

The one involving `PiNode`s was particular tricky to debug, and I'm glad to have found it.

### DIFF
--- a/src/interpreter/s2s_reverse_mode_ad.jl
+++ b/src/interpreter/s2s_reverse_mode_ad.jl
@@ -213,9 +213,10 @@ end
 __ref(P) = new_inst(Expr(:call, __make_ref, P))
 
 # Helper for reverse_data_ref_stmts.
-@inline @generated function __make_ref(::Type{P}) where {P}
-    R = zero_like_rdata_type(P)
-    return :(Ref{$R}(Tapir.zero_like_rdata_from_type(P)))
+@inline @generated function __make_ref(p::Type{P}) where {P}
+    _P = @isdefined(P) ? P : _typeof(p)
+    R = zero_like_rdata_type(_P)
+    return :(Ref{$R}(Tapir.zero_like_rdata_from_type($_P)))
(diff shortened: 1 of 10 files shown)


src/interpreter/s2s_reverse_mode_ad.jl
```diff
@@ -368,7 +369,7 @@ function make_ad_stmts!(stmt::PiNode, line::ID, info::ADInfo)
     # Assemble the above lines and construct reverse-pass.
     return ad_stmt_info(
         line,
-        PiNode(stmt.val, fcodual_type(_type(stmt.typ))),
+        PiNode(__inc(stmt.val), fcodual_type(_type(stmt.typ))),
```
willtebbutt, line 372: @yebai it's this line.

If the first element of `PiNode` is `Argument(2)` in the primal code, then it needs to correspond to `Argument(3)` in the forwards-pass code -- this is a consistent pattern that you'll see across the codebase (argument `n` in the primal is argument `n+1` in the forwards-pass -- this is handled by calls to the internal `__inc` function in various places).

I had previously only tested cases where the first argument to a `PiNode` was an `SSA`, rather than an `Argument` -- it must be that seeing an `Argument` in a `PiNode` is fairly rare in practice, otherwise I imagine I would have encountered this edge case before now.

The test function `arg_in_pi_node` that I've added in `src/test_utils.jl` is a minimal example in which this happens.

This is definitely one of those bugs that took me a while to understand, but for which the fix is entirely trivial.

yebai: Very clear -- thanks!

willtebbutt:
Update: this PR also deals with some threadsafety problems that were causing problems in the Turing.jl tests.

I'll merge once CI has passed.

PR #231, opened 2024-08-22
Add optional argument that pass in derived rule when creating a `ADGradientWrapper`

sunxd3:
This would allow us to reuse the already derived rule.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Tapir"
 uuid = "07d77754-e150-4737-8c94-cd238a1fb45b"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.2.39"
+version = "0.2.40"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"


ext/TapirLogDensityProblemsADExt.jl
```diff
@@ -31,9 +31,11 @@ end
 
 Gradient using algorithmic/automatic differentiation via Tapir.
 """
-function ADgradient(::Val{:Tapir}, ℓ; safety_on::Bool=false)
-    primal_sig = Tuple{typeof(logdensity), typeof(ℓ), Vector{Float64}}
-    rule = Tapir.build_rrule(Tapir.TapirInterpreter(), primal_sig; safety_on)
+function ADgradient(::Val{:Tapir}, ℓ; safety_on::Bool=false, rule=nothing)
```
sunxd3, line 34: not type stable, what should the type signature be?

willtebbutt: What kind of type instabilities are you seeing here?

sunxd3: thought need to give some type hint to `rule`, but not actually needed

willtebbutt (approved):
Thanks for this PR. I'm happy to accept.

Please bump the patch version, and then feel free to merge when CI passes.

PR #234, opened 2024-08-23
Add DynamicPPL extension to allow rule-reusing with `setmodel`

sunxd3:
https://github.com/TuringLang/DynamicPPL.jl/issues/640

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,13 +1,14 @@
 name = "Tapir"
 uuid = "07d77754-e150-4737-8c94-cd238a1fb45b"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.2.40"
+version = "0.2.41"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


willtebbutt:
I'm not entirely sure that this solves the Gibbs sampling problem. To replicate, try something like
```julia
using ADTypes, Random, Tapir, Turing

@model function gdemo()
    s ~ InverseGamma(2, 3)
    m ~ Normal(0, sqrt(s))
    1.5 ~ Normal(m, sqrt(s))
    2.0 ~ Normal(m, sqrt(s))
    return s, m
end

rng = Xoshiro(123);
alg = Gibbs((PG(2_000, :s), 1), (HMCDA(500, 0.8, 0.25, :m; init_ϵ=0.05, adtype=ADTypes.AutoTapir(false)), 5_000));
sample(rng, gdemo(), alg, 5, discard_initial=2);
```
Also, if you add some kind of print statement in `build_rrule` to display the `sig_or_mi` argument, you'll be able to visually track when it is that Tapir.jl is deriving a new rule.

This example highlights that it is indeed the Gibbs sampling aspect of this that's taking the time, because there are 5000 iterations of `HMCDA` at each step of Gibbs sampling, which do not appear to take much time at all (Particle Gibbs is quite expensive though, so I'm only doing a single iteration of that).

PR #237, opened 2024-08-29
WIP: Code caching (with some restrictions)

willtebbutt:

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Tapir"
 uuid = "07d77754-e150-4737-8c94-cd238a1fb45b"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.2.41"
+version = "0.2.42"
 
 [deps]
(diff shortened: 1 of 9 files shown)


willtebbutt:
I've just tested some of the Turing.jl tests locally with this PR, and it seems that the problems are resolved. Consequently, I will bump the patch version, and release this once CI passes.

willtebbutt:
> this will prevent the rule re-deriving?

Yup -- it performs very conservative rule caching. Derived rules are tied to a specific world age, and if the world age changes, then the rule gets re-derived.

However, because each rule has some state associated to it, you need to make a copy of a cached rule before using it. Consequently, I make a `deepcopy` of cached rules before returning them. This take O(microseconds) for even quite complicated functions, so while rule lookup costs aren't completely eliminated (and you're still a bit better off passing a rule around), it largely solves the problem. Certainly, if you're using a Gibbs sampler, and each step is O(milliseconds), it should be a negligible cost.

Issue #241, opened 2024-09-03
Broken Turing model


yebai:
```julia
julia> using Turing, Tapir
Precompiling Turing
  68 dependencies successfully precompiled in 48 seconds. 248 already precompiled.

julia> @model demo_lkj() = x ~ LKJCholesky(2, 1.0)
demo_lkj (generic function with 2 methods)

julia> sample(demo_lkj(), NUTS(;adtype=AutoTapir(false)), 2000)

┌ Warning: Unable to put rule in rule field. Rule should error.
└ @ Tapir ~/.julia/packages/Tapir/hBdbd/src/interpreter/s2s_reverse_mode_ad.jl:1363               |  ETA: N/A
MethodInstance is
MethodInstance for Base._string(::Char)

with signature
Tuple{typeof(Base._string), Char}

derived_rule is of type
Tapir.DerivedRule{Tuple{typeof(Base._string), Tuple{Char}}, MistyClosures.MistyClosure{Core.OpaqueClosure{Tuple{Tapir.CoDual{typeof(Base._string), NoFData}, Tapir.CoDual{Tuple{Char}, NoFData}}, Union{}}}, MistyClosures.MistyClosure{Core.OpaqueClosure{Tuple{NoRData}, Tuple{NoRData, NoRData}}}, Val{true}, Val{2}}

Expected type is
Tapir.DerivedRule{Tuple{typeof(Base._string), Tuple{Char}}, MistyClosures.MistyClosure{Core.OpaqueClosure{Tuple{Tapir.CoDual{typeof(Base._string), NoFData}, Tapir.CoDual{Tuple{Char}, NoFData}}, Tapir.CoDual{String, NoFData}}}, MistyClosures.MistyClosure{Core.OpaqueClosure{Tuple{NoRData}, Tuple{NoRData, NoRData}}}, Val{true}, Val{2}}
```

Environment: 

```
(@plato) pkg> st
Status `~/.julia/environments/plato/Project.toml`
  [07d77754] Tapir v0.2.44
  [fce5fe82] Turing v0.34.0 `https://github.com/TuringLang/Turing.jl.git#master`
```

willtebbutt:
Full error is
```julia
┌ Warning: Unable to put rule in rule field. Rule should error.
└ @ Tapir ~/.julia/packages/Tapir/hBdbd/src/interpreter/s2s_reverse_mode_ad.jl:1363
MethodInstance is
MethodInstance for Base._string(::Char)

with signature
[...]

PR #242, opened 2024-09-05
Optimise DynamicPPL Slightly and Better Zero Adjoint Functionality

willtebbutt:
This PR has now bloated to also tackle #241 , and #243 .
Ideally I would have tackled this in three separate PRs, but I got carried away fixing things.

So, it does several things:
1. add a function called `simple_zero_adjoint`, which is useful to create rrules for things which don't need differentiating through
2. use `simple_zero_adjoint` everywhere that we can in the code base -- there are quite a number of instances in which it works
3. uses `simple_zero_adjoint` to add a DynamicPPL-specific rule, in an extension, to avoid some annoying computation that was adding overhead to Tapir.jl in that context
4. a function `remove_dead_blocks(::BBCode)`, which will remove any basic blocks which cannot be reached. This was the solution to making `LKJCholesky` work properly. I don't fully understand what was going on, but basically I think that the compiler is making some assumptions about what `IRCode` it can see in practice, and I wasn't producing code which conformed to them.
5. added a coupe of `simple_zero_adjoint` rules for string and symbol related functionality that Tapir wasn't entirely happy with due to some `ccall`s.
6. add default values for all kwargs for `Tapir.TestUtils.test_rrule` and make use of them throughout the tests

ToDo:
- [x] add unit testing for `remove_dead_blocks`

I'm planning to finish this up on Monday.

edit: increased code churn is due to slow down in CI that we were observing due to not fully caching an interpreter for the current world age. This now does that, and some code has changed as a result. Additionally, the implementation of `_remove_unreachable_blocks` (renamed from `_remove_dead_blocks`) has been heavily simplified, and the docstring associated to it improved substantially.


ext/TapirDynamicPPLExt.jl
```diff
@@ -0,0 +1,19 @@
...
+    using Tapir: Tapir
+else
+    using ..DynamicPPL: DynamicPPL
+    using ..Tapir: Tapir
+end
+
+using Tapir: DefaultCtx, CoDual, NoPullback, primal, zero_fcodual
+
+# This is purely an optimisation.
+Tapir.@is_primitive DefaultCtx Tuple{typeof(DynamicPPL.istrans), Vararg}
```
yebai: Can we have a macro / function to automate the process of specifying these skipping gradients rules? IIRC, ReverseDiff has SkipOptimise which does similar things.

willtebbutt: Good point. This should now be straightforward to do. Will update the PR to do that.

willtebbutt: Okay. It's not completely straightforward to create a macro to do this, because
1. the signatures for `@is_primitive` and `rrule!!`s are different from each other, meaning that some transformation between them is required, and
2. `Vararg`s at the end of the macro are very common, and they're a little tricky to handle in full generality.

This is to say that it would be fairly easy to create a very brittle macro to do this translation, but I suspect really quite hard (and time-consuming) to create a robust one. Given that this will only save 1 or 2 lines of code, and wouldn't be super safe in general (see the health warning associated to the new `simple_zero_adjoint` function), I don't think it's worth the hassle right now.

Instead, I've improved things a little bit by adding the function `simple_zero_adjoint`, which you can use to make use of the common pattern it encodes.

willtebbutt:
CI was passing before I pushed a docstring tweak, so I should be fine to merge this in an hour or so.

willtebbutt:
It's already in there :)

PR #247, opened 2024-09-10
Change pullback stack implementation

willtebbutt:
The current implementation produces one stack per `:call` / `:invoke` instruction, onto which the pullbacks for that line are pushed. This implementation changes that the be one stack per block, onto which all pullbacks for that block are pushed. This results in less code being emitted, which (should) improve compile times. It also improves runtimes in some situations where the overhead of AD is a significant factor in the amount of time it takes to do AD.

The one thing that this PR needs that it doesn't yet have is a test which demonstrates the importance of this on an extreme case where it's really obvious - I have such an example, I just haven't added it to this PR yet. I've been seeing modest gains across a range of examples though, so it will hopefully show up in the benchmarking results table when that is generated.

Todo:
- [x] add extreme example to demonstrate improved runtime performance
- [x] improve docstring for `create_comms_stack`, and maybe rename
- [x] make naming of outputs of `create_comms_stack` consistent through the code
- [x] ~~figure out why things don't appear to have improved for `test_union_of_arrays` -- is the emitted code worse somehow?~~ majority of time is spent in the compiler, looking up the IR for the original function, so there's not really anything to be done about this.


willtebbutt:
Example added to the benchmarks that we track in PRs. It's designed to simulate an example that I was seeing with TemporalGPs.jl, where a function which uses `StaticArrays` was differentiated rather slowly with Tapir.jl. This PR seems to give at least a 2x speed up on this example -- on the TemporalGPs.jl example, I'm seeing a 3x speed up. In neither case is the performance what we would like it to be, but it's plainly much better.

PR #248, opened 2024-09-11
Remove multiple optimisation passes

willtebbutt:
I currently run multiple optimisation passes, because at some point in the past this was necessary in order to get everything to infer properly. I'm running the test suite to try and figure out if this is still true, or if it's in fact gone away, and we can get away with optimising just once.

If it is, I'll merge this and tag a release, because optimising twice feels 1) silly, and 2) sometimes causes problems during inlining.


willtebbutt:
This looks to be fine -- CI timings don't appear worse, nor do the posted benchmarks. I'll merge when CI passes again.

Issue #191, opened 2024-07-02
Improving ChainRules rule usage


willtebbutt:
The `@from_rrule` macro which lets you make use of existing `ChainRules.rrule`s to define new `Tapir.rrule!!`s is quite limited in what inputs it supports. In particular, it permits only
1. `Float64`s, and
2. types which are non-differentiable.

Moreover, the non-differentiable types must be bits types, and the `ChainRules.rrule` must return the correct type. e.g. it won't work if the `ChainRules.rrule` returns a `Float64` gradient for an `Int` primal.

One narrow way to extend this would be to assume that we can just return a `Tapir.NoRData` whenever the primal is a non-differentiable bits type, regardless what the `ChainRules.rrule` returns.

Motivated by https://github.com/TuringLang/Bijectors.jl/pull/319#discussion_r1662250153 .

willtebbutt:
This is addressed by #254

Issue #171, opened 2024-05-28
NNlib Support


willtebbutt:
#169 highlights that we need rules for a range of functionality which lives in NNlib.jl -- this is not surprising, and has largely already been done (see e.g. the Enzyme extension). Someone needs to systematically work through and test that Tapir works on everything in NNlib.

willtebbutt:
In short: yes. We don't have a nice macro for doing it at the minute, but it ought to be quite doable provided that we constrain the types that impose some of our own constraints on the types.

In the case of dropout: this one actually might be a little tricky because it uses kwargs, and I've not looked at supporting kwargs yet (this largely happen automatically), but it should otherwise be fine.

Something like [batched_transpose](https://github.com/FluxML/NNlib.jl/blob/62f6074a3944ba7db8f09f928db2121f0ea7ef07/src/batched/batchedadjtrans.jl#L95) should be quite straightforward. You would just do something along the lines of
```julia
function rrule!!(::CoDual{typeof(batched_transpose)}, A::CoDual{Array{<:Any,3}})
    B, pb = rrule(batched_transpose, primal(A)
    function pb!!(dB)
        _, dA_inc = pb(dB)
        increment!!(tangent(A), dA_inc)
        return NoRData(), NoRData()
    end
    return zero_fcodual(B), pb!!
end
```
Note that I've restricted the array type here -- we would just need to be careful with how we extend this rule to other array types, as it's only valid in Tapir.jl for arrays whose tangents are arrays, rather than `Tangent`s.

willtebbutt:
In the narrow sense that lots of `ChainRules.rrule`s are written with abstract arguments: yes.

However, recall that this gives you an overly optimistic sense of _how_ much more general a `ChainRules.rrule` is. It's not very hard to construct e.g. subtypes of `AbstractArray{<:Any, 3}` for which this won't work.

willtebbutt:
It's part motivation part oversight.

It's motivated by trying to write rules for things like Zygote, where you really want your rule system to do a lot of work, because Zygote can't differentiate much itself due to lack of mutation support.

It's an oversight in the sense that we didn't realise how hard it is to write rules which work for lots of types that are robust.

willtebbutt:
Hmm okay. Could you provide a small code example showing what you mean? I'm still not quite following what you're trying to say / ask, so I think I need some examples.

willtebbutt:
[...]

PR #254, opened 2024-09-13
Further ChainRulesCore.rrule Integration

willtebbutt:
At present the extent to which `ChainRulesCore.rrule`s can be (straightforwardly) used in this package is quite limited. The purpose of this PR is to:
- [x] make it easier to incorporate more interesting rules, thus closing #191 ,
- [x] include some failing tests to ensure that we get clear errors when conversations between Tapir and ChainRules tangent types goes wrong,
- [x]  explain the limitations of what can be done,
- [x] provide guidance to the situations in which you should consider importing an `rrule` from ChainRules, and
- [x] import rules from NNlib.jl and test that everything works #171 .
- [x] reduce quantity of code duplication between rrule wrapper with / without kwargs.
- [x] improve macro for use with kwargs `rrule`s.
- [x] improve macro with `where` terms so that diagonal dispatch can be done.
- [x] improve chain rules interop documentation.
- [x] add + test rules associated with LuxLib.jl
- [x] integration test Lux.jl
- [x] add section on quick ways to write rules to docs (method overlays, and getting rules from rrules, simple_zero_adjoint).
- [x] turn docstrings into doctests in `@from_rrule` macro

The reason for incorporating the rules from NNlib.jl in this PR, rather than a subsequent one, is that the main place we wish to import rules from in the near term is NNlib.jl. Consequently, doing that in this PR will give us a good sense of whether the new functionality we introduce here is sufficient.

edit: as a happy consequence of thinking more carefully about how we incorporate functionality from ChainRules, I've removed a lot of code from the the `@from_rrule` macro, and put it in a regular Julia function. This makes this bit of the codebase much more straightforward to understand.


willtebbutt:
@yebai could you let me know what you think about the docs?

ext/MooncakeDynamicPPLExt.jl
```diff
@@ -1,17 +1,9 @@
...
-end
-
-using Mooncake: DefaultCtx, CoDual, simple_zero_adjoint
+using DynamicPPL: DynamicPPL, istrans
+using Mooncake: Mooncake
 
 # This is purely an optimisation.
-Mooncake.@is_primitive DefaultCtx Tuple{typeof(istrans), Vararg}
-Mooncake.rrule!!(f::CoDual{typeof(istrans)}, x::CoDual...) = simple_zero_adjoint(f, x...)
+Mooncake.@zero_adjoint Mooncake.DefaultCtx Tuple{typeof(istrans), Vararg}
```
yebai, line 7: It looks good; can we remove the need for users to pass `Mooncake.DefaultCtx` explicitly, mainly if there is a `DefaultCtx` for most cases? We can still support users in specifying customised contexts if needed.

ext/MooncakeNNlibExt.jl
```diff
@@ -0,0 +1,66 @@
...
+        Tuple{typeof($conv), Array{P}, Array{P}, ConvDims} where {P<:IEEEFloat},
+        true,
+    )
+    @eval @from_rrule(
+        MinimalCtx,
+        Tuple{typeof($∇conv_data), Array{P}, Array{P}, ConvDims} where {P<:IEEEFloat},
+        true,
+    )
+end
+@eval @from_rrule(
```
yebai: I am slightly confused about the use of `@eval` in some places but not others. Can you clarify the differences?

```suggestion
@from_rrule(
```

src/tools_for_rules.jl
```diff
@@ -0,0 +1,477 @@
...
+end
+
+#
+# Functionality supporting @mooncake_overlay
+#
+
+"""
+    @mooncake_overlay method_expr
+
+Define a method of a function which only Mooncake can see. This can be used to write
```
yebai, line 36: Consider adding a reference to Julia's overlay docs for people who don't know what method overlays are yet (I didn't know until recently seeing them in Reactant).

src/tools_for_rules.jl
```diff
@@ -0,0 +1,477 @@
...
+anything in `f` or `x`. This is always the case if the result is a bits type, but more care
+may be required if it is not.
+```
+"""
+@inline function zero_adjoint(f::CoDual, x::Vararg{CoDual, N}) where {N}
+    return zero_fcodual(primal(f)(map(primal, x)...)), NoPullback(f, x...)
+end
+
+"""
+    @zero_adjoint ctx sig
```
yebai, line 142: As commented above, it is helpful to have a variant of `@zero_adjoint` with a default `ctx`, so users don't need to pass it explicitly unless in exceptional circumstances.

src/tools_for_rules.jl
```diff
@@ -0,0 +1,477 @@
...
+
+Defines `is_primitive(context_type, sig) = true`, and defines a method of
+`Mooncake.rrule!!` which returns zero for all inputs.
+Users of ChainRules.jl should be familiar with this functionality -- it is morally the same
+as `ChainRulesCore.@non_differentiable`.
```
yebai, line 147: I find the name `@non_differentiable` more intuitive. Is there any good reason for preferring the name `zero_adjoint` over `non_differentiable`?

ext/MooncakeNNlibExt.jl
willtebbutt: This was a mistake on my part. `@eval` is only needed when I'm iterating over symbols and interpolating them. So the for loops above and below this line need `@eval` for interpolation purposes. Here, I'm not interpolating anything, so it can go.

src/tools_for_rules.jl
willtebbutt: I had considered this after our chat the other day, but I couldn't find any really good docs that explain what's going on here. Were there some in particular that you think are worth linking?

src/tools_for_rules.jl
willtebbutt: The name `@non_differentiable` implies that the function cannot be differentiated / does not have a derivative (or, therefore, an adjoint). I actually think `@non_differentiable` is quite an unfortunate name in hindsight, because it suggests that a given function _can't_ be differentiated, which isn't at all what the purpose of the macro is. The macro actually _means_ "the derivative / adjoint is zero for all inputs".

Conversely, the name `zero_adjoint` aligns perfectly with what we mean.

ext/MooncakeDynamicPPLExt.jl
willtebbutt: I understand the desire to do this, but I don't think it's the right thing to do here, and it wouldn't be consistent with everything else that we do in this package. I really _do_ want users to declare whether a given rule is completely essential, or simply a performance optimisation.

docs/src/tools_for_rules.md
```diff
@@ -0,0 +1,33 @@
...
+```@docs
+Mooncake.@zero_adjoint
+Mooncake.zero_adjoint
+```
+
+## Using ChainRules.jl
+
+[ChainRules.jl](https://github.com/JuliaDiff/ChainRules.jl) provides a large number of rules for differentiating functions in reverse-mode.
+These rules are methods of the `ChainRulesCore.rrule` function.
+There are some instances where there is it most convenient to implement a `Mooncake.rrule!!` by wrapping an existing `ChainRulesCore.rrule`.
```
mhauru:
```suggestion
There are some instances where it is most convenient to implement a `Mooncake.rrule!!` by wrapping an existing `ChainRulesCore.rrule`.
```

src/interpreter/ir_utils.jl
```diff
@@ -181,18 +184,31 @@ there is no code found, or if more than one `IRCode` instance returned.
 
 Returns a tuple containing the `IRCode` and its return type.
 """
-function lookup_ir(interp::CC.AbstractInterpreter, sig::Type{<:Tuple})
-    output = Base.code_ircode_by_type(sig; interp)
-    if isempty(output)
-        throw(ArgumentError("No methods found for signature $sig"))
-    elseif length(output) > 1
-        throw(ArgumentError("$(length(output)) methods found for signature $sig"))
+function lookup_ir(interp::CC.AbstractInterpreter, tt::Type{<:Tuple}; optimize_until=nothing)
```
mhauru, line 187: The docstring says "Get the IR unique IR associated", probably a typo.

src/tools_for_rules.jl
yebai: I see both sides of the argument. I'll let you decide here.

src/tools_for_rules.jl
yebai: Unfortunately, I don't have a good reference myself.

src/tools_for_rules.jl
```diff
@@ -0,0 +1,477 @@
...
+julia> @zero_adjoint DefaultCtx Tuple{typeof(foo_varargs), Vararg}
+
+julia> is_primitive(DefaultCtx, Tuple{typeof(foo_varargs), Any, Float64, Int})
+true
+
+julia> rrule!!(zero_fcodual(foo_varargs), zero_fcodual(3.0), zero_fcodual(5))[2](NoRData())
+(NoRData(), 0.0, NoRData())
+```
+Be aware that it is not currently possible to specify any of the type parameters of the
+`Vararg`. For example, the signature `Tuple{typeof(foo), Vararg{Float64, 5}}` will not work
```
mhauru, line 181: Will it error in some understandable way?

src/tools_for_rules.jl
```diff
@@ -0,0 +1,477 @@
...
+
+"""
+    increment_and_get_rdata!(fdata, zero_rdata, cr_tangent)
+
+Increment `fdata` by the fdata component of the ChainRules.jl-style tangent, `cr_tangent`,
+and return the rdata component of `cr_tangent` by adding it to `zero_rdata`.
+"""
+increment_and_get_rdata!(::NoFData, r::T, t::T) where {T<:IEEEFloat} = r + t
+function increment_and_get_rdata!(f::Array{P}, ::NoRData, t::Array{P}) where {P<:IEEEFloat}
+    increment!!(f, t)
```
mhauru, line 247: I don't really know what this does, but it jumps at me as a call to `!!` that ignores the return value. If `increment!!` is guaranteed to mutate rather than create a new instance, could it be called `increment!`? If not, might this fail to do what is expected of it?

src/tools_for_rules.jl
```diff
@@ -0,0 +1,477 @@
...
+function increment_and_get_rdata!(f, r, t::ChainRulesCore.Thunk)
+    return increment_and_get_rdata!(f, r, ChainRulesCore.unthunk(t))
+end
+
+@doc"""
+    rrule_wrapper(f::CoDual, args::CoDual...)
+
+Used to implement `rrule!!`s via `ChainRulesCore.rrule`.
+
+Given a function `foo`, argument types `arg_types`, and a method `ChainRulesCore.rrule` of
```
mhauru:
```suggestion
Given a function `foo`, argument types `arg_types`, and a method of `ChainRulesCore.rrule`
```

docs/src/tools_for_rules.md
mhauru:
```suggestion
There are some instances where it is most convenient to implement a `Mooncake.rrule!!` by wrapping an existing `ChainRulesCore.rrule`.
```

test/tools_for_rules.jl
```diff
@@ -0,0 +1,136 @@
+overlay_tester(x) = 2x
+Mooncake.@mooncake_overlay overlay_tester(x) = 3x
+
+zero_tester(x) = 0
+Mooncake.@zero_adjoint MinimalCtx Tuple{typeof(zero_tester), Float64}
+
+vararg_zero_tester(x...) = 0
+Mooncake.@zero_adjoint MinimalCtx Tuple{typeof(vararg_zero_tester), Vararg}
+
```
mhauru: Is there a reason to have these outside the module?

mhauru (commented):
I don't understand most of the code, but I read the docstrings and skimmed the implementations and tests. Spotted a few typos and had a few questions, but no significant concerns to raise.

docs/src/tools_for_rules.md
mhauru: I seem to have managed to leave the same comment twice. Somewhat impressed by that myself, frankly.

src/interpreter/ir_utils.jl
willtebbutt: Good catch.

src/tools_for_rules.jl
willtebbutt: Hmmmm probably not. Alas, I'm not really sure how to ensure that it does either -- it's just a tricky one to design good code for.

src/tools_for_rules.jl
willtebbutt: Thanks for pointing this out. In this particular case, I know for sure that this is safe to do.

test/tools_for_rules.jl
willtebbutt: No, there is not. Good catch.

willtebbutt:
I've downgraded CI to run on 1.10 for the time being so we can merge this while I figure out the upgrades for 1.11.

src/tools_for_rules.jl
willtebbutt: @mhauru if you have any thoughts on how I might catch things like `Tuple{typeof(foo), Vararg{T}} where {T}` and `Tuple{typeof(foo), Vararg{T} where {T}}`, and `Tuple{typeof(foo), Vararg{<:Real}}` etc, I'd love to hear it. 

Users of this macro will certainly get some kind of error though, so a mistake can't silently propagate through to users or anything nasty like that.

mhauru: I doubt I'll have much useful to add, don't let this hold you back from merging. Good to know that it'll at least error in some way.

willtebbutt: Cool. Thanks!

Issue #257, opened 2024-09-18
Renaming Process


willtebbutt:
Checklist:
- [x] rename this package
- [x] open PR to registry to change the old Tapir URL to the new Mooncake one - https://github.com/JuliaRegistries/General/pull/115400
- [x] rename everything in this package + declare a breaking change - #256 
- [x] register package under new name
- [x] create new ADType which uses Mooncake, and have it use a generic Mooncake config object, so that we can change the config arguments if necessary
- [x] Update DifferentiationInterface to use AutoMooncake .
- [x] uncomment various references to `ADTypes.Mooncake` throughout this package (ctrl-f for `AutoMooncake` to find them all), and re-enable tests (these had to be commented out during the transition period)

gdalle:
Note that the new `AutoMooncake` struct cannot be something like
```julia
struct AutoMooncake{C<:Mooncake.Config}
    config::C
end
```
because that would force ADTypes to depend on Mooncake. However you can copy what we did for other backends and leave the type parameter free, while specifying exactly what it should be in the docstring.

PR #261, opened 2024-09-25
Remove useless hashtags in docs

gdalle:


gdalle:
Woops this is useless until #257 is complete

Issue #215, opened 2024-08-05
Add a `TapirMode` that we can pass to `AutoTapir(mode=TapirMode(debug=true))`


yebai:
It will provide flexible control over how Tapir is run but avoid breaking changes to `ADTypes`. If `TapirMode` is changed in a breaking manner, we simply mark the corresponding Tapir release breaking so users won't be surprised. 

The name TapirMode can be anything we agree on.

willtebbutt:
This is a good idea. We will need to do this prior to renaming / updating ADTypes.jl with the a new type.

willtebbutt:
Resolved by #262

PR #262, opened 2024-09-25
Config + DebugMode rename (breaking)

willtebbutt:
Addresses #215 and further's the renaming process.

@yebai are you happy with just calling the object `Config`?


willtebbutt:
> The name is okay to me, but that means we should probably not export it because it is too generic.

Agreed -- I won't export it.

> What is the plan for the default execution mode (i.e. true/false for debugging mode)?

The default is debug mode off, so people will just have to write
```julia
AutoMooncake(; config=Mooncake.Config())
```
to get the default fast mode.

willtebbutt:
I'm going to hold off merging this until the [ADType PR](https://github.com/SciML/ADTypes.jl/pull/89) looks good to go.

willtebbutt:
ADTypes PR is merged, so I'm going to release this.

Issue #263, opened 2024-09-25
Remarks on documentation


gdalle:
First of all, first-class AD tutorial!!! I might legit point people towards it even if they don't use Mooncake.

A few suggestions that came to me while reading:

- Add an API reference somewhere. Docstrings in the middle of text are good but hard to find.
- It's not obvious what the difference is between a tangent and a `CoDual`
- The `rrule!!` in https://compintell.github.io/Mooncake.jl/dev/mathematical_interpretation/#The-Rule-Interface-(Round-2) is not up to date, I think the attributes are no longer called `x.tangent` and `x.primal`
- The hashtag comments in https://compintell.github.io/Mooncake.jl/dev/debug_mode/ don't behave the way you want, for instance they break the TOC

willtebbutt:
Thanks for the feedback! Will look at addressing these.

willtebbutt:
Thanks for reading through the docs and your feedback @gdalle -- it's much appreciated!

PR #266, opened 2024-09-26
Improve Docs

willtebbutt:
This will address #263


gdalle:
Looks good to me. So do you want to separate the two uses of `CoDual` in the end?

willtebbutt:
Yeah, I definitely want to separate the uses -- hopefully some time in the next week. I'll merge this PR for now and open an issue about it specifically.

PR #270, opened 2024-09-26
Remove redundant blocks during optimisation

willtebbutt:


willtebbutt:
This is redundant. While this change was made in response to an error found in practice, the cause of the error was not the inlining pass, but rather a poorly-implemented `rrule!!` .

PR #279, opened 2024-10-02
Remove Method of zero_like_rdata_from_type

willtebbutt:
I'm opening this as an easy way to run the entire test suite to see what fails when I remove this. Do not merge.

update: this actually can be merged -- it looks like these method were redundant. They don't make any sense anyway because a `TypeVar` is not a type, so they really ought not to exist. I'll update, bump the patch, and merge when CI passes.


willtebbutt:
It looks like this actually could be merged. I'll do so once #254 has been merged in.

PR #281, opened 2024-10-03
Check if config is nothing before accessing fields

### LINKED ISSUE #280: `ADTypes.AutoMooncake(; config=nothing)` doesn't work
Initialising `ADTypes.AutoMooncake(; config=nothing)` doesn't seem to use a config of `Mooncake.Config()` by default.

The following MWE uses a feature branch of Turing.jl (https://github.com/TuringLang/Turing.jl/pull/2341):

```julia
using Turing
import Mooncake
backend = Turing.AutoMooncake(; config=nothing)  # reexported from ADTypes
@model f() = x ~ Normal(0, 1)
sample(f(), HMC(0.1, 3; adtype=backend), 10)

# works with backend = Turing.AutoMooncake(; config=Mooncake.Config())
```

The error:

```
ERROR: type Nothing has no field debug_mode
Stacktrace:
  [1] getproperty(x::Nothing, f::Symbol)
    @ Base ./Base.jl:37
  [2] ADgradient(x::AutoMooncake{…}, ℓ::LogDensityFunction{…})
    @ MooncakeLogDensityProblemsADExt ~/.julia/packages/Mooncake/eEGqO/ext/MooncakeLogDensityProblemsADExt.jl:63
  [3] ADgradient(ℓ::LogDensityFunction{DynamicPPL.TypedVarInfo{…}, DynamicPPL.Model{…}, DynamicPPL.SamplingContext{…}})
    @ Turing.Inference ~/ppl/lib/src/mcmc/Inference.jl:154
  [4] initialstep(rng::Random.TaskLocalRNG, model::DynamicPPL.Model{…}, spl::DynamicPPL.Sampler{…}, vi_original::DynamicPPL.TypedVarInfo{…}; initial_params::Nothing, nadapts::Int64, kwargs::@Kwargs{})
    @ Turing.Inference ~/ppl/lib/src/mcmc/hmc.jl:159
[...]
```

which points to at least this:

https://github.com/compintell/Mooncake.jl/blob/942eeade0e1d5a9b092f8a9928f8c4807a699487/ext/MooncakeLogDensityProblemsADExt.jl#L61-L72

penelopeysm:
Should fix #280 (I've tested locally), but happy for you to implement a different fix if you think it's better done otherwise. :)

### DIFF
--- a/ext/MooncakeLogDensityProblemsADExt.jl
+++ b/ext/MooncakeLogDensityProblemsADExt.jl
@@ -61,7 +61,7 @@ end
 # Interop with ADTypes.
 function getconfig(x::ADTypes.AutoMooncake)
     c = x.config
-    return isnothing(c) ? Mooncake.DEFAULT_CONFIG : c
+    return isnothing(c) ? Mooncake.Config() : c
 end
 function ADgradient(x::ADTypes.AutoMooncake, ℓ)
     debug_mode = getconfig(x).debug_mode
--- a/ext/MooncakeLogDensityProblemsADExt.jl
+++ b/ext/MooncakeLogDensityProblemsADExt.jl
@@ -59,8 +59,12 @@ function logdensity_and_gradient(∇l::MooncakeGradientLogDensity, x::Vector{Flo
 end
 
 # Interop with ADTypes.
+function getconfig(x::ADTypes.AutoMooncake)
+    c = x.config
+    isnothing(c) ? Mooncake.DEFAULT_CONFIG : c
+end
 function ADgradient(x::ADTypes.AutoMooncake, ℓ)
-    debug_mode = x.config.debug_mode
+    debug_mode = getconfig(x).debug_mode
     if debug_mode
         msg = "Running Mooncake in debug mode. This mode is computationally expensive, " *
             "should only be used when debugging a problem with AD, and turned off in " *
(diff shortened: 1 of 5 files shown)


ext/MooncakeLogDensityProblemsADExt.jl
```diff
@@ -59,8 +59,12 @@ function logdensity_and_gradient(∇l::MooncakeGradientLogDensity, x::Vector{Flo
 end
 
 # Interop with ADTypes.
+function getconfig(x::ADTypes.AutoMooncake)
+    c = x.config
+    isnothing(c) ? Mooncake.DEFAULT_CONFIG : c
```
willtebbutt:
```suggestion
    return isnothing(c) ? Mooncake.DEFAULT_CONFIG : c
```
formatting

penelopeysm:
> what's the rationale for introducing a global

To avoid constructing the same value multiple times. I don't know if it makes for any real speedup, but I feel like it makes sense :)

penelopeysm:
Not opposed to that at all. I'll switch those back.

willtebbutt:
Thanks for this. One more thing: could you add a regression test to prevent this error from creeping back in? `test/integration_testing/logdensityproblemsad_interop.jl` would be the place to put it. It could be as simple as more-or-less copying https://github.com/compintell/Mooncake.jl/blob/942eeade0e1d5a9b092f8a9928f8c4807a699487/test/integration_testing/logdensityproblemsad_interop.jl#L23-L25 and replacing the config with `nothing`, or something like that.

Other than that, I think we're good to go.

willtebbutt (approved):
Thanks for this. I'll merge + tag a release once CI has passed.

PR #292, opened 2024-10-19
Faster Stacks

willtebbutt:
It looks like I've been leaving some performance on the table by implementing the various stack data structures using a `mutable struct`, and mutating it, rather than using a `struct` and changing the way that the code works to avoid mutation.

Todo:
- [x] Use better stack data structure with the block stack
- [ ] Use better stack data structure with pullback stacks
- [ ] Remove old stack and code made redundant by this PR

This PR also fixes a bug I had in the rrule tests that caused occasional segfaults (I had written unsafe testing code -- users would only see this segfault in their own code if the code that they are trying to differentiate segfaults when run anyway).


willtebbutt:
This actually doesn't work anywhere near as well as I had hoped, so I'm closing it. A different strategy will be required in order to remove more of the remaining overhead.

PR #293, opened 2024-10-19
enum support

mohamed82008:
Hi!

I was trying this package out and I ran into an issue with `Enum` support where it wanted me to define a `tangent_type` for my `Enum` type. I think this PR fixes it according to my local tests. I don't have tests in this PR because I was not sure where the best place is to add the tests. Please let me know where you prefer I add the tests.

Great package btw!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.13"
+version = "0.4.14"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/src/tangents.jl
+++ b/src/tangents.jl
@@ -332,8 +332,9 @@ tangent_type(::Type{DimensionMismatch}) = NoTangent
 
 tangent_type(::Type{Method}) = NoTangent
 
-function tangent_type(::Type{P}) where {N, P<:Tuple{Vararg{Any, N}}}
+tangent_type(::Type{<:Enum}) = NoTangent
 
+function tangent_type(::Type{P}) where {N, P<:Tuple{Vararg{Any, N}}}
     # As with other types, tangent type of Union is Union of tangent types.
     P isa Union && return Union{tangent_type(P.a), tangent_type(P.b)}
 
@@ -374,7 +375,6 @@ function tangent_type(::Type{P}) where {N, P<:NamedTuple{N}}
 end
 
 @generated function tangent_type(::Type{P}) where {P}
-
     # This method can only handle struct types. Tell user to implement tangent type
     # directly for primitive types.
     isprimitivetype(P) && throw(error(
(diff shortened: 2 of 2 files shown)


willtebbutt:
Thanks for this @mohamed82008 -- glad you're enjoying the package. A couple of questions:
1. I'm not super familiar with the `Enum` mechanism in Julia. Am I basically correct in saying that an `Enum` is just a type which can take one of a specified number of named values? i.e. 0 -> name_0, 1 -> name_1, ... ?
2. Did you get a sensible error? In particular, did it mention specifically that there wasn't a fallback tangent type for `Enum` because it's a primitive type? (I just want to make sure that you saw the error that I'm hoping you saw haha)
3. Could you please bump the patch version? :)

Completely fair re tests -- I need to document that better!

You would just need to add another tuple to [this vector](https://github.com/compintell/Mooncake.jl/blob/997398c6189e0a609670c8492225f1c94ee05b65/src/tangents.jl#L952). The first element should be an `Enum`, and the second just `NoTangent()`. This will then automatically appear in the interface testing. If you need to create a custom `Enum` (if there aren't any public ones in e.g. `Base` that you can just use), you should add it to the [`TestResources`](https://github.com/compintell/Mooncake.jl/blob/997398c6189e0a609670c8492225f1c94ee05b65/src/test_resources.jl#L9) module, and import it from there.

willtebbutt:
Also, don't worry about the segfault in the `rrule-1` CI run -- it definitely is nothing to do with what you're doing, and I'm fixing it in the other PR I have open.

mohamed82008:
>Am I basically correct in saying that an Enum is just a type which can take one of a specified number of named values? i.e. 0 -> name_0, 1 -> name_1, ... ?

Yes it is basically a named integer type where a variable of this type can take one of a set of pre-defined values. Each enum value has both a name and an integer value associated. The integer values don't have to start from 0 though or be consecutive.

>Did you get a sensible error? In particular, did it mention specifically that there wasn't a fallback tangent type for Enum because it's a primitive type? (I just want to make sure that you saw the error that I'm hoping you saw haha)

Yes the error was nice and informative.

```julia
ERROR: EnumType is a primitive type. Implement a method of `tangent_type` for it.
```

>Could you please bump the patch version? :)

Sure thing!

>You would just need to add another tuple to [this vector](https://github.com/compintell/Mooncake.jl/blob/997398c6189e0a609670c8492225f1c94ee05b65/src/tangents.jl#L952). The first element should be an Enum, and the second just NoTangent(). This will then automatically appear in the interface testing. If you need to create a custom Enum (if there aren't any public ones in e.g. Base that you can just use), you should add it to the [TestResources](https://github.com/compintell/Mooncake.jl/blob/997398c6189e0a609670c8492225f1c94ee05b65/src/test_resources.jl#L9) module, and import it from there.

Awesome, I will try to do this. Thanks.

mohamed82008:
I think this is ready for review.

willtebbutt (approved):
LGTM -- thanks for this. I'll merge once CI has passed.

willtebbutt:
Not to worry about performance test failure -- things seem to be slightly higher variance in 1.11 for things which allocate. I'm going to merge and tag.

Thanks again for the contribution @mohamed82008

PR #295, opened 2024-10-20
Fix foreigncall test suite

willtebbutt:
The foreigncall rrule test suite reliably segfaults due to user-error (I failed to properly preserve the data underlying some pointers I'm using in some tests).


willtebbutt:
foreigncall rrules ran successfully so I'm merging + tagging.

PR #307, opened 2024-10-22
Change `sum` to `_sum` in `kron` benchmarks 

yebai:


willtebbutt:
Why?

willtebbutt:
Ah okay. `kron` and friends suffer the same problems as `sum`, so you might see a small improvement in the benchmarks, but I doubt that it will be very substantial. Interested to see though.

bench/run_benchmarks.jl
```diff
@@ -60,7 +60,7 @@ end
 
 # Zygote has rules for both sum and kron, so it's interesting to compare this against the
 # other frameworks because they don't have rules for kron, and maybe for not for sum.
-_kron_sum(x::AbstractMatrix{<:Real}, y::AbstractMatrix{<:Real}) = sum(kron(x, y))
+_kron_sum(x::AbstractMatrix{<:Real}, y::AbstractMatrix{<:Real}) = _sum(kron(x, y))
```
yebai:
```suggestion
_kron_sum(x::AbstractMatrix{<:Real}, y::AbstractMatrix{<:Real}) = _sum(identity, kron(x, y))
```

willtebbutt:
It looks like Zygote _really_ doesn't like `_sum`. You might want to consider disabling Zygote for this run. You can do so by adding this method to the benchmark suite:
```julia
should_run_benchmark(::Val{:zygote}, ::Any, x) = false
```

PR #309, opened 2024-10-22
Prevent Stack Overflow

willtebbutt:
Resolves a problem that @sunxd3 encountered when generating a large unrolled IR.

I've yet to produce a MWR, but will attempt to do so before merging. @sunxd3 could you run this branch on your problem to confirm that it indeed resolves the issue? (it does on my machine, I'm just keen to ensure that it does also on other machines)

edit: I do not understand _why_ this resolves the problem. Presumably there is some limitation of `vcat(x...)` that tends not to be hit.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.21"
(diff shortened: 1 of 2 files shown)


sunxd3:
can confirm that the fix works, thanks

willtebbutt:
Excellent. Since I really don't have any idea how to construct a MWE, and I don't have the time to dig around at the minute, I'm going to merge + tag this. If a similar problem appears, I'll try harder to great a MWE to tie it down.

Issue #326, opened 2024-10-30
`UnhandledLanguageFeatureException` when differentiating through `logerf`


andreasKroepelin:
I encountered a strange behaviour:

```julia
using DifferentiationInterface, SpecialFunctions
import Mooncake

f(x) = logerf(x[1], x[2])
gradient(f, AutoMooncake(; config=nothing), [1., 2.])
```
results in
```
ERROR: Mooncake.UnhandledLanguageFeatureException("Encountered UpsilonNode: ϒ (%2)")
Stacktrace:
  [1] unhandled_feature(msg::String)
    @ Mooncake ~/.julia/packages/Mooncake/vhMtc/src/interpreter/ir_utils.jl:260
  [2] make_ad_stmts!(stmt::Core.UpsilonNode, ::Mooncake.ID, ::Mooncake.ADInfo)
    @ Mooncake ~/.julia/packages/Mooncake/vhMtc/src/interpreter/s2s_reverse_mode_ad.jl:475
  [3] _broadcast_getindex_evalf
    @ ./broadcast.jl:673 [inlined]
  [4] _broadcast_getindex
    @ ./broadcast.jl:646 [inlined]
  [5] getindex
    @ ./broadcast.jl:605 [inlined]
  [6] macro expansion
    @ ./broadcast.jl:968 [inlined]
  [7] macro expansion
    @ ./simdloop.jl:77 [inlined]
  [8] copyto!
    @ ./broadcast.jl:967 [inlined]
  [9] copyto!
    @ ./broadcast.jl:920 [inlined]
 [10] copy
    @ ./broadcast.jl:892 [inlined]
 [11] materialize(bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{…}, Nothing, typeof(Mooncake.make_ad_stmts!), Tuple{…}})
    @ Base.Broadcast ./broadcast.jl:867
 [12] (::Mooncake.var"#212#214"{Mooncake.ADInfo})(primal_blk::Mooncake.BBlock)
    @ Mooncake ~/.julia/packages/Mooncake/vhMtc/src/interpreter/s2s_reverse_mode_ad.jl:918
 [13] iterate
    @ ./generator.jl:48 [inlined]
 [14] collect_to!(dest::Vector{Tuple{…}}, itr::Base.Generator{Vector{…}, Mooncake.var"#212#214"{…}}, offs::Int64, st::Int64)
    @ Base ./array.jl:838
[...]

PR #328, opened 2024-10-30
More SpecialFunctions Tests

willtebbutt:
Resolves the cause of the error in #326 .

I need to go through the SpecialFunctions docs and check for any more public functions that we're not doing _some_ testing on before merging this.

We definitely cannot handle most of the two-arg bessel functions still, because we do not have the ability to handle only partly-implement rules.


willtebbutt:
Additional tests have been added for various bits of functionality in SpecialFunctions, and an issue opened for a few things which aren't currently working, but really should be.

I will merge this PR and tag a release when CI finishes.

PR #344, opened 2024-11-04
Fix array tangent gen functionality

willtebbutt:
I believe the code that I already have in this PR is enough to fix the problem. I now need to double check that this is the case in all situations that I'm aware of, and write an isolated regression test. This should then be good to go.


willtebbutt:
This is basically good to go now. It seems to have fixed some problems that I was having with DiffEq integration, and fixes some functionality which was quite plainly incorrect.

It does, however, highlight he need to finish off the overhaul that we started a while ago on making tangent generation functionality work properly in the presence of aliasing and circular referencing, but that's not something to tackle right now.

If CI passes, I will merge + tag.

PR #353, opened 2024-11-07
Add Bijectors integration tests

### LINKED ISSUE #348: Add integration test for Bijectors
We should perform similar tests in https://github.com/EnzymeAD/Enzyme.jl/pull/2037 for Mooncake. 

@mhauru, can you copy the tests over?

mhauru:
Copied over from https://github.com/EnzymeAD/Enzyme.jl/pull/2037.

Closes #348

### DIFF
--- a/test/integration_testing/bijectors/bijectors.jl
+++ b/test/integration_testing/bijectors/bijectors.jl
@@ -13,9 +13,10 @@ struct TestCase
     func::Function
     arg::Any
     name::Union{String,Nothing}
+    broken::Bool
 end
 
-TestCase(f, arg; name = nothing) = TestCase(f, arg, name)
+TestCase(f, arg; name = nothing, broken=false) = TestCase(f, arg, name, broken)
 
 """
 A helper function that returns a TestCase that evaluates bijector(inverse(bijector)(x))
@@ -98,6 +99,9 @@ end
             end,
             randn(Xoshiro(23), 7);
             name = "PlanarLayer7",
+            # TODO(mhauru) Broken on v1.11 due to
+            # https://github.com/compintell/Mooncake.jl/issues/319
+            broken=(VERSION >= v"1.11"),
         ),
         TestCase(
             function (x)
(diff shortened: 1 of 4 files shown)


test/integration_testing/bijectors/bijectors.jl
```diff
@@ -0,0 +1,123 @@
...
+    arg::Any
+    name::Union{String,Nothing}
+end
+
+TestCase(f, arg; name = nothing) = TestCase(f, arg, name)
+
+"""
+A helper function that returns a TestCase that evaluates sum(bijector(inverse(bijector)(x)))
+"""
+function sum_b_binv_test_case(bijector, dim; name = nothing, rng = Xoshiro(23))
```
willtebbutt: Why is is that you're summing over the output in all cases?

willtebbutt (commented):
This is great, thank you for opening this.

Regarding the failing test: it's one that we know about already, and is related to [this](https://github.com/compintell/Mooncake.jl/issues/319) rather annoying upstream problem. We need some way of indicating that this is a problem. I propose the following:
1. exclude the offending example from the main test suite
2. call `build_rule` on the test case, which should be enough to trigger the error. Put this call inside a `@test_broken` thing, so that when we eventually fix the problem we find out about it.

Does this sound sensible to you?

test/integration_testing/bijectors/bijectors.jl
mhauru: Vestige from Enzyme complications, removed now

mhauru:
I put in the following, does this do what you wanted?
```julia
        TestCase(
            function (x)
                layer = Bijectors.PlanarLayer(x[1:2], x[3:4], x[5:5])
                flow = Bijectors.transformed(
                    Bijectors.MvNormal(zeros(2), LinearAlgebra.I),
                    layer,
                )
                x = x[6:7]
                return Bijectors.logpdf(flow.dist, x) -
                       Bijectors.logabsdetjac(flow.transform, x)
            end,
            randn(Xoshiro(23), 7);
            name = "PlanarLayer7",
            # TODO(mhauru) Broken on v1.11 due to
            # https://github.com/compintell/Mooncake.jl/issues/319
            broken=(VERSION >= v"1.11"),
        ),
```

```julia
        if case.broken
            @test_broken begin
                test_rule(Xoshiro(123456), case.func, case.arg; is_primitive=false)
                true
            end
        else
            test_rule(Xoshiro(123456), case.func, case.arg; is_primitive=false)
        end
```

willtebbutt (approved):
This is perfect. Thank you!

willtebbutt:
I'll merge when CI has passed.

PR #354, opened 2024-11-07
Improve test_rule docs

willtebbutt:
@mhauru pointed out to me that the `test_rule` docstring didn't properly explain the various kwargs. This rectifies the situation.

### DIFF
--- a/src/test_utils.jl
+++ b/src/test_utils.jl
@@ -494,9 +494,42 @@ tangent field.
 
 This function uses [`Mooncake.build_rrule`](@ref) to construct a rule. This will use an
 `rrule!!` if one exists, and derive a rule otherwise.
+
+# Arguments
+- `rng::AbstractRNG`: a random number generator
+- `x...`: the function (first element) and its arguments (the remainder)
+
+# Keyword Arguments
+- `interface_only::Bool=false`: test only that the interface is satisfied, without testing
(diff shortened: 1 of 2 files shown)


willtebbutt:
@mhauru could you take a quick look at this and let me know if it make sense?

src/test_utils.jl
```diff
@@ -494,9 +494,42 @@ tangent field.
...
 `rrule!!` if one exists, and derive a rule otherwise.
+
+# Arguments
+- `rng::AbstractRNG`: a random number generator
+- `x...`: the function (first element) and its arguments (the remainder)
+
+# Keyword Arguments
+- `interface_only::Bool=false`: test only that the interface is satisfied, without testing
+    correctness. This should generally be set to `true` (the default value), and only
+    disabled if the testing infrastructure is unable to test correctness for some reason
```
mhauru: Is the default value the wrong way around here?

mhauru (changes requested):
Looks good, answers the questions I had. I think one sentence has a true/false flip though.

src/test_utils.jl
willtebbutt: Oooo yes, very good catch.
```suggestion
    correctness. This should generally be set to `false` (the default value), and only
    enabled if the testing infrastructure is unable to test correctness for some reason
```

mhauru (approved):

PR #357, opened 2024-11-08
Update bijectors.jl

yebai:
~~Seems that https://github.com/compintell/Mooncake.jl/issues/319 is now fixed.~~

EDIT: no, it is not.


willtebbutt:
@yebai can we close this, since the problem isn't addressed?

PR #367, opened 2024-11-12
Proper FunctionWrappers Support

willtebbutt:
We need to have some partial support for `FunctionWrapper`s in order to make #320 work properly. Instead of doing a partial job that gets finished at some undetermined point in the future, this PR aims to properly support `FunctionWrapper`s in Mooncake.

edit: I also implemented support for the `Base.TwicePrecision` which is used as part of the implementation of the various range objects which get used everywhere. e.g.
```julia
julia> typeof(range(0.0, 10.0; length=50))
StepRangeLen{Float64, TwicePrecision{Float64}, TwicePrecision{Float64}, Int64}
```
This type does quite a lot of custom arithmetic which isn't algorithmically differentiable (various `bitcast` calls, for example), so quite a few rules are required to get this to work well in practice. As a result of having this, we should be able to used range-like objects throughout our code, and expect everything to work.

Note that Zygote never managed to make this work, and I'm reasonably sure that this won't work in Enzyme either.

edit: e.g.
```julia
using Enzyme, Mooncake

julia> @noinline bar(x) = x; # no-inline so that `foo` can't optimise away all of the interesting stuff.

julia> foo(x) = bar(10 * range(x[1], x[2]; step=x[3]))[5]; # do some stuff with a StepRangeLen;

julia> x = Duplicated([5.0, 10.0, 1.1], [0.0, 0.0, 0.0])
Duplicated{Vector{Float64}}([5.0, 10.0, 1.1], [0.0, 0.0, 0.0])

julia> Enzyme.autodiff(Reverse, foo, Active, x)
<a lengthy error message>

julia> rule = Mooncake.build_rrule(foo, [5.0, 10.0, 1.1]);

julia> Mooncake.value_and_gradient!!(rule, foo, [5.0, 10.0, 1.1])
(94.0, (NoTangent(), [10.0, 0.0, 40.0]))
```


willtebbutt:
I'm going to add some more tests, and then merge this, so that I can start a SciML PR.

willtebbutt:
There appears to be something strange going on with both buildkite, but it's unclear that it is at all related to this PR. Consequently, I'm going to mere + tag when CI passes.

sunxd3:
this is really fantastic 🎉

Issue #337, opened 2024-11-01
`Forward`-mode S2S autograd compiler to derive rules for some low-level functions


yebai:
Due to their lack of backward pass, forward-mode autograd often has considerably different implementation properties than reverse-mode autograd. Given its different performance tradeoffs, I wonder whether forward-mode transformation could be more friendly for autograd compilers than reverse-mode (like Mooncake/Zygote), or at least compensate for some extreme cases of reverse-mode autograd. 

For example, the `sum_1000` example is a vector-input scalar-output function, which is a perfect example for forward-mode autograd but likely hard (at least requiring significantly more compiler optimization efforts) for the reverse-mode compiler to work well. This advantage goes further if we have chunk-mode forward-mode autograd. 

I am talking about the source-to-source approach for both forward- and reverse-mode autograd implementations.

willtebbutt:
> EDIT: another interesting question is whether the forward-mode autograd compiler allows us to handle more Julia language features. For example, the try-catch-end block discovered recently in https://github.com/compintell/Mooncake.jl/issues/326#issue-2624603524

It's possible that it would, but I think there's probably more that can be done of the reverse-mode side of things to extend our current functionality to support `try-catch-end` blocks which contain `Upsilon` / `PhiCNode`s which don't wind up throwing, as is the case in the linked example, and another that was mentioned in https://github.com/compintell/Mooncake.jl/issues/31#issuecomment-2297801898 . 

> I wrote the issue to start some discussions.

Fair enough. Certainly, I think it's true that having forwards-mode which composes with reverse-mode would be nice -- I would quite like be able to compute Hessian-vector products (and, by extension, Hessians) by doing forwards-mode over reverse-mode.

In terms of what would need to be done:
- The tangent type infrastructure that we have for reverse-mode could be used as-is, with no modification. We would, however, need to consider how best to _extend_ it to make it possible to handle batches of tangents (i.e. to allow batch-mode forwards-mode).
[...]

PR #382, opened 2024-11-20
Start docs on compilation process

gdalle:
@willtebbutt following our discussion on IR, you'll find my notes here. I'm thinking some of them can go into the docs (as a substitute for proper docs of `Core.Compiler`) and some can go in #337.


gdalle:
Oh right, I think that's because I added these links starting from `docs/src` but in reality the docs are built from the `docs` working directory, so there is probably one `..` too many. Do you want to remove it and make the necessary rephrasing before merging? My notes are not really mergeable in the current state anyway, so I thought you would want to reformulate and clean them up a bit, possibly correcting inaccuracies.

willtebbutt:
Sounds good -- are you happy for me to push to this branch directly? (I assume that I can do that...)

gdalle:
Yeah take it away. But maybe save the remarks about forward mode specifically to the relevant issue before erasing them

docs/src/developer_documentation/compilation_process.md
```diff
@@ -0,0 +1,80 @@
...
+Beyond the [`interpreter`](https://github.com/compintell/Mooncake.jl/blob/main/src/interpreter/) folder, check out [`tangents.jl`](https://github.com/compintell/Mooncake.jl/blob/main/src/tangents.jl) for forward mode.
+
+`FData` and `RData` are not useful in forward mode, `Tangent` is the right representation.
+
+For testing, `generate_test_functions` from [`test_resources.jl`](https://github.com/compintell/Mooncake.jl/blob/src/test_utils.jl) should all pass.
+Recycle the functionality from reverse mode test utils.
+
+To manipulate `IRCode`, check out the fields:
+
+- `ir.argtypes` is the signature. Some are annotated with `Core.Const` to facilitate constant propagation for instance. Other annotations are `PartialStruct`, `Conditional`, `PartialTypeVar` (see `_type`)
```
willtebbutt:
```suggestion
- `ir.argtypes` is the signature. Some are annotated with `Core.Const` to facilitate constant propagation for instance. Other annotations are `PartialStruct`, `Conditional`, `PartialTypeVar`. `Core.Compiler.widenconst` is used to extract types from these.
```

docs/src/developer_documentation/compilation_process.md
```diff
@@ -0,0 +1,80 @@
...
+Do line by line transformation of the statements and then possibly refresh the CFG.
+
+Example of line-by-line transformations are in `make_ad_stmts!`.
+The `IRCode` nodes are not explicitly documented in <https://docs.julialang.org/en/v1/devdocs/ast/#Lowered-form> or <https://docs.julialang.org/en/v1/devdocs/ssair/#Main-SSA-data-structure>. Might need completion of official docs, but Mooncake docs in the meantime.
+
+Inlining pass can prevent us from using high-level rules by inlining the function (e.g. unrolling a loop).
+The contexts in [`interpreter/contexts.jl`](https://github.com/compintell/Mooncake.jl/blob/src/interpreter/contexts.jl) are `MinimalCtx` (necessary for AD to work) and `DefaultCtx` (ensure that we hit all of the rules).
+Distinction between rules is not well maintained in Mooncake at the moment.
+The function `is_primitive` defines whether we should recurse into the function during AD and break it into parts, or look for a rule.
+Typically if we define a rule we should set `is_primitive` to `true` for the corresponding function.
```
willtebbutt:
```suggestion
If we define a rule we should set `is_primitive` to `true` for the corresponding function.
```

willtebbutt:
@gdalle I'm going to merge this once CI passes so that it's merged, and not just hanging around in this PR. I've tidied it up very slightly. We can continue to edit it in other PRs as we see find things we would like to add to it.

PR #383, opened 2024-11-20
Docs and Docstrings

willtebbutt:
Following a discussion with @gdalle this morning, this PR
- [x] changes how the docs specified (rather than specifying `section_name` => `file_name`, just specifying `file_name` and letting Documenter.jl figure out the appropriate `section_name` itself from the title of the file)
- [x] making internal docstrings into comments
- [x] fixing any issues with docstrings
- [x] bump patch


gdalle:
Beware that once you add docstrings, Documenter will complain if they're not in the docs website somewhere.
I think the easiest fix is an autodocs bloc with Public=false on one of the dev documentation pages

Issue #384, opened 2024-11-20
Precompilation failure on 1.10 due to `Base.generating_output`


lcontento:
`Base.generating_output` is only available on Julia 1.11, so precompilation fail on 1.10.

https://github.com/compintell/Mooncake.jl/blob/19dae92c0ef42980fdc9299621b6d69a5a99de3a/ext/MooncakeLuxLibSLEEFPiratesExtension.jl#L10

willtebbutt:
Eurgh -- thanks for catching this. I'll resolve today and start testing more things on 1.10 as well as 1.11.

willtebbutt:
@lcontento thanks again for prompting this. Running CI on 1.10 in #385  has highlighted a couple of additional test failures that I'm currently investigating. I had hoped to get this merged + tagged today, but it looks like it will be tomorrow now -- apologies for the delay!

lcontento:
Happy to help! And thanks for quickly addressing this!

willtebbutt:
Okay. This should be addressed on v0.4.49 once it's available!

PR #385, opened 2024-11-21
Run everything on 1.10

willtebbutt:
#384 points out that the workaround I introduced in https://github.com/compintell/Mooncake.jl/pull/377 only works on 1.11. I didn't catch this because, so far, I've only been running a subset of CI on 1.11 in order to keep CI times down. This PR will

1. change our CI set up so that all tests are run on LTS and release (preventing this kind of problem from recurring), and
2. fix the problem on 1.10.

If it looks like the amount of time spent running CI is becoming prohibitive again, I'll have to look at ways to speed it up.


willtebbutt:
This is getting there, but there are a few issues on 1.10 that need dealing with. I've made some progress today, and hope to finish off tomorrow.

willtebbutt:
@lcontento it looks like we're nearly there -- down to one error. I have to call it a day now -- should be able to resolve in the morning!

PR #390, opened 2024-11-24
Use JuliaFormatter

### LINKED ISSUE #388: Using JuliaFormatter?
As I start to work on forward mode Mooncake, I realize my fingers go for `Ctrl + Shift + I` ("format document") regularly. But since the current code base is not formatted, this might affect the parts of files that my meaningful changes haven't touched.
Would you be open to setting up a JuliaFormatter style and test?

gdalle:
Fixes #388 

This PR has no semantic impact on the code, the last commit is just massive because I formatted the whole repo.

The only meaningful additions are a JuliaFormatter style file and test.


src/tools_for_rules.jl
```diff
@@ -256,31 +254,31 @@ function increment_and_get_rdata!(f, r, t::CRC.Thunk)
...
+ Assumes that methods of `to_cr_tangent` and `to_mooncake_tangent` are defined such that you
+ can convert between the different representations of tangents that Mooncake and
+ ChainRulesCore expect.
 
-Furthermore, it is _essential_ that
-1. `f(args)` does not mutate `f` or `args`, and
-2. the result of `f(args)` does not alias any data stored in `f` or `args`.
+ Furthermore, it is _essential_ that
+ 1. `f(args)` does not mutate `f` or `args`, and
+ 2. the result of `f(args)` does not alias any data stored in `f` or `args`.
```
willtebbutt, line 276: This looks like a bug to me. Does it also look like a bug to you @gdalle ?

test/runtests.jl
```diff
@@ -3,6 +3,7 @@ include("front_matter.jl")
 @testset "Mooncake.jl" begin
     if test_group == "aqua"
```
willtebbutt: We should probably rename this test group to "automated quality assurance", or really anything which highlights that it's not just Aqua.jl anymore.

willtebbutt:
Thanks for doing this. I know there aren't going to be any semantic changes, but could we bump the patch anyway please?

test/runtests.jl
gdalle: done

src/tools_for_rules.jl
gdalle: Perhaps, I've never used `@doc` in my own code so I've never encountered it. In any case, JuliaFormatter is the ground truth, so this might be grounds for opening an issue there. Not much we can do from here, except if the style file lets us skip the `@doc` docstrings.

gdalle:
Done

src/tools_for_rules.jl
willtebbutt: To be honest, I think I'm probably misusing the `@doc` macro here -- I believe that we can just remove it and use a regular docstring.

gdalle: let's fix that in a separate PR?

willtebbutt: Makes sense. Let's get this merged.

willtebbutt:
@gdalle if you're happy for this to get merged, please mark as ready for review, and I'll approve and merge.

willtebbutt (approved):

PR #408, opened 2024-12-04
corrected matrix multiplication example

RoyCCWang:
This PR fixes a typo in an example from the documentation. Instead of a large PR with multiple documentation typo fixes, I will submit separate PRs in the future if I see typos.

### DIFF
--- a/docs/src/understanding_mooncake/algorithmic_differentiation.md
+++ b/docs/src/understanding_mooncake/algorithmic_differentiation.md
@@ -189,9 +189,11 @@ where
 Df\left[X\right]:\mathbb{R}^{M\times N} & \rightarrow\mathbb{R}^{N\times N}\\
 V & \mapsto V^{T}X+X^{T}V.
 \end{align*}
+```
  By definition, an adjoint ``\mathcal{B}_{f\left(X\right)}:\mathbb{R}^{N\times N}\rightarrow\mathbb{R}^{M\times N}``
 of ``\mathcal{A}_{X}`` with respect to the inner products ``g`` and
 ``h`` must satisfy 
+```math
 \begin{align*}
 \forall V\in\mathbb{R}^{M\times N},U\in\mathbb{R}^{N\times N}, & g\left(\mathcal{B}_{f\left(X\right)}U,V\right)=h\left(U,\mathcal{A}_{X}V\right).
 \end{align*}
--- a/docs/src/understanding_mooncake/algorithmic_differentiation.md
+++ b/docs/src/understanding_mooncake/algorithmic_differentiation.md
@@ -159,41 +159,80 @@ We will put all of these problems in a single general framework later on.
 #### An Example with Matrix Calculus
 
 We have introduced some mathematical abstraction in order to simplify the calculations involved in AD.
-To this end, we consider differentiating ``f(X) := X^\top X``.
+Let's consider differentiating the function
+```math
+\begin{align*}
+f:\mathbb{R}^{M\times N} & \rightarrow\mathbb{R}^{N\times N},\\
+X & \mapsto X^{T}X.
+\end{align*}
+```
 Results for this and similar operations are given by [giles2008extended](@cite).
 A similar operation, but which maps from matrices to ``\RR`` is discussed in Lecture 4 part 2 of the MIT course mentioned previouly.
-Both [giles2008extended](@cite) and [Lecture 4 part 2](https://ocw.mit.edu/courses/18-s096-matrix-calculus-for-machine-learning-and-beyond-january-iap-2023/resources/ocw_18s096_lecture04-part2_2023jan26_mp4/) provide approaches to obtaining the derivative of this function.
-
-Following either resource will yield the derivative:
+Both [giles2008extended](@cite) and [Lecture 4 part 2](https://ocw.mit.edu/courses/18-s096-matrix-calculus-for-machine-learning-and-beyond-january-iap-2023/resources/ocw_18s096_lecture04-part2_2023jan26_mp4/) provide approaches to obtaining the derivative of this function. Following either resource will yield the derivative:
 ```math
 D f [X] (\dot{X}) = \dot{X}^\top X + X^\top \dot{X}
 ```
-Observe that this is indeed a linear operator (i.e. it is linear in its argument, ``\dot{X}``). For the rest of this example, we use the matrix A as the coordinate-form representation of the operator ``Df\left[X\right]:\mathbb{R}^{N\times M}\rightarrow\mathbb{R}^{N\times M}``, with respect to the elementary basis of ``\mathbb{R}^{N\times M}``.
+Observe that this is indeed a linear operator (i.e. it is linear in its argument, ``\dot{X}``).
 
-Our goal is to find the adjoint operator by inspection. The equation that we inspect comes from the definition of the usual inner product of matrices: 
+Before we can discuss adjoint operators, we need to choose an inner product for our problem. In this example, we use the usual dot product as the inner products for the vector spaces involved. In matrix notation, the inner products ``g`` and ``h`` are expressed as: 
 ```math
 \begin{align*}
-\forall A,B\in\mathbb{R}^{N\times M}, & \left\langle A,B\right\rangle :=\text{tr}\left(A^{T}B\right).
+\forall A,B\in\mathbb{R}^{M\times N}, & g\left(A,B\right)=\text{tr}\left(A^{T}B\right)=\sum_{i}A_{i}B_{i},\\
+\forall A,B\in\mathbb{R}^{N\times N}, & h\left(A,B\right)=\text{tr}\left(A^{T}B\right)=\sum_{i}A_{i}B_{i}.
 \end{align*}
 ```
-Let ``V:=\dot{X}`` and ``U:=\bar{Y}`` to reduce clutter. Apply this inner product to the adjoint operator condition, where the matrix ``B`` is the coordinate-form representation of the adjoint operator:
+
+For a given ``X\in\mathbb{R}^{M\times N}``, let ``\mathcal{A}_{X}:=Df\left[X\right]``
+where 
 ```math
 \begin{align*}
-\left\langle BU,V\right\rangle  & =\left\langle U,AV\right\rangle ,\\
- & =\left\langle U,V^{T}X+X^{T}V\right\rangle ,\\
+Df\left[X\right]:\mathbb{R}^{M\times N} & \rightarrow\mathbb{R}^{N\times N}\\
+V & \mapsto V^{T}X+X^{T}V.
+\end{align*}
+ By definition, an adjoint ``\mathcal{B}_{f\left(X\right)}:\mathbb{R}^{N\times N}\rightarrow\mathbb{R}^{M\times N}``
+of ``\mathcal{A}_{X}`` with respect to the inner products ``g`` and
+``h`` must satisfy 
+\begin{align*}
+\forall V\in\mathbb{R}^{M\times N},U\in\mathbb{R}^{N\times N}, & g\left(\mathcal{B}_{f\left(X\right)}U,V\right)=h\left(U,\mathcal{A}_{X}V\right).
+\end{align*}
+```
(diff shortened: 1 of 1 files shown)


docs/src/understanding_mooncake/algorithmic_differentiation.md
```diff
@@ -168,27 +168,32 @@ Following either resource will yield the derivative:
 ```math
 D f [X] (\dot{X}) = \dot{X}^\top X + X^\top \dot{X}
 ```
-Observe that this is indeed a linear operator (i.e. it is linear in its argument, ``\dot{X}``).
-(You can always plug it in to the definition of the Frechet derivative to confirm that it is indeed the derivative.)
+Observe that this is indeed a linear operator (i.e. it is linear in its argument, ``\dot{X}``). For the rest of this example, we use the matrix A as the coordinate-form representation of the operator ``Df\left[X\right]:\mathbb{R}^{N\times M}\rightarrow\mathbb{R}^{N\times M}``, with respect to the elementary basis of ``\mathbb{R}^{N\times M}``.
```
willtebbutt: Does the matrix `A` exist? Unless I'm mistaken, it's not the case that you can necessarily represent any linear operator on a matrix via multiplication with another matrix.

If you treat a matrix `X` of size `P x Q` as a vector of length `PQ`, you can certainly represent any linear transform via a matrix of size `PQ x PQ`, but (if my understanding of what you're doing here is correct), that's not what's happening here.

docs/src/understanding_mooncake/algorithmic_differentiation.md
```diff
@@ -168,27 +168,32 @@ Following either resource will yield the derivative:
...
-    \langle \bar{Y}, D f [X] (\dot{X}) \rangle &= \langle \bar{Y}, \dot{X}^\top X + X^\top \dot{X} \rangle \nonumber \\
-        &= \textrm{tr} (\bar{Y}^\top \dot{X}^\top X) + \textrm{tr}(\bar{Y}^\top X^\top \dot{X}) \nonumber \\
-        &= \textrm{tr} ( [\bar{Y} X^\top]^\top \dot{X}) + \textrm{tr}( [X \bar{Y}]^\top \dot{X}) \nonumber \\
-        &= \langle \bar{Y} X^\top + X \bar{Y}, \dot{X} \rangle. \nonumber
-\end{align}
-```
-We can read off the adjoint operator from the first argument to the inner product:
+\begin{align*}
+\left\langle BU,V\right\rangle  & =\left\langle U,AV\right\rangle ,\\
+ & =\left\langle U,V^{T}X+X^{T}V\right\rangle ,\\
```
willtebbutt: To my previous point regarding the existence of `A`, it's not obvious to me that the jump between these two lines -- substituting `V^T X + X^T V` for `AV` -- can be made sense of.

Concretely: does there exist a matrix `A` such that `AV = V^T X + X^T V` for any `X` and `V`?

docs/src/understanding_mooncake/algorithmic_differentiation.md
```diff
@@ -168,27 +168,32 @@ Following either resource will yield the derivative:
...
+ & =\text{tr}\left(U^{T}\left(V^{T}X+X^{T}V\right)\right),\\
+ & =\text{tr}\left(V^{T}XU^{T}\right)+\text{tr}\left(U^{T}X^{T}V\right),\\
+ & =\left\langle V,XU^{T}\right\rangle +\left\langle U^{T}X^{T},V\right\rangle ,\\
+ & =\left\langle XU^{T}+XU,V\right\rangle ,
+\end{align*}
+```
+ for any matrices ``A\in\mathbb{R}^{N\times M}``, ``B\in\mathbb{R}^{M\times L}``, and ``C\in\mathbb{R}^{L\times N}`` for some ``N``, ``M``, ``L``. By inspection, the coordinate-form of the adjoint operator is given by the matrix
```
willtebbutt: Unless I'm mistaken, `C` is not used anywhere?

docs/src/understanding_mooncake/algorithmic_differentiation.md
```diff
@@ -168,27 +168,32 @@ Following either resource will yield the derivative:
...
 ```math
-D f [X]^\ast (\bar{Y}) = \bar{Y} X^\top + X \bar{Y}.
+\begin{align*}
+B=Df\left[X\right]^{*}\left(U\right)= & XU^{T}+XU,
+\end{align*}
 ```
+where ``U:=\bar{Y}`` in this example.
```
willtebbutt: This is defined further up. Perhaps we can remove it?

docs/src/understanding_mooncake/algorithmic_differentiation.md
```diff
@@ -168,27 +168,32 @@ Following either resource will yield the derivative:
...
+ & =\text{tr}\left(V^{T}XU^{T}\right)+\text{tr}\left(U^{T}X^{T}V\right),\\
+ & =\left\langle V,XU^{T}\right\rangle +\left\langle U^{T}X^{T},V\right\rangle ,\\
+ & =\left\langle XU^{T}+XU,V\right\rangle ,
+\end{align*}
+```
+ for any matrices ``A\in\mathbb{R}^{N\times M}``, ``B\in\mathbb{R}^{M\times L}``, and ``C\in\mathbb{R}^{L\times N}`` for some ``N``, ``M``, ``L``. By inspection, the coordinate-form of the adjoint operator is given by the matrix
 ```math
-D f [X]^\ast (\bar{Y}) = \bar{Y} X^\top + X \bar{Y}.
+\begin{align*}
+B=Df\left[X\right]^{*}\left(U\right)= & XU^{T}+XU,
```
willtebbutt: I'm not sure how to make sense of this line.

My understanding is that you're saying we should view `B` as the coordinate form representation of the adjoint of the derivative. So it seems like we have a "type" error -- `D f [X]^\ast (U)` is the matrix resulting from applying the adjoint operator to `U`, but `B` is the operator itself. Should there be a `U` on the rhs of `B`?

(I have similar concerns about the existence of `B` as I do `A` -- so let's resolve the discussion around that before addressing this line)

willtebbutt (commented):
Thanks for opening this! It'll be great to have this fixed.

I've got a couple of concerns about the new presentation that I would like to address prior to merging though.

docs/src/understanding_mooncake/algorithmic_differentiation.md
RoyCCWang: Yes, there are a few mistakes as you pointed out. I wrote things basis-free originally (via operators), then changed it hastily as I was typing for this example without realising I made these mistakes.

I'll post a proposed change with a bit more discussion soon.

willtebbutt: Cool. FWIW, I think the basis-free approach that I originally adopted probably makes most sense here -- happy to be proven wrong though!

RoyCCWang: Okay, I'm done writing the changes up. Please see the new commit.

willtebbutt (commented):
Firstly, apologies for taking so long to take a look at this. It's been a hectic week!

I obviously really like that the PR fixes the typo in the original example, but it also introduces a range of stylistic changes that I either want to a) adopt everywhere, or b) not have at all. I'm not keen on having just this example contain the stylistic changes, because it makes for an inconsistent reading experience -- I would rather discuss style changes in a separate issue.

Could I propose the following? Change this PR to maintain the same style  as the rest of the examples (use of angle brackets to denote inner products, `D f [X]^\ast` to represent the adjoint of the derivative, retain the use of dots and bars over symbols to denote tangents and gradients rather than introducing new symbols), and just fix the typo in the original text + add in the extra detail around the cyclic trace property.

Apologies for asking you to remove a bunch of the changes you've proposed here. I really do appreciate the contribution!

RoyCCWang:
No problem, please see the latest commit.

docs/src/understanding_mooncake/algorithmic_differentiation.md
```diff
@@ -178,16 +178,17 @@ Using the usual definition of the inner product between matrices,
...
-    \langle \bar{Y}, D f [X] (\dot{X}) \rangle &= \langle \bar{Y}, \dot{X}^\top X + X^\top \dot{X} \rangle \nonumber \\
-        &= \textrm{tr} (\bar{Y}^\top \dot{X}^\top X) + \textrm{tr}(\bar{Y}^\top X^\top \dot{X}) \nonumber \\
-        &= \textrm{tr} ( [\bar{Y} X^\top]^\top \dot{X}) + \textrm{tr}( [X \bar{Y}]^\top \dot{X}) \nonumber \\
-        &= \langle \bar{Y} X^\top + X \bar{Y}, \dot{X} \rangle. \nonumber
-\end{align}
-```
-We can read off the adjoint operator from the first argument to the inner product:
+\begin{align*}
+\langle\bar{Y},Df[X](\dot{X})\rangle & =\langle\bar{Y},\dot{X}^{\top}X+X^{\top}\dot{X}\rangle\\
+ & =\textrm{tr}(\bar{Y}^{\top}\left(\dot{X}^{T}X+X^{T}\dot{X}\right))\\
```
willtebbutt:
```suggestion
 & =\textrm{tr}(\bar{Y}^{\top}\left(\dot{X}^{\top}X+X^{\top}\dot{X}\right))\\
```

docs/src/understanding_mooncake/algorithmic_differentiation.md
```diff
@@ -178,16 +178,17 @@ Using the usual definition of the inner product between matrices,
...
+ & =\textrm{tr}(\dot{X}^{\top}X\bar{Y}^{\top})+\textrm{tr}(\bar{Y}^{\top}X^{\top}\dot{X})\\
+ & =\langle\dot{X},X\bar{Y}^{\top}\rangle+\langle\bar{Y}^{\top}X^{\top},\dot{X}\rangle\\
```
willtebbutt:
```suggestion
 & =\langle\dot{X},X\bar{Y}^{\top}\rangle+\langle X\bar{Y},\dot{X}\rangle\\
```
I think the first arg to the second inner product should be transposed, since `<A, B> = tr(A^T B)`

docs/src/understanding_mooncake/algorithmic_differentiation.md
```diff
@@ -178,16 +178,17 @@ Using the usual definition of the inner product between matrices,
...
+ & =\langle X\bar{Y}^{\top}+X\bar{Y},\dot{X}\rangle.
```
willtebbutt, line 186:
```suggestion
 & =\langle X(\bar{Y}^{\top} + \bar{Y}),\dot{X}\rangle.
```
should we simplify to this?

docs/src/understanding_mooncake/algorithmic_differentiation.md
```diff
@@ -178,16 +178,17 @@ Using the usual definition of the inner product between matrices,
...
+\end{align*}
+```
+The linearity of inner products and trace, and the [cyclic property of trace](https://en.wikipedia.org/wiki/Trace_(linear_algebra)#Cyclic_property) was used in the above. We can read off the adjoint operator from the first argument to the inner product:
 ```math
-D f [X]^\ast (\bar{Y}) = \bar{Y} X^\top + X \bar{Y}.
+D f [X]^\ast (\bar{Y}) = \bar{Y}^{\top} X + X \bar{Y}.
```
willtebbutt: I think this now has a different typo haha -- the multiplication order of `X` and `\bar{Y}` has been swapped, and I don't believe it should have been.
```suggestion
D f [X]^\ast (\bar{Y}) = X (\bar{Y}^{\top} + \bar{Y}).
```

willtebbutt (commented):
Thanks for this. I think we still have a couple of typos, but are nearly there. Please correct me if I'm mistaken!

docs/src/understanding_mooncake/algorithmic_differentiation.md
RoyCCWang: Thanks for catching the typos. I put the simplification in the end. Please see latest commit.

PR #411, opened 2024-12-05
Remove hack and insist on Julia version upgrade

willtebbutt:
In https://github.com/compintell/Mooncake.jl/pull/377 we introduced a hack which ensured that code was loaded in roughly the correct order when extensions depended upon other extensions. Happily this ceased to be necessary in 1.11.2, so this PR
1. removes the hacky code
2. modifies the Julia compat bound to ensure that people still running 1.11.0 or 1.11.1 do not see this problem.


willtebbutt:
Good point -- I'll give that a try.

PR #416, opened 2024-12-12
Proper Circular Reference Handling

willtebbutt:
This PR targets finishing off https://github.com/compintell/Mooncake.jl/issues/204

This seems to be blocked by https://github.com/JuliaLang/julia/issues/56775 currently -- I make use of `deepcopy` inside of `TestUtils.test_rule_correctness`, and it falls over for circular references involving `Array`s currently. Per the linked issue, I'm not 100% certain whether it's a bug in Julia 1.11.2, or expected behaviour (it feels like a bug though).

I'm not 100% sure who's attention we should attempt to get over https://github.com/JuliaLang/julia/issues/56775 -- any idea @yebai @mhauru @penelopeysm ?

I'll also have a think about whether I can avoid the use of `deepcopy` -- I'm reasonably confident that I need it, but I might be wrong.

edit: definitely need `deepcopy`, because I need to be able to tell whether a rule has successfully returned inputs to their initial state. In order to make this comparison, I need to keep the initial state around, which requires the use of `deepcopy`. The only workaround would be to roll my own version of `deepcopy`, which I'm not keen on.

edit2: `deepcopy` seems to be fine on 1.10 though, so I could just do development on 1.10, and only run tests involving circular references on 1.10 for now.

edit3: this is now basically ready to go. Hopefully I'll be able to merge tomorrow after I do a final pass through it.


penelopeysm:
> Currently CI is failing on Turing.jl master due to the stack overflows.

Do any of you know what changed that made the tests start to fail? Comparing [the last success](https://github.com/TuringLang/Turing.jl/actions/runs/12092053453/job/33721289002) with the [first failure](https://github.com/TuringLang/Turing.jl/actions/runs/12108208232/job/33755957117) I see very little change:

(last success -> first failure)
DifferentiationInterface 0.6.23 -> 0.6.24
SciMLBase 2.65.0 -> 2.65.1
UnsafeAtomicsLLVM 0.2.1 -> 0.2.2

Both are running Julia 1.11.1.

willtebbutt:
So, what's going on is previously DI wasn't exposed to the `Mooncake.set_to_zero!!` function. Since I exposed the `prepare_pullback_cache!!` functionality, this has changed -- I `set_to_zero!!` existing memory, rather than calling `zero_tangent` and allocating new memory. In principle this ought to do exactly the same thing, but due to un-finished implementation, it happens to be the case that `set_to_zero!!` doesn't properly support circular referencing, while `zero_tangent` does. It turns out that in the test suite of Turing.jl, you wind up with a circular reference in one of the args to AD, I think due to boxing associated to constructing a Turing.jl model inside of a testset, and some weirdness there (usually, we don't have circular references in Turing.jl models, thankfully).

Anyway, I'm going to make a tiny version of this PR in which `set_to_zero!!` is fixed, and get that merged in the next hour or two -- I'd rather we didn't have to pin the version of DI in Turing.jl.

This does raise a separate question though: why are things getting boxed, and generating circular references, in the Turing.jl tests? It's possible that this kind of thing will have performance implications for the test suite, so it might actually be good to look into what's going on there if we want to get the test suite to run more quickly 🤷

willtebbutt:
This is blocked by https://github.com/JuliaLang/julia/issues/56775

willtebbutt:
A fix for https://github.com/JuliaLang/julia/issues/56775 has now been merged, and should appear in 1.11.3.

PR #422, opened 2024-12-13
StanBlocks Perf Fix

willtebbutt:
Sadly, it looks like introducing an additional generated function resolves some quite important inference issues uncovered in [this issue](https://github.com/nsiccha/StanBlocks.jl/issues/1). This should not be merged yet, because I've not managed to find an MWE which replicates the problem, so we don't yet have a regression test.

### DIFF
--- a/src/tangents.jl
+++ b/src/tangents.jl
@@ -1044,6 +1044,8 @@ function tangent_test_cases()
         (a=3, b=randn(10)),
         (a=randn(10), b=randn(10)),
         (Base.TOML.ErrorType(1), NoTangent()), # Enum
+        (((((randn(33)...,),),),),),
+        (((((((((randn(33)...,),),),),),randn(5)...),),),),
     ]
     VERSION >= v"1.11" && push!(rel_test_cases, fill!(Memory{Float64}(undef, 3), 3.0))
     return vcat(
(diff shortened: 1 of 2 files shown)


willtebbutt:
This now has a failing test associated to it.

However, its use of a generated function is quite unsafe. Specifically, it is intended that users will add methods to `tangent_type` in some instances and, moreover, I add methods to `tangent_type` in package extensions inside Mooncake. The method of `tangent_type` for `Tuple`s calls `tangent_type` internally, as part of codegen. This is a problem, because when users add methods to `tangent_type`, the changes that they introduce are not guaranteed to cause recompilation of this method of `tangent_type` in the situations that you would expect for normal Julia methods. For example, if you add a method of the form
```julia
tangent_type(::Type{Foo}) = Bar
```
and then ask for
```julia
tangent_type(Tuple{Foo})
```
you likely won't get
```julia
Tuple{Bar}
```
but rather whatever `tangent_type` returns for `Foo` by default.


Consequently, I need to refactor the implementation of the generated function associated to `tangent_type` (and a few other functions, in another PR) to avoid this behaviour.

src/tangents.jl
```diff
@@ -1044,6 +1044,8 @@ function tangent_test_cases()
         (a=3, b=randn(10)),
         (a=randn(10), b=randn(10)),
         (Base.TOML.ErrorType(1), NoTangent()), # Enum
+        (((((randn(33)...,),),),),),
```
yebai, line 1058:
```suggestion
        # Regression tests to catch type inference failures, see https://github.com/compintell/Mooncake.jl/pull/422
        (((((randn(33)...,),),),),),
```

PR #426, opened 2024-12-16
Make generated functions safe for extension

willtebbutt:
As discussed in #422 , my use of generated functions throughout Mooncake is somewhat unsafe, in the sense that they often use functions which I expect will have methods added to them as part of codegen (see https://github.com/compintell/Mooncake.jl/pull/422#issuecomment-2546304998 for further discussion). I discovered that this is a problem while trying to write the extensions necessary to handle GPUArrays properly. Since this is quite a pervasive issue, I need to resolve it asap in order to finish up our initial GPU support work.

It is helpful to have this working PR open in order to regularly run CI to check that nothing has broken as I work through the various fixes which are needed.

todo:
- [x] fix up remaining problematic generated functions
- [x] add new benchmark case for highly nested tuple -- turned into separate issue
- [x] resolve all perf problems


willtebbutt:
Note: current performance-related test failures are not replicated when running without the flags used on the runners (I'm not seeing the performance issues locally). I'll need to figure out how to fix this...

PR #430, opened 2024-12-20
Eliminate Rule Type Mismatch Errors Forever

willtebbutt:
A really annoying class of errors is handled here: https://github.com/compintell/Mooncake.jl/blob/8c478a88bbd89d12507441314f21fb193e4eceef/src/interpreter/s2s_reverse_mode_ad.jl#L1726

This error occurs because we have to guess what type inference is going infer the return for the forwards-pass IR to be, based on the primal IR (its argument types and return type). We are able to correctly guess this type the vast majority of the time on correctly derived code -- we only occasionally runs into problems when inference does quite special things for e.g. small unions of `Tuple`s. However, this is a loosing battle -- see the docstring for the new `opaque_closure` function for more details.

The bigger problem is this: when something goes wrong when deriving a rule we often see this error, rather than the code running until we hit the actual error which caused the problem. This is bad for both developers and users, because this error is almost never the root cause of a problem (I can't remember the last time it was). Instead, we have to do some detective work to figure out where the actual problem lies. I find this hard, so it will be extremely challenging for a new user.

Now, unless something has gone badly wrong, it's never the case that the type we guess for the return type is invalid, it's just that we guess a different type to the one that inference returns. A recent example of this I encountered involved us guessing that inference would return `Tuple{Union{A, B}, C}` when in fact it returned `Union{Tuple{A, C}, Tuple{B, C}}` (or something along those lines). In this case, the collection of concrete types which subtype these types is the same, so either is fine. Which one inference winds up picking is essentially an implementation detail, and (could) change between Julia versions. (I'm quite certain that this is just the tip of the iceberg in terms of weird edge cases that could cause trouble. Moreover, when this kind of thing goes wrong, it takes me at least a day to figure out what happened and how a fix might work.) It's never the case (modulo bugs) that we say that a function returns `AbstractFloat` when in fact it can return an `Int`.

To solve this problem, I've changed things so that in addition to specifying the argument types + IR for the OpaqueClosures we construct to perform the forwards- and reverse-passes, we specify the return type (previously the return type was derived from the IR + argtypes). This requires a helper function that is basically a tweaked version of some code from `Base.Experimental`. The win is that we get to specify the return type ourselves, so we can just be consistent about it, thereby completely eliminating the class of error described above. We shall never see it again.

The downside of this approach is of course that, if we have a bug and specify an upper bound on the return type of an `OpaqueClosure` we construct which is not in fact a valid upper bound, I imagine that all sorts of terrible things (e.g. segfaults) will happen. To guard against this, I've inserted `typeassert` statements into the IR that AD generates immediately before all `ReturnNode`s on both the forwards- and reverse-passes of AD. This should guard against accidentally returning bad data.

The motivation for dealing with this now is errors I've been encountering in #426 -- I'm really hoping this makes debugging them much more straightforward.

edit: this error has been causing me trouble for almost a year at this point, so I'm very excited to be rid of it.

### DIFF
--- a/src/interpreter/s2s_reverse_mode_ad.jl
+++ b/src/interpreter/s2s_reverse_mode_ad.jl
@@ -133,6 +133,8 @@ struct ADInfo
     debug_mode::Bool
     is_used_dict::Dict{ID,Bool}
     lazy_zero_rdata_ref_id::ID
+    fwd_ret_type::Type
+    rvs_ret_type::Type
 end
 
 # The constructor that you should use for ADInfo if you don't have a BBCode lying around.
@@ -144,6 +146,8 @@ function ADInfo(
     is_used_dict::Dict{ID,Bool},
     debug_mode::Bool,
     zero_lazy_rdata_ref::Ref{<:Tuple},
+    fwd_ret_type::Type,
+    rvs_ret_type::Type,
 )
     shared_data_pairs = SharedDataPairs()
     block_stack = BlockStack()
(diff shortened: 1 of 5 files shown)


src/interpreter/s2s_reverse_mode_ad.jl
```diff
@@ -1690,51 +1739,7 @@ _copy(x::P) where {P<:LazyDerivedRule} = P(x.mi, x.debug_mode)
...
-        result = derived_rule(args...)
-    else
-        err = BadRuleTypeException(rule.mi, sig, typeof(derived_rule), Trule)
-        result = try
-            derived_rule(args...)
-        catch
-            throw(err)
-        end
-    end
-    return result::_rtype(Trule)
```
willtebbutt, line 1739: Getting rid of this code is the win associated to this PR.

src/utils.jl
```diff
@@ -242,3 +242,86 @@ flat_product(xs...) = vec(collect(Iterators.product(xs...)))
 Equivalent to `map(f, flat_product(xs...))`.
 """
 map_prod(f, xs...) = map(f, flat_product(xs...))
+
+"""
```
willtebbutt, line 246: Observe that most of the increase in the number of lines of code is associated to this function, `misty_closure`, and associated tests.

PR #433, opened 2025-01-02
Nested tuple test case

willtebbutt:

### DIFF
--- a/src/test_resources.jl
+++ b/src/test_resources.jl
@@ -602,6 +602,8 @@ function inplace_invoke!(x::Vector{Float64})
     return nothing
 end
 
+highly_nested_tuple(x) = ((((x, ),), x), x)
+
 function generate_test_functions()
     return Any[
         (false, :allocs, nothing, const_tester),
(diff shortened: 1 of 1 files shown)


willtebbutt:
> I vaguely remember testing allocation is sufficient to check for type instability, but I forgot why. Can you clarify?

I think it's to do with boxing -- if the type of a piece of data is unknown, then you need to allocate memory for it on-the-fly -- under the hood this is done (I believe) by allocating a spot on the heap for the data.

> There is a nice, significant improvement in Mooncake's performance on kron_sum; do we know why?

I'm not entirely sure that this is new. The numbers seem to have quite high variance, e.g. https://github.com/compintell/Mooncake.jl/pull/431 and https://github.com/compintell/Mooncake.jl/pull/402 have quite different numbers, but I'm don't believe that I've done anything which ought to substantially improve performance. Note also that the ReverseDiff numbers appear to vary quite a bit between these examples, and I'm not aware of anything that has changed in ReverseDiff which should have a drastic effect on its numbers.

PR #434, opened 2025-01-03
Tangent Interface Docs

willtebbutt:
@jollywatt per our discussion elsewhere, here is a stab at aggregating all of the information that I have about the tangent interface into a single location. When you get a minute, could you let me know if this is the kind of thing that you need to do what you wanted to do, or if there's additional info that you'll need?


docs/src/developer_documentation/tangents.md
```diff
@@ -0,0 +1,47 @@
+# Tangents
+
+As discussed in [Representing Gradients](@ref), Mooncake requires that each "primal" type be associated to a unique "tangent" type, given by the function [tangent_type](@ref).
+Moreover, we must be able to "split" a given tangent into its _fdata_ ("forwards-data") and _rdata_ ("reverse-data"), whose types are given by [fdata_type](@ref) and `rdata_type` respectively.
```
sunxd3: maybe miss `ref` for `rdata_type`?

src/test_utils.jl
```diff
@@ -765,15 +765,22 @@ function functions_for_mutable_structs()
 end
 
 """
-    test_rule_and_type_interactions(rng::AbstractRNG, x)
+    test_rule_and_type_interactions(rng::AbstractRNG, p)
```
sunxd3: (not for this line) in the first sentence of the following doc, there is an `x` possibly should be `p` (also the first sentence of the second paragraph)?

sunxd3 (commented):
I haven't tried to implement a tangent type and use these functions, so couldn't verify claims like `If all the tests in these functions pass, then you have satisfied the interface.` All the comments are only cosmetic. 

But this is clearly a really good clarification!

docs/src/developer_documentation/tangents.md
willtebbutt: Ah, yes. For some reason, it didn't like me doing that. I'm not 100% sure why.

src/test_utils.jl
```diff
@@ -765,15 +765,22 @@ function functions_for_mutable_structs()
...
 
 Check that a collection of standard functions for which we _ought_ to have a working rrule
 for `x` work, and produce the correct answer. For example, the `rrule!!` for `typeof` should
```
willtebbutt:
```suggestion
for `p` work, and produce the correct answer. For example, the `rrule!!` for `typeof` should
```

src/test_utils.jl
```diff
@@ -765,15 +765,22 @@ function functions_for_mutable_structs()
...
 work correctly on any type, we should have a working rule for `getfield` for any
 struct-type, and we should have a rule for `setfield!` for any mutable struct type.
+See implementation for details.
 
 The purpose of this test is to ensure that, for any given `x`, the full range of primitive
```
willtebbutt:
```suggestion
The purpose of this test is to ensure that, for any given `p`, the full range of primitive
```

src/test_utils.jl
willtebbutt: Good catch, thanks!

Jollywatt:
@willtebbutt I've had a look at these docs and tried to implement the interface for a custom type. I was almost able to, but in particular I had to add methods for
- `_get_fdata_field`
- `tangent` (binary method)

in addition to those in the interface. I also ran into problems with a generated function, which I didn't find a way around.

I've put a MWE in this notebook: https://jollywatt.github.io/notes/mooncake-self-tangents

willtebbutt:
Thanks -- I'll take a look at your notebook!

willtebbutt:
@Jollywatt sorry for taking so long to get this fixed. I believe I've addressed the issues that you mentioned above.

In particular, I've added the binary version of `tangent` to the docstring for the tangent splitting tests, and stated in the docstring from the basic rules tests that you need to support `_new_` / `getfield` / `setfield!` etc (which is where the `_get_fdata_field` thing you mentioned is coming from).

Could you let me know if you think this is sufficient? If so, I'll merge this.

willtebbutt:
Ah, yes, you're definitely meant to test on the `<function>_internal` functions, but I've not updated this in the various docstrings 😬 . I forgot that I would need to reflect this change in them. I'll update them and let you know once that's done.

willtebbutt:
@Jollywatt I think that the docstring should now be accurate.

yebai (approved):

willtebbutt:
Thanks for the approval @yebai -- I'm going to wait for @Jollywatt to have another go at this before merging, as doing so seems to have caught issues quite reliably!

yebai:
I am happy to merge this PR and fix further issues (if any) in future PRs.

willtebbutt:
I'm still discussing this with @Jollywatt via email, so would prefer to keep it un-merged for now. We're nearly there, just ironing out the last couple of issues.

yebai:
As mentioned above, having these docs is helpful. Inaccuracies (if any) can be fixed via future PRs.

PR #449, opened 2025-01-27
Include primal time in benchmark table

### LINKED ISSUE #448: Show primal execution time for benchmarks
It is very helpful to collect and show the primal execution time for the CI-based benchmarks, e.g.: https://github.com/compintell/Mooncake.jl/pull/447 We could add a column labelled as "Primal time", together with the reverse pass / primal ratios.

willtebbutt:
Resolves #448 . This won't require a patch version bump. We can merge once we're confident that the table is being produced correctly.


willtebbutt:
Good idea. I'll take a look at how BenchmarkTools / Chairmarks go about doing this.

willtebbutt:
I've updated the formatting for all of the numbers in the table. Times are quoted in terms of `ns`, `ms`, etc, and printed to 2 decimal places.

willtebbutt:
It looks to me like the table is operating correctly, so I'm going to merge. We can tweak it again later if we decide it's not quite how we want it.

PR #451, opened 2025-01-30
Vargargs Bug

### LINKED ISSUE #450: Failure to compute gradient for vcat(...)
There's some funky failure to compute the gradient for the following MWE using vcat. I'm assuming similar things happen for hcat/cat.
# MWE
```julia
import Mooncake

f(x) = vcat(
    x[1],
    g(x[2:2]),
    g(x[3:3]),
    x[4:4]
)
g(x) = x .- log(sum(exp.(x)))
h(x) = sum(f(x))
x = ones(4)
prep = prepare_gradient(h, backend, x)
gradient(h, prep, backend, x)
```
gives me
```
ERROR: BoundsError: attempt to access Tuple{Mooncake.NoRData, Mooncake.NoRData, Mooncake.NoRData, Mooncake.NoRData, Mooncake.NoRData, Float64, Mooncake.NoRData, Mooncake.NoRData} at index [9]
Stacktrace:
  [1] f
    @ ~/github/nsiccha/mwe/Mooncake/index.jl:4 [inlined]
  [2] (::Tuple{…})(none::Mooncake.NoRData)
  [3] Pullback
    @ ~/.julia/packages/Mooncake/nSw4x/src/interpreter/s2s_reverse_mode_ad.jl:888 [inlined]
  [4] h
    @ ~/github/nsiccha/mwe/Mooncake/index.jl:11 [inlined]
  [5] (::Tuple{…})(none::Float64)
  [6] Pullback
    @ ~/.julia/packages/Mooncake/nSw4x/src/interpreter/s2s_reverse_mode_ad.jl:888 [inlined]
  [7] __value_and_pullback!!(::Mooncake.DerivedRule{…}, ::Float64, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}; y_cache::Float64)
    @ Mooncake ~/.julia/packages/Mooncake/nSw4x/src/interface.jl:20
  [8] __value_and_pullback!!
    @ ~/.julia/packages/Mooncake/nSw4x/src/interface.jl:11 [inlined]
  [9] value_and_pullback!!
    @ ~/.julia/packages/Mooncake/nSw4x/src/interface.jl:219 [inlined]
 [10] value_and_pullback(::typeof(h), ::DifferentiationInterfaceMooncakeExt.MooncakeOneArgPullbackPrep{…}, ::AutoMooncake{…}, ::Vector{…}, ::Tuple{…})
    @ DifferentiationInterfaceMooncakeExt ~/.julia/packages/DifferentiationInterface/Tls8M/ext/DifferentiationInterfaceMooncakeExt/onearg.jl:30
 [11] prepare_pullback(::Function, ::AutoMooncake{Nothing}, ::Vector{Float64}, ::Tuple{Bool})
    @ DifferentiationInterfaceMooncakeExt ~/.julia/packages/DifferentiationInterface/Tls8M/ext/DifferentiationInterfaceMooncakeExt/onearg.jl:16
 [12] prepare_gradient(::typeof(h), ::AutoMooncake{Nothing}, ::Vector{Float64})
    @ DifferentiationInterface ~/.julia/packages/DifferentiationInterface/Tls8M/src/first_order/gradient.jl:70
 [13] top-level scope
    @ ~/github/nsiccha/mwe/Mooncake/index.jl:14
```

Varying the arguments given to `vcat` sometimes removes the error. E.g. the following all works:
* `f(x) = vcat(x[1], g(x[3:3]), x[4:4])`
* `f(x) = vcat(x[1], g(x[2:2]), g(x[3:3]), x[4])`
* `f(x) = vcat(x[1:1], g(x[2:2]), g(x[3:3]), x[4:4])`

# Version info:
  julia version 1.10.8
  [a0c0ee7d] DifferentiationInterface v0.6.34
  [da2b9cff] Mooncake v0.4.78

willtebbutt:
Fixes #450 -- this was a systematic problem in the way that we were counting the number of arguments provided to a call site when the call involved a `Vararg`. More explanation to follow later today. The seemingly substantial changes to the plumbing of rule construction are really just needed to change the way in which the number of arguments which are passed to a function are counted (looking at the actual number, rather than looking at the signature, which sometimes doesn't contain sufficient information, as it turns out).

todo:
- [x] explain what was going wrong
- [x] add regression test
- [x] fix failing tests
- [x] tidy up the change in implementation

edit: what was going wrong?

If you have a function of the form
```julia
f(x, y...) = ...
```
you'll see non-splatted calls to it appear in `IRCode` as `f(a, b, c, d)`. For book-keeping-related reasons, we need to know how many elements appear in `y` inside a call to `f` (splatted arguments appear as tuples).

Previously, we were doing this by looking at the signature associated to the `Core.MethodInstance` which appears at a call site. This signature is usually something like
```julia
Tuple{typeof(f), Float64, Float64, Float64, Float64}
```
However, it can also be something like
```julia
Tuple{typeof(f), Float64, Vararg{Float64}}
```
if type inference / the compiler decides that is the best thing to do for some reason. In the first case, everything is fine, but in the second case, I was getting the wrong answer -- plainly, there's not enough information in the second signature to know how many arguments are actually provided at run-time.

The new implementation dispenses with this way of counting arguments, and just encodes in the signature of the `Pullback`, the number of arguments which are _actually_ passed in. If this is known statically, then this will be free -- otherwise it will involve dynamic dispatch, which is what you should expect in this situation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.78"
+version = "0.4.79"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/src/interpreter/s2s_reverse_mode_ad.jl
+++ b/src/interpreter/s2s_reverse_mode_ad.jl
@@ -915,9 +915,8 @@ _copy(::Nothing) = nothing
 function _copy(x::P) where {P<:DerivedRule}
     new_captures = _copy(x.fwds_oc.oc.captures)
     new_fwds_oc = replace_captures(x.fwds_oc, new_captures)
-    new_pb_oc_ref = Ref(replace_captures(x.pb.pb_oc[], new_captures))
-    new_pb = typeof(x.pb)(new_pb_oc_ref)
-    return P(new_fwds_oc, new_pb, x.nargs)
+    new_pb_oc_ref = Ref(replace_captures(x.pb_oc_ref[], new_captures))
+    return P(new_fwds_oc, new_pb_oc_ref, x.nargs)
 end
 
 _copy(x::Symbol) = x
--- a/src/interpreter/s2s_reverse_mode_ad.jl
+++ b/src/interpreter/s2s_reverse_mode_ad.jl
@@ -878,9 +878,7 @@ struct Pullback{Tprimal,Tpb_args,Tpb_ret,isva,nargs}
     pb_oc::Base.RefValue{RuleMC{Tpb_args,Tpb_ret}}
 end
 
-function Pullback(
-    sig, pb_oc::Tpb_oc, isva::Bool, nargs::Int
-) where {A,R,Tpb_oc<:Ref{RuleMC{A,R}}}
+function Pullback(sig, pb_oc::Ref{<:RuleMC{A,R}}, isva::Bool, nargs::Int) where {A,R}
     return Pullback{sig,A,R,isva,nargs}(pb_oc)
 end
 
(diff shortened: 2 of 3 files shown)


willtebbutt:
This is ready to go -- I'll merge + tag once CI has passed.

PR #454, opened 2025-02-02
Ensure the primal time for `gp_lml` is consistent across CI runs. 

yebai:
The primal time for `gp_lml` seems to vary greatly for different CI runs, see, e.g.: 

- 571.0 μs in [here](https://github.com/compintell/Mooncake.jl/pull/447#issuecomment-2612997584)
- 186.0 μs in [here](https://github.com/compintell/Mooncake.jl/pull/451#issuecomment-2624451265)

causing a large variation in the reverse / primal ratio for Mooncake and others. I suspect the variation in primal run time is caused by the different number of working threads for BLAS. 

EDIT: it is strange that with a single thread, the `gp_lml` primal run time is smaller (198.0 μs) below. This result indicates some other factor could significantly influence `gp_lml` primal run time. I don't have a clear idea what that might be.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.79"
+version = "0.4.80"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/src/rrules/blas.jl
+++ b/src/rrules/blas.jl
@@ -299,57 +299,69 @@ function rrule!!(
     return y_dy, symv!_adjoint
 end
 
-for (trmv, elty) in ((:dtrmv_, :Float64), (:strmv_, :Float32))
-    @eval @inline function rrule!!(
-        ::CoDual{typeof(_foreigncall_)},
-        ::CoDual{Val{$(blas_name(trmv))}},
-        ::CoDual,
-        ::CoDual,
-        ::CoDual,
-        ::CoDual,
-        _uplo::CoDual{Ptr{UInt8}},
-        _trans::CoDual{Ptr{UInt8}},
-        _diag::CoDual{Ptr{UInt8}},
-        _N::CoDual{Ptr{BLAS.BlasInt}},
-        _A::CoDual{Ptr{$elty}},
-        _lda::CoDual{Ptr{BLAS.BlasInt}},
-        _x::CoDual{Ptr{$elty}},
-        _incx::CoDual{Ptr{BLAS.BlasInt}},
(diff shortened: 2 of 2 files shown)


willtebbutt:
Interesting — it’s surprising that the AD time doesn’t vary massively between runs. I wonder if there’s something weird happening with GC? I would be surprised if the problem is thread-count related, because AD for GPs also makes heavy use of multi-threading via LAPACK / BLAS when there are multiple threads available.

willtebbutt:
> Any chance that this is due to a slow rule implementation similar to https://github.com/compintell/Mooncake.jl/pull/453

Quite possibly -- I looked into this briefly over Christmas and the rule for the Cholesky factorisation didn't seem to be particularly optimal. I've got a better implementation lying around somewhere, which I'll turn into a PR tomorrow.

willtebbutt:
Okay. So I've had a dig around in this, and two things.
Firstly, this problem seems to be quite machine-dependent. On my mac, I'm seeing a ratio of around 6, which seems fine.

Secondly, the problem actually isn't the cholesky factorisation, it's low-level loops of the sort that we're looking to address in https://github.com/compintell/Mooncake.jl/issues/156 .

To see this, consider this flamegraph, which demonstrates how time is spent in AD:
gp_lml_ad_benchmark_flamegraph

It's roughly divided into two, the left half being the forwards-pass, and the right half the reverse-pass.
You'll see that there are basically three functions which dominate the forwards pass: `kernel_matrix`, `+`, and `copy_similar`.

`kernel_matrix`: looking at this, the time spent is roughly half in the rule for `syrk!` (a BLAS function), and half on overhead. We might be able to optimise the `syrk!` rule a bit, but I'm not sure how far we'll be able to push it. Either way, the big story here is that 50% of the time spent computing the kernel matrix is AD overhead.
`copy_similar`: this is an internal function in the LinearAlgebra standard library. The full signature is something like `Tuple{typeof(LinearAlgebra.copy_similar), Symmetric{Float64, Matrix{Float64}}, Float64}`. As you can see, this call is nearly 100% overhead. Moreover, this isn't really something that we want to be writing a rule for.
`+`: the full signature for this is `Tuple{typeof(+), Matrix{Float64}, Diagonal{Float64}}`. As with `copy_similar`, nearly 100% of the time is overhead. Again, this is a very particular combination of types, and we don't really want to have to write a rule for it.

Reverse-pass:
- `kernel_matrix`, `copy_similar`, and `+`: the performance of these functions mirror that of the forwards-pass. `kernel_matrix` is half `syrk!` and half overhead, while the other two are basically entirely overhead.
- `potrf!` (cholesky factorisation) does actually make an appearance here. It looks to me like it's occupying around 1/3rd of the reverse-pass. Some code I have suggests that this could be made maybe 50% faster.

My summary of this is that the performance issues here are maybe 80% to do with loop overhead, and 20% to do with linear algebra rules which could be optimised a bit.

Issue #249, opened 2024-09-11
Specific Rules for Performance


yebai:
Tapir implements rules at a very low level at the moment. This design choice helps reduce the burden of writing and maintaining rules, thanks to the small number of primitives that require manually written rules. The good news is that Tapir seems to have excellent performance even with most of its rules derived automatically by Tapir's autodiff transform (known as `DerivedRule`). 

However, this design choice occasionally (it seems most cases are due to one common root cause, i.e. vectorisation of specific loops) gets in the way of Julia's and LLVM's compiler optimisation passes. Our CI benchmarks (e.g., `sum`, `kron`, `kron_view_sum`, and `gp_lml`) reflect this. Meanwhile, Zygote, which imports most (or all) rules from `ChainRules`, performs well in these test cases. 

Given that `ChainRules` is pretty well tested, should we import a more significant number of its rules into Tapir by default? If so, what specific rules should we import?

willtebbutt:
I'm broadly in favour of selectively importing more rules. I'm working on the interface to import rules for ChainRules.jl at the minute, but the broad picture will remain as it currently is: importing rules from ChainRules.jl, or `ChainRules.rrule`s implemented in other packages is fine, but we'll need to restrict the argument types to ones that we are happy with. So, for example, we if CR has a rule `rrule(::typeof(*), ::AbstractMatrix, ::AbstractMatrix)`, there is no chance that we'll define a `Tapir.rrule!!` with signature
```julia
Tapir.rrule!!(::CoDual{typeof(*)}, ::CoDual{<:AbstractMatrix}, ::CoDual{<:AbstractMatrix})
```
Rather, we would define rules for concrete types, or finite unions of concrete types, and call out to `rrule` inside. For example, we might define a method of `rrule!!` with argument types
```julia
Tapir.rrule!!(::CoDual{typeof(*)}, ::CoDual{Matrix{P}}, ::CoDual{Matrix{P}}) where {P<:IEEEFloat}
```

> If so, what specific rules should we import?

I can think of two questions for any given signature:
1. is a rule for this sufficiently beneficial to warrant adding a rule for it?
2. is making use of the rule in CR the best way to implement it?

[...]

PR #458, opened 2025-02-04
Initial Work on Performance Patches

willtebbutt:
This PR starts to address #249 . In particular it
1. sets up conventions which ought to be followed for how these rules are to be incorporated into Mooncake, and
2. adds some initial rules.

This will make a couple of specific cases more performant, and make it possible for others to contribute rules that they would like to see in Mooncake.jl.

### DIFF
--- a/src/rrules/performance_patches.jl
+++ b/src/rrules/performance_patches.jl
@@ -1,7 +1,6 @@
 # All of the code here purely exists to work around current performance limitations of
 # Mooncake.jl. In order to prevent this from getting out of hand, there are several
 # conventions to which we adhere when writing these rules:
-#
 # 1. for each rule, a comment is added containing a link to the issue or issues that are
 #   believed to describe the deficiencies of Mooncake.jl which cause the rule to be needed.
 # 2. the number of concrete types for which the signature is valid is finite, and all are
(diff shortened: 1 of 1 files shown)


willtebbutt:
Something unfortunate is happening with some allocation checking functionality. It's not really related to the main bit of this PR, but I'll need to fix it before merging.

src/rrules/performance_patches.jl
```diff
@@ -0,0 +1,93 @@
...
+function rrule!!(
+    ::CoDual{typeof(Base._mapreduce)},
+    ::CoDual{typeof(abs2)},
+    ::CoDual{typeof(Base.add_sum)},
+    ::CoDual{Base.IndexLinear},
+    x::CoDual{<:Array{P}}
+) where {P<:IEEEFloat}
+    dx = x.dx
+    function sum_pb!!(dz::P)
+        x.dx .+= 2 .* dz .* x.x
```
yebai: Perhaps
```suggestion
        dx .+= 2 .* dz .* x.x
```

willtebbutt: I've tweaked the implementation since this -- good point though.

Issue #192, opened 2024-07-03
Simplify the README file now that we have docs


yebai:
Let's keep the README file brief and move stuff to docs where possible.

willtebbutt:
Closed via #463

PR #463, opened 2025-02-07
Tidy Up README

willtebbutt:
Moves most of the README to the docs. Also ensures that there's an example using DI in the docs so that it can't be missed. This addresses #192 .


willtebbutt:
The docs build, so this is good to go.

Issue #470, opened 2025-02-10
Rule for atomic_pointerset


willtebbutt:
https://github.com/TuringLang/DynamicPPL.jl/pull/799 re-highlighted a problem that causes Mooncake.jl to fail in specific situations in DynamicPPL (specific `VarInfo`s).  Happily, this just appears to be a problem with the `rrule!!` for `atomic_pointerset`, not anything more widespread. See https://github.com/TuringLang/DynamicPPL.jl/actions/runs/13218146414/job/36899908042?pr=799 . In particular, it looks like I basically hadn't considered the possibility that you might be setting a pointer-to-a-pointer when I wrote the rule, and therefore failed to consider that `zero_tangent` would not work. Moreover, this is implementation is simply incorrect for pointers (they have fdata), so it is a good thing that it errors loudly.

The fix would be to modify `atomic_pointerset` to account for the fact that the thing being pointed to might have fdata associated to it, and to not construct a `zero_tangent` if so. I should probably also check this case for `pointerset`, as I imagine that I made the same mistake there.

PR #389, opened 2024-11-24
Forward-mode AD

gdalle:
This is a very rough backbone of forward mode AD, based on #386 and the existing reverse mode implementation.

Will's edits (apologies for editing your thing @gdalle  -- I just want to make sure that the todo list is at the top of the PR):

Todo:
- [x] ~~make FunctionWrappers work correctly~~ not going to do this in this PR
- [x] add support for MistyClosures
- [x] add tests for Hessian vector products
- [x] define `is_primitive` separately for forwards and reverse pass.
- [x] do a complete pass to review design -- are there any high-level things we ought to modify?
- [x] improve DRY-ness of code, particularly in testing infrastructure in particular.
- [x] check GPU compatibility, make sure no major design issues prevent future GPU compatibility, and be explicit about what needs to be done in the future. 
- [x] what name should we use for `@from_rule`: `@from_chainrules` or `@from_chain_rule`, see comments below.
- [x] add support for `UpsilonNode`s and `PhiCNode`s.
- [x] get all tests passing
- [x] ~~bump to version 0.5~~ actually not needed

Once the above are complete, I'll request reviews.

PR #471, opened 2025-02-10
pointerset + friends with pointers-to-pointers

willtebbutt:
Will address #470 .
Todo:
- [x] add failing primitive tests for `pointerset` and `atomic_pointerset`
- [ ] add failing derived tests for `pointerset` and `atomic_pointerset` (to enable testing of numerical correctness)
- [ ] fix problem

### DIFF
--- a/src/rrules/builtins.jl
+++ b/src/rrules/builtins.jl
@@ -741,9 +741,29 @@ function generate_hand_written_rrule!!_test_cases(rng_ctor, ::Val{:builtins})
     dy = zero_tangent(y)
     dq = pointer(dy)
 
+    # Pointer to pointer.
+    c_1 = [5.0]
+    c_2 = [3.0, 4.0]
+    c = [pointer(c_1), pointer(c_2)]
+    c_ptr = pointer(c)
+
+    c_new_val = [6.0, 5.0, 4.0]
+    c_new_val_ptr = pointer(c_new_val)
+    cs = (c_1, c_2, c, c_new_val)
+
+    # Tangents of pointers to pointers.
+    dc_1 = copy(c_1)
+    dc_2 = copy(c_2)
+    dc = [pointer(dc_1), pointer(dc_2)]
+    dc_ptr = pointer(dc)
+    dc_new_val = randn(3)
+    dc_new_val_ptr = pointer(dc_new_val)
+    dcs = (dc_1, dc_2, dc, dc_new_val)
+
+
     # Slightly wider range for builtins whose performance is known not to be great.
     _range = (lb=1e-3, ub=200.0)
-    memory = Any[_x, _dx, _a, x, p, dx, dp, y, q, dy, dq]
+    memory = Any[_x, _dx, _a, x, p, dx, dp, y, q, dy, dq, cs..., dcs...]
 
     test_cases = Any[
 
@@ -775,6 +795,15 @@ function generate_hand_written_rrule!!_test_cases(rng_ctor, ::Val{:builtins})
             IntrinsicsWrappers.atomic_pointerset,
             CoDual(p, dp),
             1.0,
+            :monotonic,
+        ),
+        (
+            true,
+            :stability,
+            nothing,
+            IntrinsicsWrappers.atomic_pointerset,
+            CoDual(c_ptr, dc_ptr),
+            CoDual(c_new_val_ptr, dc_new_val_ptr),
             :monotonic,
         ),
         # atomic_pointerswap -- NEEDS IMPLEMENTING AND TESTING
@@ -881,6 +910,16 @@ function generate_hand_written_rrule!!_test_cases(rng_ctor, ::Val{:builtins})
             1,
         ),
         (true, :stability, nothing, IntrinsicsWrappers.pointerset, CoDual(q, dq), 1, 2, 1),
+        (
+            true,
+            :stability,
+            nothing,
+            IntrinsicsWrappers.pointerset,
+            CoDual(c_ptr, dc_ptr),
+            CoDual(c_new_val_ptr, dc_new_val_ptr),
+            1,
+            1,
+        ),
         # rem_float -- untested and unimplemented because seemingly unused on master
         # rem_float_fast -- untested and unimplemented because seemingly unused on master
         (false, :stability, nothing, IntrinsicsWrappers.rint_llvm, 5),


yebai:
@willtebbutt, can you address this issue to unblock DynamicPPL work?

willtebbutt:
Ah, sorry, I forgot / didn't realise that this was blocking. The issues here are mainly to do with our testing infrastructure's handling of pointers being a little tricky to work with. I need to sort that out before this can be safely merged -- I think I've figured out what needs doing while working on the forwards-mode PR, so I'll take a look this afternoon.

yebai:
I remember this PR will benefit from merging from #389 for better testing. Do we have a clear description of why and what remains to be done here?

yebai:
@willtebbutt, I'd appreciate it if you could push this to completion, as it is non-trivial for someone else to resolve.

AstitvaAggarwal:
tests should pass now, ill add a description of the changes soon.

AstitvaAggarwal:
**MWE** for rrule failure of `pointerset`, `atomic_pointerset` : 

```
using Mooncake
using Core.Intrinsics: pointerset, atomic_pointerset

f_pointerset = x -> begin
    c_1 = Ref(x)
    c_2 = Ref(x * 2.0)
    p = Ref(Base.unsafe_convert(Ptr{Float64}, c_1))
    GC.@preserve c_1 c_2 p begin
        pointerset(
            Base.unsafe_convert(Ptr{Ptr{Float64}}, p),
            Base.unsafe_convert(Ptr{Float64}, c_2),
            1,
            1,
        )
        unsafe_load(p[])
    end
end

f_atomic_pointerset = x -> begin
    c_1 = Ref(x)
    c_2 = Ref(x * 2.0)
    p = Ref(Base.unsafe_convert(Ptr{Float64}, c_1))
    GC.@preserve c_1 c_2 p begin
        atomic_pointerset(
            Base.unsafe_convert(Ptr{Ptr{Float64}}, p),
            Base.unsafe_convert(Ptr{Float64}, c_2),
            :monotonic,
        )
        unsafe_load(p[])
    end
end

value_and_gradient!!(build_rrule(f_pointerset, 3.0), f_pointerset, 3.0)
value_and_gradient!!(build_rrule(f_atomic_pointerset, 3.0), f_atomic_pointerset, 3.0)
```

- This is due to these parts of the rules for [`pointerset`](https://github.com/chalk-lab/Mooncake.jl/blob/9ab55b6f19d8636e8b001ba503201b7522371cd3/src/rules/builtins.jl#L641) and [`atomic_pointerset`](https://github.com/chalk-lab/Mooncake.jl/blob/9ab55b6f19d8636e8b001ba503201b7522371cd3/src/rules/builtins.jl#L232).

- This can be solved by passing `zero_tangent(primal(x), tangent(x))` so that we use [this](https://github.com/chalk-lab/Mooncake.jl/blob/9ab55b6f19d8636e8b001ba503201b7522371cd3/src/tangents/fwds_rvs_data.jl#L1031) dispatch of `zero_tangent`. Now when called for a primal `Ptr{T}`, we get the tangent itself a `Ptr{T}` back (`Ptr` only has `fdata`).
 Other primals are handled as they were being done so earlier, no change.

- Added a `Mooncake.frule!!` for `atomic_pointerset`, this should fix #876. And the rules for `unsafe_wrap`, which were missing.

- Testing the handwritten rules with `test_rrule_correctness/test_frule_correctness` is not possible because FD vs AD correctness tests dont handle numerical gradients for pointer primal types. Correctness tests for non pointer-to-pointer `pointerset` (only pointer to value) cases are already present and pass.

- However to test correctness, ive added some derived rule tests in `test/rules/builtins.jl`, these are present as a seperate manual testset.

AstitvaAggarwal:
Importantly note that the below MWE shows that there is another issue beyond `pointerset`.

```
using Mooncake: build_rrule, value_and_gradient!!
using Core.Intrinsics: pointerset, atomic_pointerset

f_pointerset = x -> begin
    c_1 = [x]
    c_2 = [x * 2.0]
    p = [pointer(c_1)] 
    pointerset(pointer(p), pointer(c_2), 1, 1)
    unsafe_load(p[1])
end

f_atomic_pointerset = x -> begin
    c_1 = [x]
    c_2 = [x * 2.0]
    p = [pointer(c_1)]
    atomic_pointerset(pointer(p), pointer(c_2), :monotonic)
    unsafe_load(p[1])
end

value_and_gradient!!(build_rrule(f_pointerset, 3.0), f_pointerset, 3.0)
value_and_gradient!!(build_rrule(f_atomic_pointerset, 3.0), f_atomic_pointerset, 3.0)
```

1. All handwritten rules that call `zero_tangent`  inside their definition will error out when any of the arguments contain or are pointers. (eg: `arrayset` [rule](https://github.com/chalk-lab/Mooncake.jl/blob/9ab55b6f19d8636e8b001ba503201b7522371cd3/src/rules/array_legacy.jl#L460))

2. Derived rules that use `@zero_adjoint` will fail when the function returns a pointer-containing value. Since `@zero_adjoint` calls `zero_fcodual` on the primal output internally (e.g. `Base.vect` in the [MWE](https://github.com/chalk-lab/Mooncake.jl/pull/471#issuecomment-3927054658)).

3. In general, direct usage of [zero_fcodual](https://github.com/chalk-lab/Mooncake.jl/blob/9ab55b6f19d8636e8b001ba503201b7522371cd3/src/tangents/codual.jl#L94)/[zero_codual](https://github.com/chalk-lab/Mooncake.jl/blob/9ab55b6f19d8636e8b001ba503201b7522371cd3/src/tangents/codual.jl#L33) will fail for pointers.

Finally, all the above is due to the fact that these call the single arg `zero_tangent` that always fails for pointers. The solution is to always call the 2 arg `zero_tangent`, which fallbacks to 1 arg `rdata` aware dispatch for `zero_tangent`. 
I have create an issue #999 to seperate it from the current Issue, PR.

src/rules/builtins.jl
```diff
@@ -210,12 +210,42 @@ end
 @inactive_intrinsic and_int
 @inactive_intrinsic ashr_int
 
+# unsafe_wrap() gives an array view for the memory pointed by p.
+# Tangent propagation happens through memory aliasing rather than explicit
+# computation in the pullback. Downstream rules write directly into 
+# the tangent memory pointed to by tangent_arr.
+@is_primitive MinimalCtx Tuple{typeof(unsafe_wrap),Type{Array},Ptr,Any}
```
sunxd3:
```suggestion
@is_primitive MinimalCtx Tuple{typeof(unsafe_wrap),Type{<:Array},Ptr,Any}
```
if this makes sense, also changes the rest?

AstitvaAggarwal: It will actually be :
```
@is_primitive MinimalCtx Tuple{typeof(unsafe_wrap),<:Type{<:Array},Ptr,Any}

```

Added this, thanks

sunxd3 (approved):
As far as I can tell, this is good

PR #474, opened 2025-02-11
Fix Stack Overflow

### LINKED ISSUE #473: Bug Report: Stack Overflow in `gradient` with `Tuple` Input
When using `Mooncake.jl` to compute the gradient of a function with a `Tuple` input, Julia encounters a stack overflow during type inference and crashes. However, when using a standard `Vector` instead of a `Tuple`, the function executes as expected.

## Reproduction

The issue can be reproduced with the following minimal example:

```julia
using Mooncake

f(x) = sum(cos, x)

x = ones(1_000) |> Tuple  # Using Tuple instead of Vector

gradient(f, backend, x)  # Causes stack overflow
```

```julia
Julia Version 1.11.3
Commit d63adeda50d (2025-01-21 19:42 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin24.0.0)
  CPU: 12 × Apple M3 Pro
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, apple-m2)
Threads: 6 default, 0 interactive, 3 GC (on 6 virtual cores)

julia> Pkg.installed()
Dict{String, VersionNumber} with 2 entries:
  "DifferentiationInterface" => v"0.6.40"
  "Mooncake"                 => v"0.4.84"
```

### Error Output (Truncated)

```
Internal error: during type inference of
gradient(typeof(Main.f), ADTypes.AutoMooncake{Nothing}, NTuple{1000, Float64})
Encountered stack overflow.
This might be caused by recursion over very long tuples or argument lists.
...
```

If `x` is defined as `x = ones(1_000)`, the function runs correctly.

## Additional Notes

- The issue also occurs with custom array types, such as those from `StaticArrays.jl`.
- The error message suggests an issue with recursion over long tuples.

## Question

Is this a known limitation, an expected behavior, or a bug in `Mooncake.jl`? If it is expected behavior, is there a recommended workaround for users who need to work with `Tuple` or `StaticArrays`?

willtebbutt:
Resolves #473 .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.84"
+version = "0.4.85"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/src/codual.jl
+++ b/src/codual.jl
@@ -38,7 +38,7 @@ function _codual_internal(::Type{P}, f::F, extractor::E) where {P,F,E}
 
     if P <: Tuple && !all(isconcretetype, (P.parameters...,))
         field_types = (P.parameters...,)
-        union_fields = _findall(Base.Fix2(isa, Union), 1, field_types)
+        union_fields = _findall(Base.Fix2(isa, Union), field_types)
         if length(union_fields) == 1 &&
             all(p -> p isa Union || isconcretetype(p), field_types)
             P_split = split_union_tuple_type(field_types)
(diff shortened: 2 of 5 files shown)


willtebbutt:
The modifications I've made here seem to have produced a really substantial compile-time performance regression. I'm going to figure this out before merging.

willtebbutt:
I _believe_ that I have resolve the completely ridiculous compile time problem, and will wait to see what this has done to average compile times. If they haven't inflated, I'll merge. If it looks like they've gone up noticeably, I'll try to reduce them by tweaking the implementations. I'm hoping that this is basically good to go now though.

willtebbutt:
I'm fairly sure that CI is lying to me about the coverage misses, so will merge provided that all other tests pass.

PR #477, opened 2025-02-11
Clarify public symbols

### LINKED ISSUE #475: What is public?
It would be useful to specify which symbols are public and which are not, by using the `@public` keyword introduced for Julia 1.11 (and a [hancrafted macro](https://discourse.julialang.org/t/is-compat-jl-worth-it-for-the-public-keyword/119041) for earlier versions).
We could also use the opportunity to remove private names from exports, and gather public docstrings on a single "API Reference" docs page. What do you think @willtebbutt?

gdalle:
First attempt to fix #475 

My rationale was:

- stuff starting with an underscore should be private
- stuff with generic names like `set_to_zero!!` should be unexported
- whatever isn't in the "Developer Documentation" should be public

But note that this is not easy to walk back because public symbols are part of the official API, so we can't remove them without a breaking release. Therefore, we need to get this right on the first try, and it would be safer to start with a reduced list of public symbols.

### DIFF
--- a/docs/src/interface.md
+++ b/docs/src/interface.md
@@ -1,0 +1,14 @@
+# Interface
+
+This is the public interface that day-to-day users of AD are expected to interact with if
+for some reason DifferentiationInterface.jl does not suffice.
+If you have not tried using Mooncake.jl via DifferentiationInterface.jl, please do so.
+See [Tutorial](@ref) for more info.
+
+```@docs; canonical=false
+Mooncake.Config
+Mooncake.value_and_gradient!!(::Mooncake.Cache, f::F, x::Vararg{Any, N}) where {F, N}
+Mooncake.value_and_pullback!!(::Mooncake.Cache, ȳ, f::F, x::Vararg{Any, N}) where {F, N}
+Mooncake.prepare_gradient_cache
+Mooncake.prepare_pullback_cache
+```
--- a/docs/src/interface.md
+++ b/docs/src/interface.md
@@ -5,7 +5,7 @@ for some reason DifferentiationInterface.jl does not suffice.
 If you have not tried using Mooncake.jl via DifferentiationInterface.jl, please do so.
 See [Tutorial](@ref) for more info.
 
-```@docs; canonical=false
+```@docs; canonical=true
 Mooncake.Config
 Mooncake.value_and_gradient!!(::Mooncake.Cache, f::F, x::Vararg{Any, N}) where {F, N}
 Mooncake.value_and_pullback!!(::Mooncake.Cache, ȳ, f::F, x::Vararg{Any, N}) where {F, N}
--- a/src/Mooncake.jl
+++ b/src/Mooncake.jl
@@ -142,14 +142,10 @@ include("config.jl")
 include("developer_tools.jl")
 
 # Public, not exported
-
 include("public.jl")
-@public CoDual
+@public value_and_pullback!!, prepare_pullback_cache
 
 # Public, exported
-
-export Config
-export value_and_pullback!!, prepare_pullback_cache
-export value_and_gradient!!, prepare_gradient_cache
+export Config, value_and_gradient!!, prepare_gradient_cache
 
 end
--- a/src/Mooncake.jl
+++ b/src/Mooncake.jl
@@ -158,7 +158,7 @@ export primal
 export Tangent, MutableTangent, PossiblyUninitTangent, NoTangent
 export tangent_type, tangent, zero_tangent
 export NoFData, fdata_type, fdata
-export NoRData,rdata_type, rdata
+export NoRData, rdata_type, rdata
 export rrule!!, build_rrule
 export value_and_pullback!!
 export value_and_gradient!!
--- a/src/Mooncake.jl
+++ b/src/Mooncake.jl
@@ -144,9 +144,9 @@ include("developer_tools.jl")
 
 # Public, not exported
 include("public.jl")
-@public value_and_pullback!!, prepare_pullback_cache
+@public Config, value_and_pullback!!, prepare_pullback_cache
 
 # Public, exported
-export Config, value_and_gradient!!, prepare_gradient_cache
+export value_and_gradient!!, prepare_gradient_cache
 
 end
--- a/src/Mooncake.jl
+++ b/src/Mooncake.jl
@@ -147,9 +147,9 @@ include("public.jl")
 
 @public set_to_zero!!, increment!!
 @public get_interpreter
-@public zero_adjoint
+@public zero_adjoint, var"@zero_adjoint"
 @public is_primitive, build_primitive_rrule
-export @mooncake_overlay, @zero_adjoint, @from_rrule  # how to make macros public with the @public workaround??
+@public var"@mooncake_overlay", var"@from_rrule"
 @public Config
 
 # Public, exported
--- a/src/Mooncake.jl
+++ b/src/Mooncake.jl
@@ -144,23 +144,11 @@ include("developer_tools.jl")
 # Public, not exported
 
 include("public.jl")
-
-@public set_to_zero!!, increment!!
-@public get_interpreter
-@public zero_adjoint, var"@zero_adjoint"
-@public is_primitive, build_primitive_rrule
-@public var"@mooncake_overlay", var"@from_rrule"
-@public Config
+@public CoDual
 
 # Public, exported
 
-export CoDual, codual_type, zero_codual
-export primal
-export Tangent, MutableTangent, PossiblyUninitTangent, NoTangent
-export tangent_type, tangent, zero_tangent
-export NoFData, fdata_type, fdata
-export NoRData, rdata_type, rdata
-export rrule!!, build_rrule
+export Config
 export value_and_pullback!!, prepare_pullback_cache
 export value_and_gradient!!, prepare_gradient_cache
 
--- a/src/Mooncake.jl
+++ b/src/Mooncake.jl
@@ -150,6 +150,7 @@ include("public.jl")
 @public zero_adjoint
 @public is_primitive, build_primitive_rrule
 export @mooncake_overlay, @zero_adjoint, @from_rrule  # how to make macros public with the @public workaround??
+@public Config
 
 # Public, exported
 
@@ -160,7 +161,7 @@ export tangent_type, tangent, zero_tangent
 export NoFData, fdata_type, fdata
 export NoRData, rdata_type, rdata
 export rrule!!, build_rrule
-export value_and_pullback!!
-export value_and_gradient!!
+export value_and_pullback!!, prepare_pullback_cache
+export value_and_gradient!!, prepare_gradient_cache
 
 end
--- a/src/Mooncake.jl
+++ b/src/Mooncake.jl
@@ -141,33 +141,27 @@ include("interface.jl")
 include("config.jl")
 include("developer_tools.jl")
 
-export primal,
-    tangent,
-    randn_tangent,
-    increment!!,
-    NoTangent,
-    Tangent,
-    MutableTangent,
-    PossiblyUninitTangent,
-    set_to_zero!!,
-    tangent_type,
-    zero_tangent,
-    _scale,
-    _add_to_primal,
-    _diff,
-    _dot,
-    zero_codual,
-    codual_type,
-    rrule!!,
-    build_rrule,
-    value_and_gradient!!,
-    value_and_pullback!!,
-    NoFData,
-    NoRData,
-    fdata_type,
-    rdata_type,
-    fdata,
-    rdata,
-    get_interpreter
+# Public, not exported
+
+include("public.jl")
+
+@public set_to_zero!!, increment!!
+@public get_interpreter
+@public mooncake_overlay
+@public zero_adjoint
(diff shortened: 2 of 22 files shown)


src/Mooncake.jl
```diff
@@ -141,33 +141,27 @@ include("interface.jl")
...
+@public get_interpreter
+@public zero_adjoint, var"@zero_adjoint"
+@public is_primitive, build_primitive_rrule
+@public var"@mooncake_overlay", var"@from_rrule"
+@public Config
+
+# Public, exported
+
+export CoDual, codual_type, zero_codual
+export primal
```
willtebbutt:
```suggestion
```

src/Mooncake.jl
```diff
@@ -141,33 +141,27 @@ include("interface.jl")
...
+export Tangent, MutableTangent, PossiblyUninitTangent, NoTangent
+export tangent_type, tangent, zero_tangent
```
willtebbutt:
```suggestion
export tangent_type
```

src/Mooncake.jl
```diff
@@ -141,33 +141,27 @@ include("interface.jl")
...
-    rdata_type,
-    fdata,
-    rdata,
-    get_interpreter
+# Public, not exported
+
+include("public.jl")
+
+@public set_to_zero!!, increment!!
+@public get_interpreter
```
willtebbutt:
```suggestion
@public set_to_zero!!, increment!!
```

src/Mooncake.jl
```diff
@@ -141,33 +141,27 @@ include("interface.jl")
...
+
+# Public, exported
+
+export CoDual, codual_type, zero_codual
+export primal
+export Tangent, MutableTangent, PossiblyUninitTangent, NoTangent
+export tangent_type, tangent, zero_tangent
+export NoFData, fdata_type, fdata
+export NoRData, rdata_type, rdata
+export rrule!!, build_rrule
```
willtebbutt:
```suggestion
export rrule!!
```

willtebbutt (commented):
I'm very much in favour making the public interface as minimal as possible for now, as you suggest.

Could we perhaps start by figuring out what it is that is needed to make DI work + handy tools for writing rules, and go from there? We might add to it later but, as you say, it's probably better to err on the side of caution here.

gdalle:
For DI to work, if I didn't screw things up in https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/728, we only need:

```julia
using Mooncake:
    CoDual,
    Config,
    prepare_gradient_cache,
    prepare_pullback_cache,
    tangent_type,
    value_and_gradient!!,
    value_and_pullback!!,
    zero_tangent
```

willtebbutt:
Yes, absolutely. I'll take a look at it now.

willtebbutt:
@gdalle I've refined the export list quite substantially. I've essentially stripped it back to the bare minimum. My proposal is to make public only those bits of functionality which AD users (as opposed to rule authors) need in order to work with Mooncake.

The one additional function that I believe DI requires is `tangent_type`. This is a tricky one, because while `tangent_type` is very stable for many possible inputs, it's the kind of thing which is constantly being extended to new types. e.g. people are free to add methods to it which change its behaviour for types that they own, for which it already provides a result. e.g. if you define a type `Foo`, `tangent_type` will return _something_ for it (and this thing will be "correct"), but you are at liberty to define another method of `tangent_type` for `Foo` which returns something arbitrary.

So, for types in Base / the standard libraries, this behaviour ought never to change. However, the behaviour can change in general, meaning that it's quite hard to make it part of the public interface. Do you have thoughts on this?

willtebbutt:
Additionally, since `Config` is now public and we reference it in the docs, in order to get the docs to build properly we're going to have to add the public API page in this PR, rather than in a later one, which is fine.

src/Mooncake.jl
```diff
@@ -142,33 +142,11 @@ include("interface.jl")
...
-    rdata_type,
-    fdata,
-    rdata,
-    get_interpreter
+# Public, not exported
+include("public.jl")
+@public value_and_pullback!!, prepare_pullback_cache
+
+# Public, exported
+export Config, value_and_gradient!!, prepare_gradient_cache
```
gdalle: do we want to export a name as generic as `Config`?

src/Mooncake.jl
```diff
@@ -142,33 +142,11 @@ include("interface.jl")
...
-    NoRData,
-    fdata_type,
-    rdata_type,
-    fdata,
-    rdata,
-    get_interpreter
+# Public, not exported
+include("public.jl")
+@public value_and_pullback!!, prepare_pullback_cache
+
```
gdalle, line 149:
```suggestion
@public tangent_type, zero_tangent
```
DI needs these two as well

gdalle:
> we're going to have to add the public API page in this PR

I suggest we make two pages: public API and private API.
I would recommend setting `canonical = true` on these pages and `canonical = false` everywhere else, because some of the docstrings are also displayed along the pages of explanations and this would lead to a "duplicate docstring" error.
Note that right now, `@autodocs` isn't aware of the `public` keyword, but I'm trying to change that https://github.com/JuliaDocs/Documenter.jl/pull/2629

src/Mooncake.jl
willtebbutt: Very good point. No, we almost certainly do not.

src/Mooncake.jl
willtebbutt: Why is it that DI needs `zero_tangent`? I know that it need `tangent_type` to check whether the user has provided the correct thing before calling `Mooncake.value_and_pullback!!`, but I'm not sure why we need `zero_tangent`.

willtebbutt:
Good idea -- I'll do that.

> I would recommend setting canonical = true on these pages and canonical = false everywhere else, because some 

TIL. I'll go through and do that systematically.

src/Mooncake.jl
gdalle: Maybe we can do without, you tell me

https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/58605dc2c91922c20cbefd24197bd6cd809a2383/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/onearg.jl#L16

https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/58605dc2c91922c20cbefd24197bd6cd809a2383/DifferentiationInterface/ext/DifferentiationInterfaceMooncakeExt/twoarg.jl#L24

gdalle: So can we remove this?

willtebbutt:
I've added a proper public-facing interface page. Now going to look at DI's use of Mooncake.

src/Mooncake.jl
willtebbutt: Hmmm my feeling is that we're currently doing this the wrong way in DI. I should really just make it possible to access the memory allocated by `prepare_pullback_cache` for the tangent of `y`, so that you don't have to call `zero_tangent`.

To be honest, I think I would like to make a breaking release of Mooncake in the next couple of days with refined definitions of `value_and_gradient!!` and `value_and_pullback!!`. Would you object to depending on internals for another couple of days?

gdalle: No worries! But maybe this shouldn't stop us from merging this PR?

willtebbutt: Ah, yes, sorry. My point was going to be that I would release this PR without making either of these functions public. Would that be acceptable?

gdalle: Definitely. Since Mooncake's interface was undefined until now, I've been relying on internals this whole time after all ^^

willtebbutt: Cool. In that case I'm going to mark this as resolved, and not include these functions in the public interface.

willtebbutt:
@gdalle what are your thoughts on where this is now? I think I'm basically happy with it.

docs/src/interface.md
```diff
@@ -0,0 +1,14 @@
+# Interface
+
+This is the public interface that day-to-day users of AD are expected to interact with if
+for some reason DifferentiationInterface.jl does not suffice.
+If you have not tried using Mooncake.jl via DifferentiationInterface.jl, please do so.
+See [Tutorial](@ref) for more info.
+
+```@docs; canonical=false
```
gdalle:
```suggestion
```@docs; canonical=true
```
There needs to be at least one canonical location for the docstring, otherwise I think references won't work properly.
This also applies to private symbols by the way. My suggestion to put `canonical = false` for docstrings inside text was based on the assumption that there would also be a "list of private docstrings" somewhere, mirroring the public API, with `canonical = true`. Does that make sense?

willtebbutt: Right you are. I had to apply a filter to handle the unexported bits of the public interface, but I think I've now got it working.

gdalle:
Regarding rule authors, why not make these utilities public? If you expect them to change, you can just tag a major release, but making rule writing private would essentially prevent anyone except you from adding new rules.

About tangent types, I think there's a difference between making a function public and guaranteeing that its behavior will never change. By definition, users are free to extend Base.sum on their types, but that doesn't mean the function itself shouldn't be public (on the contrary)

willtebbutt:
> About tangent types, I think there's a difference between making a function public and guaranteeing that its behavior will never change. By definition, users are free to extend Base.sum on their types, but that doesn't mean the function itself shouldn't be public (on the contrary)

I agree re `Base.sum`, but in my view there's a subtle difference. If you implement a method of `Base.sum` for `MyArray{Float64}`, it ought to return the same thing as if you wrote `sum(collect(x))` where `x isa MyArray{Float64}`, or whatever `Base.sum` would return when applied to `x` if you hadn't implemented a method for it directly (e.g. you had just implemented the AbstractArray interface, so a default method of `Base.sum` could be applied).

`tangent_type` has no such constraint. Supposing that, for a given type, the fallback method of `tangent_type(P)` works correctly, and returns some type `T`. The owner of `P` is at liberty to add a method of `tangent_type` which returns some other type, `V`.

I could specify that the interface for `tangent_type` is "it'll return a type, but it can be any type, and this type can change without even a non-breaking release Mooncake if a type-owner decides to add a method to it", or I can declare it private. The former is highly unsatisfactory, but so is the latter.

My medium-term ambition is to identify a large subset of types (e.g. all types in Base / LinearAlgebra, or maybe something a little bit more restrictive than than) and to assert that `tangent_type` won't change for them between non-breaking releases of Mooncake. For now, I think it's better not to make `tangent_type` public.

willtebbutt:
> Regarding rule authors, why not make these utilities public? If you expect them to change, you can just tag a major release, but making rule writing private would essentially prevent anyone except you from adding new rules.

You're right here, of course. I'm not opposed to making these public as part of version 0.5 of Mooncake, I just want to sort out the public interface one thing at a time to try and minimise the chances that we accidentally make something public which ought not to be (it's always a bit disruptive making breaking releases, and I'd rather minimise effort for the various people I'm aware of who depend on Mooncake via an extension / in their test suites.

willtebbutt:
Hmm. Maybe my previous comment regarding `tangent_type` is misguided -- I guess we could just assert that, while type-owners are free to choose what `tangent_type` returns for their types, making it differ from the default method is a breaking change (if you've already exposed the types to users)?

willtebbutt:
> I think there's a difference between making a function public and guaranteeing that its behavior will never change.

Separately from the previous discussion, I'd just like to highlight that my view on public functionality is exactly the opposite of this -- to my mind, declaring that something is public is _exactly_ a promise that its behaviour won't change unless also accompanied by a breaking release. You might extend the functionality (make it support new types etc), or change implementation details without modifying behaviour (up to acceptable things like small numerical differences), but not see anything change.

Is this not your perspective?

willtebbutt:
@gdalle I've gone on a bit of a rant above. What are your thoughts on merging this?

docs/src/interface.md
gdalle: Yeah, I meant to show you, I had to do the same in DI

https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/58605dc2c91922c20cbefd24197bd6cd809a2383/DifferentiationInterface/docs/src/api.md?plain=1#L134-L142

gdalle:
> Hmm. Maybe my previous comment regarding tangent_type is misguided -- I guess we could just assert that, while type-owners are free to choose what tangent_type returns for their types, making it differ from the default method is a breaking change (if you've already exposed the types to users)?

I think that is reasonable. But I also think we can keep it private for now, even though DI depends on it, and renegociate at the next breaking release. It is still unclear to me why people would want to redefine `tangent_type`. The one use case I see is for e.g. Flux gradients to return a `Layer` too, but I assume that would require overloading some more stuff (like `increment!!`)?

gdalle (commented):
lgtm

gdalle:
@willtebbutt can we merge this one?

willtebbutt:
Sorry, yes. I've been procrastinating merging this because it scares me slightly, but we should just merge it and see what happens. I've resolved the merge conflict + bumped the patch version, and will merge once CI passes.

willtebbutt:
Will actually merge when 0.4.97 reaches main.

yebai:
@willtebbutt, let's get this merged.

Issue #397, opened 2024-11-28
Issue with terminators in `normalise!`


gdalle:
@vchuravy spotted an issue with Mooncake's custom normalisation. Here's an MWE:

```julia
const CC = Core.Compiler
using Mooncake

function func(x)
    y = sin(x)
    if x[1] > 0
        z = cos(y)
    else
        z = sin(y)
    end
    return z
end

interp = Mooncake.get_interpreter()
ir = Base.code_ircode(func, (Float64,))[1][1]
sig = Tuple{typeof(func),Float64}
primal_ir, _ = Mooncake.lookup_ir(interp, sig)
_, spnames = Mooncake.is_vararg_and_sparam_names(sig)
primal_ir = Mooncake.normalise!(primal_ir, spnames)
CC.verify_ir(primal_ir)
```

Here's the error

```julia
julia> CC.verify_ir(primal_ir)
Block 1 successors (Array{Int64, 1}(dims=(2,), mem=Memory{Int64}(2, 0x739f1089a780)[3, 2])), does not match fall-through terminator %1 (Expr(:invoke, sin(Float64) from sin(T) where {T<:Union{Float32, Float64}}, Main.sin, Core.Argument(n=2)))::Float64
5  1 ─ %1  = invoke Main.sin(_2::Float64)::Float64
   2 ─       nothing::Nothing
   3 ┄       goto #4
6  4 ─ %4  = (Mooncake.IntrinsicsWrappers.lt_float)(0.0, _2)::Bool
   │   %5  = (Mooncake.IntrinsicsWrappers.or_int)(%4, false)::Bool
   └──       goto #6 if not %5
7  5 ─ %7  = invoke Main.cos(%1::Float64)::Float64
   └──       goto #7
9  6 ─ %9  = invoke Main.sin(%1::Float64)::Float64
11 7 ┄ %10 = φ (#5 => %7, #6 => %9)::Float64
   └──       return %10
   ERROR: 
Stacktrace:
 [1] error(s::String)
   @ Core.Compiler ./error.jl:35
 [2] verify_ir(ir::Core.Compiler.IRCode, print::Bool, allow_frontend_forms::Bool, 𝕃ₒ::Core.Compiler.PartialsLattice{…})
   @ Core.Compiler ./compiler/ssair/verify.jl:233
 [3] verify_ir(ir::Core.Compiler.IRCode)
   @ Core.Compiler ./compiler/ssair/verify.jl:98
 [4] top-level scope
[...]

PR #478, opened 2025-02-12
Fix Terminator Issue

willtebbutt:
Per discussions in #389 and #397 , this changes the normalisation pass to ensure that the CFG doesn't change. Another approach would have been to continue with how things are currently done and to properly reflect the changes in the CFG, but that would have been more work, and not obviously improved things further.


willtebbutt:
It looks like this might have caused a small performance regression. I'm going to try a couple of things locally to resolve.

willtebbutt:
Coverage failure is due to a chunk of code which is should only throw if something goes badly wrong -- everything else appears to be covered. I'm going to try to add a unit test for `kill_edge!`, but then I'll merge + tag.

willtebbutt:
Unit test added. Will merge when CI (other than coverage) passes.

Issue #484, opened 2025-02-13
Cstring is a primitive type. Implement a method of `tangent_type` for it.


penelopeysm:
I was trying to get to the bottom of the DPPL .~ error (https://github.com/compintell/Mooncake.jl/issues/319#issuecomment-2656053215), then ran into this which gives a separate error message:

## mwe

```julia
using DynamicPPL: DynamicPPL, @model, LogDensityFunction
using Distributions: Normal
using DifferentiationInterface
import Mooncake

@model function m()
    x ~ Normal()
    # ^ Need this so that the model has at least one parameter.
    1.5 ~ Normal()
end

function f(x)
    ldf = LogDensityFunction(m())
    vi = DynamicPPL.unflatten(ldf.varinfo, x)
    return DynamicPPL.getlogp(last(DynamicPPL.evaluate!!(ldf.model, vi)))
end

prep = prepare_gradient(f, AutoMooncake(; config=nothing), [1.0])
```

## versions

```julia
(ppl) pkg> st
Status `~/ppl/Project.toml`
  [a0c0ee7d] DifferentiationInterface v0.6.40
  [31c24e10] Distributions v0.25.117
  [366bfd00] DynamicPPL v0.34.2
  [da2b9cff] Mooncake v0.4.86

julia> versioninfo()
Julia Version 1.11.3
Commit d63adeda50d (2025-01-21 19:42 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin24.0.0)
  CPU: 10 × Apple M1 Pro
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, apple-m1)
Threads: 1 default, 0 interactive, 1 GC (on 8 virtual cores)
```

## stacktrace

<details><summary>expand</summary>

```
julia> prep = prepare_gradient(f, adtype, x)
ERROR: MooncakeRuleCompilationError: an error occured while Mooncake was compiling a rule to differentiate something. If the `caused by` error message below does not make it clear to you how the problem can be fixed, please open an issue at github.com/compintell/Mooncake.jl describing your problem.
To replicate this error run the following:

Mooncake.build_rrule(Mooncake.MooncakeInterpreter(), Tuple{typeof(issorted), Vector{Symbol}, Base.Order.ForwardOrdering}; debug_mode=false)

Note that you may need to `using` some additional packages if not all of the names printed in the above signature are available currently in your environment.

[...]

Issue #319, opened 2024-10-29
AssertionError: useidx !== nothing


willtebbutt:
There is a known issue with `Core.Compiler.populate_use_def_map!`, as discussed here:  https://github.com/JuliaLang/julia/issues/56193

This can result in problems like that encountered by @penelopeysm in this PR: https://github.com/TuringLang/Bijectors.jl/pull/338

For posterity, here is the kind of error that you see if you encounter it:
```julia
  Got exception outside of a @test
  AssertionError: useidx !== nothing
  Stacktrace:
    [1] kill_def_use!(tpdum::Core.Compiler.TwoPhaseDefUseMap, def::Int64, use::Int64)
      @ Core.Compiler ./compiler/inferencestate.jl:147
    [2] kill_def_use!
      @ ./compiler/inferencestate.jl:158 [inlined]
    [3] reprocess_instruction!(interp::Core.Compiler.NativeInterpreter, inst::Core.Compiler.Instruction, idx::Int64, bb::Nothing, irsv::Core.Compiler.IRInterpretationState)
      @ Core.Compiler ./compiler/ssair/irinterp.jl:119
    [4] _ir_abstract_constant_propagation(interp::Core.Compiler.NativeInterpreter, irsv::Core.Compiler.IRInterpretationState; externally_refined::Nothing)
      @ Core.Compiler ./compiler/ssair/irinterp.jl:391
    [5] _ir_abstract_constant_propagation
      @ ./compiler/ssair/irinterp.jl:280 [inlined]
    [6] __infer_ir!
      @ ~/.julia/packages/Mooncake/LKJK9/src/interpreter/ir_utils.jl:117 [inlined]
    [7] optimise_ir!(ir::Core.Compiler.IRCode; show_ir::Bool, do_inline::Bool)
      @ Mooncake ~/.julia/packages/Mooncake/LKJK9/src/interpreter/ir_utils.jl:150
    [8] optimise_ir!
      @ ~/.julia/packages/Mooncake/LKJK9/src/interpreter/ir_utils.jl:139 [inlined]
    [9] build_rrule(interp::Mooncake.MooncakeInterpreter{Mooncake.DefaultCtx}, sig_or_mi::Type; debug_mode::Bool, silence_debug_messages::Bool)
      @ Mooncake ~/.julia/packages/Mooncake/LKJK9/src/interpreter/s2s_reverse_mode_ad.jl:925
   [10] build_rrule
[...]

PR #485, opened 2025-02-13
Cstring and Cwstring

willtebbutt:
Just adds `NoTangent` tangent types for `Cstring` and `Cwstring`. I've confirmed locally that this ensures that @penelopeysm 's example in #484 successfully errors on the bug from #319 instead of asking you to define `tangent_type` for `Cstring`!


penelopeysm:
Aha, okay, so I _did_ actually minimise it ;)

PR #490, opened 2025-02-15
Some House Keeping

willtebbutt:
There are a few small things which could do with being tidied up. This PR does that.


willtebbutt:
The only uncovered lines are those which only get hit if you don't load AllocCheck / JET before calling `check_allocs` etc.

PR #491, opened 2025-02-15
Performance of binary method of tangent_type

willtebbutt:
The binary method of `tangent_type` could sometimes fail to constant-fold properly for tricky examples, in the same way that the unary method previously could.

The solution is to
1. use Mooncake's new `@foldable` macro to just declare that it's foldable, and
2. insert a `@test is_foldable(...)` in the test suite for `test_fwds_rvs_data` to ensure that it always holds.


willtebbutt:
As before, coverage failures appear to be spurious.

PR #502, opened 2025-02-28
Check on 1.11.4

willtebbutt:
v1.11.4 is now available. In principle it ought not to cause us any trouble, but it's best to be sure by running CI. This PR should definitely not be merged.


willtebbutt:
This isn't testing the thing I wanted to test. Will have to check locally.

PR #511, opened 2025-03-04
GC between benchmarks

willtebbutt:
I think it's possible that the variance we often see in benchmarks is because we're not properly GC-ing between benchmarks. This PR hopefully addresses this problem, although it's obviously rather hard to know whether this has worked from a single run.


willtebbutt:
I honestly have no idea whether this is better or worse than before. @yebai what are your thoughts? I'm tempted to merge as-is, and see how things work for a while.

yebai:
Sounds good.

PR #512, opened 2025-03-04
Put BBCode in module

willtebbutt:
For the sake of hygiene, put BBCode in its own module.

This requires a little bit of formatting / shuffling of code around, but CI needs to run to check that I've refactored everything correctly.

### DIFF
--- a/src/interpreter/bbcode.jl
+++ b/src/interpreter/bbcode.jl
@@ -32,7 +32,6 @@ export ID,
     collect_stmts,
     compute_all_predecessors,
     BBCode,
-    inc_args,
     remove_unreachable_blocks!,
     characterise_used_ids,
     characterise_unique_predecessor_blocks,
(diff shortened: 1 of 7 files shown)


src/interpreter/bbcode.jl
```diff
@@ -1,7 +1,47 @@
-# See the docstring for `BBCode` for some context on this file.
+"""
+    module BBCode
```
yebai: Gemini [suggests](https://g.co/gemini/share/261dc2571263) 

```suggestion
    module BasicBlockIR
```

willtebbutt: I'll tweak the module name to `BasicBlockCode`, since that's what the acronym `BBCode` stands for. We can rename to e.g. `NamedIRCode` or `NIRCode` at a later point in time.

src/interpreter/bbcode.jl
```diff
@@ -1,7 +1,72 @@
...
```
yebai:
```suggestion
    module BasicBlockCode
```

willtebbutt:
Will merge when CI passes.

penelopeysm:
I know that BBCode is something different, but I'm always reminded of https://en.wikipedia.org/wiki/BBCode!

willtebbutt:
TIL!

Issue #466, opened 2025-02-09
Mooncake less efficient than Zygote or Enzyme on Flux layers


m-laprise:
(This is a follow-up to a slack thread, cc @willtebbutt & @gdalle )

I have been comparing performance of different autodiff backends for training Flux models. I get one order of magnitude worse performance from Mooncake compared to Enzyme and even Zygote in most cases. This may be due to the Fluxperimental Moonduo implementation ([https://github.com/FluxML/Fluxperimental.jl/blob/master/ext/FluxMooncakeExt.jl](https://github.com/FluxML/Fluxperimental.jl/blob/master/ext/FluxMooncakeExt.jl)) rather than something with Mooncake? here is a MWE if useful.

Simple case of two standard feed forward layers. Benchmarks below with Flux v0.15.2, Zygote v0.6.75 (constrained from updating further by Fluxperimental), Enzyme v0.13.30, Mooncake v0.4.83, on Julia 1.10.5.

```
using Flux
using BenchmarkTools

# Create random inputs and targets
const MINIBATCHSIZE = 64
X = rand(Float32, 100, MINIBATCHSIZE)
Y = rand(Float32, 20, MINIBATCHSIZE)

# Create trivial Flux NN and loss 
model = Chain(Dense(100, 50, relu), 
              Dense(50, 20))
myloss(m, x, y) = Flux.mse(m(x), y)

# Compare time to first gradient (restarting the session for each example):

using Zygote
@btime loss, grads = Flux.withgradient($myloss, $model, $X, $Y)
# 81.875 μs (87 allocations: 126.46 KiB)

using Enzyme
@btime loss, grads = Flux.withgradient($myloss, $Duplicated(model), $X, $Y)
# 82.875 μs (129 allocations: 84.27 KiB)

using Fluxperimental, Mooncake
@btime loss, grads = Flux.withgradient($myloss, $Moonduo(model), $Moonduo(X), $Moonduo(Y))
# 837.625 μs (16045 allocations: 1.89 MiB)

fclosure(m) = myloss(m, X, Y)
@btime loss, grads = Flux.withgradient($fclosure, $Moonduo(model))
# 919.000 μs (16048 allocations: 1.89 MiB)
```

willtebbutt:
Thank you very much for this example, I really appreciate you following up on the slack thread! I'll have a look to see if I can figure out what's going on.

willtebbutt:
[...]

PR #514, opened 2025-03-08
Rule for Flux.MSE

AstitvaAggarwal:
Add a Rule for Flux.Losses.MSE. Addressing (partially) issue  #466.

On Julia 1.10.8 performance AD with Zygote, Enzyme: 
image

current Mooncake version:
image
image
image

With the new rule: 
image
image
image


AstitvaAggarwal:
All the tests should pass now. Note that the correctness tests have been flagged as true (interface_only). I verified the results manually with AD from Zygote and Enzyme and it does give correct gradients.

Project.toml
```diff
@@ -10,6 +10,7 @@ ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 DiffRules = "b552c78f-8df3-52c6-915a-8e097449b14b"
 DiffTests = "de460e47-3fe3-5279-bb4a-814414816d5d"
 ExprTools = "e2ba6199-217a-4e67-a87a-7c52f15ade04"
+Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
```
willtebbutt: Could we please make this a package extension, rather than making Flux a dep of Mooncake?

src/rrules/linear_algebra.jl
```diff
@@ -33,3 +33,55 @@ end
...
+}
+
+function rrule!!(
+    ::CoDual{typeof(Losses.mse)},
+    X::CoDual{<:AbstractArray{P}},
+    Y::CoDual{<:AbstractArray{P}},
+) where {P<:IEEEFloat}
+    N = P(2) / P(length(X.x)) .* (X.x - Y.x)
+
+    function FluxMSE_pullback(dloss::P)
```
willtebbutt:
```suggestion
    function flux_mse_pullback(dloss::P)
```
style

src/rrules/linear_algebra.jl
```diff
@@ -33,3 +33,55 @@ end
...
+        # adjoints got by VJP reverse pass equations.
+        temp = N .* dloss
+        @. X.dx += temp
+        @. Y.dx -= temp
+        return NoRData(), NoRData(), NoRData()
+    end
+
+    # in forward pass it returns codual float64, hence coduals dx is NoFData().
+    return CoDual(Losses.mse(X.x, Y.x), NoFData()), FluxMSE_pullback
```
willtebbutt:
```suggestion
    return zero_fcodual(Losses.mse(X.x, Y.x)), flux_mse_pullback
```

src/rrules/linear_algebra.jl
```diff
@@ -33,3 +33,55 @@ end
...
+@is_primitive DefaultCtx Tuple{
+    typeof(Losses.mse),AbstractArray{<:IEEEFloat},AbstractArray{<:IEEEFloat}
+}
+
+function rrule!!(
+    ::CoDual{typeof(Losses.mse)},
+    X::CoDual{<:AbstractArray{P}},
+    Y::CoDual{<:AbstractArray{P}},
+) where {P<:IEEEFloat}
+    N = P(2) / P(length(X.x)) .* (X.x - Y.x)
```
willtebbutt:
```suggestion
    N = P(2) / P(length(X.x)) .* (X.x .- Y.x)
```
Two things here:
1. could we please use a variable name other than `N`? It's typically used for counts and lengths of things
2. per the suggestion, I think we probably want to fuse the broadcasting here to avoid extra allocations

src/rrules/linear_algebra.jl
```diff
@@ -33,3 +33,55 @@ end
...
 end
+
+@is_primitive DefaultCtx Tuple{
+    typeof(Losses.mse),AbstractArray{<:IEEEFloat},AbstractArray{<:IEEEFloat}
+}
+
+function rrule!!(
+    ::CoDual{typeof(Losses.mse)},
+    X::CoDual{<:AbstractArray{P}},
+    Y::CoDual{<:AbstractArray{P}},
```
willtebbutt:
```suggestion
    X::CoDual{<:Array{P}},
    Y::CoDual{<:Array{P}},
```
the convention in Mooncake is to provide very strict typing for rules. I'm realising now that I don't have the rationale for this written down anywhere -- in short it's because if you write rules for abstract types, they'll usually be subtly wrong for most input types.

src/rrules/linear_algebra.jl
```diff
@@ -33,3 +33,55 @@ end
...
+            )
+        end,
+        map([Float16, Float32, Float64]) do P
+            return (
+                true,
+                :none,
+                nothing,
+                Losses.mse,
+                P.([1e-3, 1e-4, 0]),
+                P.([1e-1, 1e-2, 1e-3]),
```
willtebbutt: Could you explain why such small numbers are being used here? Rather than e.g. just sampling them via `randn`?

willtebbutt (changes requested):
Thanks for this PR @AstitvaAggarwal . I think the substance of the PR is largely good, there's just a few technical / style issues that we need to resolve. Apologies in advance if this comes across as being pedantic, I just need to make sure that any code that we merge is consistent in style with the rest of the package.

src/rrules/linear_algebra.jl
AstitvaAggarwal: no reason in particular, I just wanted to see how FDM and the cache etc in _dot_internal() worked for super small differences in testing. I did notice sometimes that these pass performance tests though, but i have kept them interface_only.

src/rrules/linear_algebra.jl
AstitvaAggarwal: oh yeah, missed out on that!

src/rrules/linear_algebra.jl
willtebbutt: Ahh okay, cool. It probably makes sense to make these numbers a more standard scale, otherwise someone reading through might think there's some significance to their scale.

test/ext/flux/flux.jl
```diff
@@ -0,0 +1,40 @@
...
+    @testset "$f, $(typeof(fargs))" for (
+        interface_only, perf_flag, is_primitive, f, fargs...
+    ) in vcat(
+        # Testing specific floating point precisions as FDM is unstable for certain tests.
+        # Larger differences can compare well with AD for for higher precisions as FDM is relatively stable.
+        # for smaller differences we start getting better FDM gradients for smaller floating points.
+
+        map([Float64]) do P
+            return (
+                false, :none, false, Losses.mse, P.([1e1, 1e2, 1e3]), P.([1e3, 1e4, 0])
```
willtebbutt:
```suggestion
                false, :none, true, Losses.mse, randn(StableRNG(1), P, 3), randn(StableRNG(2), P, 3))
```
Since this is now a primitive, we should tweak the corresponding flag to ensure that the `test_rule` checks to see whether it's hitting the primitive. The same applies to the other test cases in this file. Also, I'd really recommend just generating random numbers here, but using a seeded `StableRNG`, in order to be consistent with the other tests in this package.

The same line of reasoning applies to the other test cases in this file.

Incidentally, it looks like this extension isn't loading properly (see the Flux extension). I suspect that CI will pick up on this once you change the `is_primitive` flag to `true`.

AstitvaAggarwal: i think all the test cases would be in just one map() then (randn generates smaller number), but it wouldnt show the Float precisions failing at certain cases. Would that be okay?

willtebbutt: Yeah, that's completely fine -- a single `map` makes sense to me.

ext/MooncakeFluxExt.jl
```diff
@@ -0,0 +1,29 @@
+module MooncakeFluxExt
+
+using Mooncake, Flux
+using Base: IEEEFloat
+import Mooncake:
+    DefaultCtx, rrule!!, @is_primitive, @mooncake_overlay, CoDual, zero_fcodual, NoRData
```
willtebbutt:
```suggestion
    DefaultCtx, rrule!!, @is_primitive, CoDual, zero_fcodual, NoRData
```
Unless I'm mistaken, mooncake_overlays aren't being used here.

AstitvaAggarwal: oh yeahh thanks, forgot to remove it.

test/ext/flux/flux.jl
willtebbutt: To be honest, I would just test on `Float32` and `Float64`.

AstitvaAggarwal: perfect ill do it. Thanks.

AstitvaAggarwal:
@willtebbutt good to merge?

ext/MooncakeFluxExt.jl
```diff
@@ -0,0 +1,28 @@
...
+
+    function flux_mse_pullback(dloss::P)
+        # adjoints got by VJP reverse pass equations.
+        temp = norm_factor .* dloss
+        @. X.dx += temp
+        @. Y.dx -= temp
+        return NoRData(), NoRData(), NoRData()
+    end
+
+    # in forward pass it returns codual float64, hence coduals dx is NoFData().
```
willtebbutt:
```suggestion
```
I think this comment can go, since the line below it no longer contains a `NoFData` directly.

test/ext/flux/flux.jl
```diff
@@ -0,0 +1,29 @@
...
+using Mooncake.TestUtils: test_rule
+
+@testset "flux" begin
+    @testset "$f, $(typeof(fargs))" for (
+        interface_only, perf_flag, is_primitive, f, fargs...
+    ) in vcat(
+        # Testing specific floating point precisions as FDM is unstable for certain tests.
+        # Larger differences can compare well with AD only for higher precisions as FDM is relatively stable.
+        # for smaller differences we start getting better FDM gradients for smaller floating points as well.
+
```
willtebbutt:
```suggestion
```
Can this comment also go now that we're not changing how we choose the input vector depending on the precision?

test/ext/flux/flux.jl
```diff
@@ -0,0 +1,29 @@
...
+        map([Float32, Float64]) do P
+            return (
+                false,
+                :none,
+                false,
```
willtebbutt:
```suggestion
                true,
```
I think maybe this still needs to be changed to `true`?

test/ext/flux/flux.jl
```diff
@@ -0,0 +1,29 @@
...
+        interface_only, perf_flag, is_primitive, f, fargs...
+    ) in vcat(
+        # Testing specific floating point precisions as FDM is unstable for certain tests.
+        # Larger differences can compare well with AD only for higher precisions as FDM is relatively stable.
+        # for smaller differences we start getting better FDM gradients for smaller floating points as well.
+
+        map([Float32, Float64]) do P
+            return (
+                false,
+                :none,
```
willtebbutt, line 15:
```suggestion
                :stability,
```
Could we please change this flag so that we always check that the code is type-stable? (it definitely looks like it ought to be, but best to be sure).

test/rrules/linear_algebra.jl
```diff
@@ -1,3 +1,3 @@
 @testset "linear_algebra" begin
     TestUtils.run_rrule!!_test_cases(StableRNG, Val(:linear_algebra))
-end
+end
```
willtebbutt:
```suggestion
end
```
Do you know what has changed here? Would be good to get this out of the diff.

ext/MooncakeFluxExt.jl
```diff
@@ -0,0 +1,28 @@
...
+    return zero_fcodual(Flux.Losses.mse(X.x, Y.x)), flux_mse_pullback
+end
+
+end
```
willtebbutt:
```suggestion
end

```
I think we maybe need a newline at the end of this file to make github happy.

willtebbutt: Ah, you addressed this as I was reviewing haha

test/rrules/linear_algebra.jl
willtebbutt: Same here. Thanks for resolving.

Project.toml
```diff
@@ -49,6 +51,7 @@ ChainRulesCore = "1"
 DiffRules = "1"
 DiffTests = "0.1"
 ExprTools = "0.1"
+Flux = "0.15.2"
```
willtebbutt: I just dev-ed Flux locally, and it looks like there's a new version, 0.16.3 available. Is there a reason we're pinning ourselves to an old patch version of Flux here, or could we update to the latest version?

ext/MooncakeFluxExt.jl
```diff
@@ -0,0 +1,28 @@
...
+function rrule!!(
+    ::CoDual{typeof(Flux.Losses.mse)}, X::CoDual{<:Array{P}}, Y::CoDual{<:Array{P}}
+) where {P<:IEEEFloat}
+    norm_factor = P(2) / P(length(X.x)) .* (X.x .- Y.x)
+
+    function flux_mse_pullback(dloss::P)
+        # adjoints got by VJP reverse pass equations.
+        temp = norm_factor .* dloss
+        @. X.dx += temp
+        @. Y.dx -= temp
```
willtebbutt:
```suggestion
        tmp = dloss * P(2 / length(X.x))
        @inbounds for n in eachindex(X.x)
            d = X.x[n] - Y.x[n]
            X.dx[n] += tmp * d
            Y.dx[n] -= tmp * d
        end
```
I did some benchmarking locally, and I think this might be the best way to implement the reverse-pass. It avoids creating a new `Array{P}`, and still only loads the memory from each of the arrays involved into memory a single time.

Could you let me know whether you are also seeing that this gives better performance than the current implementation?

test/rrules/linear_algebra.jl
AstitvaAggarwal: i ran a format(".")

Project.toml
AstitvaAggarwal: we can use the latest version, sorry i missed this out.

willtebbutt: No problem!

ext/MooncakeFluxExt.jl
AstitvaAggarwal: yup you are correct. ive changed it now, thanks!

ext/MooncakeFluxExt.jl
```diff
@@ -0,0 +1,31 @@
...
+import Mooncake: DefaultCtx, rrule!!, @is_primitive, CoDual, zero_fcodual, NoRData
+
+@is_primitive DefaultCtx Tuple{
+    typeof(Flux.Losses.mse),Array{P},Array{P}
+} where {P<:IEEEFloat}
+
+function rrule!!(
+    ::CoDual{typeof(Flux.Losses.mse)}, X::CoDual{<:Array{P}}, Y::CoDual{<:Array{P}}
+) where {P<:IEEEFloat}
+    norm_factor = P(2) / P(length(X.x)) .* (X.x .- Y.x)
```
willtebbutt:
```suggestion
```

AstitvaAggarwal:
Tests fail as it's not type stable. Shall i set the flag back to false?

AstitvaAggarwal:
The function is not type stable and so the tests fail. I will set the flag back to :none.

willtebbutt:
> The function is not type stable and so the tests fail. I will set the flag back to :none.

Are you sure? It just looks to me like JET needs to be added to the Project.toml associated to the Flux extension environment.

AstitvaAggarwal:
I actually tested it out locally and it dint pass as well.

AstitvaAggarwal:
turns out the aggregating inside Flux.Losses.mse is Flux.mean which creates a runtime dispatch.

willtebbutt (approved):
Oh, wow, okay, thanks for looking into this. I guess we have to set the flag to `false`.

Once you've done that + CI is passing, I think we're ready to merge!

AstitvaAggarwal:
surprisingly it dosent pass for the extension only on lts. It passes on 1.

AstitvaAggarwal:
nvm ill just remove it.

ext/MooncakeFluxExt.jl
```diff
@@ -0,0 +1,29 @@
...
+
+using Mooncake, Flux
+using Base: IEEEFloat
+import Mooncake: DefaultCtx, rrule!!, @is_primitive, CoDual, zero_fcodual, NoRData
+
+@is_primitive DefaultCtx Tuple{
+    typeof(Flux.Losses.mse),Array{P},Array{P}
+} where {P<:IEEEFloat}
+
+function rrule!!(
```
yebai, line 12:
```suggestion

# This is a performance-specific rule motivated by https://github.com/compintell/Mooncake.jl/issues/466
function rrule!!(
```

willtebbutt:
Okay. It looks like something is going on wrong artifact download in buildkite. Since this PR plainly doesn't involve any GPU stuff, I'm going to assume this is an issue out of our control, and merge this PR.

willtebbutt:
Thanks for the contribution @AstitvaAggarwal !

Issue #513, opened 2025-03-07
Access to undefined reference


mhauru:
Repro:
```julia
module MWE

using DynamicPPL
using Distributions
using Mooncake
using ADTypes
using LogDensityProblems

@model function lda(K, d, w)
    V = length(unique(w))
    D = length(unique(d))
    N = length(d)
    @assert length(w) == N

    ϕ = Vector{Vector{Real}}(undef, K)
    for i in 1:K
        ϕ[i] ~ Dirichlet(ones(V) / V)
    end

    θ = Vector{Vector{Real}}(undef, D)
    for i in 1:D
        θ[i] ~ Dirichlet(ones(K) / K)
    end

    z = zeros(Int, N)

    for i in 1:N
        z[i] ~ Categorical(θ[d[i]])
        w[i] ~ Categorical(ϕ[d[i]])
    end
    return (; ϕ=ϕ, θ=θ, z=z)
end

model = begin
    w = [1, 2, 3, 2, 1, 1]
    d = [1, 1, 1, 2, 2, 2]
    lda(2, d, w)
end
vi = DynamicPPL.link(VarInfo(model), model)
x = randn(length(vi[:]))

ldf_mooncake = LogDensityFunction(
    model, vi, DefaultContext(); adtype=AutoMooncake(; config=nothing)
)

LogDensityProblems.logdensity_and_gradient(ldf_mooncake, x)

end
```

```
ERROR: LoadError: UndefRefError: access to undefined reference
Stacktrace:
  [1] getindex
    @ ./essentials.jl:13 [inlined]
  [2] getindex
    @ ./subarray.jl:323 [inlined]
  [3] _broadcast_getindex
    @ ./broadcast.jl:675 [inlined]
  [4] _getindex
    @ ./broadcast.jl:705 [inlined]
  [5] _broadcast_getindex
    @ ./broadcast.jl:681 [inlined]
  [6] getindex
    @ ./broadcast.jl:636 [inlined]
  [7] macro expansion
    @ ./broadcast.jl:1004 [inlined]
  [8] macro expansion
    @ ./simdloop.jl:77 [inlined]
  [9] copyto!
    @ ./broadcast.jl:1003 [inlined]
 [10] copyto!
    @ ./broadcast.jl:956 [inlined]
 [11] materialize!
    @ ./broadcast.jl:914 [inlined]
 [12] materialize!
    @ ./broadcast.jl:911 [inlined]
[...]

PR #505, opened 2025-03-04
Fix unsafe_copyto with memoryref

willtebbutt:
Resolves #504 .

PR #516, opened 2025-03-10
Fix unsafe_copyto on 1.10

willtebbutt:
#513 highlights that #505 only addressed `Memory` / `MemoryRef`s, and therefore didn't address an equivalent issue on 1.10. This PR just fixes that.

It also ensures that we run the tests for the legacy rules. Concerningly, we were not previously doing so...

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -48,10 +48,10 @@ jobs:
           - x64
         include:
           - test_group: 'basic'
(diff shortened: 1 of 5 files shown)


willtebbutt:
I've run @mhauru example in #513 with / without this PR, and can confirm that this seems to resolve it. Will merge + tag a release as soon as CI passes.

PR #519, opened 2025-03-11
Check patch versions of Julia

willtebbutt:
New patch versions of Julia are available. I want to check that nothing has broken.


willtebbutt:
The latest patch versions of Julia seems to be working nicely, so I'm going to close this PR.

Issue #517, opened 2025-03-10
More clever `copy!!`


gdalle:
This line is causing trouble when StaticArrays are involved:

https://github.com/compintell/Mooncake.jl/blob/93a977894687a593a984375242515f86ddf50789/src/interface.jl#L172

Any idea if we can make it more generic without a StaticArrays extension?

willtebbutt:
Hmmm yeah. We really need to do something about this, because it just doesn't work for anything other than a select few types (those which have `copy!` defined, or are numbers). Similarly, the call to `copy` at https://github.com/compintell/Mooncake.jl/blob/93a977894687a593a984375242515f86ddf50789/src/interface.jl#L193 is really limiting.

I _think_ what we ought to be doing is:
1. inside Mooncake, hunt for any circular references or aliasing in `primal(y)`, and error if we find any,
2. add a unary version of `_copy!!` to recursively copy `primal(y)`, safe in the knowledge that there isn't any aliasing,
3. extend the existing binary methods to handle arbitrary (alias and circular reference-free) data structures.

I don't think these things are especially hard to do, we just need someone to do them.

willtebbutt:
Also, this relates to #486 in the sense that I want to expose `y_cache` and `_copy!!` to users so that they can write to it directly to it, but this is a breaking change.

willtebbutt:
> Yeah, this code was originally in DI, which only guarantees support for arrays and numbers, so there's your reason.

Ahh okay, yeah, this makes sense.

> Isn't there something in Base which can help? This sounds tedious

I'm not aware of anything. Happily, it's really not as bad as you might imagine. All of our functions like `zero_tangent` etc are now alias + circular-reference friendly, and the type of code that you have to implement to check that you don't have any aliasing / circular references is very similar to the code you have to write to handle them correctly. See e.g. the implementation of `zero_tangent` to see a flavour of the kind of thing we that do.

willtebbutt:
A more detailed sketch of the kind of functionality that we would need in order to solve this.

## Overall Goal

Make `value_and_pullback!!`
1. handle a much more general range of return values,
2. provide an informative contextualised error message for any outputs that cannot be handled,
3. explain clearly to users what kinds of outputs are supported, and
4. be robust and correct.

This is really _very_ important to get right, because it's part of our public interface, and something that we expect users to make use of with regularity. My proposed plan to achieve this is to
1. carefully define what outputs we want to be able to handle, and document this in `value_and_pullback!!`,
[...]

PR #525, opened 2025-03-16
Primal(output) exception handling.

AstitvaAggarwal:
First PR - clearly defined types for output. Refer #517 .


src/interface.jl
```diff
@@ -47,6 +47,23 @@ function throw_val_and_grad_ret_type_error(y)
     )
 end
 
+function throw_forward_ret_type_error(y)
+    throw(
+        ValueAndGradientReturnTypeError(
+            "Primal of output cannot contain or be of type $(typeof(y)), the amount of memory referred to must be known.",
```
willtebbutt: This is a good error + message, but I have a couple of thoughts.

Since this is a user-facing error, it would be good to avoid using the word "primal" here, because users who are just Mooncake through e.g. DifferentiationInterface and have never read any Mooncake docs are unlikely to know what we mean by it.

Also, it might be good to specifically highlight the fact that this really just applies to pointers as part of the error message, so that users learn that we're specifically trying to say that you can't use pointers.

Perhaps something like:
```suggestion
            "Found a value of type $(typeof(y)) in output, but output is not permitted to be or contain a pointer. This is because the amount of memory to which it refers is unknown, and Mooncake.jl is therefore unable to allocate appropriate memory for its gradients.",
```

src/interface.jl
```diff
@@ -47,6 +47,23 @@ function throw_val_and_grad_ret_type_error(y)
...
+        ),
+    )
+end
+
+function throw_circular_reference_or_alias_error(y)
+    throw(
+        ValueAndGradientReturnTypeError(
+            "Error: Object with address $(objectid(y)) and type $(typeof(y)) has been seen before." *
+            " Primal of output cannot contain Circular references or aliases",
```
willtebbutt:
```suggestion
            "Object with address $(objectid(y)) and type $(typeof(y)) appears more than once." *
            " Output cannot contain circular references or aliases.",
```

src/interface.jl
```diff
@@ -172,14 +189,62 @@ end
...
+    # iterate not defined for custom struct, so must use getfield() and fieldnames().
+    for y_sub in y
+        __exclude_unsupported_output_internal!(y_sub, address_set)
+    end
+    return nothing
+end
+
+# if __exclude_unsupported_output() is called over an immutable isbitstype.
+# custom structs objects are not bitstype, but thier fields can be.
+function __exclude_unsupported_output_internal!(
```
willtebbutt, line 261: I would suggest removing this method, in favour of just checking whether `P` is a bitstype in the method above, since bits types (except for pointers), cannot contain pointers / aliasing / circular references.

src/interface.jl
```diff
@@ -172,14 +189,62 @@ end
...
 Returns a cache used with [`value_and_pullback!!`](@ref). See that function for more info.
 """
 function prepare_pullback_cache(fx...; kwargs...)
-
     # Take a copy before mutating.
+    fx_temp = deepcopy(fx)
+
+    # Handle forward pass primal exceptions
+    __exclude_unsupported_output(fx_temp[1](fx_temp[2:end]...))
+    # In case any of f_new... have gone through mutations
```
willtebbutt:
```suggestion
    fx_temp = deepcopy(fx)

    # Handle forward pass primal exceptions
    __exclude_unsupported_output(fx_temp[1](fx_temp[2:end]...))
    # In case any of f_new... have gone through mutations
```
I would suggest removing this, and putting the call to `__exclude_unsupported_output` immediately after the call to `rule` on line 255 (below). This will save us having to take an extra `deepcopy`, and from running the function twice.

test/interface.jl
```diff
@@ -98,4 +98,28 @@ end
             end
         end
     end
+    @testset "prepare_pullback_cache" begin
```
willtebbutt:
```suggestion
    @testset "prepare_pullback_cache errors" begin
```

src/interface.jl
```diff
@@ -172,14 +189,62 @@ end
...
 
+"""
+    __exclude_unsupported_output(y)
+
+Required for the robust design of value_and_pullback(), prepare_pullback_cache().
+Handles aliasing, circular references and excludes the Ptr datatype.
+In the forward pass f(args...) output can only return a "Tree" like datastructure with leaf nodes as primitive types.
+Refer https://github.com/compintell/Mooncake.jl/issues/517#issuecomment-2715202789 and related issue for details.
+Internally calls __exclude_unsupported_output_internal!().
+The design is modelled after `zero_tangent`.
```
willtebbutt:
```suggestion
Required for the robust design of [`value_and_pullback`](@ref), [`prepare_pullback_cache`](@ref).
Ensures that `y` contains no aliasing, circular references or `Ptr`s.
In the forward pass f(args...) output can only return a "Tree" like datastructure with leaf nodes as primitive types.
Refer https://github.com/compintell/Mooncake.jl/issues/517#issuecomment-2715202789 and related issue for details.
Internally calls [`__exclude_unsupported_output_internal!`](@ref).
The design is modelled after `zero_tangent`.
```
Just some style notes

willtebbutt (commented):
This is a great first pass at this PR.

I think we need a _lot_ more tests for this new functionality though. Happily, you can generate a really wide range of test cases really easily using https://github.com/compintell/Mooncake.jl/blob/88dd94f0412d306f241ac89a11e9858cdbd782fb/src/tangents.jl#L1048 . If you just iterate over the vector returned by that function, pulling out the second element in each case (which is a primal value), you could just check that when you run this new functionality on it, you either get `nothing` back, or it throws a `ValueAndGradientReturnTypeError`.

src/interface.jl
AstitvaAggarwal: Yes ill do that, also im not sure how Complex or Rational bitstype must be handled, can you confirm if they are differentiable via Mooncake's current design?

willtebbutt: Yup, they ought to be fine. I'm not 100% sure whether they're currently supported by our copying behaviour, but I don't think we should handle them here. In my view it's best to keep this focused on aliasing + circular references + pointers.

src/interface.jl
AstitvaAggarwal: Yeah that makes sense, ill change it up.

src/interface.jl
AstitvaAggarwal: oh okay, got it.

src/interface.jl
AstitvaAggarwal: yeah i avoided that because if a user calls prepare_pullback_cache() over an invalid output (something like Mooncake.prepare_pullback_cache(x->Ptr{Float64}(x), 1)), the call to the rule fails (typeassert error as the CoDual tangent type comes out as NoFData) before we can check for output types. Let me know if Im seeing this correctly.

willtebbutt: Oh, that's an interesting error. I think that's actually a bug though, rather than something that is meant to happen. For example, `Mooncake.prepare_pullback_cache(x->pointer(x), randn(5))` seems to error at the appropriate moment.

(Could you open an issue about the error you've found? I suspect that a rule has a bad implementation somewhere)

AstitvaAggarwal: cool ill do that.

AstitvaAggarwal:
@willtebbutt those additional tests were actually much better than what I thought I could write. I've included some changes thanks to that.

src/interface.jl
```diff
@@ -169,8 +205,86 @@ struct Cache{Trule,Ty_cache,Ttangents<:Tuple}
...
+Ensures that `y` contains no aliasing, circular references, `Ptr`s or non differentiable datatypes. 
+In the forward pass f(args...) output can only return a "Tree" like datastructure with leaf nodes as primitive types.  
+Refer https://github.com/compintell/Mooncake.jl/issues/517#issuecomment-2715202789 and related issue for details.  
+Internally calls [`__exclude_unsupported_output_internal!`](@ref).
+The design is modelled after `zero_tangent`.
+"""
+function __exclude_unsupported_output(y::T) where {T}
+    if (T <: Union{Dict,Set})
+        throw_datastructure_output_error(y)
+    end
```
willtebbutt: Could you explain why we need to exclude `Dict`s and `Set`s?

src/interface.jl
```diff
@@ -169,8 +205,86 @@ struct Cache{Trule,Ty_cache,Ttangents<:Tuple}
...
+        throw_circular_reference_or_alias_error(y)
+    end
+
+    push!(address_set, objectid(y))
+
+    if is_user_defined_struct(T)
+        # recurse over a composite type.
+        for y_sub in fieldnames(T)
+            # isassigned, isdefined are not defined for Tuples, NamedTuples.
+            if !(T <: supportedcollections) && !isdefined(y, y_sub)
```
willtebbutt: Would it be easier just to add a method of `__exclude_unsupported_output_internal!` that explicitly handles `Tuple`s and `NamedTuple`s? i.e. something like
```julia
function __exclude_unsupported_output_internal!(y::Union{Tuple,NamedTuple}, address_set::Set{UInt})
    map(Base.Fix2(__exclude_unsupported_output_internal!, address_set), y)
    return nothing
end
```
I'm just thinking that it's more in line with the way that we handle other things.

src/interface.jl
```diff
@@ -169,8 +205,86 @@ struct Cache{Trule,Ty_cache,Ttangents<:Tuple}
     tangents::Ttangents
 end
 
-_copy!!(dst, src) = copy!(dst, src)
-_copy!!(::Number, src::Number) = src
+const supportedcollections = Union{Tuple,NamedTuple}
```
willtebbutt:
```suggestion
```
See below.

src/interface.jl
```diff
@@ -47,6 +47,39 @@ function throw_val_and_grad_ret_type_error(y)
...
+    )
+end
+
+function throw_undef_err(y)
+    throw(
+        ValueAndGradientReturnTypeError(
+            "Error: Object $(y) cannot be or contain undefined memory addresses. "
+        ),
+    )
+end
```
willtebbutt:
```suggestion
```
I think we can get rid of this. See below.

src/interface.jl
```diff
@@ -47,6 +47,39 @@ function throw_val_and_grad_ret_type_error(y)
...
+
+function throw_datastructure_output_error(y)
+    throw(
+        ValueAndGradientReturnTypeError(
+            "Error: Object of type $(typeof(y)) does not have a Real finite Hilbert space. "
+        ),
+    )
+end
```
willtebbutt:
```suggestion
```
I think we can also get rid of this. See below

src/interface.jl
```diff
@@ -169,8 +205,86 @@ struct Cache{Trule,Ty_cache,Ttangents<:Tuple}
...
+            else
+                __exclude_unsupported_output_internal!(getfield(y, y_sub), address_set)
+            end
+        end
+    else
+        # recurse over built in collections.
+        for i in eachindex(y)
+            # isassigned is valid for Arrays.
+            if !(T <: supportedcollections) && !isassigned(y, i)
+                throw_undef_err(y)
```
willtebbutt: Is there a particular problem that arises if the output contains unassigned elements?

src/interface.jl
willtebbutt: Also, I think that this handles undefined fields nicely. What's the reason that we need to exclude

willtebbutt (commented):
This is looking good, but I think there are a couple of ways we can simplify it / make it more in-keeping with the existing code in Mooncake. See comments for my thoughts.

In particular, I think we should really try to support undefined fields and assigned elements of `Array`s and `Memory`s, because they don't cost us a lot to support / prevent us from having a tree-like data structure.

src/interface.jl
AstitvaAggarwal: So in the docs in the Rule system section, its taken that the output for the model function is X x A where both X and A are real hilbert spaces. So i am under the impression that Dictionaries and Sets can be excluded as they are unordered and dont the have the properties of the real line (unlike other structures that are indexed). Am I seeing this correctly?

src/interface.jl
AstitvaAggarwal: Yes, as we are explicitly passing parts of the output as  `__exclude_unsupported_output_internal!(y[i], address_set)`
. If any of there part that contains an #undef we get an `ERROR: UndefRefError: access to undefined reference` .

src/interface.jl
AstitvaAggarwal: same reason as in above comment.

src/interface.jl
AstitvaAggarwal: I was thinking, that maybe recursion over the typeof(y) would be a way out. But to push memory addresses into our address_set we need to visit each subpart of the output.

src/interface.jl
willtebbutt: Ahhh I see. Yes, this is a fun one.

The trick to interpret `Dict`s and `Set`s as (at least locally) as forming vector spaces, is to just treat them as composite Julia types, and model them as usual.

For example, on 1.11 we have
```julia
julia> fieldtypes(Dict{Int,Float64})
(Memory{UInt8}, Memory{Int64}, Memory{Float64}, Int64, Int64, UInt64, Int64, Int64)
```
So the way that Mooncake thinks about `Dict`s is really just as a composite type with the above fields, each of which have sensibly-defined tangents. You can see that this really is the case from
```julia
julia> tangent_type(Dict{Int,Float64})
MutableTangent{@NamedTuple{slots::Memory{NoTangent}, keys::Memory{NoTangent}, vals::Memory{Float64}, ndel::NoTangent, count::NoTangent, age::NoTangent, idxfloor::NoTangent, maxprobe::NoTangent}}
```
i.e. the tangent type for a `Dict` really is just a `MutableTangent`. `Set`s are handled in a similar manner.

So it's not so much that I expect users to try to differentiate `Dict`-valued and `Set`-valued functions, as it is that there's not a particular reason to exclude them.

src/interface.jl
willtebbutt: Ah, right, but if an element is not defined, then there's no address to record, or to search over. This means that we can just skip the element if it's unassigned.

src/interface.jl
```diff
@@ -169,8 +205,86 @@ struct Cache{Trule,Ty_cache,Ttangents<:Tuple}
...
+Also errors out if `y` is or contains a Pointer.
+It is called internally by [`__exclude_unsupported_output(y)`](@ref).
+"""
+function __exclude_unsupported_output_internal!(y::T, address_set::Set{UInt}) where {T}
+    isbitstype(T) && return nothing
+    if objectid(y) in address_set
+        throw_circular_reference_or_alias_error(y)
+    end
+
+    push!(address_set, objectid(y))
```
willtebbutt, line 249: Ah, also, we only need to add items to the `address_set` if their identity is given by their address in memory, so we can restrict this to only run if `ismutabletype(T)` is `true` I think.

src/interface.jl
AstitvaAggarwal: im not sure how this can be done properly though, i would have to call isassigned() and continue in all the for loops?

AstitvaAggarwal: OMG i missed out on this, Thanks!!

src/interface.jl
AstitvaAggarwal: nice catch!!

src/interface.jl
AstitvaAggarwal: oh okay its clear for the forward pass, but then how is each element's gradient in the reverse pass (pullback) being kept track of correctly? wouldnt the unordered elements in a Set for example maybe cause mixed up gradient updates?

src/interface.jl
AstitvaAggarwal: yup removed it.

src/interface.jl
willtebbutt: Happily, because the `Set`s themselves are just implemented as `Dict`s under the hood, we don't have any problems.

I don't know if Julia actually does e.g. change up the locations of things in the `vals` field for any reason during execution but, if it does, Mooncake will see it and act accordingly.

It's an interesting one though -- although it's true that Mooncake has no problem differentiating through code which makes use of `Dict`s, but I think a user would have a hard time trying to actually define the tangent of a `Dict` in such a way that they could pull it back on the reverse-pass. However, because Mooncake will complain if they don't provide a the correct `MutableTangent` type, they can't accidentally pass in a bad tangent and have Mooncake not provide the correct result. In any case, this isn't something we need to think too much about in this PR haha.

AstitvaAggarwal: Ahh okay 😂, Thanks for the responses Will.

test/interface.jl
```diff
@@ -98,4 +98,81 @@ end
             end
         end
     end
+
+    @testset "prepare_pullback_cache errors" begin
+        # Test when function outputs a valid type.
+        struct userdefinedstruct
```
willtebbutt:
```suggestion
        struct UserDefinedStruct
```
style

test/interface.jl
```diff
@@ -98,4 +98,81 @@ end
...
+            a::Int64
+            b::Vector{Float64}
+            c::Vector{Vector{Float64}}
+        end
+
+        mutable struct userdefinedmutablestruct
```
willtebbutt:
```suggestion
        mutable struct UserDefinedMutableStruct
```
style

test/interface.jl
```diff
@@ -98,4 +98,81 @@ end
...
+            a::Int64
+            b::Vector{Float64}
+            c::Vector{Vector{Float64}}
+        end
+
+        test_topass_cases = [
```
willtebbutt:
```suggestion
        test_to_pass_cases = [
```

src/interface.jl
```diff
@@ -47,6 +47,23 @@ function throw_val_and_grad_ret_type_error(y)
     )
 end
 
+function throw_forward_ret_type_error(y)
+    throw(
+        ValueAndGradientReturnTypeError(
```
willtebbutt:
```suggestion
        ValueAndPullbackReturnTypeError(
```
This is super picky, but I wonder whether we could have a separate error type for pullback return-type errors?

AstitvaAggarwal: Yeah makes more sense as its literally for `prepare_pullback_cache()` lol.

AstitvaAggarwal: ive changed both the errors to this type.

src/interface.jl
```diff
@@ -169,8 +193,82 @@ struct Cache{Trule,Ty_cache,Ttangents<:Tuple}
...
+
+Required for checking if datatype `T` is a immutable Composite, Mutable Composite type (returns true) or a built in collection type (returns false).
+Helps in identifying user defined struct for recursing over fields correctly. 
+"""
+function is_user_defined_struct(T)
+    return isconcretetype(T) &&
+           !isprimitivetype(T) &&
+           !(T <: Union{Tuple,NamedTuple,Array}) &&
+           (@static VERSION >= v"1.11" ? !(T <: Memory) : true)
+end
```
willtebbutt: (Hopefully) the last thing I'd like to discuss is this. I'm wondering whether it makes more sense to just add methods of `__exclude_unsupported_output` for `Array` and `Memory` (the latter only on 1.11)? This would mean that we could get rid of this function entirely, and just keep the loop over `fieldnames(T)` in the method of `__exclude_unsupported_output_internal` below.

The advantage of doing this is again that it would be more in line with what we've done elsewhere in the package, and would reduce the symbol count by 1 (so there's one fewer docstring to maintain, and one fewer functions for people reading the code to have to understand).

AstitvaAggarwal: But then how would Structs/Tuple/Ntuples get checked ?

AstitvaAggarwal: Tuples/NamedTuples/Structs would still need checking though? i dont get how that would be done? (consider something like Tuple((1,1,1,[1])).

AstitvaAggarwal: if any of these has a field which has to be checked unless we call the `__exclude_unsupported_output` we cannot check for validity. And if we anyways include these, we still need to check for #undef fields differently (arrays vs Tuples etc), so the internal if condition using `is_user_defined_struct` ends up being needed.

AstitvaAggarwal: Also, 
```
julia> a=[1,2,4]
3-element Vector{Int64}:
 1
 2
 4

julia> fieldnames(typeof(a))
()
```
. This means we must only `iterate()` (`fieldnames()` wouldnt work) over such data structures. (`iterate()` is not defined for user defined structs)

willtebbutt: > Tuples/NamedTuples/Structs would still need checking though? i dont get how that would be done? (consider something like Tuple((1,1,1,[1])).

I think that `Tuple`s and `NamedTuple`s are now handled by the method that you've added that's specific to them, no?

I agree that `struct`s and `mutable struct`s still need to be handled by the existing method of `__exclude_unsupported_output_internal` though.

> This means we must only iterate() (fieldnames() wouldnt work) over such data structures. (iterate() is not defined for user defined structs)

Ah, right, yes. This is why I'm suggesting defining methods specifically for `Array` and `Memory`, which iterate over the elements.

I guess the overall picture is that the method of `__exclude_unsupported_output_internal` would now only ever handle `primitive` types, and "usual" `struct`s, and `mutable struct`s (i.e. not `Array`s and `Memory` etc), so the call to `is_user_defined_struct` would always return `true`, and could be removed.

Does this make sense, or am I missing something?

AstitvaAggarwal: ahhh got it, yes this is optimal. thanks! (i missed out the presence of the method dispatch for tuples case, and also misunderstood using the fieldnames handling for arrays.)

willtebbutt (approved):
This is a really excellent piece of work @AstitvaAggarwal -- I'll merge once CI is passing!

AstitvaAggarwal:
Thank you Will, for the reviews and the kind words.

src/interface.jl
```diff
@@ -169,8 +193,82 @@ struct Cache{Trule,Ty_cache,Ttangents<:Tuple}
...
+    end
+
+    return nothing
+end
+
+const IterableCollections = @static VERSION >= v"1.11" ? Union{Array,Memory} : Array
+
+function __exclude_unsupported_output_internal!(
+    y::T, address_set::Set{UInt}
+) where {T<:IterableCollections}
```
willtebbutt:
```suggestion
const __BuiltinArrays = @static VERSION >= v"1.11" ? Union{Array,Memory} : Array

function __exclude_unsupported_output_internal!(
    y::T, address_set::Set{UInt}
) where {T<:__BuiltinArrays}
```
so sorry, would you mind if we renamed `IterableCollections`? It's just a very general name, and I think we would be better off with something more specific. (I otherwise very much like what you've done here to avoid having to duplicate code for `Array`s and `Memory`s)

AstitvaAggarwal: No worries, ive added the change. I agree it does sound too generic, Thanks!

PR #529, opened 2025-03-21
_copy_temp() to replace Base.copy() usage.

AstitvaAggarwal:
Second PR - Avoid type piracy and define a custom function to copy outputs of supported types. Refer #517 .


AstitvaAggarwal:
@willtebbutt I'll start small with this function as `Base.copy()` is used in multiple places. If it works for all test cases in `prepare_pullback_cache()` and `value_and_pullback()` . I will switch it in all the rrules, src files.

AstitvaAggarwal:
I think the test failures are unrelated, they even spring up on PR #531 . All my tests pass locally.

willtebbutt:
Once https://github.com/compintell/Mooncake.jl/pull/531 is merged, you should just need to merge the changes from that PR into this one, and CI should start passing again.

willtebbutt:
If you sync this branch with main, it should clear up any errors caused by the issue with AllocCheck.

AstitvaAggarwal:
Although all tests pass here (as expected) I'm still working on the edge case PR for `Base.deepcopy()` so once that is finalized i will add that logic in as well, for now this code should pass all the given test cases. Also the number of commits in between kinda helped me figure out the bug in `deepcopy` 😂.

yebai:
Do we still miss anything here?

AstitvaAggarwal:
@yebai No not really, just waiting for @willtebbutt to take a look.

src/interface.jl
```diff
@@ -238,6 +238,63 @@ end
 
 const _BuiltinArrays = @static VERSION >= v"1.11" ? Union{Array,Memory} : Array
 
+# explicit for svec
+function _copy_temp(x::P) where {P<:SimpleVector}
+    return Core.svec([map(_copy_temp, x_sub) for x_sub in x]...)
+end
```
willtebbutt:
```suggestion
_copy_temp(x::SimpleVector) = Core.svec([map(_copy_temp, x_sub) for x_sub in x]...)
```
It doesn't look like the parameter `P` is being used in the method body.

src/interface.jl
```diff
@@ -238,6 +238,63 @@ end
...
+    @inbounds for i in eachindex(temp)
+        isassigned(x, i) && (temp[i] = _copy_temp(x[i]))
+    end
+    return temp
+end
+
+# Tuple, NamedTuple
+function _copy_temp(x::P) where {P<:Union{Tuple,NamedTuple}}
+    return map(_copy_temp, x)
+end
```
willtebbutt:
```suggestion
_copy_temp(x::Union{Tuple,NamedTuple}) = map(_copy_temp, x)
```

test/interface.jl
```diff
@@ -175,5 +175,61 @@ end
                 @test isa(err, Mooncake.ValueAndPullbackReturnTypeError)
             end
         end
+
+        # Test for new copy function
+        @testset for i in eachindex(additional_test_set)
```
willtebbutt: rather than having a comment, it might be better to include a string in the testset. Something like
```suggestion
        @testset "_copy_temp, $(typeof(i))" for i in eachindex(additional_test_set)
```

willtebbutt: Oh, wait, sorry. Why are we looping over the indices of `additional_test_set`, rather than just its items?

test/interface.jl
```diff
@@ -175,5 +175,61 @@ end
...
+            original = additional_test_set[i][2]
+            try
+                if isnothing(Mooncake.__exclude_unsupported_output(original))
+                    test_copy = Mooncake._copy_temp(original)
+
+                    function comparisons(
```
willtebbutt: Does `Mooncake.TestUtils.has_equal_data` not give us the correct definition of equality here?

test/interface.jl
AstitvaAggarwal: sorry, using `eachindex` is almost a habit for me now lol.

test/interface.jl
AstitvaAggarwal: ooh, i havent checked that out, will do. thanks

src/interface.jl
```diff
@@ -238,6 +238,63 @@ end
...
+# mutable composite types, bitstype
+function _copy_temp(x::P) where {P}
+    isbitstype(P) && return x
+    nf = nfields(P)
+
+    if ismutable(x)
+        temp = ccall(:jl_new_struct_uninit, Any, (Any,), P)
+        for x_sub in 1:nf
+            if isdefined(x, x_sub)
+                ccall(
```
willtebbutt, line 265: Possibly I'm missing something obvious, but do we have to use this `ccall`? Could we not call `setfield!` instead?

willtebbutt: Oh, is it that `setfield!` not work if a particular field of a `mutable struct` is declared `const`?

AstitvaAggarwal: yes that is the reason.

willtebbutt:
I think the only thing left to do here is to sort the name of the function out. As you say, something other than `_copy_temp` is probably needed.

What do you think about `_copy_output`, since we only ever use it to copy the output of a function? It's not an amazing name, but we can always change it later since we won't be making it part of the public interface.

AstitvaAggarwal:
yeah it sounds okay.

willtebbutt (approved):
LGTM. Thanks for this!

Issue #534, opened 2025-03-26
NaN gradient when raising 0 to the power of a parameter which is estimated in Turing


PTWaade:
Forgive me if this is a known issue! And thank you for the incredible work you are doing.

I encounter errors due to NaN gradients being returned by Mooncake when estimating a parameter "a" in Turing. Values are raised to the power of a, and if any of them is 0, Mooncake will return a NaN gradient. 

It works well in ForwardDiff.

I'm not sure what a workaround would be, although I am sure I could find one. I also do not know why the error happens.

```
using Turing
import Mooncake

inputs = [1., 2., 0., 4.]
outputs = [1., 2., 0., 4.] .+ randn(4)

@model function m(inputs, outputs)
    a ~ LogitNormal()
    
    for (input, output) in zip(inputs, outputs)
        output ~ to_submodel(subm(input, output, a))
    end
end

@model function subm(input, output, a)
    output ~ Normal(input^a, 1)
end

model = m(inputs, outputs)


sample(model, NUTS(; adtype = AutoMooncake(; config = nothing)), 10) #Hangs and errors
sample(model, NUTS(), 10) #works fine

using Turing: LogDensityProblems

ldf = LogDensityFunction(model; adtype=AutoMooncake(; config = nothing))
@show LogDensityProblems.logdensity_and_gradient(ldf, [0.3]) #NaN gradient

ldf = LogDensityFunction(model; adtype=AutoForwardDiff())
@show LogDensityProblems.logdensity_and_gradient(ldf, [0.3]) #Normal gradient
```

PTWaade:
And a small follow-up:
The NaN gradient leads to a NaN value in "a". When this is used with a distribution which doesn't accept NaN parameters (here it is used as the standard deviation of the Normal instead of the mean), this gives a DomainError. Because Mooncake doesn't seem to have a rrule!! for DomainError calls, it returns an error message which is hard to interpret. It took me a long time to first find the exact place in my (somewhat complex) model which gave the DomainError, then realize the parameter had a NaN value, and then to realize it was due to a NaN gradient, before finally understanding it was due to the 0^a call. 

Perhaps there is a way to return a more informative error message when a DomainError is thrown in the model? 

Otherwise - just again, my gratitude for the work you do. Let me know if I can be helpful somehow! 

```
using Turing
import Mooncake

inputs = [1., 2., 0., 4.]
outputs = [1., 2., 0., 4.] .+ randn(4)

@model function m(inputs, outputs)
[...]

PR #535, opened 2025-03-26
Rule for throw

willtebbutt:
Adds an `rrule!!` for throw, which has the behaviour that it just throws the original exception. I might tweak this to add a little bit more context at a later date, because currently it's not currently quite as informative as it might be.

Addresses part of #534

### DIFF
--- a/test/rrules/builtins.jl
+++ b/test/rrules/builtins.jl
@@ -1,3 +1,5 @@
+foo_throws(e) = throw(e)
+
 @testset "builtins" begin
     @test_throws(
         ErrorException,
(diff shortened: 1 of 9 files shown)


test/rrules/builtins.jl
```diff
@@ -1,12 +1,11 @@
 @testset "builtins" begin
     @test_throws(
         ErrorException,
-        Mooncake.rrule!!(CoDual(IntrinsicsWrappers.add_ptr, NoTangent()), 5.0, 4.0),
+        Mooncake.rrule!!(zero_fcodual(IntrinsicsWrappers.add_ptr), 5.0, 4.0),
```
willtebbutt:
```suggestion
        Mooncake.rrule!!(CoDual(IntrinsicsWrappers.add_ptr, NoTangent()), 5.0, 4.0),
```

test/rrules/builtins.jl
```diff
@@ -1,12 +1,11 @@
...
     )
-
     @test_throws(
         ErrorException,
-        Mooncake.rrule!!(CoDual(IntrinsicsWrappers.sub_ptr, NoTangent()), 5.0, 4.0),
+        Mooncake.rrule!!(zero_fcodual(IntrinsicsWrappers.sub_ptr), 5.0, 4.0),
```
willtebbutt:
```suggestion
        Mooncake.rrule!!(CoDual(IntrinsicsWrappers.sub_ptr, NoTangent()), 5.0, 4.0),
```

PR #545, opened 2025-04-05
Add tests on pre-release version of Julia (currently 1.12)

gdalle:

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -42,6 +42,7 @@ jobs:
         version:
           - 'lts'
           - '1'
+          - 'pre'
         arch:
           - x64
         include:


willtebbutt:
Certainly, we need to get Mooncake running on the pre-release version of Julia. I'm just a little hesitant to permanently commit to running everything on pre-release, due to the extra CI load it will add (I'm assuming that it will add around 10-20 mins to a typical CI run, because we're already maxing out the number of parallel runners we can use).

What are your thoughts on this @gdalle ?

gdalle:
Yeah it may not be worth it to always run tests on pre-release, especially when the pre-release is the same as the current one. You could detect this overlap with [a script](https://discourse.julialang.org/t/how-can-a-script-determine-the-latest-julia-version-and-latest-pre-release-in-ci/127446/6) to cut the CI short though.
For a library such as Mooncake, which is very sensitive to Julia language internals, I think it does make sense to adapt as early as possible to new versions of the language though. 
There are other things we can do to cut CI budget. For instance, add a switch so that draft PRs only run CI on Julia's latest release, and then once they are ready for review you add Julia LTS and pre-release? That's what I do for DI since I often run more CIs on drafts than on ready PRs

willtebbutt:
What if we just ran the `basic` test suite on pre? I'm pretty sure that this will be sensitive to changes to the compiler internals, because this is the test suite which checks the 100ish test cases for rule-derivation, which necessarily involves IR transformation.

timholy:
@willtebbutt I've been doing quite a few updates to Julia 1.12, it's certainly a big set of changes: for many packages that delve into internals, I've found the changes are so breaking that I find it best to drop support for versions before 1.12. I recognize this might be undesirable for Mooncake.

One tip: if you're sensitive to line numbers, then [`CodeTracking.linetable_scopes`](https://github.com/timholy/CodeTracking.jl/blob/332c21fbe0c69adb97d43e7bd21e6a6db6b1a0e2/src/utils.jl#L197-L240) may be of interest because it provides a consistent API for both pre- and post-Julia 1.12. I suspect you might be most interested in the variants that accept a `CodeInfo` although these don't currently get a mention in the docstring (I will fix that if you find it useful).

willtebbutt:
Thanks for your thoughts @timholy -- sadly I really do need to maintain compatibility with 1.10 until we get a new LTS. I'll take a look at `linetable_scopes`!

ChrisRackauckas:
RC1 is out so it would be good to enable pre.

.github/workflows/CI.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         version:
           - 'lts'
           - '1'
+          - 'pre'
```
yebai, line 54:
```suggestion
          - '1.12'
```

.github/workflows/CI.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         version:
           - 'lts'
           - '1'
+          - '1.12'
```
yebai:
```suggestion
          - 'rc'
```

.github/workflows/CI.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         version:
           - 'lts'
           - '1'
+          - 'rc'
```
yebai:
```suggestion
          - 'pre'
```

serenity4:
You can have a look at some of the fixes we did for Diffractor, e.g.:
- https://github.com/JuliaDiff/Diffractor.jl/pull/300 (contains a batch of fixes organized in separate commits)
- https://github.com/JuliaDiff/Diffractor.jl/pull/301
- https://github.com/JuliaDiff/Diffractor.jl/pull/303
- https://github.com/JuliaDiff/Diffractor.jl/pull/304

yebai:
@copilot, take a look at the proposed changes in https://github.com/TuringLang/Libtask.jl/pull/196 and apply the same updates here. Note that Libtask and Mooncake share the same underlying `basicblockcode` implementation.

yebai:
@copilot, take a look at the proposed changes in https://github.com/TuringLang/Libtask.jl/pull/196 and apply the same updates here. Note that Libtask and Mooncake share the same underlying `basicblockcode` implementation.

yebai:
Closed in favour of https://github.com/chalk-lab/Mooncake.jl/pull/714

PR #559, opened 2025-04-24
in place, _copy_to_output!

AstitvaAggarwal:
Second PR - Avoid type piracy and define a custom function to copy! outputs of supported types in place. Refer #517

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -16,7 +16,11 @@ function __value_and_pullback!!(
     out, pb!! = rule(fx_fwds...)
     @assert _typeof(tangent(out)) == fdata_type(T)
     increment!!(tangent(out), fdata(ȳ))
-    v = y_cache === nothing ? _copy_output(primal(out)) : _copy!!(y_cache, primal(out))
+    v = if y_cache === nothing
+        _copy_output(primal(out))
+    else
+        _copy_to_output!(y_cache, primal(out))
+    end
     return v, tuple_map((f, r) -> tangent(fdata(tangent(f)), r), fx, pb!!(rdata(ȳ)))
 end
 
@@ -166,9 +170,6 @@ function value_and_gradient!!(rule::R, fx::Vararg{Any,N}) where {R,N}
     return __value_and_gradient!!(rule, __create_coduals(fx)...)
 end
 
-_copy!!(dst, src) = copy!(dst, src)
-_copy!!(::Number, src::Number) = src
-
 function __create_coduals(args)
     try
         return tuple_map(zero_codual, args)
@@ -238,11 +239,20 @@ end
 
 const _BuiltinArrays = @static VERSION >= v"1.11" ? Union{Array,Memory} : Array
 
-# explicit for svec
+_copy_to_output!(dst::Number, src::Number) = src
+
+# explicit copy for Core.svec
+function _copy_to_output!(dst::SimpleVector, src::SimpleVector)
+    Core.svec(map(_copy_to_output!, dst, src)...)
+end
 _copy_output(x::SimpleVector) = Core.svec([map(_copy_output, x_sub) for x_sub in x]...)
 
-# Array, Memory
-function _copy_output(x::P) where {P<:_BuiltinArrays}
+# copy for Array, Memory
+function _copy_to_output!(dst::P, src::P) where {P <: _BuiltinArrays}
+    map!(_copy_to_output!, dst, dst, src)
+end
+
+function _copy_output(x::P) where {P <: _BuiltinArrays}
     temp = P(undef, size(x)...)
     @inbounds for i in eachindex(temp)
         isassigned(x, i) && (temp[i] = _copy_output(x[i]))
@@ -251,9 +261,19 @@ function _copy_output(x::P) where {P<:_BuiltinArrays}
 end
 
 # Tuple, NamedTuple
-_copy_output(x::Union{Tuple,NamedTuple}) = map(_copy_output, x)
+function _copy_to_output!(dst::P, src::P) where {P <: Union{Tuple, NamedTuple}}
+    isbitstype(P) && return src
+    map(_copy_to_output!, dst, src)
+end
 
-# mutable composite types, bitstype
+_copy_output(x::Union{Tuple, NamedTuple}) = map(_copy_output, x)
+
+# Handling structs
+function _copy_to_output!(dst::P, src::P) where {P}
+    isbitstype(P) && return src
+    P((_copy_to_output!(getfield(dst, src_sub), getfield(src, src_sub)) for src_sub in fieldnames(P))...)
+end
+
 function _copy_output(x::P) where {P}
     isbitstype(P) && return x
     nf = nfields(P)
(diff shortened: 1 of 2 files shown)


yebai:
@AstitvaAggarwal, can you fix the CI errors?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.118"
+version = "0.4.119"
```
yebai:
```suggestion
version = "0.4.118"
```

src/interface.jl
```diff
@@ -251,9 +266,52 @@ function _copy_output(x::P) where {P<:_BuiltinArrays}
...
+                    dst,
+                    src_sub - 1,
+                    _copy_to_output!(getfield(dst, src_sub), getfield(src, src_sub)),
+                )
+            end
+        end
+
+        return dst
+    else
+        flds = Vector{Any}(undef, nf)
```
sunxd3, line 294: Is this allocation on performance critical path?

sunxd3 (approved):
The PR makes a lot of sense following through the two previous ones. Great work.

Caveat: I don't know the code base that well to comment on to what degree this solves the original issue. I have to assume that some of the previous PR will catch types that we don't support. This also applies to test - I am not entirely certain about the test because I don't understand it in detail. But my confidence in the soundness of this PR is high. So approve.

There is a small concern w.r.t. the allocation, specially in potentially deeply nested recursive function. I probably won't affect the correctness, but warrant some measurement of performance IMO.

yebai:
Since this PR is part of the interface improvement relevant to DI, @gdalle might want to review it too.

src/interface.jl
AstitvaAggarwal: yes, without this allocation the logic for handling fields in immutable structs fails. (undef handling)

src/interface.jl
```diff
@@ -251,9 +266,52 @@ function _copy_output(x::P) where {P<:_BuiltinArrays}
...
+    if ismutable(src)
+        for src_sub in 1:nf
+            if isdefined(src, src_sub)
+                ccall(
+                    :jl_set_nth_field,
+                    Cvoid,
+                    (Any, Csize_t, Any),
+                    dst,
+                    src_sub - 1,
+                    _copy_to_output!(getfield(dst, src_sub), getfield(src, src_sub)),
```
penelopeysm: I was wondering why you need to do a ccall, then realised that this has already been discussed at https://github.com/chalk-lab/Mooncake.jl/pull/529/files#r2022849548 - could you add a comment in the code to explain this?

penelopeysm (commented):
Apologies if you were waiting on my review, it's been a very busy month.

The code looks great to me, after all it is mostly a carbon copy of #529 (which was also quite fine). However, it would be really useful if some of this information could be captured as comments in the code itself. It's a tricky codebase to navigate through and this will help people in the future.

More generally, an explanation of _why_ you need the `_copy_output` and now `_copy_to_output!` functions would be useful (as opposed to just using `Base.copy` and `Base.copy!`). I know some of this discussion is on GitHub and it's fine to just put the link and not go into so much detail, but even that link would be helpful, as otherwise I'd have to `git blame` to figure out the motivation for this.

AstitvaAggarwal:
@penelopeysm please refer to https://github.com/chalk-lab/Mooncake.jl/issues/517#issuecomment-2715202789 and to will's first comment on the issue as well. The need for the new copy functions is mainly as preexisting copy functions dont work on some types (eg: StaticArrays).

src/interface.jl
yebai: Let's add this clarification as a comment too.

yebai:
> @penelopeysm please refer to https://github.com/chalk-lab/Mooncake.jl/issues/517#issuecomment-2715202789 and to will's first comment on the issue as well. The need for the new copy functions is mainly as preexisting copy functions dont work on some types (eg: StaticArrays).

@AstitvaAggarwal, adding this background information as comments or doc strings in the code is good practice, so let's do it more systematically.

AstitvaAggarwal:
@yebai  regarding the new copy! function, I think will had suggested having that function exposed directly to users.

yebai:
> @yebai regarding the new copy! function, I think will had suggested having that function exposed directly to users.

Is `_copy_to_output!` used outside Mooncake? If so, I suggest we rename it to `copy_to_output!`

AstitvaAggarwal:
[here](https://github.com/chalk-lab/Mooncake.jl/issues/517#issuecomment-2711210106) he mentions `y_cache` and `copy!`  should be available to the user directly. I'm not sure I understand what he exactly here yet. I could deal with this separately as well.

src/interface.jl
```diff
@@ -238,10 +239,84 @@ end
...
+"""
+    _copy_output(x::T)
+
+Returns a copy of `x`, of the same type `T`. Allocates new memory for the copy.
+Required as Base.copy() does not work for all supported primal types. For example, `Base.copy` does not work for `Core.svec`.
+"""
 _copy_output(x::SimpleVector) = Core.svec([map(_copy_output, x_sub) for x_sub in x]...)
 
-# Array, Memory
 function _copy_output(x::P) where {P<:_BuiltinArrays}
```
```suggestion
# Array, Memory
function _copy_output(x::P) where {P<:_BuiltinArrays}
```

src/interface.jl
```diff
@@ -250,10 +325,8 @@ function _copy_output(x::P) where {P<:_BuiltinArrays}
     return temp
 end
 
-# Tuple, NamedTuple
 _copy_output(x::Union{Tuple,NamedTuple}) = map(_copy_output, x)
```
```suggestion
# Tuple, NamedTuple
_copy_output(x::Union{Tuple,NamedTuple}) = map(_copy_output, x)
```

src/interface.jl
```diff
@@ -250,10 +325,8 @@ function _copy_output(x::P) where {P<:_BuiltinArrays}
...
 
-# mutable composite types, bitstype
 function _copy_output(x::P) where {P}
```
```suggestion
# mutable composite types, bitstype
function _copy_output(x::P) where {P}
```

src/interface.jl
```diff
@@ -238,10 +239,84 @@ end
...
 const _BuiltinArrays = @static VERSION >= v"1.11" ? Union{Array,Memory} : Array
 
-# explicit for svec
+"""
+    _copy_to_output!(dst::T, src::T)
+
+Copy the contents of `src` to `dst`, with zero or minimal new memory allocation. The type of `dst` and `src` must be the same.
+Required as Base.copy!() does not work for all supported primal types. For example, `Base.copy!` does not work for `Core.svec`.
+"""
+_copy_to_output!(dst::Number, src::Number) = src
```
yebai, line 248: This function won't change `dst`.  Such semantics is unusual for `_copy_to_output!` functions. 

Nothing needs to be done here, but these unusual semantics are worth noting to avoid any sort of misuse.

PR #574, opened 2025-05-09
Update CI.yml

shravanngoswamii:
By default, the CI workflow was finding a comment posted by the GitHub Actions bot, which had the same comment ID as one generated by the Documentation workflow. This caused a permissions error because the CI workflow couldn’t update a comment created by another workflow, despite both using the same bot. We fixed this by ensuring each workflow creates and manages its own distinct comment using unique identifiers, like `<!-- perf-results -->` for CI and `<!-- docs-preview-url-Mooncake.jl -->` for Documentation, preventing overlap and resolving the issue.

Fixes https://github.com/TuringLang/actions/issues/22

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -186,17 +186,27 @@ jobs:
             cat bench/benchmark_results.txt
             echo "EOF"
           } >> $GITHUB_OUTPUT
+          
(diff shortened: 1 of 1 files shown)


yebai:
Thanks @shravanngoswamii!

Issue #428, opened 2024-12-19
Stack overflow from recursive function in DynamicExpressions.jl


MilesCranmer:
Hey all,

Thanks for working on this! I'm trying out Mooncake 0.4.65 on DynamicExpressions.jl 1.8.0 (via DifferentiationInterface 0.6.27) and ran into a stack overflow from this example:

```julia
import Mooncake
using DynamicExpressions
using DifferentiationInterface

# Build up expression:
operators = OperatorEnum(; binary_operators=(+, -, *, /), unary_operators=(cos, sin))
variable_names = ["x1", "x2", "x3"]
x1, x2, x3 = map(i -> Expression(Node{Float64}(; feature=i); operators, variable_names), 1:3)
f = x1 + cos(x2 - 0.2)

eval_sum = let f = f
    X -> sum(f(X)[1])
end
backend = AutoMooncake(; config=nothing)

# Example data
X = randn(3, 100)
dX = gradient(f, backend, X)
```

This hits the following error:

```julia
ERROR: LoadError: MooncakeRuleCompilationError: an error occured while Mooncake was compiling a rule to differentiate something. If the `caused by` error message below does not make it clear to you how the problem can be fixed, please open an issue at github.com/compintell/Mooncake.jl describing your problem.
To replicate this error run the following:

Mooncake.build_rrule(Mooncake.MooncakeInterpreter(), Tuple{Expression{Float64, Node{Float64}, @NamedTuple{operators::OperatorEnum{Tuple{typeof(+), typeof(-), typeof(*), typeof(/)}, Tuple{typeof(cos), typeof(sin)}}, variable_names::Vector{String}}}, Matrix{Float64}}; debug_mode=false)

Note that you may need to `using` some additional packages if not all of the names printed in the above signature are available currently in your environment.

Stacktrace:
 [1] build_rrule(interp::Mooncake.MooncakeInterpreter{Mooncake.DefaultCtx}, sig_or_mi::Type; debug_mode::Bool, silence_debug_messages::Bool)
   @ Mooncake ~/.julia/packages/Mooncake/N9iX9/src/interpreter/s2s_reverse_mode_ad.jl:1074
 [2] build_rrule
   @ ~/.julia/packages/Mooncake/N9iX9/src/interpreter/s2s_reverse_mode_ad.jl:1017 [inlined]
[...]

PR #588, opened 2025-05-27
Add documentation for recursive type demo

sunxd3:
Partial solution to https://github.com/chalk-lab/Mooncake.jl/issues/428


docs/src/custom_tangent_type.md
```diff
@@ -0,0 +1,228 @@
...
+        new{Tx}(x_tangent, Mooncake.NoTangent())
+    end
+
+    function TangentForA{Tx}(x_tangent::Tx, a_tangent::Union{TangentForA{Tx}, Mooncake.NoTangent}) where {Tx}
+        new{Tx}(x_tangent, a_tangent)
+    end
+
+    function TangentForA{Tx}(nt::@NamedTuple{x::Tx, a::Union{Mooncake.NoTangent, TangentForA{Tx}}}) where Tx
+        return new{Tx}(nt.x, nt.a)
+    end
```
penelopeysm: Is this method necessary, or just for convenience?

If it's just for convenience, I think I'd prefer to leave it out, otherwise a reader may assume that it's part of the interface that's needed. If it's needed, then that should be mentioned IMO.

docs/src/custom_tangent_type.md
```diff
@@ -0,0 +1,228 @@
...
+    Tx = Mooncake.tangent_type(T)
+    return Tx == Mooncake.NoTangent ? Mooncake.NoTangent : TangentForA{Tx}
+end
+```
+
+This overrides the default mechanism and associates `A` with your custom tangent type.
+
+### 4.2. `fdata_type` and `rdata_type`
+
+Define the types of forward and reverse data for `TangentForA`. In this example, since both `A` and `TangentForA` are mutable, all updates can be done in-place, so the `fdata` is the tangent itself and `rdata` is [`NoRData`](@ref). In other cases, you may need to split these more carefully.
```
penelopeysm: Do you need to declare methods for `fdata` and `rdata`?

docs/src/custom_tangent_type.md
```diff
@@ -0,0 +1,228 @@
...
+
+## 5. Bottom-Up Integration: Implement Only What You Need
+
+Mooncake provides extensive coverage and thorough testing. To get started, you can implement just enough to differentiate simple functions and add more as needed. For example, consider:
+
+```@example custom_tangent_type
+f1(a::A) = 2.0 * a.x
+```
+
+When you try to differentiate this, Mooncake will complain it lacks an `rrule!!` for `lgetfield`. Implement it:
```
penelopeysm: Maybe a brief mention of what `lgetfield` is might be useful

docs/src/custom_tangent_type.md
```diff
@@ -0,0 +1,228 @@
...
+@show val_f5
+@show grad_f5.x
+@show grad_f5.a.x
+@show grad_f5.a.a.x
+```
+
+Depending on your use case, this may be sufficient.
+
+## 6. From "It Works!" to Passing [`test_data`](@ref)
+
```
penelopeysm: Would it be possible to have a snippet of using `test_data`? (It's ok if it fails -- that's the point of this section, but I think it'd be nice to have a brief example of 'this is broken, below is how you fix it'.)

docs/src/custom_tangent_type.md
sunxd3: Not necessary for this PR, but necessary to pass `TestUtils.test_data`. Removing.

penelopeysm: Ah ok in that case actually it could be moved below too!

sunxd3: Sorry, said too soon. It's indeed necessary here.

docs/src/custom_tangent_type.md
sunxd3: Added a sentence for that.

docs/src/custom_tangent_type.md
sunxd3: I added all the code need to pass the test. Probably should live somewhere else, but okay for now.

docs/src/custom_tangent_type.md
sunxd3: Added a ref link

docs/src/custom_tangent_type.md
```diff
@@ -0,0 +1,703 @@
...
+| [`_diff_internal`](@ref)          | Structural diff between two primals → tangent                    |
+| [`_dot_internal`](@ref)           | Inner-product between tangents (dual-number consistency)         |
+| [`_scale_internal`](@ref)         | Scalar × tangent scaling                                         |
+
+#### Test Utilities
+
+| Override                                             | What it proves                                             |
+| ---------------------------------------------------- | ---------------------------------------------------------- |
+| [`TestUtils.populate_address_map_internal`](@ref)              | Tangent-to-primal pointer correspondence (cycle safety)    |
+| [`TestUtils.has_equal_data_internal`](@ref Mooncake.TestUtils.has_equal_data) (primal & tangent) | Deep equality ignoring pointer identity; handles recursion |
```
AstitvaAggarwal: This ref does not work, you could add the link to [has_equal_data](https://chalk-lab.github.io/Mooncake.jl/stable/developer_documentation/internal_docstrings/#Mooncake.TestUtils.has_equal_data-Tuple{Any,%20Any}) directly.

docs/src/custom_tangent_type.md
```diff
@@ -0,0 +1,703 @@
...
+
+By following this process—starting with a minimal set of methods and expanding as Mooncake requests more—you can support recursive types robustly in Mooncake.jl.
+
+## Appendix: Full Implementations
+
+Before defining the full implementation, `TestUtils.test_data` will fail.
```
AstitvaAggarwal:
```suggestion
Before defining the full implementation, [`TestUtils.test_data`](@ref) will fail.
```

docs/src/custom_tangent_type.md
```diff
@@ -0,0 +1,703 @@
...
+Below is a checklist of most functions you need to make [`Mooncake.TestUtils.test_data`](@ref) pass for the recursive struct `A` and its tangent `TangentForA`. They are grouped by their role in Mooncake’s test suite.
+
+#### Primitive rrules (Mandatory Differentiation Hooks)
+
+You must provide adjoints for every `getfield`/`lgetfield` variant that appears in tests.
+
+| Primitive            | Variants to implement                                                      |
+| -------------------- | -------------------------------------------------------------------------- |
+| [`lgetfield`](@ref)  | `(A, Val{:x})`, `(A, Val{:a})`, plus Symbol, Int, and (Val, Val) fallbacks |
+| `Base.getfield`      | Same coverage as `lgetfield`                                               |
```
AstitvaAggarwal: you could link `Base.getfield` to https://docs.julialang.org/en/v1/base/base/#Core.getfield

docs/src/custom_tangent_type.md
```diff
@@ -0,0 +1,703 @@
...
+            @assert Δy_rdata isa Mooncake.NoRData
+        end
+        return ntuple(_ -> Mooncake.NoRData(), n_args)
+    end
+
+    return y_cd, pb
+end
+
+Mooncake.@is_primitive Mooncake.MinimalCtx Tuple{typeof(Mooncake.lgetfield),A{T},Val{S}} where {T,S<:Symbol}
+
```
AstitvaAggarwal:
```suggestion
```

docs/src/custom_tangent_type.md
```diff
@@ -0,0 +1,703 @@
...
+Before defining the full implementation, `TestUtils.test_data` will fail.
+
+```@example custom_tangent_type
+try
+    Mooncake.TestUtils.test_data(Random.default_rng(), A(1.0, A(2.0, A(3.0))))
+catch e
+    @show e
+end
+```
+
```
AstitvaAggarwal: Overall this is awesome. But maybe break the appendix below into smaller @example blocks, with a one liner before each block referencing the function being specified + minimal tldr?

docs/src/custom_tangent_type.md
```diff
@@ -0,0 +1,703 @@
+# Writing Custom Tangent Types
+
+```@meta
```
AstitvaAggarwal: Im not sure why is this needed?

docs/src/custom_tangent_type.md
```diff
@@ -0,0 +1,703 @@
...
+        return false
+    elseif t.a isa Mooncake.NoTangent
+        return eq
+    else
+        return eq && Mooncake.TestUtils.has_equal_data_internal(t.a, s.a, equal_undefs, d)
+    end
+end
+```
+
+Now we can run it again,
```
AstitvaAggarwal:
```suggestion
Now we can run it again and successfully check if all the tangent / fdata / rdata and other required functionality works correctly for the self-referential type A.
```

docs/src/custom_tangent_type.md
sunxd3: I think it works?

docs/src/custom_tangent_type.md
sunxd3: Tried to remove it, but give me errors. I also wondered about this, given https://github.com/chalk-lab/Mooncake.jl/blob/6e95c287284b683793488f24ae1ec52d10cc398b/docs/make.jl#L29

sunxd3:
Thanks both for the review! Really helpful.

docs/src/custom_tangent_type.md
AstitvaAggarwal: https://github.com/chalk-lab/Mooncake.jl/blob/71525ecfd3ea40bbf67f7f0de17a8ca29cd2f294/docs/src/custom_tangent_type.md#test-utilities

sunxd3: Interesting, it worked for a moment. Let me fix it.

yebai:
@sunxd3, you need to (re-)format the code to address the "quality" CI failures.

docs/src/custom_tangent_type.md
```diff
@@ -0,0 +1,744 @@
...
+        return eq
+    else
+        return eq && Mooncake.TestUtils.has_equal_data_internal(t.a, s.a, equal_undefs, d)
+    end
+end
+```
+
+Now we can run it again and successfully check if all the tangent / fdata / rdata and other required functionality works correctly for the self-referential type A.
+
+```@example custom_tangent_type
```
yebai: @sunxd3, is it possible to turn these example blocks into doctests? That would help us catch any broken changes from PRs.

docs/src/custom_tangent_type.md
AstitvaAggarwal: you could add something like (as in the @meta blocks in known_limitations.md): 

> 
> ```@meta
> DocTestSetup = quote
>     using Mooncake
> end
> ```
>

docs/src/custom_tangent_type.md
sunxd3: actually it worked locally but not remotely, might be a version issue. now it should be fixed (after the doc are built.)

AstitvaAggarwal: cool

yebai:
I moved the new doc page to the "Developer" section. I'm happy to merge!

docs/src/known_limitations.md
```diff
@@ -87,6 +87,7 @@ end
 ```
 The point here is that you can manually resolve the circular dependency using a data structure which mimics the primal type.
 You will, however, need to implement similar methods for `zero_tangent`, `randn_tangent`, etc, and presumably need to implement additional `getfield` and `setfield` rules which are specific to this type.
+An example implementation of this is provided in [custom_tangent_type.md](custom_tangent_type.md).
```
yebai:
```suggestion
An example implementation of this is provided in [custom_tangent_type.md](developer_documentation/custom_tangent_type.md).
```

docs/src/custom_tangent_type.md
sunxd3: I think meta block should be equivalent.

PR #594, opened 2025-06-15
Compatibility with DynamicExpressions.jl 

MilesCranmer:
This is an attempt to define a extension for DynamicExpressions.jl recursive tree objects.

This assumes the v2 API which is slightly different than the one in #428. In particular, v2 of DynamicExpressions introduces the following change to `Node{T}` to permit arbitrary nodes of arity `D` rather than only binary nodes:

```diff
- mutable struct Node{T} <: AbstractExpressionNode{T}
+ mutable struct Node{T,D} <: AbstractExpressionNode{T,D}
       degree::UInt8
       constant::Bool
       val::T
       feature::UInt16
       op::UInt8
-      l::Node{T,D}
-      r::Node{T,D}
+      children::NTuple{D,Nullable{Node{T,D}}}
  end
```

where `Nullable{T}` is a simple container `null::Bool; x::T` to prevent invalid accesses. 

I have been struggling to get this working based off of the example in #428 and would really appreciate help getting the final part completed as I have spent about as much time as I could dedicate to it.

Here is the current error:

```julia
julia> using Mooncake, DynamicExpressions, DifferentiationInterface, Random

julia> operators = OperatorEnum(1 => (cos, sin), 2 => (+, -, *, /));

julia> x1, x2 = (
           Expression(Node{Float64}(; feature=i); operators) for i in 1:2
       );

julia> f = x1 + cos(x2 - 0.2) + 0.5
(x1 + cos(x2 - 0.2)) + 0.5

julia> X = randn(MersenneTwister(0), 3, 100);

julia> eval_sum = let f = f
           X -> sum(f(X))
       end;

julia> backend = AutoMooncake(; config=nothing);

julia> prep = prepare_gradient(eval_sum, backend, X)
ERROR: TypeError: in typeassert, expected Mooncake.CoDual{Tuple{DynamicExpressions.UtilsModule.Nullable{Node{Float64, 2}}, DynamicExpressions.UtilsModule.Nullable{Node{Float64, 2}}}, Tuple{Mooncake.FData{@NamedTuple{null::Mooncake.NoFData, x::MooncakeDynamicExpressionsExt.TangentNode{Float64, 2}}}, Mooncake.FData{@NamedTuple{null::Mooncake.NoFData, x::MooncakeDynamicExpressionsExt.TangentNode{Float64, 2}}}}}, got a value of type Mooncake.CoDual{Tuple{DynamicExpressions.UtilsModule.Nullable{Node{Float64, 2}}, DynamicExpressions.UtilsModule.Nullable{Node{Float64, 2}}}, Tuple{@NamedTuple{null::Mooncake.NoFData, x::MooncakeDynamicExpressionsExt.TangentNode{Float64, 2}}, @NamedTuple{null::Mooncake.NoFData, x::MooncakeDynamicExpressionsExt.TangentNode{Float64, 2}}}}
Stacktrace:
  [1] #_#23
    @ ~/.julia/packages/DynamicExpressions/Kn3mj/src/Expression.jl:518 [inlined]
  [2] (::Tuple{…})(none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…})
  [3] DerivedRule
    @ ~/PermaDocuments/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:966 [inlined]
  [4] _build_rule!(rule::Mooncake.LazyDerivedRule{…}, args::Tuple{…})
    @ Mooncake ~/PermaDocuments/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:1827
  [5] LazyDerivedRule
    @ ~/PermaDocuments/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:1822 [inlined]
  [6] #13
    @ ./REPL[6]:2 [inlined]
  [7] (::Tuple{…})(none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…})
  [8] (::Mooncake.DerivedRule{…})(::Mooncake.CoDual{…}, ::Mooncake.CoDual{…})
    @ Mooncake ~/PermaDocuments/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:966
  [9] prepare_gradient_cache(::Function, ::Vararg{Any}; kwargs::@Kwargs{debug_mode::Bool, silence_debug_messages::Bool})
    @ Mooncake ~/PermaDocuments/Mooncake.jl/src/interface.jl:485
 [10] prepare_gradient_cache
    @ ~/PermaDocuments/Mooncake.jl/src/interface.jl:482 [inlined]
 [11] prepare_gradient_nokwarg(::Val{true}, ::var"#13#14"{Expression{…}}, ::AutoMooncake{Nothing}, ::Matrix{Float64})
    @ DifferentiationInterfaceMooncakeExt ~/.julia/packages/DifferentiationInterface/alBlj/ext/DifferentiationInterfaceMooncakeExt/onearg.jl:114
 [12] #prepare_gradient#46
    @ ~/.julia/packages/DifferentiationInterface/alBlj/src/first_order/gradient.jl:11 [inlined]
 [13] prepare_gradient(::var"#13#14"{Expression{…}}, ::AutoMooncake{Nothing}, ::Matrix{Float64})
    @ DifferentiationInterface ~/.julia/packages/DifferentiationInterface/alBlj/src/first_order/gradient.jl:8
 [14] top-level scope
    @ REPL[8]:1
```

 cc @yebai @sunxd3


sunxd3:
Thanks Miles, I'll take a look as soon as I could.

MilesCranmer:
Thx. Actually I think I just figured out how to fix it. It just needed the following in the _rrule_getfield_common:

```julia
    fdata_for_output = if field_sym === :val
        Mooncake.fdata(pt.val)
    elseif field_sym === :children
        map(value_primal, pt.children) do child_p, child_t
            if child_t isa Mooncake.NoTangent
                Mooncake.uninit_fdata(child_p)
            else
                Mooncake.FData(Mooncake.fdata(child_t))
            end
        end
    else
        Mooncake.NoFData()
    end
```


So should be ready for a proper review now!

MilesCranmer:
I tried to integrate `test_rule` but am getting a few remaining stack overflows.... Might need some help

ext/MooncakeDynamicExpressionsExt.jl
```diff
@@ -0,0 +1,486 @@
...
+        ntuple(i -> i <= deg ? _wrap_nullable(children[i]) : NoTangent(), Val(D)),
+    )
+end
+
+function Mooncake.tangent_type(::Type{<:AbstractExpressionNode{T,D}}) where {T,D}
+    Tv = Mooncake.tangent_type(T)
+    return Tv === NoTangent ? NoTangent : TangentNode{Tv,D}
+end
+function Mooncake.tangent_type(::Type{TangentNode{Tv,D}}) where {Tv,D}
+    return TangentNode{Tv,D}
```
MilesCranmer, line 44: Not sure about this. Wouldn't this be used in higher-order derivatives, and therefore be incorrect?

(Should I make the TangentNode{Tv,D} <: AbstractExpressionNode{Tv,D} so that it can be used as its own higher order derivative? Tricky...)

MilesCranmer: (Done. I think now in principle you should be able to do derivatives _OF_ the TangentNode)

MilesCranmer: Probably too annoying to get higher-order derivatives working now. I think on that front it might be better to wait until\ recursive types are solved generally.

MilesCranmer:
I am deeply confused by the error:

```julia
 caused by: StackOverflowError:
  Stacktrace:
        [1] macro expansion
          @ ~/work/Mooncake.jl/Mooncake.jl/src/tangents.jl:435 [inlined]
        [2] macro expansion
          @ ./none:0 [inlined]
        [3] tangent_type(::Type{Node{Float64, 3}})
          @ Mooncake ./none:0
  --- the above 3 lines are repeated 1 more time ---
        [7] macro expansion
          @ ~/work/Mooncake.jl/Mooncake.jl/src/tangents.jl:388 [inlined]
        [8] macro expansion
          @ ./none:0 [inlined]
        [9] tangent_type(::Type{Tuple{DynamicExpressions.UtilsModule.Nullable{Node{Float64, 3}}, DynamicExpressions.UtilsModule.Nullable{Node{Float64, 3}}, DynamicExpressions.UtilsModule.Nullable{Node{Float64, 3}}}})
          @ Mooncake ./none:0
  --- the above 9 lines are repeated 26660 more times ---
   [239950] macro expansion
          @ ~/work/Mooncake.jl/Mooncake.jl/src/tangents.jl:435 [inlined]
   [239951] macro expansion
          @ ./none:0 [inlined]
```

Is this an internal stack overflow? The `tangent_type` call itself should not hit a stackoverflow anymore... So I am very confused by this.

MilesCranmer:
MWE

```julia
using Mooncake, DynamicExpressions, DifferentiationInterface

operators = OperatorEnum(1 => (cos, sin, exp, log, abs), 2 => (+, -, *, /), 3 => (fma, max))

# Define components of expression
x1 = Expression(Node{Float64,3}(; feature=1); operators)

# Build expression object
expr = max(x1, x1, x1)

make_eval_sum(expr) = X -> sum(expr(X))  # (To avoid full recompilation)

gradient(make_eval_sum(expr), backend, randn(1, 20))
```

which gets an unexpected stack overflow from the macro expansion.

sunxd3:
Having spent the last couple of days looking into this, I haven't fully resolved the issue, but I have a lead.

https://github.com/MilesCranmer/Mooncake.jl/blob/7f7f7ff3106759394454a1866c719d8fdf4f19e1/ext/MooncakeDynamicExpressionsExt.jl#L41-L44 wasn't called because the type is not matched correctly.

When I changed the function to
```julia
function Mooncake.tangent_type(::Type{T}) where {T<:Nullable{<:AbstractExpressionNode}}
    N = T.parameters[1]
    @assert N <: AbstractExpressionNode
    D = N.parameters[2]
    Tparam = N.parameters[1]
    Tv = Mooncake.tangent_type(Tparam)
    return Union{@NamedTuple{null::NoTangent, x::TangentNode{Tv,D}},NoTangent}
end
```
the stackoverflow went away.

---

The error becomes
<details><summary>error message</summary>

```
ERROR: MethodError: no method matching Union{…}(::@NamedTuple{…})
The type `Union{Mooncake.NoTangent, @NamedTuple{null::Mooncake.NoTangent, x::MooncakeDynamicExpressionsExt.TangentNode{Float64, 3}}}` exists, but no method is defined for this combination of argument types when trying to construct it.
Stacktrace:
  [1] macro expansion
    @ ~/TuringLang/Mooncake.jl/src/tangents.jl:549 [inlined]
  [2] macro expansion
    @ ./none:0 [inlined]
  [3] zero_tangent_internal(x::DynamicExpressions.UtilsModule.Nullable{Node{Float64, 3}}, d::IdDict{Any, Any})
    @ Mooncake ./none:0
  [4] zero_tangent(x::DynamicExpressions.UtilsModule.Nullable{Node{Float64, 3}})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/tangents.jl:464
  [5] uninit_tangent
    @ ~/TuringLang/Mooncake.jl/src/tangents.jl:561 [inlined]
  [6] uninit_fdata(p::DynamicExpressions.UtilsModule.Nullable{Node{Float64, 3}})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/fwds_rvs_data.jl:275
  [7] (::MooncakeDynamicExpressionsExt.var"#25#26")(child_p::DynamicExpressions.UtilsModule.Nullable{…}, child_t::Mooncake.NoTangent)
    @ MooncakeDynamicExpressionsExt ~/TuringLang/Mooncake.jl/ext/MooncakeDynamicExpressionsExt.jl:370
  [8] map
    @ ./tuple.jl:383 [inlined]
  [9] map
    @ ./tuple.jl:386 [inlined]
 [10] _rrule_getfield_common
    @ ~/TuringLang/Mooncake.jl/ext/MooncakeDynamicExpressionsExt.jl:368 [inlined]
 [11] rrule!!
    @ ~/TuringLang/Mooncake.jl/ext/MooncakeDynamicExpressionsExt.jl:391 [inlined]
 [12] call_mapreducer
    @ ~/.julia/packages/DynamicExpressions/Kn3mj/src/base.jl:123 [inlined]
 [13] (::Tuple{…})(none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…})
 [14] DerivedRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:966 [inlined]
 [15] _build_rule!(rule::Mooncake.LazyDerivedRule{…}, args::Tuple{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:1827
 [16] LazyDerivedRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:1822 [inlined]
 [17] AbstractExpression
    @ ~/.julia/packages/DynamicExpressions/Kn3mj/src/Expression.jl:515 [inlined]
 [18] (::Tuple{…})(none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…})
 [19] (::Mooncake.DerivedRule{…})(::Mooncake.CoDual{…}, ::Mooncake.CoDual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:966
 [20] (::Mooncake.DynamicDerivedRule{…})(::Mooncake.CoDual{…}, ::Mooncake.CoDual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:1739
 [21] (::Mooncake.RRuleZeroWrapper{…})(f::Mooncake.CoDual{…}, args::Mooncake.CoDual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:302
 [22] #11
    @ ~/TuringLang/Mooncake.jl/test_pr594/recreating_error_test.jl:21 [inlined]
 [23] (::Tuple{…})(none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…})
 [24] (::Mooncake.DerivedRule{…})(::Mooncake.CoDual{…}, ::Mooncake.CoDual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:966
 [25] prepare_gradient_cache(::Function, ::Vararg{Any}; kwargs::@Kwargs{debug_mode::Bool, silence_debug_messages::Bool})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interface.jl:485
 [26] prepare_gradient_cache
    @ ~/TuringLang/Mooncake.jl/src/interface.jl:482 [inlined]
 [27] prepare_gradient_nokwarg(::Val{true}, ::var"#11#12", ::AutoMooncake{Nothing}, ::Matrix{Float64})
    @ DifferentiationInterfaceMooncakeExt ~/.julia/packages/DifferentiationInterface/alBlj/ext/DifferentiationInterfaceMooncakeExt/onearg.jl:114
 [28] gradient(::var"#11#12", ::AutoMooncake{Nothing}, ::Matrix{Float64})
    @ DifferentiationInterface ~/.julia/packages/DifferentiationInterface/alBlj/src/first_order/gradient.jl:62
 [29] top-level scope
    @ ~/TuringLang/Mooncake.jl/test_pr594/recreating_error_test.jl:22
```
</details>

So I added
```julia
function Mooncake.zero_tangent_internal(
    x::Nullable{N}, dict::Mooncake.MaybeCache
) where {N<:AbstractExpressionNode}
    if x.null
        return NoTangent()
    else
        node_tangent = Mooncake.zero_tangent_internal(x.x, dict)
        return (; null=NoTangent(), x=node_tangent)
    end
end
```

---

Then the error becomes

<details><summary>error message</summary>

```julia
ERROR: TypeError: in typeassert, expected Mooncake.CoDual{Tuple{DynamicExpressions.UtilsModule.Nullable{Node{Float64, 3}}, DynamicExpressions.UtilsModule.Nullable{Node{Float64, 3}}, DynamicExpressions.UtilsModule.Nullable{Node{Float64, 3}}}, Union{Mooncake.NoFData, Tuple{Union{Mooncake.NoFData, @NamedTuple{null::Mooncake.NoFData, x::MooncakeDynamicExpressionsExt.TangentNode{Float64, 3}}}, Union{Mooncake.NoFData, @NamedTuple{null::Mooncake.NoFData, x::MooncakeDynamicExpressionsExt.TangentNode{Float64, 3}}}, Union{Mooncake.NoFData, @NamedTuple{null::Mooncake.NoFData, x::MooncakeDynamicExpressionsExt.TangentNode{Float64, 3}}}}}}, got a value of type Mooncake.CoDual{Tuple{DynamicExpressions.UtilsModule.Nullable{Node{Float64, 3}}, DynamicExpressions.UtilsModule.Nullable{Node{Float64, 3}}, DynamicExpressions.UtilsModule.Nullable{Node{Float64, 3}}}, Tuple{Mooncake.FData{@NamedTuple{null::Mooncake.NoFData, x::MooncakeDynamicExpressionsExt.TangentNode{Float64, 3}}}, Mooncake.FData{@NamedTuple{null::Mooncake.NoFData, x::MooncakeDynamicExpressionsExt.TangentNode{Float64, 3}}}, Mooncake.FData{@NamedTuple{null::Mooncake.NoFData, x::MooncakeDynamicExpressionsExt.TangentNode{Float64, 3}}}}}
Stacktrace:
  [1] call_mapreducer
    @ ~/.julia/packages/DynamicExpressions/Kn3mj/src/base.jl:123 [inlined]
  [2] (::Tuple{…})(none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…})
  [3] DerivedRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:966 [inlined]
  [4] _build_rule!(rule::Mooncake.LazyDerivedRule{…}, args::Tuple{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:1827
  [5] LazyDerivedRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:1822 [inlined]
  [6] AbstractExpression
    @ ~/.julia/packages/DynamicExpressions/Kn3mj/src/Expression.jl:515 [inlined]
  [7] (::Tuple{…})(none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…})
  [8] (::Mooncake.DerivedRule{…})(::Mooncake.CoDual{…}, ::Mooncake.CoDual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:966
  [9] (::Mooncake.DynamicDerivedRule{…})(::Mooncake.CoDual{…}, ::Mooncake.CoDual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:1739
 [10] (::Mooncake.RRuleZeroWrapper{…})(f::Mooncake.CoDual{…}, args::Mooncake.CoDual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:302
 [11] #13
    @ ~/TuringLang/Mooncake.jl/test_pr594/recreating_error_test.jl:24 [inlined]
 [12] (::Tuple{…})(none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…})
 [13] (::Mooncake.DerivedRule{…})(::Mooncake.CoDual{…}, ::Mooncake.CoDual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/s2s_reverse_mode_ad.jl:966
 [14] prepare_gradient_cache(::Function, ::Vararg{Any}; kwargs::@Kwargs{debug_mode::Bool, silence_debug_messages::Bool})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interface.jl:485
 [15] prepare_gradient_cache
    @ ~/TuringLang/Mooncake.jl/src/interface.jl:482 [inlined]
 [16] prepare_gradient_nokwarg(::Val{true}, ::var"#13#14", ::AutoMooncake{Nothing}, ::Matrix{Float64})
    @ DifferentiationInterfaceMooncakeExt ~/.julia/packages/DifferentiationInterface/alBlj/ext/DifferentiationInterfaceMooncakeExt/onearg.jl:114
 [17] gradient(::var"#13#14", ::AutoMooncake{Nothing}, ::Matrix{Float64})
    @ DifferentiationInterface ~/.julia/packages/DifferentiationInterface/alBlj/src/first_order/gradient.jl:62
 [18] top-level scope
    @ ~/TuringLang/Mooncake.jl/test_pr594/recreating_error_test.jl:25
```
</details>

(in a more human readable form)
```julia
# Expected type:
  Mooncake.CoDual{
      Tuple{Nullable{Node{Float64,3}}, Nullable{Node{Float64,3}}, Nullable{Node{Float64,3}}},
      Union{
          Mooncake.NoFData,
          Tuple{
              Union{Mooncake.NoFData, @NamedTuple{null::Mooncake.NoFData, x::TangentNode{Float64,3}}},
              Union{Mooncake.NoFData, @NamedTuple{null::Mooncake.NoFData, x::TangentNode{Float64,3}}},
              Union{Mooncake.NoFData, @NamedTuple{null::Mooncake.NoFData, x::TangentNode{Float64,3}}}
          }
      }
  }

# Actual type received:
  Mooncake.CoDual{
      Tuple{Nullable{Node{Float64,3}}, Nullable{Node{Float64,3}}, Nullable{Node{Float64,3}}},
      Tuple{
          Mooncake.FData{@NamedTuple{null::Mooncake.NoFData, x::TangentNode{Float64,3}}},
          Mooncake.FData{@NamedTuple{null::Mooncake.NoFData, x::TangentNode{Float64,3}}},
          Mooncake.FData{@NamedTuple{null::Mooncake.NoFData, x::TangentNode{Float64,3}}}
      }
  }
```

which I am not able to resolve at the moment. (I suspect that Mooncake doesn't particularly like the `tangent_type` [being a Union type](https://github.com/MilesCranmer/Mooncake.jl/blob/7f7f7ff3106759394454a1866c719d8fdf4f19e1/ext/MooncakeDynamicExpressionsExt.jl#L43). @willtebbutt)

---

Interestingly, the code without my modifications would work for
```julia
using Mooncake, DynamicExpressions, DifferentiationInterface

operators = OperatorEnum(
    1 => (cos, sin, exp, log, abs), 2 => (+, -, *, /, max, min), 3 => (fma, max, min)
)
x1 = Expression(Node{Float64,3}(; feature=1); operators)
test_input = randn(1, 20)
expr_binary = max(x1, x1)
f = X -> sum(expr_binary(X))
result = gradient(f, backend, test_input)
```
which tripped me for a bit. I don't have a good answer why, but it might be a helpful piece of info.

MilesCranmer:
> Having spent the last couple of days looking into this, I haven't fully resolved the issue, but I have a lead.

Thanks so much for looking into this. It is really appreciated given that I don't have much knowledge about Mooncake's internals!

> Interestingly, the code without my modifications would work for
> 
> ```julia
> using Mooncake, DynamicExpressions, DifferentiationInterface
> 
> operators = OperatorEnum(
>     1 => (cos, sin, exp, log, abs), 2 => (+, -, *, /, max, min), 3 => (fma, max, min)
> )
> x1 = Expression(Node{Float64,3}(; feature=1); operators)
> test_input = randn(1, 20)
> backend = AutoMooncake(; config=nothing)
> expr_binary = max(x1, x1)
> f = X -> sum(expr_binary(X))
> result = gradient(f, backend, test_input)
> ```
> 
> which tripped me for a bit. I don't have a good answer why, but it might be a helpful piece of info.

It's very weird. It seems the stack overflow error only shows up when using a degree 3 node like `fma` (or the 3 arg versions of max/min). Here you are using max with only 2 arguments so it works.

Which is strange because technically they all go through the exact same function call regardless of degree: https://github.com/SymbolicML/DynamicExpressions.jl/blob/b80be2a3158e8175752ae15972962a5d1cc8a53c/src/Evaluate.jl#L292-L303

Is there a specific branch in the tangent calculation that gets triggered for 3+ argument functions that differs from others?

sunxd3:
> Is there a specific branch in the tangent calculation that gets triggered for 3+ argument functions that differs from others?

This is the question I wish I had a good answer for. Sorry to say, I haven't fully internalized Mooncake's code, so couldn't say much more on the top of my head. But it's worth looking into once more free time shows up!

MilesCranmer:
I think I might (?) have just solved it. I applied this patch to DynamicExpressions.jl:

```diff
   @unstable function _get_return_type(tree, cX, operators, eval_options)
      # public Julia API version of `Core.Compiler.return_type(_eval_tree_array, typeof((tree, cX, operators, eval_options)))`
-     return eltype([_eval_tree_array(tree, cX, operators, eval_options) for _ in 1:0])
+     return Core.Compiler.return_type(_eval_tree_array, typeof((tree, cX, operators, eval_options)))
   end
```

Then the stack overflow goes away.

Basically, this `_get_return_type` – whose only purpose is to help the compiler with inference – only gets triggered for degree 3+. I do this because higher degrees tend to be harder for the compiler to reliably infer the precise array return type, and this seems to help.

While this does result in a recursive type inference, the forward pass handles it, so I am not sure why the reverse pass is unable to.

Also note that I use the `eltype([f() for _ in 1:0])` simply as a public API version of `Core.Compiler.return_type(f)` though it should be identical. I have no idea why switching to explicit `Core.Compiler.return_type` is enough to fix this. Maybe Mooncake has special treatment of either Core.Compiler.return_type _or_ the list comprehension? @willtebbutt @yebai

yebai:
> I have no idea why switching to explicit Core.Compiler.return_type is enough to fix this. Maybe Mooncake has special treatment of either Core.Compiler.return_type or the list comprehension? @willtebbutt @yebai

I am unaware of any; this is more likely a Julia compiler's inference edge case. Better to ask on the Julia repo.

willtebbutt:
We have special handling for `Core.Compiler.return_type` -- see https://github.com/chalk-lab/Mooncake.jl/blob/044fae510a68156723378c871e5cd4e072e41c95/src/rrules/foreigncall.jl#L88 . This is necessary because the implementation of `CC.return_type` involves ccalls. Hope that's helpful @MilesCranmer

MilesCranmer:
Thanks. 

> We have special handling for `Core.Compiler.return_type` -- see
> 
> https://github.com/chalk-lab/Mooncake.jl/blob/044fae510a68156723378c871e5cd4e072e41c95/src/rrules/foreigncall.jl#L88
> 
> . This is necessary because the implementation of `CC.return_type` involves ccalls. Hope that's helpful @MilesCranmer

I wonder if there is a chance list comprehensions is using a different (internal) call than `CC.return_type`, and that internal call also needs the `@zero_adjoint`?

willtebbutt:
The chance seems low to me -- list comprehensions are usually just fine, so it must be something specific to this example. Are you able to provide the IRCode associated to this function using `Mooncake.primal_ir`? There might be something there that gives more information.

MilesCranmer:
I see, thanks. It's no problem, it seems to work now anyways.

This PR is ready for review.

ext/MooncakeDynamicExpressionsExt.jl
```diff
@@ -0,0 +1,490 @@
...
+end
+
+function Mooncake.tangent_type(::Type{<:AbstractExpressionNode{T,D}}) where {T,D}
+    Tv = Mooncake.tangent_type(T)
+    return Tv === NoTangent ? NoTangent : TangentNode{Tv,D}
+end
+function Mooncake.tangent_type(::Type{Nullable{<:AbstractExpressionNode{T,D}}}) where {T,D}
+    Tv = Mooncake.tangent_type(T)
+    return Union{@NamedTuple{null::NoTangent,x::TangentNode{Tv,D}},NoTangent}
+end
```
sunxd3: This is one of the functions that are not being hit.

MilesCranmer:
A lot of the missed coverage is from inlined functions, I think.?

ext/MooncakeDynamicExpressionsExt.jl
MilesCranmer: This one seems real! I'll delete it.

ext/MooncakeDynamicExpressionsExt.jl
```diff
@@ -0,0 +1,490 @@
...
+function Mooncake.tangent_type(::Type{TangentNode{Tv,D}}) where {Tv,D}
+    return TangentNode{Tv,D}
+end
+function Mooncake.tangent_type(
+    ::Type{TangentNode{Tv,D}}, ::Type{Mooncake.NoRData}
+) where {Tv,D}
+    return TangentNode{Tv,D}
+end
```
MilesCranmer: e.g., this one - if I remove it, I get an error. But the coverage is red.

sunxd3:
Yeah, the coverage is not super reliable. I was only suggesting.

MilesCranmer:
Thx no problem.

Btw I ran into this in my integration testing within SymbolicRegression.jl. Is this real or did I do something silly?

```julia
ERROR: LoadError: TaskFailedException

    nested task error: MethodError: no method matching increment_internal!!(::Mooncake.NoCache, ::Mooncake.NoRData, ::Float64)
    The function `increment_internal!!` exists, but no method is defined for this combination of argument types.
```

Here are the full types: https://gist.github.com/MilesCranmer/ab415b39fd83073d7e424662e1c220f0

Aside: one of these types is 54,000 characters long. Is that normal?

willtebbutt:
> Aside: one of these types is 54,000 characters long. Is that normal?

😬 this is not usual haha. It's commonly to get fairly long types, but 54000 does seem excessive.

> Btw I ran into this in my integration testing within SymbolicRegression.jl. Is this real or did I do something silly?

This looks like it might be something in your code. `increment_internal!!` should only ever see two objects of the same type, since the rdata-type for any given primal is unique, and it should never happen that the rdata for two different objects is added together.

I would first determine what the correct rdata type is for whatever the primal is -- should it be `NoRData` or `Float64`? Once that's known, hopefully it won't be too tricky to figure out where the incorrect rdata is coming from.

MilesCranmer:
Sorry, what primal should I be looking at?

MilesCranmer:
Ok, fixed!

The key change was:

```diff
mutable struct TangentNode{Tv,D}
     const degree::UInt8
-    val::Union{Tv,NoTangent}
+    const constant::Bool
+    val::Tv
```

And then leave `val` undefined when there is no tangent.

I think the issue was because `(::Node).val::T` regardless of whether the val field is actually used or not. In other words, I think the `Union{Tv,NoTangent}` is only appropriate if the underlying type were something like `(::Node).val::Union{T,Nothing}`.

With this, #599, and the fixes for #597, it seems SymbolicRegressoin.jl works with Mooncake.jl as a backend now

MilesCranmer:
Hm... For some things downstream I'm still getting stack overflows. 

Are there any missing pieces of the interface? I saw `_new_` mentioned in a few places. Do I need to write a custom method for that to ensure that recursive fields aren't initialised?

It's been quite hard to get this working partly because the stackoverflow appears in `@generated` functions, and the stack trace seems unrelated from the actual recursive call. Tips appreciated. Or if there's a way to get a better error that would save me a lot of time here.

---

In particular, when I call `test_data`, I do see a stack overflow for `_new_`. See new test case.

MilesCranmer:
@willtebbutt @yebai @sunxd3 I wonder if the way `tangent_type` is computed right now is not fully compatible with recursive types? I seem some lines like this:

```julia
    tangent_type_exprs = map(n -> :(tangent_type(fieldtype(P, $n))), 1:N)
```

I worry if any of these might skip around the custom `tangent_type` (though not this particular line).

I messed around with a version of the tangent_type in src/tangents.jl where it writes to a local cache to indicate it is being computed. I then pass this cache forward to field type's tangent_type calculation – these will do an early exit if they see the tangent_type is pending. And this seems to kind of fix the issue...

yebai:
> I worry if any of these might skip around the custom tangent_type (though not this particular line).

Can you clarify via an MWE example? 

> I messed around with a version of the tangent_type in src/tangents.jl where it writes to a local cache to indicate it is being computed. I then pass this cache forward to field type's tangent_type calculation – these will do an early exit if they see the tangent_type is pending. And this seems to kind of fix the issue...

I can see why that might help with circular reference types. Can you show a prototype so we understand this more concretely?

yebai:
> when I call test_data, I do see a stack overflow for _new_. See new test case.

We need customised `_new_` methods for recursive types. See [here](https://chalk-lab.github.io/Mooncake.jl/stable/developer_documentation/custom_tangent_type/#Primitive-rrules-(Mandatory-Differentiation-Hooks)).

MilesCranmer:
Thanks. That's probably the issue. I will try now.

> I can see why that might help with circular reference types. Can you show a prototype so we understand this more concretely?

So I think to handle recursive types automatically, you would ultimately need to have a second argument for `tangent_type` (actually I suppose it should be a callee of tangent_type so it doesn't have the constraints of a generated function) in the form of a type _cache_. Something like:

```julia
function _resolve_tangent_type(::Type{P}, cache=IdDict{Any,Any}) where {P}
```

This cache would need to be passed to downstream calls. It's kind of similar to your existing cache-like mechanisms, only here it would be for _types_ rather than values.

Then what you would do is have a special sentinel struct for catching recursions:

```julia
struct UnderAssembly end
const UNDER_ASSEMBLY = UnderAssembly()
```

so that in your `tangent_type` call, you could have something like:

```julia
cached_type = get(cache, P, nothing)

if cached_type === nothing
    cache[P] = UNDER_ASSEMBLY
    
    result = #= Regular code =#

    cache[P] = result  # Indicate it is finished
    return result
elseif cached_type === UNDER_ASSEMBLY

    ######################################
    #= This means we have a recursion!! =#
    ######################################
    
    result = #= handle of recursive structs =#

    cache[P] = result  # Regular caching
    return result
else
    return cached_type  # Already computed
end
```

Now, you want this to be a local cache rather than global, to prevent any world age problems. Which is why I think having a second argument to `tangent_type` makes the most sense.

This can follow on #600 which means you would do all of this in the codegen step, so you wouldn't get any allocations at runtime.

MilesCranmer:
> @yebai 
> > when I call test_data, I do see a stack overflow for _new_. See new test case.
> 
> We need customised `_new_` methods for recursive types. See [here](https://chalk-lab.github.io/Mooncake.jl/stable/developer_documentation/custom_tangent_type/#Primitive-rrules-(Mandatory-Differentiation-Hooks)).

So, in DynamicExpressions.jl, there is only a single constructor that calls `new` - it is `new{T,D}()` for a node with eltype `T` and max degree `D`. The fields are mutable so that is what gets redefined later.

There are other convenience constructors, like `Node{T,D}(; feature=1)`. But those are done with:

```julia
    n = Node{T,D}()  # calls into `new{T,D}()`
    n.degree = 0
    n.constant = false
    n.feature = 1
```

Therefore, I don't think I need any custom `rrule!!` for `_new_`, right?  (I did try doing one but I got the same stack overflow as usual)

MilesCranmer:
Do I need to overload `setfield!`'s rrule as well? (I already did `lsetfield!`) I'm confused why I'm still getting stackoverflows after all of this

yebai:
> I don't think I need any custom rrule!! for _new_, right? 

I think you need `A(x)`, `A(x, a::A)`, `A(x, nothing)`—three separate `rrule!!` methods for different input argument types.  

> Do I need to overload setfield!'s rrule as well? (I already did lsetfield!)

I don't think you need to `setfield!`. @sunxd3 is that correct?

MilesCranmer:
> > I don't think I need any custom rrule!! for _new_, right?
> 
> I think you need `A(x)`, `A(x, a::A)`, `A(x, nothing)`—three separate `rrule!!` methods for different input argument types.

This is specific to the `A{T}` in the example. Unlike `A{T}`, the DynamicExpressions type `<:AbstractExpressionNode{T,D}` only has a single constructor: `new{T,D}()`.

yebai:
> https://github.com/chalk-lab/Mooncake.jl/pull/594#issuecomment-2994238772

Looks promising. Can you open a new issue for it? We can continue discussions there.

MilesCranmer:
Will think more about the general case later... but first I would really like to get DE.jl + Mooncake.jl working since I have an immediate need for it

ext/MooncakeDynamicExpressionsExt.jl
yebai: @MilesCranmer, following discussions in https://github.com/chalk-lab/Mooncake.jl/pull/601#issuecomment-2994361257, it seems odd that this function isn't hit.

yebai:
Update: the following works correctly on this branch: 

```julia
julia> using Mooncake, DynamicExpressions

julia> Mooncake.tangent_type(DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}})
Mooncake.Tangent{@NamedTuple{null::Mooncake.NoTangent, x::MooncakeDynamicExpressionsExt.TangentNode{Float32, 3}}}

julia> Mooncake.tangent_type(Tuple{DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}},DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}},DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}}})
Tuple{Mooncake.Tangent{@NamedTuple{null::Mooncake.NoTangent, x::MooncakeDynamicExpressionsExt.TangentNode{Float32, 3}}}, Mooncake.Tangent{@NamedTuple{null::Mooncake.NoTangent, x::MooncakeDynamicExpressionsExt.TangentNode{Float32, 3}}}, Mooncake.Tangent{@NamedTuple{null::Mooncake.NoTangent, x::MooncakeDynamicExpressionsExt.TangentNode{Float32, 3}}}}
``` 

This suggests that the `tangent_type` overloads in the extension work as expected. The remaining issue is from another source.

MilesCranmer:
It makes me think that there is some `@generated` function that is internally calling the old version of `tangent_type`... Otherwise I don't understand the stacktrace.

yebai:
Note the second example I gave is constructing a tangent type for a tuple of `Nullables`s, so it already uses the internal generated function-based `tangent_type`. A few more examples:

```julia
julia> P = DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}}
DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}}

# Tuple type 
julia> Mooncake.tangent_type(Tuple{P,P,P})
Tuple{Mooncake.Tangent{@NamedTuple{null::Mooncake.NoTangent, x::MooncakeDynamicExpressionsExt.TangentNode{Float32, 3}}}, Mooncake.Tangent{@NamedTuple{null::Mooncake.NoTangent, x::MooncakeDynamicExpressionsExt.TangentNode{Float32, 3}}}, Mooncake.Tangent{@NamedTuple{null::Mooncake.NoTangent, x::MooncakeDynamicExpressionsExt.TangentNode{Float32, 3}}}}

# NamedTupe type
julia> Mooncake.tangent_type(NamedTuple{(:x, :y, :z), Tuple{P,P,P}})
@NamedTuple{x::Mooncake.Tangent{@NamedTuple{null::Mooncake.NoTangent, x::MooncakeDynamicExpressionsExt.TangentNode{Float32, 3}}}, y::Mooncake.Tangent{@NamedTuple{null::Mooncake.NoTangent, x::MooncakeDynamicExpressionsExt.TangentNode{Float32, 3}}}, z::Mooncake.Tangent{@NamedTuple{null::Mooncake.NoTangent, x::MooncakeDynamicExpressionsExt.TangentNode{Float32, 3}}}}
```

Something else (could be trivial) is causing the stack overflow.

MilesCranmer:
Thanks. Ok I think I figured out the rest of the interface. I switched to testing `AbstractExpressionNode` directly which now passes all of `test_data`.

However, the stack overflow is still showing up for tests on `AbstractExpression`. Which I don't understand, because it's not a recursive type by itself - it just wraps the `AbstractExpressionNode`. Why would there by a stack overflow for `AbstractExpression` and not for `AbstractExpressionNode`... Any idea?

yebai:
There seems to be a mistake in the `_new_` overload. Here is an MWE. 

```julia
julia> using Mooncake, DynamicExpressions, StableRNGs

julia> T = Float32;

julia> operators = OperatorEnum(
           1 => (cos, sin, exp, log, abs), 2 => (+, -, *, /), 3 => (fma, max)
       );

julia> x1 = Expression(Node{T,3}(; feature=1); operators);

julia> x2 = Expression(Node{T,3}(; feature=2); operators);

julia> expressions = [
           Expression(Node{T,3}(; val=T(1.0)); operators),
           x1,
           x1 + T(1.0),
           cos(x1),
           x1 * x2 + sin(x1 - T(0.5)),
           fma(x1, x2, x2 + x2),
           fma(max(x1, x2, T(2.0) * x1), x2 * T(2.1), x1 * T(3.2)),
       ];

julia> arg_sets = Mooncake.TestUtils.generate_args(Mooncake._new_, expressions[end]); 

julia> Mooncake.TestUtils.test_rule(
                       StableRNG(3),
                       Mooncake._new_,
                       arg_sets[end]...;
                       interface_only=true,
                       is_primitive=true,
                       perf_flag=:none,
                       interp=Mooncake.get_interpreter(),
                   )
```


Error message 

```julia
StackOverflowError()

ERROR: ArgumentError: rule for Mooncake.CoDual{typeof(Mooncake._new_), Mooncake.NoFData} with argument types Tuple{Mooncake.CoDual{Type{Expression{Float32, Node{Float32, 3}, @NamedTuple{operators::OperatorEnum{Tuple{Tuple{typeof(cos), typeof(sin), typeof(exp), typeof(log), typeof(abs)}, Tuple{typeof(+), typeof(-), typeof(*), typeof(/)}, Tuple{typeof(fma), typeof(max)}}}}}}, Mooncake.NoFData}, Mooncake.CoDual{Node{Float32, 3}, MooncakeDynamicExpressionsExt.TangentNode{Float32, 3}}, Mooncake.CoDual{DynamicExpressions.ExpressionModule.Metadata{@NamedTuple{operators::OperatorEnum{Tuple{Tuple{typeof(cos), typeof(sin), typeof(exp), typeof(log), typeof(abs)}, Tuple{typeof(+), typeof(-), typeof(*), typeof(/)}, Tuple{typeof(fma), typeof(max)}}}}}, Mooncake.NoFData}} does not run.
Stacktrace:
 [1] test_rrule_interface(::Any, ::Any, ::Vararg{Any}; rule::Function)
   @ Mooncake.TestUtils ~/.julia/packages/Mooncake/UJPUJ/src/test_utils.jl:532
 [2] test_rule(::StableRNGs.LehmerRNG, ::typeof(Mooncake._new_), ::Vararg{…}; interface_only::Bool, is_primitive::Bool, perf_flag::Symbol, interp::Mooncake.MooncakeInterpreter{…}, debug_mode::Bool, unsafe_perturb::Bool)
   @ Mooncake.TestUtils ~/.julia/packages/Mooncake/UJPUJ/src/test_utils.jl:711
 [3] top-level scope
   @ REPL[8]:1

caused by: StackOverflowError:
Stacktrace:
      [1] tangent_type(::Type{Tuple{…}})
        @ Mooncake ./none:0
      [2] macro expansion
        @ ~/.julia/packages/Mooncake/UJPUJ/src/tangents.jl:435 [inlined]
      [3] macro expansion
        @ ./none:0 [inlined]
--- the above 3 lines are repeated 1 more time ---
      [7] tangent_type(::Type{DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}}})
        @ Mooncake ./none:0
      [8] macro expansion
        @ ~/.julia/packages/Mooncake/UJPUJ/src/tangents.jl:388 [inlined]
      [9] macro expansion
        @ ./none:0 [inlined]--- the above 9 lines are repeated 26660 more times ---
```

sunxd3:
_If_ I read correctly, I had a similar issue https://github.com/chalk-lab/Mooncake.jl/pull/594#issuecomment-2990539249 (in the middle section, after I added `zero_tangent_internal`).

My naive guess is that when the tangent_type is a Union (https://github.com/milescranmer/mooncake.jl/blob/60a5b086f652f448587f5152c302f1e6f75a834c/ext/MooncakeDynamicExpressionsExt.jl#L46), there might be some subtle dispatch issue. Not sure this is correct or helpful, but I think it's worth mentioning.

MilesCranmer:
> There seems to be a mistake in the `_new_` overload. Here is an MWE.

The weird thing is that it works with a plain `Node{T,3}` type though. Like if you test the rule on `expressions[end].tree`.

So, why would a wrapper of a recursive type hit a stack overflow, while the recursive type itself is fine? Does this still point to an issue in `_new_`’s overload, even though I’ve only overloaded it for `Node{T,3}` - which works?

MilesCranmer:
> My naive guess is that when the tangent_type is a Union (https://github.com/milescranmer/mooncake.jl/blob/60a5b086f652f448587f5152c302f1e6f75a834c/ext/MooncakeDynamicExpressionsExt.jl#L46), there might be some subtle dispatch issue. Not sure this is correct or helpful, but I think it's worth mentioning.

Turns out that this tangent_type wasn't required. I think I had added it in an attempt to fix one of the stack overflows. But it's not needed.

Same stack overflow error without it unfortunately.

```julia
  caused by: StackOverflowError:
  Stacktrace:
        [1] tangent_type(::Type{DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}}})
          @ Mooncake ./none:0
        [2] macro expansion
          @ ~/PermaDocuments/Mooncake.jl/src/tangents.jl:388 [inlined]
        [3] macro expansion
          @ ./none:0 [inlined]
        [4] tangent_type(::Type{Tuple{DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}}, DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}}, DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}}}})
          @ Mooncake ./none:0

```

It's such a weird error because clearly if you run this manually, there is no such stack overflow.

MilesCranmer:
Here's a smaller MWE:

```julia
using DynamicExpressions, Mooncake, DifferentiationInterface
operators = OperatorEnum(1 => (cos, sin), 2 => (+, -, *, /))
x1 = Expression(Node{Float64}(; feature=1); operators)

#  simple closed‑form ground truth: ∂/∂c sum(x1 + c) = N
N = 100
X = randn(3, N)
expr = x1 + 0.0      # constant in the tree

# Propagate gradients through a tree copy
eval_sum_copy = let X = X
    f -> sum(copy(f)(X))
end
gradient(eval_sum_copy, AutoMooncake(; config=nothing), expr)
```

So it seems it struggles to differentiate through tree traversal somehow


```julia
ERROR: StackOverflowError:
Stacktrace:
      [1] tangent_type(::Type{Node{Float64, 2}})
        @ Mooncake ./none:0
      [2] macro expansion
        @ ~/PermaDocuments/Mooncake.jl/src/tangents.jl:435 [inlined]
      [3] macro expansion
        @ ./none:0 [inlined]
      [4] tangent_type(::Type{DynamicExpressions.UtilsModule.Nullable{Node{Float64, 2}}})
        @ Mooncake ./none:0
      [5] macro expansion
        @ ~/PermaDocuments/Mooncake.jl/src/tangents.jl:388 [inlined]
      [6] macro expansion
        @ ./none:0 [inlined]
      [7] tangent_type(::Type{Tuple{DynamicExpressions.UtilsModule.Nullable{…}, DynamicExpressions.UtilsModule.Nullable{…}}})
        @ Mooncake ./none:0
      [8] macro expansion
        @ ~/PermaDocuments/Mooncake.jl/src/tangents.jl:435 [inlined]
--- the above 3 lines are repeated 1 more time ---
--- the above 9 lines are repeated 26659 more times ---
 [239943] macro expansion
        @ ./none:0 [inlined]
 [239944] tangent_type(::Type{DynamicExpressions.UtilsModule.Nullable{Node{Float64, 2}}})
        @ Mooncake ./none:0
 [239945] macro expansion
        @ ~/PermaDocuments/Mooncake.jl/src/tangents.jl:388 [inlined]
 [239946] macro expansion
        @ ./none:0 [inlined]
 [239947] tangent_type(::Type{Tuple{DynamicExpressions.UtilsModule.Nullable{…}, DynamicExpressions.UtilsModule.Nullable{…}}})
        @ Mooncake ./none:0
 [239948] macro expansion
        @ ~/PermaDocuments/Mooncake.jl/src/tangents.jl:435 [inlined]
 [239949] macro expansion
        @ ./none:0 [inlined]
```

MilesCranmer:
@yebai Here is an even more minimal MWE.

First, we can see that `Node{Float64}(; feature=1)` works perfectly:

```julia
julia> test_data(Random.default_rng(), Node{Float64}(; feature=1))
#= (everything passes) =#
```

However, if we embed this into any other struct, things start to break:

```julia
julia> struct A{T}; x::T; end;

julia> test_data(Random.default_rng(), A(Node{Float64}(; feature=1)))
Test Summary: | Pass  Total  Time
===           |   42     42  0.0s
Test Summary: | Pass  Total  Time
ifelse        |   92     92  0.0s
Test Summary: | Pass  Total  Time
sizeof        |   19     19  0.0s
Test Summary: | Pass  Total  Time
isa           |   63     63  0.0s
Test Summary: | Pass  Total  Time
tuple         |   63     63  0.0s
Test Summary: | Pass  Total  Time
typeassert    |   21     21  0.0s
Test Summary: | Pass  Total  Time
typeof        |   19     19  0.0s
Test Summary: | Pass  Total  Time
getfield      |   88     88  0.0s
Test Summary: | Pass  Total  Time
lgetfield     |   88     88  0.0s
StackOverflowError()

args = (A{Node{Float64, 2}}, x1): Error During Test at /Users/mcranmer/PermaDocuments/Mooncake.jl/src/test_utils.jl:1263
  Got exception outside of a @test
  ArgumentError: rule for Mooncake.CoDual{typeof(Mooncake._new_), Mooncake.NoFData} with argument types Tuple{Mooncake.CoDual{Type{A{Node{Float64, 2}}}, Mooncake.NoFData}, Mooncake.CoDual{Node{Float64, 2}, MooncakeDynamicExpressionsExt.TangentNode{Float64, 2}}} does not run.
  Stacktrace:
    [1] test_rrule_interface(::Any, ::Any, ::Vararg{Any}; rule::Function)
      @ Mooncake.TestUtils ~/PermaDocuments/Mooncake.jl/src/test_utils.jl:532
    [2] test_rule(::TaskLocalRNG, ::typeof(Mooncake._new_), ::Vararg{Any}; interface_only::Bool, is_primitive::Bool, perf_flag::Symbol, interp::Mooncake.MooncakeInterpreter{Mooncake.DefaultCtx}, debug_mode::Bool, unsafe_perturb::Bool)
      @ Mooncake.TestUtils ~/PermaDocuments/Mooncake.jl/src/test_utils.jl:711
    [3] macro expansion
      @ ~/PermaDocuments/Mooncake.jl/src/test_utils.jl:1264 [inlined]
    [4] macro expansion
      @ ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/Test/src/Test.jl:1793 [inlined]
    [5] macro expansion
      @ ~/PermaDocuments/Mooncake.jl/src/test_utils.jl:1263 [inlined]
    [6] macro expansion
      @ ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/Test/src/Test.jl:1793 [inlined]
    [7] test_rule_and_type_interactions(rng::AbstractRNG, p::A{Node{Float64, 2}})
      @ Mooncake.TestUtils ~/PermaDocuments/Mooncake.jl/src/test_utils.jl:1261
    [8] #test_data#81
      @ ~/PermaDocuments/Mooncake.jl/src/test_utils.jl:1370 [inlined]
    [9] test_data(rng::TaskLocalRNG, p::A{Node{Float64, 2}})
      @ Mooncake.TestUtils ~/PermaDocuments/Mooncake.jl/src/test_utils.jl:1367
   [10] top-level scope
      @ REPL[35]:1
   [11] eval
      @ ./boot.jl:430 [inlined]
   [12] eval_user_input(ast::Any, backend::REPL.REPLBackend, mod::Module)
      @ REPL ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/REPL/src/REPL.jl:261
   [13] repl_backend_loop(backend::REPL.REPLBackend, get_module::Function)
      @ REPL ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/REPL/src/REPL.jl:368
   [14] start_repl_backend(backend::REPL.REPLBackend, consumer::Any; get_module::Function)
      @ REPL ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/REPL/src/REPL.jl:343
   [15] run_repl(repl::REPL.AbstractREPL, consumer::Any; backend_on_current_task::Bool, backend::Any)
      @ REPL ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/REPL/src/REPL.jl:500
   [16] run_repl(repl::REPL.AbstractREPL, consumer::Any)
      @ REPL ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/REPL/src/REPL.jl:486
   [17] (::Base.var"#1150#1152"{Bool, Symbol, Bool})(REPL::Module)
      @ Base ./client.jl:446
   [18] #invokelatest#2
      @ ./essentials.jl:1055 [inlined]
   [19] invokelatest
      @ ./essentials.jl:1052 [inlined]
   [20] run_main_repl(interactive::Bool, quiet::Bool, banner::Symbol, history_file::Bool, color_set::Bool)
      @ Base ./client.jl:430
   [21] repl_main
      @ ./client.jl:567 [inlined]
   [22] _start()
      @ Base ./client.jl:541
  
  caused by: StackOverflowError:
  Stacktrace:
        [1] tangent_type(::Type{Nullable{Node{Float64, 2}}})
          @ Mooncake ./none:0
        [2] macro expansion
          @ ~/PermaDocuments/Mooncake.jl/src/tangents.jl:388 [inlined]
        [3] macro expansion
          @ ./none:0 [inlined]
        [4] tangent_type(::Type{Tuple{Nullable{Node{Float64, 2}}, Nullable{Node{Float64, 2}}}})
          @ Mooncake ./none:0
        [5] macro expansion
          @ ~/PermaDocuments/Mooncake.jl/src/tangents.jl:435 [inlined]
  --- the above 3 lines are repeated 1 more time ---
        [9] macro expansion
          @ ./none:0 [inlined]--- the above 9 lines are repeated 26660 more times ---
Test Summary:                      | Pass  Error  Total  Time
_new_                              |    4      1      5  3.0s
  args = (A{Node{Float64, 2}}, x1) |    4      1      5  3.0s
ERROR: Some tests did not pass: 4 passed, 0 failed, 1 errored, 0 broken.
```

Even though `A{T}` itself is not recursive, it seems to break `_new_` when its own type parameter references a recursive struct... Which I guess explains why `Tuple{}` wrapping node breaks things.

yebai:
I think there is a world age issue with this generic `_new_` implementation for types like `A{T}`. This world age issue implies that new functions defined after the definition of this `_new_` function are invisible. Thus, when it tries to build tangent types for field types, it will fall back to Mooncake's generic `tangent_type`

https://github.com/chalk-lab/Mooncake.jl/blob/ddef072b106ded46894cc766c174a1e837629078/src/utils.jl#L260-L262

This generated function `_new_` needs a simliar trick to 

https://github.com/chalk-lab/Mooncake.jl/blob/ddef072b106ded46894cc766c174a1e837629078/src/tangents.jl#L364-L408

I might not have enough time today.  Do you want to give it a try?

MilesCranmer:
I think part of the issue was this: https://github.com/chalk-lab/Mooncake.jl/pull/606. Which might help explain why I was getting confusing stack traces this whole time

MilesCranmer:
Ok with #606 (cherry picked here) and a few missing methods implemented, this seems to all be working! Ready for new review.

MilesCranmer:
Done

yebai (approved):

yebai:
@MilesCranmer, free to merge if CI passes.

MilesCranmer:
Note that this also includes the commits in #606. Do you want to merge that separately or with this?

yebai:
> Note that this also includes the commits in https://github.com/chalk-lab/Mooncake.jl/pull/606. Do you want to merge that separately or with this?

It's okay to merge this; @sunxd3 will add a MWE-based regression test later.

MilesCranmer:
Squash or regular merge? Seems my account can only create squashes

yebai:
Squash is preferred to keep the history simple.

PR #600, opened 2025-06-22
refactor: simpler stacktraces for tangents

MilesCranmer:
Rather than splitting `tangent_type` between codegen and compute time, this moves it all to codegen time. The result is much simpler stack traces

Helps address my complaint about hard-to-debug stack traces here: https://github.com/chalk-lab/Mooncake.jl/pull/594#issuecomment-2993801064.

Now the stack overflows will be a bit clearer.

Since the changed functions are all `@foldable`, moving all the calculation to codegen time does not change the assumptions made (I think).

### DIFF
--- a/src/tangents.jl
+++ b/src/tangents.jl
@@ -114,11 +114,7 @@ function _possibly_uninit(::Type{T}, init::Bool) where {T}
 end
 
 function _resolve_tangent_field_types(::Type{P}) where {P}
-    return map(
-        _possibly_uninit,
-        map(tangent_type, fieldtypes(P)),
-        always_initialised(P)
-    )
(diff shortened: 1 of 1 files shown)


MilesCranmer:
The GitHub diff seems to make it look more substantial than it is. Here's the VSCode diff:

image

yebai:
@MilesCranmer, can you open a new PR with a branch directly within this repo so we can see the performance benchmark results? Benchmark tasks only run for branches in this repo by default.

MilesCranmer:
```
remote: Permission to chalk-lab/Mooncake.jl.git denied to MilesCranmer.
fatal: unable to access 'https://github.com/chalk-lab/Mooncake.jl.git/': The requested URL returned error: 403
```

I think I only have triage access

yebai:
> I think I only have triage access

Fixed.

PR #601, opened 2025-06-22
refactor `tangent_type` to consolidate processing to codegen time

MilesCranmer:
Rather than splitting `tangent_type` between codegen and compute time, this moves it all to codegen time. The result is much simpler stack traces

Helps address my complaint about hard-to-debug stack traces here: https://github.com/chalk-lab/Mooncake.jl/pull/594#issuecomment-2993801064.

Now the stack overflows will be a bit clearer.

Since the changed functions are all `@foldable`, moving all the calculation to codegen time does not change the assumptions made (I think).

### DIFF
--- a/src/tangents.jl
+++ b/src/tangents.jl
@@ -114,11 +114,7 @@ function _possibly_uninit(::Type{T}, init::Bool) where {T}
 end
 
 function _resolve_tangent_field_types(::Type{P}) where {P}
-    return map(
-        _possibly_uninit,
-        map(tangent_type, fieldtypes(P)),
-        always_initialised(P)
-    )
+    return map(_possibly_uninit, map(tangent_type, fieldtypes(P)), always_initialised(P))
 end
 
 # It is essential that this gets inlined. If it does not, then we run into performance
@@ -414,7 +410,6 @@ end
     TT == NoTangent && return NoTangent
     return isconcretetype(TT) ? NamedTuple{N,TT} : Any
 end
-
 
 function _resolve_tangent_type_any(::Type{P}) where {P}
     # This method can only handle struct types. Something has gone wrong if P is primitive.
--- a/src/tangents.jl
+++ b/src/tangents.jl
@@ -437,7 +437,9 @@ function _resolve_tangent_type_any(::Type{P}) where {P}
     return ismutabletype(P) ? MutableTangent{bt} : Tangent{bt}
 end
 @foldable @generated function tangent_type(::Type{P}) where {P}
-    return _resolve_tangent_type_any(P)
+    return quote 
+        return _resolve_tangent_type_any($P)
+     end
 end
 
 backing_type(P::Type) = NamedTuple{fieldnames(P),Tuple{tangent_field_types(P)...}}
--- a/src/tangents.jl
+++ b/src/tangents.jl
@@ -439,7 +439,7 @@ end
 @foldable @generated function tangent_type(::Type{P}) where {P}
     return quote 
         return _resolve_tangent_type_any($P)
-     end
+    end
 end
 
 backing_type(P::Type) = NamedTuple{fieldnames(P),Tuple{tangent_field_types(P)...}}
--- a/src/tangents.jl
+++ b/src/tangents.jl
@@ -109,18 +109,22 @@ end
     return findfirst(==(s), fieldnames(Tfields))
 end
 
-function tangent_field_types_exprs(P::Type)
-    tangent_type_exprs = map(fieldtypes(P), always_initialised(P)) do _P, init
-        T_expr = Expr(:call, :tangent_type, _P)
-        return init ? T_expr : Expr(:curly, PossiblyUninitTangent, T_expr)
-    end
-    return tangent_type_exprs
+function _possibly_uninit(::Type{T}, init::Bool) where {T}
+    return init ? T : PossiblyUninitTangent{T}
+end
+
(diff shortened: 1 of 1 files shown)


MilesCranmer:
@yebai I'm wondering if there was a specific reason why `tangent_type` is written like it is? Is this refactor okay?

I don't completely understand the potential effects of `@foldable @generated`, especially with respect to world ages, invalidation, and extensions.

yebai:
There are some notes on the design of `tangent_type`. See [here](https://chalk-lab.github.io/Mooncake.jl/stable/developer_documentation/misc_internals_notes/#tangent_type-and-friends). 

I don't see any immediate issues other than this PR currently breaks the docs page on customised tangents by @sunxd3.

src/tangents.jl
```diff
@@ -413,34 +411,33 @@ end
...
-        # Derive tangent type.
-        bt = NamedTuple{$(fieldnames(P)),tangent_field_types_tuple}
-        return $(ismutabletype(P) ? MutableTangent : Tangent){bt}
-    end
+    # Derive tangent type.
+    bt = NamedTuple{fieldnames(P),tangent_field_types_tuple}
+    return ismutabletype(P) ? MutableTangent{bt} : Tangent{bt}
+end
+@foldable @generated function tangent_type(::Type{P}) where {P}
+    return _resolve_tangent_type_any(P)
```
yebai: We need to guarantee this won't cause world ageing issues. The `tangent_type` [note](https://chalk-lab.github.io/Mooncake.jl/stable/developer_documentation/misc_internals_notes/#tangent_type-and-friends) provides more context.

MilesCranmer:
Thanks.

Yes I am a bit worried about this (before this PR too) because of this part of the Julia docs: https://docs.julialang.org/en/v1/manual/metaprogramming/.

> Generated functions are only permitted to call functions that were defined before the definition of the generated function. (Failure to follow this may result in getting MethodErrors referring to functions from a future world-age.)

Does this explain why I can't seem to get rid of the stack overflows in #594? Because the extension is overloaded on `tangent_type` _after_ the `tangent_type` on, e.g., `Tuple`, is already compiled.

I think it might result in `tangent_type(Tuple{Node{T,D}})`  calling out to the _old_ version of `tangent_type` on its innermost arguments... Rather than the new world age one created by the extension.

MilesCranmer:
e.g., looking at the stack overflow in #594:

```julia
  caused by: StackOverflowError:
  Stacktrace:
        [1] macro expansion
          @ ~/work/Mooncake.jl/Mooncake.jl/src/tangents.jl:390 [inlined]
        [2] macro expansion
          @ ./none:0 [inlined]
        [3] tangent_type(::Type{Tuple{DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}}, DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}}, DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}}}})
          @ Mooncake ./none:0
        [4] macro expansion
          @ ~/work/Mooncake.jl/Mooncake.jl/src/tangents.jl:437 [inlined]
  --- the above 3 lines are repeated 1 more time ---
        [8] macro expansion
          @ ./none:0 [inlined]
        [9] tangent_type(::Type{DynamicExpressions.UtilsModule.Nullable{Node{Float32, 3}}})
          @ Mooncake ./none:0
  --- the above 9 lines are repeated 26660 more times ---
   [239950] macro expansion
          @ ~/work/Mooncake.jl/Mooncake.jl/src/tangents.jl:390 [inlined]
   [239951] macro expansion
          @ ./none:0 [inlined]
```

It doesn't seem to call the extension code anywhere. It seems certain branches of `tangent_type` are stuck in the old world age?

yebai:
> I think it might result in tangent_type(Tuple{Node{T,D}}) calling out to the old version of tangent_type on its innermost arguments... Rather than the new world age one created by the extension.

~~You are correct here. Due to world ageing issues, the new `tangent_type` methods in the DEMooncake extension are not callable by the generated function `tangent_type`. But, for exactly matched argument types, `tangent_type` overloads in the DEMooncake extension can still be called.~~

Hmm, I am confused. [Here](https://chalk-lab.github.io/Mooncake.jl/stable/developer_documentation/misc_internals_notes/#Good-Generated-Function-Implementation) says this issue is already fixed.

src/tangents.jl
yebai: I think this would avoid world-ageing issues since the returned expression of the generated function can call functions in the current world age. 

```suggestion
    return quote 
        return _resolve_tangent_type_any($P)
     end
```

src/tangents.jl
```diff
@@ -413,34 +411,35 @@ end
...
-        return $(ismutabletype(P) ? MutableTangent : Tangent){bt}
-    end
+    # Derive tangent type.
+    bt = NamedTuple{fieldnames(P),tangent_field_types_tuple}
+    return ismutabletype(P) ? MutableTangent{bt} : Tangent{bt}
+end
+@foldable @generated function tangent_type(::Type{P}) where {P}
+    return quote 
+        return _resolve_tangent_type_any($P)
+     end
```
yebai:
```suggestion
    return quote 
        return _resolve_tangent_type_any($P)
    end
```

yebai:
> Rather than splitting tangent_type between codegen and compute time, this moves it all to codegen time. The result is much simpler stack traces

After a more careful look, this could not work due to world age issues.

PR #603, opened 2025-06-23
overlay for Base.eps

AstitvaAggarwal:

PR #605, opened 2025-06-23
Overlay for Base.eps

AstitvaAggarwal:
related to

### DIFF
--- a/src/rrules/low_level_maths.jl
+++ b/src/rrules/low_level_maths.jl
@@ -76,6 +76,9 @@ rand_inputs(rng, P::Type{<:IEEEFloat}, ::typeof(acosd), _) = (2 * 0.9 * rand(rng
 rand_inputs(rng, P::Type{<:IEEEFloat}, ::typeof(acos), _) = (2 * 0.9 * rand(rng) - 0.9,)
 rand_inputs(rng, P::Type{<:IEEEFloat}, ::typeof(sqrt), _) = (rand(rng) + 1e-3,)
 
+# required method overlays
+Mooncake.@mooncake_overlay Base.eps(t) = zero(t)
+
 function generate_hand_written_rrule!!_test_cases(rng_ctor, ::Val{:low_level_maths})
     rng = Xoshiro(123)
     test_cases = Any[]


src/rrules/low_level_maths.jl
```diff
@@ -76,6 +76,9 @@ rand_inputs(rng, P::Type{<:IEEEFloat}, ::typeof(acosd), _) = (2 * 0.9 * rand(rng
 rand_inputs(rng, P::Type{<:IEEEFloat}, ::typeof(acos), _) = (2 * 0.9 * rand(rng) - 0.9,)
 rand_inputs(rng, P::Type{<:IEEEFloat}, ::typeof(sqrt), _) = (rand(rng) + 1e-3,)
 
+# required method overlays
+Mooncake.@mooncake_overlay Base.eps(t) = zero(t)
```
yebai, line 80: This seems to cause numerical stability issues for the `LKJChol` distribution. See CI. 

More generally, what are the consequences of using `zero(t)` over `eps(t)`. @ChrisRackauckas

ChrisRackauckas: It's only supposed to overlay it for the derivatives? Basically I'm just trying to say that using zero instead for the derivative is fine since it's only going to be off by a few floating point values. It should not change the primal though.

ChrisRackauckas: d eps(t) / dt is approximately 1 is maybe a decent enough approximation to work in any scenario. I think a lot of AD engines list it as zero but 1 is generally closer, since it's 1 sans values that don't exist in floating point.

yebai: Can we do `Mooncake.@zero_adjoint Mooncake.DefaultCtx Tuple{typeof(Base.eps), T} where T<:Base.IEEEFloat` instead?

ChrisRackauckas: https://github.com/JuliaDiff/ForwardDiff.jl/blob/d81302372bc812fbc7f3e2eafa17b4b8f570d8ae/src/dual.jl#L299 for reference ForwardDiff does `Base.eps(d::Dual) = eps(value(d))`

yebai:
@AstitvaAggarwal, this PR is still based on a branch from a fork. As suggested in #603, can you open new PRs by creating branches from this repo directly so benchmarks can run?

PR #606, opened 2025-06-23
fix: world age issue in build_fdata

MilesCranmer:
@yebai I think this fixes the world age issue causing me trouble in https://github.com/chalk-lab/Mooncake.jl/pull/594

### DIFF
--- a/src/rrules/new.jl
+++ b/src/rrules/new.jl
@@ -34,21 +34,14 @@ function rrule!!(
 end
 
 @inline function build_fdata(::Type{P}, x::Tuple, fdata::Tuple) where {P}
-    return _build_fdata_cartesian(
-        P,
-        x,
-        fdata,
-        Val(fieldcount(P)),
-        Val(fieldnames(P))
-    )
+    return _build_fdata_cartesian(P, x, fdata, Val(fieldcount(P)), Val(fieldnames(P)))
 end
 @generated function _build_fdata_cartesian(
     ::Type{P}, x::Tuple, fdata::Tuple{Vararg{Any,N}}, ::Val{nfield}, ::Val{names}
 ) where {P,N,nfield,names}
     quote
         processed_fdata = Base.Cartesian.@ntuple(
-            $nfield,
-            n -> let
+            $nfield, n -> let
                 F = fdata_field_type(P, n)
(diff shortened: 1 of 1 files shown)


yebai (commented):
Your fix is correct. Can you add a regression test for this?

MilesCranmer:
I fixed another one also on this PR. This new one has a greater risk of performance regression.

MilesCranmer:
> Your fix is correct. Can you add a regression test for this?

This is a case where I think the test would be 10x more difficult than the fix! Especially because the test can actually pass if you compile in the right world age.

I would prefer to just leave #594 as a "test" for this. (I have cherry picked the commits there)

yebai:
> This is a case where I think the test would be 10x more difficult than the fix! Especially because the test can actually pass if you compile in the right world age.

I was disappointed by Julia's autodiff situation. So, from the early stages, much effort was spent on the testing infrastructure (together with a few other considerations, e.g. docs and maintainability). That is one of the core strengths of Mooncake.

MilesCranmer:
Yes, makes a lot of sense to me and I think that is great! (Part of why I'm eager to integrate Mooncake 😉 ) Indeed, all I meant by my comment was that I have no idea how to test this particular example. If you have an idea then please feel free to add. In any case, since the current `build_fdata` breaks assumptions of `@generated`, I think it is a good thing to refactor this anyways, regardless of there being a specific MWE that demonstrates a downstream bug. Though we are lucky we have an example in #594

MilesCranmer:
Ok I think I fixed the performance regressions. Should be pretty much identical to before but now without world age issues

yebai:
Thanks for the fix. I'll see if I can nudge a team member to create a regression test based on https://github.com/chalk-lab/Mooncake.jl/pull/594#issuecomment-2997614871. 

I was more of the "demanding" users who want a robust autodiff tool. Much of the credit goes to the TuringLang team.

sunxd3:
I'll try to write a test. We can turn this into an issue if you both are keen to merge https://github.com/chalk-lab/Mooncake.jl/pull/594. (Because the code change here is used in #594).

MilesCranmer:
Merged with #594

Issue #611, opened 2025-06-24
Type stability of `prepare_gradient` in DifferentiationInterface


MilesCranmer:
I think this is more of a question rather than an issue. I was curious if the type stability of `prepare_gradient` could be improved somehow. Here's an example of where I see some type stability issues come up:

```julia
julia> foo(x) = sum(x .^ 2)
foo (generic function with 1 method)

julia> Test.@inferred prepare_gradient(f, AutoMooncake(;config=nothing), rand(32))
ERROR: return type DifferentiationInterfaceMooncakeExt.MooncakeGradientPrep{Tuple{var"#13#14", AutoMooncake{Nothing}, Vector{Float64}, Tuple{}}, Mooncake.Cache{Mooncake.DerivedRule{Tuple{var"#13#14", Vector{Float64}}, Tuple{Mooncake.CoDual{var"#13#14", Mooncake.NoFData}, Mooncake.CoDual{Vector{Float64}, Vector{Float64}}}, Mooncake.CoDual{Float64, Mooncake.NoFData}, Tuple{Float64}, Tuple{Mooncake.NoRData, Mooncake.NoRData}, false, Val{2}}, Nothing, Tuple{Mooncake.NoTangent, Vector{Float64}}}} does not match inferred return type DifferentiationInterfaceMooncakeExt.MooncakeGradientPrep{Tuple{var"#13#14", AutoMooncake{Nothing}, Vector{Float64}, Tuple{}}, Tcache} where Tcache<:(Mooncake.Cache{_A, Nothing, Tuple{Mooncake.NoTangent, Vector{Float64}}} where _A)
Stacktrace:
```

Here's a formatted version of those types:

Return type:
```julia
DifferentiationInterfaceMooncakeExt.MooncakeGradientPrep{
    Tuple{var"#13#14",AutoMooncake{Nothing},Vector{Float64},Tuple{}},
    Mooncake.Cache{
        Mooncake.DerivedRule{
            Tuple{var"#13#14",Vector{Float64}},
            Tuple{
                Mooncake.CoDual{var"#13#14",Mooncake.NoFData},
                Mooncake.CoDual{Vector{Float64},Vector{Float64}},
            },
            Mooncake.CoDual{Float64,Mooncake.NoFData},
            Tuple{Float64},
            Tuple{Mooncake.NoRData,Mooncake.NoRData},
            false,
            Val{2},
        },
        Nothing,
[...]

PR #612, opened 2025-06-24
Experimental type stability improvements

gdalle:
Mostly an experiment to answer #611. These improvements probably can't hurt but they won't make rule-building type-stable either


yebai:
The code here has drifted quite a bit from main. Please feel free to open a new PR.

PR #614, opened 2025-06-24
fix: IOStream types

MilesCranmer:
Similar issue to Base.TTY but for open file streams.


src/tangents.jl
```diff
@@ -1138,6 +1140,7 @@ function tangent_test_cases()
         (UnionAll, NoTangent),
         (typeof(<:), NoTangent),
         (Base.TTY, NoTangent),
+        (IOStream(""), NoTangent),
```
yebai, line 1143: A type is needed here: `tangent_type(IOStream)`. 

```suggestion
        (IOStream, NoTangent),
``` 

or 

```suggestion
        (typeof(IOStream("")), NoTangent),
```

MilesCranmer: I'm confused, why are there lines above like

```julia
  ([1, 2], Vector{NoTangent}),
```

? I thought it was object => tangent tests?

yebai: My bad. It is testing `(pimal, Tangent)` where `Tangent = tangent_type(typeof(pimal))`.

yebai (approved):

PR #616, opened 2025-06-25
Define overlay for Base.eps

AstitvaAggarwal:
refer [comment](https://github.com/SciML/OrdinaryDiffEq.jl/pull/2742#issuecomment-2996664400) and discussion.


src/rrules/low_level_maths.jl
```diff
@@ -57,6 +57,12 @@ end
 
 @from_rrule MinimalCtx Tuple{typeof(^),P,P} where {P<:IEEEFloat}
 
+@is_primitive MinimalCtx Tuple{typeof(Base.eps),<:IEEEFloat}
+function rrule!!(::CoDual{typeof(Base.eps)}, x::CoDual{P}) where {P<:IEEEFloat}
```
yebai, line 61: Can we have a test for this handwritten rule?

AstitvaAggarwal: Yup currently working on that.

src/rrules/low_level_maths.jl
```diff
@@ -57,6 +57,13 @@ end
...
+    y = Base.eps(primal(x))
+    eps_pb!!(dy::P) = NoRData(), isinteger(log2(primal(x))) ? P(Inf) : zero(y)
```
yebai: Is this correct? The pushback would be `Inf` for `isinteger(log2(1.))`.

```julia
julia> DifferentiationInterface.derivative(Base.eps, AutoMooncake(config=nothing), 1.0f0)
Inf32

julia> DifferentiationInterface.derivative(Base.eps, AutoForwardDiff(), 1.0f0)
0.0f0

julia> DifferentiationInterface.derivative(Base.eps, AutoFiniteDifferences(;fdm=central_fdm(5, 1)), 1.0f0)
6.3352627f-6
```

yebai:
```suggestion
    eps_pb!!(dy::P) = NoRData(), zero(y)
```

AstitvaAggarwal: Yes as Base.eps steps in value for LHS at 0.999 and RHS at 1.11 at the point 1.0. so the slope at 1. numerically Would be Inf/undef

AstitvaAggarwal: More specifically it should be NaN

MilesCranmer: (Passerby comment; feel free to ignore) – I don't understand why this should be defined as Inf32. `eps` is specific to floating point arithmetic. It is not defined in continuous $\mathbb{R}$ space; otherwise it would be 0.0 everywhere. But assuming it to be infinite at integers seems to move back to continuous space just for the derivative - where eps isn't even defined?

I think it should just be 0.25, which is what you get if you compute it with finite differences:

```julia
julia> dy = eps(1.0 + eps(1.0)) - eps(1.0 - eps(1.0))
1.1102230246251565e-16

julia> dx = (1.0 + eps(1.0)) - (1.0 - eps(1.0))
4.440892098500626e-16

julia> dy / dx
0.25
```

Same number for Float32:

```julia
julia> dy = eps(1f0 + eps(1f0)) - eps(1f0 - eps(1f0))
5.9604645f-8

julia> dx = (1f0 + eps(1f0)) - (1f0 - eps(1f0))
2.3841858f-7

julia> dy / dx
0.25f0
```

You get the same at other integers too:

```julia
julia> dy = eps(4.0 + eps(4.0)) - eps(4.0 - eps(4.0))
4.440892098500626e-16

julia> dx = (4.0 + eps(4.0)) - (4.0 - eps(4.0))
1.7763568394002505e-15

julia> dy / dx
0.25
```

MilesCranmer: Note that other AD backends actually just give zero instead:

```julia
julia> derivative(eps, AutoForwardDiff(), 1.0)
0.0

julia> derivative(eps, AutoEnzyme(), 1.0)
0.0

julia> derivative(eps, AutoReverseDiff(), 1.0)
0.0

julia> derivative(eps, AutoFiniteDiff(), 1.0)
0.0
```

MilesCranmer: (But I think giving inf return is a bad idea. Often people just use `eps` to get an offset for `1/(x+eps(x))`. I don't think that should turn into `-(1 + inf) / (x + eps(x))^2` at integers! That would produce infinities in random places.

MilesCranmer: Oh wait sorry I see that @yebai already implemented this change!

AstitvaAggarwal: Oh okay, thanks for the input!

AstitvaAggarwal: I had actually meant NaN as it is non differentiable at certain points though

PR #618, opened 2025-06-25
docs update for @mooncake_overlay

### LINKED ISSUE #587: `@mooncake_overlay` does not seem to activate if in an extension
I put `Mooncake.@mooncake_overlay DiffEqBase.set_mooncakeoriginator_if_mooncake(x::ChainRulesOriginator) = MooncakeOriginator()` in an extension, and it does not seem to trigger in the core package. See https://github.com/SciML/DiffEqBase.jl/pull/1158

AstitvaAggarwal:
closes #587 . `Mooncake.@mooncake_overlay` can be extended to work consistently globally if needed in the future.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.127"
+version = "0.4.126"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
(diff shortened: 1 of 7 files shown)


src/tools_for_rules.jl
```diff
@@ -37,6 +37,11 @@ Define a method of a function which only Mooncake can see. This can be used to w
 versions of methods which can be successfully differentiated by Mooncake if the original
 cannot be.
 
+!!! note
+    This macro works only within the module's namespace from which it is called from.
+    In case it is called within a package's extension to overlay a method, the overlay effect is only valid within that extension module.
+    To use it in the `Main` module you must explicitly apply `Mooncake.@mooncake_overlay` again.
+
```
yebai, line 44: @AstitvaAggarwal This is not clear. Can you add an example where this doesn't work as expected, and also how it gets fixed via applying overlay again in the `Main` module

AstitvaAggarwal: I looked into this in detail, overlay seems to be working just fine even with extensions.

Issue #597, opened 2025-06-20
Missing `tangent_type` for `Union{Mooncake.NoRData, Float32}`


MilesCranmer:
This is related to my attempt to get Mooncake.jl working as an autodiff backend for SymbolicRegression.jl and by extension PySR. The main issue was #428 which might be solved by #594.

This builds off of #594 so I'm not sure whether it's from an incorrect implementation there or something else. The error is about a missing `tangent_type` for `::Type{NoFData}`, but when the rightmost argument is `::Type{Union{Mooncake.NoRData, Float32}}`, which I thought was a little weird.

Here's the reproducer:

```julia
julia> using SymbolicRegression, Mooncake, MLJBase, DifferentiationInterface;

julia> Mooncake.zero_tangent_internal(::Base.TTY, ::IdDict{Any, Any}) = Mooncake.NoTangent();

julia> X = (; x=randn(100), y=randn(100), z=randn(100));

julia> y = @. 2 * cos(X.x) + X.y ^ 2 - 4 * X.z;

julia> model = SRRegressor(
           binary_operators=(+, *, /, -),
           unary_operators=(cos, exp),
           autodiff_backend=AutoMooncake(; config=nothing),
           parallelism=:serial,
       );

julia> mach = machine(model, X, y);

julia> fit!(mach)
[ Info: Training machine(SRRegressor(defaults = nothing, …), …).
┌ Error: Problem fitting the machine machine(SRRegressor(defaults = nothing, …), …). 
└ @ MLJBase ~/.julia/packages/MLJBase/F1Eh6/src/machines.jl:694
[ Info: Running type checks... 
[ Info: Type checks okay. 
ERROR: TaskFailedException

    nested task error: MethodError: no method matching tangent_type(::Type{Mooncake.NoFData}, ::Type{Union{Mooncake.NoRData, Float32}})
    The function `tangent_type` exists, but no method is defined for this combination of argument types.
    
    Closest candidates are:
      tangent_type(::Type{Mooncake.NoFData}, ::Type{Mooncake.NoRData})
       @ Mooncake ~/.julia/packages/Mooncake/zsutF/src/fwds_rvs_data.jl:859
      tangent_type(::Type{Mooncake.NoFData}, ::Type{<:Base.TwicePrecision})
       @ Mooncake ~/.julia/packages/Mooncake/zsutF/src/rrules/twice_precision.jl:52
[...]

PR #620, opened 2025-06-26
Support for `Union{Nothing, <:IEEEFloat}`

yebai:
Fix https://github.com/chalk-lab/Mooncake.jl/issues/597 (again).

### DIFF
--- a/src/test_resources.jl
+++ b/src/test_resources.jl
@@ -147,15 +147,23 @@ end
 # https://github.com/chalk-lab/Mooncake.jl/issues/597 for the reason.
 struct P_union_nothing_float{T<:Base.IEEEFloat}
     x::Union{T,Nothing}
+    x2::Union{T,Nothing}
     y::T
+    z::Union{Array{T,1},Nothing}
+    z2::Union{Array{T,1},Nothing}
+    w::Union{Array{T,2},Nothing}
+    w2::Union{Array{T,2},Nothing}
 end
-make_P_union_nothing() = P_union_nothing_float{Float32}(Float32(1.0), Float32(1.0))
-struct P_union_nothing_array{T<:Base.IEEEFloat}
-    x::Union{Array{T,1},Nothing}
-    y::T
-end
-function make_P_union_nothing_array()
-    P_union_nothing_array{Float32}(randn(Xoshiro(1), Float32, 2), Float32(1.0))
+function make_P_union_nothing(T=Float32)
+    P_union_nothing_float{T}(
+        T(1.0),
+        nothing,
+        T(1.0),
+        randn(Xoshiro(1), T, 2),
+        nothing,
+        randn(Xoshiro(1), T, 2, 2),
+        nothing,
+    )
 end
 
 function build_big_isbits_struct()
--- a/test/fwds_rvs_data.jl
+++ b/test/fwds_rvs_data.jl
@@ -30,9 +30,6 @@ end
         TestUtils.test_tangent_splitting(
             Xoshiro(123456), TestResources.make_P_union_nothing(); test_opt_flag=false
         )
-        TestUtils.test_tangent_splitting(
-            Xoshiro(123456), TestResources.make_P_union_nothing_array(); test_opt_flag=false
-        )
     end
 
     @testset "zero_rdata_from_type checks" begin


yebai:
I don't understand the `Aqua` complaint about unbound type parameters. @MilesCranmer @penelopeysm any ideas? 

```julia
Unbound type parameters detected:
[1] tangent_type(::Type{NoFData}, ::Type{Union{NoRData, T}}) where T<:Union{Float16, Float32, Float64} @ Mooncake ~/work/Mooncake.jl/Mooncake.jl/src/fwds_rvs_data.jl:864
[2] tangent_type(::Type{Union{NoFData, T}}, ::Type{NoRData}) where T<:(Array{<:Union{Float16, Float32, Float64}}) @ Mooncake ~/work/Mooncake.jl/Mooncake.jl/src/fwds_rvs_data.jl:867
Unbound type parameters: Test Failed at /home/runner/.julia/packages/Aqua/1UuaV/src/unbound_args.jl:37
  Expression: isempty(unbounds)
   Evaluated: isempty(Method[tangent_type(::Type{NoFData}, ::Type{Union{NoRData, T}}) where T<:Union{Float16, Float32, Float64} @ Mooncake ~/work/Mooncake.jl/Mooncake.jl/src/fwds_rvs_data.jl:864, tangent_type(::Type{Union{NoFData, T}}, ::Type{NoRData}) where T<:(Array{<:Union{Float16, Float32, Float64}}) @ Mooncake ~/work/Mooncake.jl/Mooncake.jl/src/fwds_rvs_data.jl:867])
```

MilesCranmer:
I wonder if that is an Aqua.jl bug? Usually that error would be for things like

```julia
f(x::Union{Float16,T}) where {T} = T
```

which _is_ a real issue. But when inside `Type{Union{Float16,T}}` it's different, I suppose if `T` could be `Float16`, that would be an issue, but here `T` can never be `NoRData` so it seems fine to me.

penelopeysm:
```julia
module Foo
struct Wat end
function f(::Type{Union{Wat,T}}) where {T <: Base.IEEEFloat}
    return T
end
end

using Aqua
Aqua.test_unbound_args(Foo)
# ERROR: ...

Foo.f(Foo.Wat)
# Union {}

Foo.Wat isa Type{Union{Foo.Wat,T}} where {T <: Base.IEEEFloat}
# true
```

I have to confess that the last line is entirely unintuitive to me (but I suppose given this behaviour, Aqua is _technically_ correct...?). (Edit: And then:)

```julia
Float64 isa Type{Union{Foo.Wat,T}} where {T <: Base.IEEEFloat}
# false
```

MilesCranmer:
Ugh. Maybe it's special handling of `Type{Union{A,B}}` gets converted into `Type{<:Union{A,B}}` which is similar to `Vararg`...

MilesCranmer:
```julia
julia> Float64 isa Type{Union{Foo.Wat,T}} where {T <: Base.IEEEFloat}
false

julia> Foo.Wat isa Type{Union{Foo.Wat,T}} where {T <: Base.IEEEFloat}
true
```

Very weird indeed

MilesCranmer:
Even weirder!

```julia
julia> Union{Foo.Wat,Float32} isa Type{Union{Foo.Wat,T}} where {T <: Base.IEEEFloat}
true
```

Ok @yebai I think the best solution is to just manually declare for each element of Base.IEEEFloat rather than use the union.

yebai:
It appears to be an edge behaviour of Julia that we don't understand... 

```julia
julia> Float64 isa Type{Union{Foo.Wat,T}} where {T <: Base.IEEEFloat}
false

julia> Foo.Wat isa Type{Union{Foo.Wat,T}} where {T <: Base.IEEEFloat}
true

julia> Union{Float16} isa Type{Union{Foo.Wat,T}} where {T <: Base.IEEEFloat}
false

julia> Union{Float16, Float32} isa Type{Union{Foo.Wat,T}} where {T <: Base.IEEEFloat}
false

julia> Union{Float16, Float32, Float64} isa Type{Union{Foo.Wat,T}} where {T <: Base.IEEEFloat}
false

julia> Union{Float16, Float32, Foo.Wat} isa Type{Union{Foo.Wat,T}} where {T <: Base.IEEEFloat}
true
```

@ChrisRackauckas, @ViralBShah, maybe you have additional insights?

yebai:
@MilesCranmer, this is ready for another look.

MilesCranmer:
Thanks. Got some explanation here: https://discourse.julialang.org/t/weirdness-of-type-union-a-b-where-b-superb/130234/2?u=milescranmer.

But at the moment I'm not sure any efficient way of writing this so let's just go with the current

src/fwds_rvs_data.jl
```diff
@@ -861,12 +861,14 @@ tangent type. This method must be equivalent to `tangent_type(_typeof(primal))`.
...
-@foldable tangent_type(F::Type{Union{NoFData,Array{<:Base.IEEEFloat}}}, R::Type{NoRData}) = Union{
-    NoTangent,tangent_type(Array{<:Base.IEEEFloat})
-}
+for T in [Float16, Float32, Float64]
+    @eval @foldable tangent_type(::Type{NoFData}, ::Type{Union{NoRData,$T}}) = Union{
+        NoTangent,tangent_type($T)
+    }
+    @eval @foldable tangent_type(::Type{Union{NoFData,Array{$T}}}, ::Type{NoRData}) = Union{
+        NoTangent,tangent_type(Array{$T})
+    }
```
MilesCranmer:
```suggestion
    for N in 0:5  # Just go up to N=5 until general solution to https://github.com/chalk-lab/Mooncake.jl/pull/620 available
        @eval @foldable tangent_type(::Type{Union{NoFData,Array{$T,$N}}}, ::Type{NoRData}) = Union{
            NoTangent,tangent_type(Array{$T,$N})
        }
    end
```

src/test_resources.jl
```diff
@@ -145,10 +145,14 @@ end
...
 # https://github.com/chalk-lab/Mooncake.jl/issues/597 for the reason.
-struct P_union_nothing
-    x::Union{Base.IEEEFloat,Nothing}
+struct P_union_nothing_float{T<:Base.IEEEFloat}
+    x::Union{T,Nothing}
 end
-T_union_nothing = Mooncake.Tangent{@NamedTuple{x::Union{Mooncake.NoTangent,Base.IEEEFloat}}}
+make_P_union_nothing() = P_union_nothing_float(Float32(1.0))
+struct P_union_nothing_array{T<:Base.IEEEFloat}
+    x::Union{Array{T},Nothing}
```
MilesCranmer:
```suggestion
    x::Union{Array{T,1},Nothing}
```

MilesCranmer:
Actually, I think it's fine as-is. Because that unbounded form is never actually matched!

```julia
julia> struct Foo end

julia> f(::Type{Union{Foo,T}}) where {T<:Base.IEEEFloat} = T

julia> f(Foo)  # If we hadn't defined another method - we WOULD see the unbounded form
Union{}

julia> f(::Type{Foo}) = Foo
f (generic function with 2 methods)

julia> f(Foo)
Foo

julia> f(Union{Foo,Float32})
Float32

julia> f(Union{Foo})
Foo
```

So just need to skip that single Aqua error.

I think it is preferable to match all `Array{T,N}` than avoid the appearance of an unbounded (but not real, due to method dispatch) method.

yebai:
Should be okay now -- Aqua has been annoying (though useful)!

yebai:
@MilesCranmer, I have invited you to this organisation so you can make new releases using @JuliaRegistrator.

ViralBShah:
@oscardssmith

src/fwds_rvs_data.jl
```diff
@@ -861,12 +861,16 @@ tangent type. This method must be equivalent to `tangent_type(_typeof(primal))`.
...
+for T in [Float16, Float32, Float64]
+    @eval @foldable tangent_type(::Type{NoFData}, ::Type{Union{NoRData,$T}}) = Union{
+        NoTangent,tangent_type($T)
+    }
+    for N in 0:5  # Just go up to N=5 until general solution to https://github.com/chalk-lab/Mooncake.jl/pull/620 available
+        @eval @foldable tangent_type(::Type{Union{NoFData,Array{$T,$N}}}, ::Type{NoRData}) = Union{
+            NoTangent,tangent_type(Array{$T,$N})
+        }
+    end
+end
```
nsajko, line 873: Pretty sure it would be nicer to the compiler not to eval three duplicate methods. Something like this, for example:

```suggestion
let f(@nospecialize x::Type) = Type{Union{NoRData, x}},
    f16 = f(Float16),
    f32 = f(Float32),
    f64 = f(Float64)
    global tangent_type
    @foldable function tangent_type(::Type{NoFData}, u::Union{f16, f32, f64})
        if u === f16
            t = Float16
        elseif u === f32
            t = Float32
        elseif u === f64
            t = Float64
        end
        Union{NoTangent, tangent_type(t)}
    end
end

for T in [Float16, Float32, Float64]
    for N in 0:5  # Just go up to N=5 until general solution to https://github.com/chalk-lab/Mooncake.jl/pull/620 available
        @eval @foldable tangent_type(::Type{Union{NoFData,Array{$T,$N}}}, ::Type{NoRData}) = Union{
            NoTangent,tangent_type(Array{$T,$N})
        }
    end
end
```

nsajko: I will apply the same deduplication to the `for N in 0:5` loop, if you wish.

MilesCranmer: I think we could just use the unbounded form right? It's only an issue if there is ambiguity. But since `Type{NoFData}, Type{NoRData}` are defined, it never gets matched.

```julia
julia> struct Foo end

julia> foo(::Type{Union{Foo,B}}) where {B<:AbstractFloat} = B
foo (generic function with 1 method)

julia> foo(Foo)  # Only issue here
Union{}

julia> foo(::Type{Foo}) = Foo
foo (generic function with 2 methods)

julia> foo(Foo)
Foo

julia> foo(Union{Foo,Float32})
Float32
```

which would let us match to any `N` for the arrays rather than only up to 5.

MilesCranmer, line 873:
```suggestion
tangent_type(::Type{NoFData}, ::Type{Union{NoRData,T}}) where {T<:Base.IEEEFloat} = Union{
    NoTangent,tangent_type(T)
}

tangent_type(::Type{Union{NoFData,Array{T,N}}}, ::Type{NoRData}) where {T<:Base.IEEEFloat,N} = Union{
    NoTangent,tangent_type(Array{T,N})
}
```

since `NoFData` and `NoRData` already have a method, the unbounded method issue never actually occurs. So just need to somehow inform Aqua that this particular instance is not an issue.

yebai: When `T = Union{}`, `Array{T,N}` would be problematic. But since this case is never called, as you noticed, it might be okay.

yebai:
Please feel free to take over this PR or replace it with a new one!

Issue #590, opened 2025-05-30
Can't differentiate a function that copies a Dict


mhauru:
```julia
julia> h(x, y) = (z = copy(y); sum(x))
h (generic function with 1 method)

julia> DifferentiationInterface.gradient(h, AutoMooncake(; config=nothing), [0.1], Constant(Dict()))
ERROR: No rrule!! available for foreigncall with primal argument types Tuple{Val{:jl_genericmemory_copy}, Val{Ref{Memory{UInt8}}}, Tuple{Val{Any}}, Val{0}, Val{:ccall}, Memory{UInt8}}. This problem has most likely arisen because there is a ccall somewhere in the function you are trying to differentiate, for which an rrule!! has not been explicitly written.You have three options: write an rrule!! for this foreigncall, write an rrule!! for a Julia function that calls this foreigncall, or re-write your code to avoid this foreigncall entirely. If you believe that this error has arisen for some other reason than the above, or the above does not help you to workaround this problem, please open an issue.
Stacktrace:
 [1] rrule!!(::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…})
   @ Mooncake ~/.julia/packages/Mooncake/Lwrnz/src/rrules/foreigncall.jl:12
 [2] h
   @ ./REPL[59]:1 [inlined]
 [3] (::Tuple{Mooncake.Stack{…}, Base.RefValue{…}, Mooncake.Stack{…}})(none::Mooncake.CoDual{typeof(h), Mooncake.NoFData}, none::Mooncake.CoDual{Vector{…}, Vector{…}}, none::Mooncake.CoDual{Dict{…}, Mooncake.MutableTangent{…}})
   @ Base.Experimental ./<missing>:0
 [4] (::Mooncake.DerivedRule{…})(::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…})
   @ Mooncake ~/.julia/packages/Mooncake/Lwrnz/src/interpreter/s2s_reverse_mode_ad.jl:966
 [5] prepare_gradient_cache(::Function, ::Vararg{Any}; kwargs::@Kwargs{debug_mode::Bool, silence_debug_messages::Bool})
[...]

PR #621, opened 2025-06-26
Add rule for copy with Dict

ErikQQY:
Fix: #590


src/rrules/array_legacy.jl
```diff
@@ -379,6 +380,16 @@ function rrule!!(::CoDual{typeof(copy)}, a::CoDual{<:Array})
     end
     return y, copy_pullback!!
 end
+function rrule!!(::CoDual{typeof(copy)}, a::CoDual{<:Dict})
+    dx = tangent(a)
+    dy = deepcopy(dx)
```
willtebbutt: I think `deepcopy`ing here will most likely lead to incorrect results if the fields of the `Dict` are not bits types. `deepcopy` will produce independent copies of the tangents of each field, while `copy` will not.

Sadly, I think the correct solution here is somewhat verbose -- I think you need to manually go in, take copies of the `slots`, `keys`, and `vals` fields, and manually construct a new `MutableTangent` of the appropriate type.

It would be good to add a (failing with the current implementation) test, so that we're confident that the semantics have been respected. The following `Dict` might be a good candidate:
```julia
Dict{Any,Any}("A"=>[5.0], [3.0] => 5.0)
```
because it involves non bits types as both keys and values, and makes use of abstract element types (this is always a good general case to cover).

willtebbutt (commented):
Thanks for this PR -- I think it's almost there, just requires a couple of small tweaks.

src/rrules/array_legacy.jl
ErikQQY: I found out we can directly construct a new `MutableTangent` with the original `CoDual.tangent.fields`, is this the correct ways of doing this?

willtebbutt: In this instance, it is :)

src/rrules/array_legacy.jl
```diff
@@ -379,6 +380,16 @@ function rrule!!(::CoDual{typeof(copy)}, a::CoDual{<:Array})
     end
     return y, copy_pullback!!
 end
+function rrule!!(::CoDual{typeof(copy)}, a::CoDual{<:Dict})
+    dx = tangent(a)
+    dy = MutableTangent(dx.fields)
```
willtebbutt: You'll need to take a copy of the fields :) Something like `map(copy, dx.fields)` ought to suffice, or simply writing it out.

src/rrules/foreigncall.jl
```diff
@@ -156,6 +156,24 @@ function rrule!!(
...
+        ::CoDual,
+        ::CoDual{Tuple{Val{Any}}},
+        ::CoDual{Val{0}},
+        ::CoDual{Val{:ccall}},
+        x::CoDual{<:Memory},
+    )
+        y = CoDual(
+            ccall(:jl_genericmemory_copy, Ref{Memory}, (Any,), primal(x)),
+            ccall(:jl_genericmemory_copy, Ref{Memory}, (Any,), tangent(x)),
+        )
```
willtebbutt:
```suggestion
        y = CoDual(primal(x), tangent(x))
```
It's probably fine to simplify this to simply `copy`ing the memory, since the implementation of `copy` is a one-liner which calls `:jl_genericmemory_copy`.

src/rrules/array_legacy.jl
ErikQQY: I see, IIUC to construct a tangent, we need all the informations in fields, but structure like `NoTangent()` doesn't support `copy` right?

```julia
julia> copy(Mooncake.NoTangent())
ERROR: MethodError: no method matching copy(::Mooncake.NoTangent)
The function `copy` exists, but no method is defined for this combination of argument types.
```

willtebbutt: Ohh good point. I'd suggest manually writing out the tangent tuple. I guess it would be something like this
```julia
t = dx.fields
new_fields = typeof(t)(copy(t.slots), copy(t.keys), copy(t.vals), tuple_fill(NoTangent(), Val(5))...)
```
(I've not tested it locally, but I think it should do the right thing)

ErikQQY: Thanks, this did work

src/rrules/array_legacy.jl
```diff
@@ -414,6 +429,7 @@ function generate_hand_written_rrule!!_test_cases(rng_ctor, ::Val{:array_legacy}
         (true, :stability, nothing, Array{Float64,4}, undef, (2, 3, 4, 5)),
         (true, :stability, nothing, Array{Float64,5}, undef, (2, 3, 4, 5, 6)),
         (false, :stability, nothing, copy, randn(5, 4)),
+        (false, :stability, nothing, copy, Dict{Any,Any}("A"=>[5.0], [3.0] => 5.0)),
```
willtebbutt: I wouldn't expect type stability here, so I think you just need to change this to 
```suggestion
        (false, :none, nothing, copy, Dict{Any,Any}("A"=>[5.0], [3.0] => 5.0)),
```
in order to get CI to pass. Might make sense to also add a test in which the field types are concrete to, and the `:stability` flag is set, to ensure that we have type stability in that situation.

willtebbutt:
Modulo the above comment, I think this just needs a patch version bump in order to merge.

ErikQQY:
I am a little confused, these rules are about `copy` with Dict, why CI complain numerical data-equal issues?

willtebbutt:
As part of the tests for rules, we check that the primal value that they produce is the same value as would be produced by running the function. In this case, it looks like this is going wrong.

These checks are implemented via `has_equal_data`, and `has_equal_data_up_to_undefs`. They recurse into the fields of objects, and check that they're all equal. This makes sense for many types, but for `Dict`s we probably only want to check that the keys / values are equal.

Could you please try adding this method:
```julia
function has_equal_data_internal(
    x::T, y::T, equal_undefs::Bool, d::Dict{Tuple{UInt,UInt},Bool}
) where {T<:Dict}
    f(x, y) = has_equal_data_internal(x, y, equal_undefs, d)
    return length(x) == length(y) &&
        all(map(f, keys(x), keys(y))) &&
        all(map(f, values(x), values(y)))
end
```
at this location:
https://github.com/chalk-lab/Mooncake.jl/blob/9122368a8b25f3ccdc4c369c6227596731d11f1b/src/test_utils.jl#L302

It should make everything pass.

ErikQQY:
Ohh I see, when dealing with structure like Dict, we should only check if the internal data are equal. Thanks for the detailed explanation!

ErikQQY:
I don’t have merge rights, feel free to merge this PR. Thanks again for the patient guidance!

willtebbutt:
My pleasure -- thanks for the PR!

PR #624, opened 2025-06-26
Generated function based `tangent_type` for `Union{Nothing, IEEEFloat, Array{<:IEEEFloat}`. 

yebai:


yebai:
@MilesCranmer, ready for a look.

src/fwds_rvs_data.jl
```diff
@@ -861,14 +861,28 @@ tangent type. This method must be equivalent to `tangent_type(_typeof(primal))`.
...
-            NoTangent,tangent_type(Array{$T,$N})
-        }
+@foldable function tangent_type(
+    ::Type{NoFData}, ::Type{R}
+) where {R<:Union{NoRData,T} where {T<:Base.IEEEFloat}}
+    if R == NoRData || R == Union{NoRData} || R == Union{}
+        return NoTangent
+    elseif R isa Union # R <: Union{NoRData,<:Base.IEEEFloat}
+        T = R.a isa Base.IEEEFloat ? R.a : R.b
+        return Union{NoTangent,tangent_type(T)}
```
MilesCranmer: More readable if just `Union{tangent_type(R.a), tangent_type(R.b)}`. Is there a reason to not do this?

src/fwds_rvs_data.jl
```diff
@@ -861,14 +861,28 @@ tangent type. This method must be equivalent to `tangent_type(_typeof(primal))`.
...
-    @eval @foldable tangent_type(::Type{NoFData}, ::Type{Union{NoRData,$T}}) = Union{
-        NoTangent,tangent_type($T)
-    }
-    for N in 0:5  # Just go up to N=5 until general solution to https://github.com/chalk-lab/Mooncake.jl/pull/620 available
-        @eval @foldable tangent_type(::Type{Union{NoFData,Array{$T,$N}}}, ::Type{NoRData}) = Union{
-            NoTangent,tangent_type(Array{$T,$N})
-        }
+@foldable function tangent_type(
+    ::Type{NoFData}, ::Type{R}
+) where {R<:Union{NoRData,T} where {T<:Base.IEEEFloat}}
```
MilesCranmer, line 866: Instead of having explicit `Base.IEEEFloat` throughout the library maybe you could define something like

```julia
const LeafTypes = Base.IEEEFloat
```

But you would need to reference this throughout the library.

I feel it's a bit easier to read than explicitly referencing specific float types. Also if there's ever a need to include some sort of `Float8` it would be really easy to add it in.

src/fwds_rvs_data.jl
yebai: Good catch. It could be further simplified: `tangent_type(R)` in fact.

src/fwds_rvs_data.jl
yebai: I think it would be better to do it another time. I have spent too much time on this issue -- I thought it would be a 30-minute task.

MilesCranmer: Yes yes of course. This is a more general comment for the library, not specific to this issue

PR #625, opened 2025-06-27
DispatchDoctor.jl integration

MilesCranmer:
Just to help catch more type instabilities.

As an FYI:
- The `@stable` macro essentially wraps every function in a code block with something analogous to `Test.@inferred`.
- The `@unstable` macro disables it for known type unstable functions.
- This behavior can be disabled, turning `@stable` into a no-op, with the `default_mode = "disable"` parameter, which I have done.
- At testing time, one would use Preferences.jl to set `dispatch_doctor_mode = "error"` or `dispatch_doctor_mode = "warn"` which will turn it on. (This setting goes into `test/Project.toml`)

There will be a lot of warnings initially. Some of these are (hopefully) real. Marking code blocks `@unstable` or using per-test `allow_unstable() do ... end` is how to get around these.


MilesCranmer:
Needs https://github.com/MilesCranmer/DispatchDoctor.jl/pull/75 first apparently

MilesCranmer:
Any ideas where this is coming from?

```julia
┌ Warning: DispatchDoctor.TypeInstabilityWarning: Instability detected in `_map` defined at 
/home/runner/work/Mooncake.jl/Mooncake.jl/src/utils.jl:151 with arguments `(Mooncake.var"#133#135"{IdDict{Any, Any}}, 
@NamedTuple{x1::Float64, x2::Float64, x3::Float64, x4::Float64, x5::Float64, x6::Float64, x7::Float64, x8::Float64, x9::Float64, x10::Float64, x11::Float64, x12::Float64, x13::Float64, x14::Float64, x15::Float64, x16::Float64, x17::Float64, x18::Float64, x19::Float64, x20::Float64, x21::Float64, x22::Float64, x23::Float64, x24::Float64, x25::Float64, x26::Float64, x27::Float64, x28::Float64, x29::Float64, x30::Float64, x31::Float64, x32::Float64, x33::Float64}, @NamedTuple{x1::Float64, x2::Float64, x3::Float64, x4::Float64, x5::Float64, x6::Float64, x7::Float64, x8::Float64, x9::Float64, x10::Float64, x11::Float64, x12::Float64, x13::Float64, x14::Float64, x15::Float64, x16::Float64, x17::Float64, x18::Float64, x19::Float64, x20::Float64, x21::Float64, x22::Float64, x23::Float64, x24::Float64, x25::Float64, x26::Float64, x27::Float64, x28::Float64, x29::Float64, x30::Float64, x31::Float64, x32::Float64, x33::Float64})` and parameters `(:F => Mooncake.var"#133#135"{IdDict{Any, Any}}, :N => 2)`. 
Inferred to be `NamedTuple{(:x1, :x2, :x3, :x4, :x5, :x6, :x7, :x8, :x9, :x10, :x11, :x12, :x13, :x14, :x15, :x16, :x17, :x18, :x19, :x20, :x21, :x22, :x23, :x24, :x25, :x26, :x27, :x28, :x29, :x30, :x31, :x32, :x33)}`, which is not a concrete type.
```

I'd like to just mark it `allow_instability` since I think various internal Julia utilities give up inference with this size of tuple

MilesCranmer:
I think a lot of the other instabilities are real. I think it's too much to fix all at once, so let me know how you'd like to proceed @willtebbutt @yebai.

An example, in `_add_to_primal_internal`, which does have a type instability:

```julia
    fields = map(fieldnames(P)) do f
        tf = getfield(t.fields, f)
        isdefined(p, f) &&
            is_init(tf) &&
            return _add_to_primal_internal(c, getfield(p, f), val(tf), unsafe)
        !isdefined(p, f) && !is_init(tf) && return FieldUndefined()
        throw(error("unable to handle undefined-ness"))
    end
```

this causes a type instability as `f` in the closure ends up being a symbol. Therefore `fields` is a generic `Tuple{Vararg{Any,nf}}` and it can't specialize.

I think the cleanest way to fix things like this is use a generated function with `Base.Cartesian.@ntuple`. An example is given on https://github.com/chalk-lab/Mooncake.jl/pull/634.

This means you get a specific branch with the number of initialised fields known at compile time. And you can define this with the `Base.Cartesian.@ntuple`. This macro will generate a literal `i` -> `3`, which means you will get explicit `getfield(t.fields, 3)` in one branch of the code, so the compiler can actually get the precise types. (I think something similar is done internally a couple of times, such as the `tangent_type_exprs` variable for tuple types.)

I've ported a few already including `_dot_internal` (cherry picked to #630), `_diff_internal`, `_copy_output` (cherry picked to #634) , and `build_tangent`. I think if I keep adding them all here it will end up being too much at once (and is a lot of work) so is probably best left to gradually fix these in the future. 

Right now I have DispatchDoctor set up in "warn" mode which means it won't fail the CI. I think this is probably best for now, as otherwise it would be too much work at once to fix everything.

willtebbutt:
In my view, it's fine to leave it like this. Once this gets merged, could you please open an issue about it so that we don't forget though?

src/tangents.jl
```diff
@@ -749,7 +776,7 @@ Implementation for [`_scale`](@ref). Use `c` to handle circular references and a
 _scale_internal(::MaybeCache, ::Float64, ::NoTangent) = NoTangent()
 _scale_internal(::MaybeCache, a::Float64, t::T) where {T<:IEEEFloat} = T(a * t)
 function _scale_internal(c::MaybeCache, a::Float64, t::Union{Tuple,NamedTuple})
-    return map(t -> _scale_internal(c, a, t), t)
+    return map(ti -> _scale_internal(c, a, ti)::typeof(ti), t)::typeof(t)
 end
```
MilesCranmer: For some reason this assertion is failing with the error

```julia
  TypeError: in typeassert, expected @NamedTuple{a, b::PossiblyUninitTangent{Vector{Float64}}}, got a value of type @NamedTuple{a::Float64, b::PossiblyUninitTangent{Vector{Float64}}}
```

Is this expected? Should I remove the assertion?

MilesCranmer: Temporarily removing this assertion and marking _scale_internal unstable. But let me know otherwise

MilesCranmer:
Ready for review @willtebbutt @yebai. 

I have only enabled DD on a subset of `"basic"` and verified stability of all remaining functions. Unstable methods are now explicitly marked `@unstable` for future refactoring.

Note that the Aqua.jl error in `"quality"` is real - there is currently a circular dependency between DD and Mooncake. However, we will need to fix it _after_ the next release by moving DispatchDoctorMooncakeExt to here. In other words, that error will continue to error until the release. Hopefully that is okay with you.

MilesCranmer:
Actually I will just delete DispatchDoctorMooncakeExt and move it here. Should be much easier.

https://github.com/MilesCranmer/DispatchDoctor.jl/pull/77

MilesCranmer:
Also, I have set DispatchDoctor to error in the `basic-dd` test now. This means that it will catch regressions in type stability and fail the CI.

We can continue to expand the covered methods by adding more of basic to basic-dd.

MilesCranmer:
This should be good to go. @yebai @willtebbutt would you be up to review this?

This also forms a regression test against #634

src/rrules/performance_patches.jl
```diff
@@ -38,7 +38,9 @@ function rrule!!(
     return zero_fcodual(sum(abs2, x.x)), sum_abs2_pb!!
 end
 
-function generate_hand_written_rrule!!_test_cases(rng_ctor, ::Val{:performance_patches})
+@unstable @unstable function generate_hand_written_rrule!!_test_cases(
```
yebai: Is there any special meaning of double `@unstable`? 

```suggestion
@unstable function generate_hand_written_rrule!!_test_cases(
```

src/rrules/performance_patches.jl
```diff
@@ -60,4 +62,8 @@ function generate_hand_written_rrule!!_test_cases(rng_ctor, ::Val{:performance_p
     return test_cases, memory
 end
 
-generate_derived_rrule!!_test_cases(rng_ctor, ::Val{:performance_patches}) = Any[], Any[]
+@unstable @unstable function generate_derived_rrule!!_test_cases(
```
yebai:
```suggestion
@unstable function generate_derived_rrule!!_test_cases(
```

src/rrules/misc.jl
```diff
@@ -78,10 +78,12 @@ lgetfield(x, ::Val{f}) where {f} = getfield(x, f)
...
-@inline _get_fdata_field(_, t::Union{Tuple,NamedTuple}, f) = getfield(t, f)
-@inline _get_fdata_field(_, data::FData, f) = val(getfield(data.data, f))
-@inline _get_fdata_field(primal, ::NoFData, f) = uninit_fdata(getfield(primal, f))
-@inline _get_fdata_field(_, t::MutableTangent, f) = fdata(val(getfield(t.fields, f)))
+@unstable begin
+    @inline _get_fdata_field(_, t::Union{Tuple,NamedTuple}, f) = getfield(t, f)
+    @inline _get_fdata_field(_, data::FData, f) = val(getfield(data.data, f))
+    @inline _get_fdata_field(primal, ::NoFData, f) = uninit_fdata(getfield(primal, f))
+    @inline _get_fdata_field(_, t::MutableTangent, f) = fdata(val(getfield(t.fields, f)))
+end
```
yebai:
```suggestion
@unstable @inline _get_fdata_field(_, t::Union{Tuple,NamedTuple}, f) = getfield(t, f)
@unstable @inline _get_fdata_field(_, data::FData, f) = val(getfield(data.data, f))
@unstable @inline _get_fdata_field(primal, ::NoFData, f) = uninit_fdata(getfield(primal, f))
@unstable @inline _get_fdata_field(_, t::MutableTangent, f) = fdata(val(getfield(t.fields, f)))
```

src/test_utils.jl
```diff
@@ -160,13 +160,21 @@ using Mooncake:
...
 struct Shim end
 
-test_opt(x...) = test_opt_internal(Shim(), x...)
+const DD_ENABLED = let uuid = get_uuid(@__MODULE__)
+    mode = load_preference(uuid, "dispatch_doctor_mode")
+
+    mode ∉ (nothing, "disable")
+end
+
+test_opt(x...) = DD_ENABLED ? nothing : test_opt_internal(Shim(), x...)
```
yebai: Why disable JET when DD_ENABLED is false?

yebai (commented):
I finished reading half of this PR today and left a few questions and comments above. Will continue tomorrow. 

It appears that DispatchDoctor is fairly invasive (i.e., requiring extensive annotation of source code) compared to other tools like `AllocChecks` and `Aqua`. Is there any alternative approach for DispatchDoctor that is less invasive (e.g. work with SSAIR instead of AST)?

src/rrules/performance_patches.jl
MilesCranmer: Hm maybe this was from a merge or something.

src/test_utils.jl
MilesCranmer: When you run JET on code that actively has DispatchDoctor checking for type instabilities, there are a lot of "errors" that get raised but which aren't real errors. Something about their analysis seems to currently conflict with eachother. 

I do the same thing in my own packages and just disable DD whenever I'm running JET on it.

MilesCranmer:
> requiring extensive annotation of source code

Yes. A couple of notes on this:

(1) When disabled, the macros are no-ops: [code](https://github.com/MilesCranmer/DispatchDoctor.jl/blob/99e89c2cc5af31dc0bc7f3f957dcfb1ed816f71d/src/stabilization.jl#L50). So only if someone has explicitly set `preferences.Mooncake.dispatch_doctor_mode = "error"` will there by any code processing at all. Otherwise nothing will happen. The Preferences.jl library is maintained by core Julia devs and is robust about invalidating precompiled code, given changes to the preferences.

(2) While it is subjective, I actually view the explicit annotations as a plus rather than a drawback. Julia is very loose about type instabilities, to the point the user needs to explicitly run `Test.@inferred` in CI to verify there is no type instability. DispatchDoctor.jl flips this – type stability is _expected_, and any type unstable function needs to be explicitly declared `@unstable`. This is because there can be massive performance repercussions if you miss just one (and sometimes only discover it later). Usually people will get hit by some performance regression and then analyse deep call stacks to try to find it. I don't think this is a practical way to maintain performant code. It's better to detect type instabilities at the individual function level, and stabilise them before they leak into the main code.

It's similar (and actually inspired by) how in rust there is an `unsafe { }` keyword when doing memory unsafe things, which I think is a really nice design and encourages people to write safe code by default, and only do unsafe things if they absolutely need to.

(3) In any case, DispatchDoctor works quite differently than JET, but is most similar to AllocCheck.jl: you explicitly annotate it at a per-function level. However, whereas allocations can be measured from the top level, this is harder for type instabilities: 

- you actually want to permit _some_ type instabilities (like logging code that isn't inlined), and don't want to fail CI because of that. You only fail CI if that type instability creeps up to a function where it is unexpected. 
- DispatchDoctor.jl just uses `Core.Compiler.return_type` which is quite fast (I guess because it's already being computed), and since its inferred at compile time, the stability check can get completely removed:

    ```julia
    julia> @stable f(x) = x;
    
    julia> @code_llvm f(1)
    define i64 @julia_f_12055(i64 signext %"x::Int64") #0 {
    top:
      ret i64 %"x::Int64"
    }
    ```
    
    so it is much more practical to apply DD to every function in your library and run your full test suite, whereas with AllocCheck.jl this might be prohibitively expensive.


(Note that I don't recommend actually leaving the `@stable` on in libraries _without_ the `default_mode = "disable"`, because there are instances where the code does not get compiled away)

<details> <summary> (Full comparison) </summary>

```julia
julia> @stable f(x) = x

julia> @check_allocs g(x) = x
g (generic function with 1 method)

julia> @code_llvm f(1)
; Function Signature: f(Int64)
;  @ /Users/mcranmer/.julia/packages/DispatchDoctor/18wG3/src/stabilization.jl:309 within `f`
define i64 @julia_f_13622(i64 signext %"x::Int64") #0 {
top:
  ret i64 %"x::Int64"
}

julia> @code_llvm g(1)
; Function Signature: g(Int64)
;  @ REPL[5]:1 within `g`
define nonnull ptr @julia_g_14204(i64 signext %"x::Int64") #0 {
top:
  %jlcallframe1 = alloca [2 x ptr], align 8
  %gcframe2 = alloca [4 x ptr], align 16
  call void @llvm.memset.p0.i64(ptr align 16 %gcframe2, i8 0, i64 32, i1 true)
  %pgcstack = call ptr inttoptr (i64 6518463472 to ptr)(i64 261) #4
  store i64 8, ptr %gcframe2, align 16
  %task.gcstack = load ptr, ptr %pgcstack, align 8
  %frame.prev = getelementptr inbounds ptr, ptr %gcframe2, i64 1
  store ptr %task.gcstack, ptr %frame.prev, align 8
  store ptr %gcframe2, ptr %pgcstack, align 8
;  @ REPL[5]:138 within `g`
; ┌ @ /Users/mcranmer/.julia/packages/AllocCheck/gZlvT/src/compiler.jl:82 within `compile_callable`
   %0 = call nonnull ptr @"j_#compile_callable#22_14207"(i8 zeroext 1)
   %gc_slot_addr_1 = getelementptr inbounds ptr, ptr %gcframe2, i64 3
   store ptr %0, ptr %gc_slot_addr_1, align 8
; └
;  @ REPL[5]:139 within `g`
; ┌ @ Base.jl:49 within `getproperty`
   store ptr %0, ptr %jlcallframe1, align 8
   %1 = getelementptr inbounds ptr, ptr %jlcallframe1, i64 1
   store ptr @"jl_sym#analysis#14209.jit", ptr %1, align 8
   %jl_f_getfield_ret = call nonnull ptr @jl_f_getfield(ptr null, ptr nonnull %jlcallframe1, i32 2)
   %gc_slot_addr_0 = getelementptr inbounds ptr, ptr %gcframe2, i64 2
   store ptr %jl_f_getfield_ret, ptr %gc_slot_addr_0, align 16
; └
  store ptr %jl_f_getfield_ret, ptr %jlcallframe1, align 8
  %2 = call nonnull ptr @ijl_apply_generic(ptr nonnull @"jl_global#14211.jit", ptr nonnull %jlcallframe1, i32 1)
  store ptr %2, ptr %gc_slot_addr_0, align 16
  store ptr %2, ptr %jlcallframe1, align 8
  store ptr @"jl_global#14214.jit", ptr %1, align 8
  %3 = call nonnull ptr @ijl_apply_generic(ptr nonnull @"jl_global#14213.jit", ptr nonnull %jlcallframe1, i32 2)
  %.tag_addr = getelementptr inbounds i64, ptr %3, i64 -1
  %.tag = load atomic i64, ptr %.tag_addr unordered, align 8
  %4 = and i64 %.tag, -16
  %5 = inttoptr i64 %4 to ptr
  %exactly_isa = icmp eq ptr %5, inttoptr (i64 192 to ptr)
  br i1 %exactly_isa, label %pass, label %fail

L6:                                               ; preds = %pass
; ┌ @ /Users/mcranmer/.julia/packages/AllocCheck/gZlvT/src/macro.jl:140 within `macro expansion`
   store ptr %0, ptr %jlcallframe1, align 8
   store ptr @"jl_sym#analysis#14209.jit", ptr %1, align 8
   %6 = call nonnull ptr @ijl_apply_generic(ptr nonnull @"jl_global#14216.jit", ptr nonnull %jlcallframe1, i32 2)
   store ptr %6, ptr %gc_slot_addr_0, align 16
   store ptr %6, ptr %jlcallframe1, align 8
   %7 = call nonnull ptr @ijl_apply_generic(ptr nonnull @"+AllocCheck.AllocCheckFailure#14218.jit", ptr nonnull %jlcallframe1, i32 1)
   call void @ijl_throw(ptr nonnull %7)
   unreachable

L10:                                              ; preds = %pass
; └
;  @ REPL[5]:142 within `g`
  %box_Int64 = call nonnull align 8 dereferenceable(8) ptr @ijl_box_int64(i64 signext %"x::Int64") #6
  store ptr %box_Int64, ptr %gc_slot_addr_0, align 16
  store ptr %box_Int64, ptr %jlcallframe1, align 8
  %8 = call nonnull ptr @ijl_apply_generic(ptr nonnull %0, ptr nonnull %jlcallframe1, i32 1)
  %frame.prev15 = load ptr, ptr %frame.prev, align 8
  store ptr %frame.prev15, ptr %pgcstack, align 8
  ret ptr %8

fail:                                             ; preds = %top
;  @ REPL[5]:139 within `g`
  %9 = load ptr, ptr getelementptr (i8, ptr @jl_small_typeof, i64 192), align 8
  store ptr %9, ptr %gc_slot_addr_0, align 16
  call void @ijl_type_error(ptr nonnull @"_j_str_if#1", ptr nonnull %9, ptr nonnull %3)
  unreachable

pass:                                             ; preds = %top
  %jl_false = load ptr, ptr @jl_false, align 8
  %10 = icmp eq ptr %3, %jl_false
  br i1 %10, label %L10, label %L6
}
```

</details>

MilesCranmer:
@yebai I realised that the `generate_derived_rrule!!_test_cases` and `generate_hand_written_rrule!!_test_cases` functions are only called from a single location. I've moved the `unstable` marker to that callsite, using `allow_unstable(f)`, rather than every individual source. As a result, the PR is significantly smaller!

MilesCranmer:
P.S., here is roughly how DispatchDoctor works. The following code:

```julia
@stable function f(x)
    return Val(x)
end
```

gets transformed into something similar to

```julia
function f_inner(x)
    return Val(x)
end

function f(x)
    T = Core.Compiler.return_type(f_inner, typeof((x,)))
    !Base.isconcretetype(T) && error()
    return f_inner(x)
end
```

When you put `@stable` on a block with multiple functions, it applies the same transform to _all_ functions in that block – walking through `include()`, `module`, etc. And any code block with an `@unstable` is skipped. There are various other functions that get skipped automatically, such as generated functions, anything with `@nospecialize`, closure functions, and some other cases. 

And these transforms only happen when enabled. But `@stable default_mode = "disable"` would by default result in no transforms taking place.

MilesCranmer:
Any other issues with this PR I can work on @yebai?

I should emphasise again that the current behavior is a no-op; `@stable` is just an identity macro. There is no effect on Mooncake outside of the specific test that enables it.

If you would feel more comfortable with it, we can also remove the `@stable` from the codebase and have the specific DispatchDoctor test run with:

```bash
sed -i 's|include("utils.jl")|@stable default_mode = "disable" default_union_limit = 2 begin\ninclude("utils.jl")|g' src/Mooncake.jl
```

but even without this, there won't be any effects outside of the specific test.

yebai:
I'll try to finish my review today. I was busy most of this week.

src/tangents.jl
```diff
@@ -123,18 +128,40 @@ end
...
+        TP, fields, ftypes, Val(fnames), Val(length(_ftypes))
+    )::TP
+end
+@generated function _build_tangent_cartesian(
+    ::Type{TP}, fields::Tuple{Vararg{Any,N}}, ::Type{ftypes}, ::Val{fnames}, ::Val{nfields}
+) where {TP,N,ftypes,fnames,nfields}
+    quote
+        full_fields = Base.Cartesian.@ntuple(
+            $nfields, n -> let
+                tt = ftypes.types[n]
```
yebai, line 146: Should this be? 

```suggestion
                tt = ftypes[n]
```

test/rrules/dispatch_doctor.jl
```diff
@@ -0,0 +1,68 @@
+using DispatchDoctor: @stable, @unstable, allow_unstable, TypeInstabilityError
+using Test
+using Mooncake
```
yebai:
```suggestion
```

test/rrules/dispatch_doctor.jl
```diff
@@ -0,0 +1,68 @@
...
+    @testset "TypeInstabilityError without allow_unstable" begin
+        x = 2.0
+
+        @test_throws TypeInstabilityError type_unstable_square(x)
+
+        rule = Mooncake.build_rrule(Tuple{typeof(type_unstable_square),Float64})
+        @test rule !== nothing
+
+        # However, note that type_unstable_square2 is _not_ marked unstable.
+        # This should be preserved through the rrule.
```
yebai: This doesn't seem to test `rrule`?

src/test_utils.jl
```diff
@@ -753,6 +765,40 @@ function run_rrule!!_test_cases(rng_ctor, v::Val)
     return run_derived_rrule!!_test_cases(rng_ctor, v)
 end
 
+"""
```
yebai: Should these functions be moved to DD?

src/test_utils.jl
yebai: > When you run JET on code that actively has DispatchDoctor checking for type instabilities, there are a lot of "errors" that get raised but which aren't real errors. Something about their analysis seems to currently conflict with eachother.

Can you add some comments in the code to explain this?

src/tangents.jl
```diff
@@ -997,30 +1025,58 @@ function _diff_internal(c::MaybeCache, p::P, q::P) where {P<:SimpleVector}
...
+            c[(p, q)] = _t
+            _t
         else
-            throw(error("Unhandleable undefinedness"))
+            nothing
         end
+        Base.Cartesian.@nif(
+            $(nfield + 1),
+            n -> let
+                defined_p = isdefined(p, n)
```
yebai, line 1051: This new implementation implicitly assumes the ordering of fields is the same, whereas the current code doesn't. So, this might not be correct in all cases.

yebai (changes requested):
I have completed a thorough review; it took a lot of time. 

I left a few questions and suggestions below. 

I think I see the value proposition here, but a bit on the edge whether this is helpful, since the rules determine the speed of AD, while lots of type stability checks aim at the compiler itself.

src/tangents.jl
MilesCranmer: ftypes is Tuple{_ftypes...} so the current form is correct. I did try passing a Val of tuple of types but it isn't actually valid.

src/tangents.jl
MilesCranmer: They are both `::P` though right? Or maybe you could clarify what you mean as I don't understand the exact issue

src/test_utils.jl
MilesCranmer: This utility hasn't been needed in other libraries yet. But maybe eventually this could be put there as an option to `allow_unstable` or something?

For the moment, these are specific to Mooncake, to weaken the type stability checks, because a lot of the tests operate on unstable types to begin with and the type instabilities "make sense". But normally you would want to still throw an error for such circumstances.

test/rrules/dispatch_doctor.jl
MilesCranmer: We test it below with `test_rule`. This was just a "quick fail" test to make sure it builds at all. But I can delete it as it's redundant. Wdyt?

src/test_utils.jl
MilesCranmer: Maybe we could put it directly in the dispatch doctor test group though? It's tricky as we do need to inject this behaviour.

src/tangents.jl
MilesCranmer: For example:
```julia
julia> Val((Float32,))
ERROR: TypeError: in Type, in parameter, expected Type, got a value of type Tuple{DataType}
Stacktrace:
 [1] Val(x::Tuple{DataType})
   @ Base ./essentials.jl:1037
 [2] top-level scope
   @ REPL[1]:1
```

hence why it is passed as

```julia
julia> Type{Tuple{Float32}}
Type{Tuple{Float32}}
```

test/rrules/dispatch_doctor.jl
MilesCranmer: Ill proceed with deleting this

MilesCranmer:
Thanks a lot for the review @yebai.

For the new review just a tip that you can look at the new changes by clicking this:
image

MilesCranmer:
> since the rules determine the speed of AD, while lots of type stability checks aim at the compiler itself.

So the `@stable` check also covers the rrules and (some) pullbacks although there weren't any instabilities detected (which is good!). Although I suppose it was only the basic tests..

Although I see that a lot of the pullbacks are closure functions - those are not automatically covered. So in the future the pullbacks could either be turned into callable structs (preferred - like `ExpPullback`) or the `@stable` could be applied internally on the rrules. Maybe an internal `@stable` macro could carry the same options as the current toplevel call, similar to `@foldable`.

src/tangents.jl
```diff
@@ -981,12 +1008,13 @@ aliasing. If `c` is a `NoCache` then assume no circular references or aliasing i
 `p` or `q`.
 """
 function _diff_internal(c::MaybeCache, p::P, q::P) where {P}
-    tangent_type(P) === NoTangent && return NoTangent()
+    TP = tangent_type(P)
```
yebai, line 1012:
```suggestion
    @assert typeof(p) == typeof(q) # this function implicitly assumes p and q have identical type
    TP = tangent_type(P)
```

src/tangents.jl
yebai: I overlooked that. We should insert a check to prevent future changes from accidentally breaking the assumption.

src/test_utils.jl
```diff
@@ -160,13 +160,26 @@ using Mooncake:
...
+using DispatchDoctor: type_instability, allow_unstable as _allow_unstable
 
 struct Shim end
 
-test_opt(x...) = test_opt_internal(Shim(), x...)
+const DD_ENABLED = let uuid = get_uuid(@__MODULE__)
+    mode = load_preference(uuid, "dispatch_doctor_mode")
+
+    mode ∉ (nothing, "disable")
+end
```
yebai: I am concerned about 

1. DD might quietly switch off the JET and AllocChecks tests. 
2. The changes make this test-related code hard to read.  

Instead, for clarity, can we introduce 

```julia
struct JETTests end
struct DispatchDoctorTests end
struct AllocChecksTests end

test_opt(test_ctx::JETTests, x...) = test_opt_internal(Shim(), x...)
test_opt(test_ctx::DispatchDoctorTests, x...) =  nothing
check_allocs(test_ctx::DispatchDoctorTests, f, x...) = nothing
check_allocs(test_ctx, f, x...) = check_allocs_internal(f, x...)
```

Then, users must explicitly pass the `test_ctx` context to calls to `test_opt` and `check_allocs`.

src/test_utils.jl
yebai: They don't look like Mooncake-specific, other than Mooncake uses them.

src/test_utils.jl
MilesCranmer: I like that design, sure.

src/tangents.jl
MilesCranmer: Out of curiosity why isn't this already imposed by `p::P, q::P`? I would have thought that constrains such things. Maybe I have an incorrect mental model for signatures.

yebai: Yes, but future devs might (accidentally) change the type signature and introduce unexpected behaviour.

src/test_utils.jl
MilesCranmer: @yebai I just tried this. It makes the TestUtils implementation extremely complex, as you have to pass the context through many layers of functions, and track it. Keep in mind that the DispatchDoctor tests are running the same test code as usual.

I think this refactor would only be worth it if there were multiple additional interacting contexts.

MilesCranmer: Note that `DD_ENABLED` is set if and only if

```toml
[Mooncake]
dispatch_doctor_mode = "error"
```

is set in your LocalPreferences.toml file. 

So I think it would be hard to accidentally trigger this. Wdyt?

src/test_utils.jl
MilesCranmer: The specific-to-mooncake part is the fact that Mooncake has a massive amount of tests that are inherently type unstable. In other packages you would have only a handful of these, and use `allow_unstable() do; ...; end` at each individual test after manually confirming it's not an issue. But since Mooncake operates on general code, there are, as expected, tons of these. Using `allow_unstable` manually would be a nightmare. Note that things like `x -> x.a` when used on input `@NamedTuple{a::Any}` is technically an unstable function. In other packages you generally wouldn't want to filter such cases, and, instead, let the failure guide you to avoid using the type `@NamedTuple{a::Any}` in the first place.

So, the `skip_instability_check` is a Mooncake-specific approximate filter for these. I don't think this is appropriate to put in DispatchDoctor though, because I would discourage such filtering in other contexts. And it's hard to determine what a sensible general filter is. It's probably better left to specific packages to decide when to `allow_unstable` and when to not.

Let me know if this makes a bit more sense?

src/test_utils.jl
yebai: In that case, can we move DD tests into an integration test? The advantages of a separate integration test for DD are: 

1. It is possible to have a separate `Project.toml` file, thus making it easier to run DD tests locally. The current setup, relying on CI environment variables, makes it difficult. 
2. Move all DD-specific test code and overloads into the integration test, so they live in one place. 

This would remove the interaction between the current JET and AllocChecks tests with DD.

src/test_utils.jl
yebai: Sounds good to keep them here if we move all this code into a new integration test for DD, as suggested below.

src/test_utils.jl
MilesCranmer: Good idea. What should I do for the tests though? Should they be made to `include(joinpath(@__DIR__, "..", "utils.jl"))` and so forth? The tricky thing is that they would still need all the same imports as for `basic`.

yebai: > Should they be made to include(joinpath(@__DIR__, "..", "utils.jl")) and so forth? The tricky thing is that they would still need all the same imports as for basic.

It's okay to include these files from integration tests. The key part is a separate environment.

MilesCranmer: Makes sense!

Ok i have made the change.

MilesCranmer:
Ah, I think there's a really nice way to rewrite this. Give me a moment.

MilesCranmer:
@yebai How about this? I like it a lot.

src/test_utils.jl
```diff
@@ -160,13 +160,27 @@ using Mooncake:
...
+    test_hook(f, caller, ::Any...; kws...)
+
+A function which can alter the behavior of a given test. `f` is the function
+that is normally executed, `caller` is the function this is being called from,
+and `args` and `kws` are the arguments that were passed to `caller`.
+
+Normally, `f()` is executed. However, this can be overridden by an integration test
+defining custom behavior.
+"""
+test_hook(f, caller, args...; kws...) = f()
```
MilesCranmer, line 176: @yebai Basically, the DispatchDoctor.jl integration test will overload this `test_hook` to inject logic for skipping type instabilities.

MilesCranmer:
@yebai `rrules/misc` is failing, but it's also failing on main. Should I still merge?

yebai:
yes

PR #628, opened 2025-06-27
Fix type instability in `_dot_internal`

MilesCranmer:
Identified in #625

Also reduced some allocations from `sum(map(...))` to `sum(...)`

### DIFF
--- a/src/rrules/memory.jl
+++ b/src/rrules/memory.jl
@@ -190,9 +190,9 @@ for A in (Array, Memory)
         key = (t, s)
         haskey(c, key) && return c[key]::Float64
         c[key] = 0.0
-        bitstype = isbitstype(eltype(T))
-        return sum(eachindex(t); init=0.0) do i
-            if bitstype || (isassigned(t, i) && isassigned(s, i))
+        bitstype = Val(isbitstype(eltype(T)))
+        return sum(eachindex(t, s); init=0.0) do i
+            if bitstype isa Val{true} || (isassigned(t, i) && isassigned(s, i))
                 _dot_internal(c, t[i], s[i])::Float64
             else
                 0.0
--- a/src/rrules/memory.jl
+++ b/src/rrules/memory.jl
@@ -95,19 +95,6 @@ function _diff_internal(c::MaybeCache, p::Memory{P}, q::Memory{P}) where {P}
     return _map_if_assigned!((p, q) -> _diff_internal(c, p, q), t, p, q)
 end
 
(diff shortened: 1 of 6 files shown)


src/rrules/memory.jl
```diff
@@ -198,17 +185,20 @@ function _scale_internal(c::MaybeCache, a::Float64, t::T) where {T<:Array}
...
-        end;
-        init=0.0,
-    )
+for A in (Array, Memory)
+    @eval function _dot_internal(c::MaybeCache, t::T, s::T) where {T<:$A}
+        key = (t, s)
+        haskey(c, key) && return c[key]::Float64
+        c[key] = 0.0
+        bitstype = Val(isbitstype(eltype(T)))
+        return sum(eachindex(t, s); init=0.0) do i
```
MilesCranmer, line 194: (Note that `eachindex(t, s)` does the same `allequal(length)` check done by `_map`)

src/tangents.jl
```diff
@@ -788,17 +788,19 @@ or aliasing.
...
+    is_init(t) && is_init(s) && return _dot_internal(c, val(t), val(s))::Float64
     return 0.0
 end
 function _dot_internal(c::MaybeCache, t::T, s::T) where {T<:Union{Tangent,MutableTangent}}
     key = (t, s)
     haskey(c, key) && return c[key]::Float64
     c[key] = 0.0
-    return sum(_map((t, s) -> _dot_internal(c, t, s), t.fields, s.fields); init=0.0)
+    return sum(
+        _map((t, s) -> _dot_internal(c, t, s)::Float64, t.fields, s.fields); init=0.0
```
MilesCranmer, line 802: (This `_map` can be left as-is because it doesn't allocate)

yebai:
@MilesCranmer, thanks. Can you replace this PR with a new one based on a branch from this repository (instead of a fork)? That would allow benchmark CI to run.  It's helpful to see benchmarks for these changes.

MilesCranmer:
Is that limitation from the GitHub comment action? You can just use `pull_request_target` instead of `pull_request` (and enable pull request write permissions) and then it will run.

Better yet, use [AirspeedVelocity.jl](https://github.com/MilesCranmer/AirspeedVelocity.jl) to handle this in a safe way, because it automatically masks the GITHUB_TOKEN from user scripts. There’s a GitHub Summary option too. It should be out-of-the-box compatible since it just uses the same format as BenchmarkTool.jl

MilesCranmer:
Moved to #630

PR #630, opened 2025-06-28
Fix type instability in `_dot_internal`

MilesCranmer:
Identified in #625

Also reduced some allocations from `sum(map(...))` to `sum(...)`

(Re-do of #628)

### DIFF
--- a/src/fwds_rvs_data.jl
+++ b/src/fwds_rvs_data.jl
@@ -868,7 +868,7 @@ tangent type. This method must be equivalent to `tangent_type(_typeof(primal))`.
 end
 @foldable function tangent_type(
     ::Type{F}, ::Type{NoRData}
-) where {F<:Union{NoFData,T} where {T<:Array{<:Any,N} where N}}
+) where {F<:Union{NoFData,T} where {T<:Array{<:Any,N} where {N}}}
     return tangent_type(F)
 end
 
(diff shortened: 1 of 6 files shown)


willtebbutt:
Very much in favour of this, but would like to add a test to ensure that it doesn't regress before merging.

I wonder if we could just add an inference test as part of the standardised test suite? I.e. putting a call to `__dot` behind an `@inferred` call somewhere in [this function](https://github.com/chalk-lab/Mooncake.jl/blob/9122368a8b25f3ccdc4c369c6227596731d11f1b/src/test_utils.jl#L832).

willtebbutt:
Or does https://github.com/chalk-lab/Mooncake.jl/pull/625 cover the testing? (I've not had a chance to go through that PR, so I'm not entirely sure of its scope, apologies!)

MilesCranmer:
Yes #625 would cover this. Currently it is set to `dispatch_doctor_mode="warn"` to help print more debugging info but in my packages I test with `dispatch_doctor_mode="error"` to prevent such regressions.

willtebbutt (approved):
Cool. In that case, I'm happy to merge this. Could you please bump the patch version (since this PR involves changes in `src`, it gets a version bump, even though it pertains to testing functionality).

MilesCranmer:
Done

Issue #608, opened 2025-06-24
Add a regression test for `build_fdata` world age issue


yebai:
See, https://github.com/chalk-lab/Mooncake.jl/pull/606

sunxd3:
Form @yebai on slack

## World Age Issue with `build_fdata` and Recursive Types

You can read this [note](https://chalk-lab.github.io/Mooncake.jl/stable/developer_documentation/misc_internals_notes/#tangent_type-and-friends) for a good explanation of the problem. The issue with `build_fdata` is identical. You can create a regression test by wrapping this [example](https://chalk-lab.github.io/Mooncake.jl/stable/developer_documentation/custom_tangent_type/#3.-Defining-a-Custom-Tangent-Type-for-Recursion) in another parent type, e.g.

```julia
mutable struct A{TA}
    x::TA
    a::Union{A{TA},Nothing}

    A(x::TA) where {TA} = new{TA}(x, nothing)
    A(x::TA, child::A{TA}) where {TA} = new{TA}(x, child)
end

struct Wrapper{TW}; x::TW; end;

# Constructing a tangent type for `Wrapper` will trigger the issue in `build_fdata`
# so this line will break master, but run correctly after fixing the world age issue 
# for `build_fdata`. 

T = tangent_type(Wrapper{A{Float32}}) 

T(args...) # this doesn't see new `tangent_type` definitions for A{TA} due to world age issue
```

The issue is that the new constructor of tangent type `tangent_type(Wrapper{A{Float32}})` doesn't see new `tangent_type` definitions due to the world age issue. This has been neglected because we haven't encountered recursive types before.

PR #633, opened 2025-06-29
Add regression test

sunxd3:
Address https://github.com/chalk-lab/Mooncake.jl/issues/608

The tests should error on code before #594, for instance `6dfbe394`, but pass now.


test/rrules/build_fdata_world_age_regression.jl
```diff
@@ -0,0 +1,108 @@
+# Regression test for build_fdata world age issue (#606, #608)
+# Tests that @generated functions can see custom tangent_type definitions for recursive types
+
+using Mooncake
+using Test
```
yebai:
```suggestion
```

PR #634, opened 2025-06-30
refactor: type stable _copy_output

MilesCranmer:
Detected in the DispatchDoctor PR.

It's due to the various dynamic `for i in 1:nf; getfield(x, i); end`. To fix it I use `Base.Cartesian` in a generated function.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -320,14 +320,17 @@ _copy_output(x::SimpleVector) = Core.svec([map(_copy_output, x_sub) for x_sub in
 # Array, Memory
 function _copy_output(x::P) where {P<:_BuiltinArrays}
     temp = P(undef, size(x)...)
+    Tx = eltype(P)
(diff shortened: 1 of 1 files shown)


src/interface.jl
```diff
@@ -320,50 +320,66 @@ _copy_output(x::SimpleVector) = Core.svec([map(_copy_output, x_sub) for x_sub in
 # Array, Memory
 function _copy_output(x::P) where {P<:_BuiltinArrays}
     temp = P(undef, size(x)...)
```
MilesCranmer: P.S., shouldn't this use `similar`?

willtebbutt: That would make sense to me.

willtebbutt (approved):
Thanks for this -- please bump the patch version, and then I'm happy for this to be merged.

Btw, do you have merge rights?

PR #635, opened 2025-06-30
Fix typo in algorithmic_differentiation.md

cncastillo:
This is just a small typo I found while reading the docs, that btw are awesome! Huge congrats :smile:. 

 As the idea is to approximate $f(x + \text{d} h) \approx f(x) + D f [x] (\text{d} h)$, then

```math
\lim_{\text{d} h \to 0} \frac{\| f(x + \text{d} h) - f(x) {\textcolor{red}-} D f [x] (\text{d} h)  \|_\mathcal{Y}}{\| \text{d}h \|_\mathcal{X}} = 0.
```
Wikipedia also agrees: https://en.wikipedia.org/wiki/Fr%C3%A9chet_derivative

Thanks for this amazing package, we are definitely trying it out :).


willtebbutt:
Thanks for catching this typo -- it's great to know that the docs are being read! I'll merge this once CI passes :)

PR #638, opened 2025-06-30
Fix jl_genericmemory_copy

willtebbutt:
#621 introduced a correctness bug -- when reviewing, I missed that the rule wasn't actually tested against typical examples, so it made it through. The previous implementation of the rule didn't take a copy of the data or the tangent during the forwards-pass, and did not increment the tangent of the original with the tangent of the copy on the reverse-pass.

This PR
- moves the rule to reside in the `memory.jl` file, which feels like a better home (to me) for the rule
- adds a couple of test cases involving `copy`ing `Memory` directly to provide correctness tests
- leaves the previous "can I make a copy of this empty `Dict`" test in place

@ErikQQY I'll request a review from you, since it's your rule that this modifies.

edit: this PR also introduces the ability to run `copy` on `Memory`, which is a nice bonus.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -69,7 +69,7 @@ SLEEFPirates = "0.6.43"
 SpecialFunctions = "2"
 StableRNGs = "1"
 Test = "1"
-julia = "~1.10, 1.11.5"
+julia = "~1.10, 1.11.3"
 
 [extras]
 AllocCheck = "9b6a8646-10ed-4001-bbdc-1d2f46dfbb1a"


src/rrules/memory.jl
```diff
@@ -880,6 +899,8 @@ function generate_derived_rrule!!_test_cases(rng_ctor, ::Val{:memory})
         (true, :none, nothing, Array{Float64,0}, undef, ()),
         (true, :none, nothing, Array{Float64,4}, undef, (2, 3, 4, 5)),
         (true, :none, nothing, Array{Float64,5}, undef, (2, 3, 4, 5, 6)),
+        (false, :none, nothing, copy, Memory{Float64}(randn(5))),
+        (false, :none, nothing, copy, Memory{Any}([randn(5), 5.0])),
```
ErikQQY, line 903: Maybe we can add more test cases like copying with `Set`? From local testing, differentiating copying with `Set` works fine.

willtebbutt: Yup, will add a test with a `Set` to check that it runs :) Will merge once I've done that and CI passes.

yebai:
@ErikQQY, the test for `copy(Dict{String, Float43})` occasionally fails on CI, indicating some remaining issues with the new rule. Can you create an issue to track it? See, https://github.com/chalk-lab/Mooncake.jl/actions/runs/16032852474/job/45237456779#step:6:1063

Issue #528, opened 2025-03-17
Possible incorrect rule implementation.


AstitvaAggarwal:
When i call `Mooncake.prepare_pullback_cache(x->Ptr{Float64}(x), 1)`

I get the following Error and Stacktrace :

```

ERROR: TypeError: in typeassert, expected Mooncake.CoDual{Ptr{Float64}, Ptr{Float64}}, got a value of type Mooncake.CoDual{Ptr{Float64}, Mooncake.NoFData}

Stacktrace:
 [1] #25
   @ .\REPL[208]:1 [inlined]
 [2] (::Tuple{…})(none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…})
   @ Base.Experimental .\<missing>:0
 [3] (::Mooncake.DerivedRule{…})(::Mooncake.CoDual{…}, ::Mooncake.CoDual{…})
   @ Mooncake C:\Users\astit\OneDrive\Desktop\GSOC\Julia\contri\Mooncake.jl\src\interpreter\s2s_reverse_mode_ad.jl:963
 [4] prepare_pullback_cache(::Function, ::Vararg{Any}; kwargs::@Kwargs{})
   @ Mooncake C:\Users\astit\OneDrive\Desktop\GSOC\Julia\contri\Mooncake.jl\src\interface.jl:279
 [5] prepare_pullback_cache(::Function, ::Vararg{Any})
   @ Mooncake C:\Users\astit\OneDrive\Desktop\GSOC\Julia\contri\Mooncake.jl\src\interface.jl:270
 [6] top-level scope
   @ REPL[208]:1
Some type information was truncated. Use `show(err)` to see complete types.

julia> show(err)
1-element ExceptionStack:
TypeError: in typeassert, expected Mooncake.CoDual{Ptr{Float64}, Ptr{Float64}}, got a value of type Mooncake.CoDual{Ptr{Float64}, Mooncake.NoFData}
Stacktrace:
  [1] #13
    @ .\REPL[147]:1 [inlined]
  [2] (::Tuple{Mooncake.Stack{Int32}, Base.RefValue{Tuple{Mooncake.LazyZeroRData{var"#13#14", Nothing}, Mooncake.LazyZeroRData{Int64, Nothing}}}})(none::Mooncake.CoDual{var"#13#14", Mooncake.NoFData}, none::Mooncake.CoDual{Int64, Mooncake.NoFData})
    @ Base.Experimental .\<missing>:0
[...]

PR #639, opened 2025-07-01
bitcast rrule fix

AstitvaAggarwal:
fixes issue #528 . The issue arises with functions like `Mooncake.prepare_pullback_cache(x->Ptr{Float64}(x), 1)`, when constructing the forward pass primal CoDual object [here](https://github.com/chalk-lab/Mooncake.jl/blob/c858c37e7c0714c25a1ecd2b6e2d56802dc7807e/src/rrules/builtins.jl#L221), we get a `Type Error` in the IR generated for the forward pass due to incorrect tangent generated for the Pointer primal type. The rrule for `Core.bitcast` was setting the forward tangent data to  `Ptr{<:IEEEFloat}` (assuming we can only bitcast a pointer to a pointer) otherwise to `NoFData()` - ignoring the case where an `Int`/`UInt` is bitcast into a `Ptr{<:Base.IEEEFloat}`.

Summary:
Mooncake's Interface (`prepare_pullback_cache`, `value_and_pullback` etc) does not support functions with primals of type `Ptr`.
However, the primitive rrule!!s for differentiating functions with `Ptr` primal type do exist (`Core.IntrinsicFunctions` -> `Core.bitcast`). Currently the primitive rrule works correctly if a program being differentiated has logic for reinterpreting a pointer type (`bitcast(Ptr{Float64}, Ptr{Float32}(12))`), via this PR we now error out with an appropriate message when an `Int/UInt` is being converted/cast to a `Ptr` type. This kind of operation is unsafe as accessing the memory address pointed by this pointer value might turns out to be nonexistent/preoccupied leading to segfaults/crashes or corrupting pre-existing stored memory.

### DIFF
--- a/test/rrules/builtins.jl
+++ b/test/rrules/builtins.jl
@@ -49,7 +49,7 @@ foo_throws(e) = throw(e)
         invoke(Mooncake.IntrinsicsWrappers.translate, Tuple{Any}, Val(:foo)),
     )
 
-    @testset "Disable bitcast to differentiable type & Int->Ptr" begin
+    @testset "Disable bitcast to differentiable type, or bitcast from Int/UInt to Ptr" begin
         @test_throws(
             ArgumentError,
             rrule!!(zero_fcodual(bitcast), zero_fcodual(Float64), zero_fcodual(5))


src/interface.jl
```diff
@@ -430,6 +430,7 @@ function prepare_pullback_cache(fx...; kwargs...)
     y, rvs!! = rule(map((x, dx) -> CoDual(x, fdata(dx)), fx, tangents)...)
 
     # Handle forward pass's primal exceptions
+    # __exclude_unsupported_output(primal(y))
```
AstitvaAggarwal: Considering `prepare_pullback_cache` does not support `Ptr` primal types, here i think we must pass only the primal values for checking.

test/interface.jl
```diff
@@ -110,6 +110,11 @@ end
     end
 
     @testset "prepare_pullback_cache errors" begin
+        test_func = x -> Ptr{Float64}(x)
```
AstitvaAggarwal: considering my above comment, this part is also subject to change.

src/rrules/builtins.jl
```diff
@@ -219,6 +219,9 @@ function rrule!!(f::CoDual{typeof(bitcast)}, t::CoDual{Type{T}}, x) where {T}
     v = bitcast(T, _x)
     if T <: Ptr && _x isa Ptr
         dv = bitcast(Ptr{tangent_type(eltype(T))}, tangent(x))
+    elseif T <: Ptr && _x isa Union{Int,UInt}
+        @warn "Creating Core.bitcast's primal tangent from a raw integer to a pointer is potentially unsafe."
+        dv = bitcast(Ptr{tangent_type(eltype(T))}, _x)
```
yebai:
```suggestion
        throw(error("Creating `Core.bitcast`'s primal tangent from a raw integer is unsafe."))
```

src/interface.jl
```diff
@@ -430,7 +430,7 @@ function prepare_pullback_cache(fx...; kwargs...)
     y, rvs!! = rule(map((x, dx) -> CoDual(x, fdata(dx)), fx, tangents)...)
 
     # Handle forward pass's primal exceptions
-    __exclude_unsupported_output(y)
+    __exclude_unsupported_output(primal(y))
```
yebai: Please fix this in a separate PR with a clear explanation.

src/interpreter/s2s_reverse_mode_ad.jl
```diff
@@ -258,7 +258,7 @@ initial rdata directly into the statement, which is safe because it is always a
 """
 function reverse_data_ref_stmts(info::ADInfo)
     function make_ref_stmt(id, P)
-        ref_type = Base.RefValue{P<:Type ? NoRData : zero_like_rdata_type(P)}
+        ref_type = Base.RefValue{P <: Type ? NoRData : zero_like_rdata_type(P)}
```
yebai: Let's avoid unnecessary styling changes in PRs. Same comments for a few other places. Please revert these styling changes.

src/tangents.jl
```diff
@@ -672,7 +672,7 @@ counting". If `c` is a `NoCache`, assume no aliasing or circular referencing.
 increment_internal!!(::IncCache, ::NoTangent, ::NoTangent) = NoTangent()
 increment_internal!!(::IncCache, x::T, y::T) where {T<:IEEEFloat} = x + y
 function increment_internal!!(::IncCache, x::Ptr{T}, y::Ptr{T}) where {T}
-    return x === y ? x : throw(error("eurgh"))
+    return x === y ? x : throw(error("incrementing Pointers is not supported"))
```
yebai:
```suggestion
    return x === y ? x : throw(error("Incrementing pointers is not supported!"))
```

src/rrules/builtins.jl
AstitvaAggarwal: Could removing the forward pass object creation for int->Ptr bitcast exclude some possible applications?

yebai: IIUC, it is not supported in its current form, right?

AstitvaAggarwal: Mooncake's Interface (`prepare_pullback_cache`, `value_and_pullback` etc) does not support functions with primals of type `Ptr`. 
However, the primitive `rrule!!`s for differentiating functions with `Ptr` primal type do exist (`Core.IntrinsicFunctions` -> `Core.bitcast`). Currently the rrule works correctly if a program being differentiated has logic for reinterpreting a pointer type (`bitcast(Ptr{Float64}, Ptr{Float32}(12))`) , but with this PR we can handle AD for programs where a **valid address** given as an `Int`/`UInt` form is being converted to a `Ptr` to use.

yebai: > but with this PR we can handle AD for programs where a valid address given as an Int/UInt form is being converted to a Ptr to use.

I don't think there's a use case for this. I suggest we forbid examples like 

1. `Mooncake.prepare_pullback_cache(x->Ptr{Float64}(x), 1)` 
2. `Mooncake.value_and_pullback(x->Ptr{Float64}(x), 1)`

with an informative error message. 

We can revisit this if there is a motivating example.

AstitvaAggarwal:
@yebai I'm not sure why Mooncake.is_homogeneous_and_immutable test fail. I think it's unrelated to the PR. Edit: fixed.

src/rrules/builtins.jl
```diff
@@ -219,6 +219,12 @@ function rrule!!(f::CoDual{typeof(bitcast)}, t::CoDual{Type{T}}, x) where {T}
     v = bitcast(T, _x)
     if T <: Ptr && _x isa Ptr
         dv = bitcast(Ptr{tangent_type(eltype(T))}, tangent(x))
+    elseif T <: Ptr && _x isa Union{Int,UInt}
+        throw(
+            ArgumentError(
+                "Creating `Core.bitcast`'s primal tangent from a raw integer is unsafe."
+            ),
+        )
```
yebai:
```suggestion
        int2ptr_err_msg =
            "It is not permissible to bitcast to an Int/UInt type to a Ptr type during AD, as " *
            "this risks giving the wrong answer, or causing Julia to segfault. " *
            "If this call to bitcast appears as part of the implementation of a " *
            "differentiable function, you should write a rule for this function, or modify " *
            "its implementation to avoid the bitcast."
        throw(ArgumentError(int2ptr_err_msg))
```

test/rrules/builtins.jl
```diff
@@ -49,11 +49,25 @@ foo_throws(e) = throw(e)
         invoke(Mooncake.IntrinsicsWrappers.translate, Tuple{Any}, Val(:foo)),
     )
 
-    @testset "Disable bitcast to differentiable type" begin
+    @testset "Disable bitcast to differentiable type & Int->Ptr" begin
```
yebai:
```suggestion
    @testset "Disable bitcast to differentiable type, or bitcast from Int/UInt to Ptr" begin
```

yebai (approved):

yebai:
@AstitvaAggarwal -- looks correct to me. Can you also update the PR summary box to reflect the changes in this PR? Feel free to merge after that if CI passes.

AstitvaAggarwal:
build failures are unrelated. Im merging this.

PR #642, opened 2025-07-02
fix supported primal checks in prepare_pullback_cache

AstitvaAggarwal:
Checking for `Ptr` types etc. can only be done if we primal check before running the forward pass for the derived rules/generate zero_tangent data for the rule's forward pass as it results in `ArgumentError: zero_tangent not defined for Pointers`.
refer : 
image

Also fixes alias tuple test [case](https://github.com/chalk-lab/Mooncake.jl/blob/f0e3f9a2886782415fd8e2b13c1bc5008b5672d7/test/interface.jl#L151)  (tuples are stored by value, aliasing is not possible unless the elements are address based objects). Broadcasting `__exclude_unsupported_output` [here](https://github.com/chalk-lab/Mooncake.jl/blob/f0e3f9a2886782415fd8e2b13c1bc5008b5672d7/test/interface.jl#L176) gives a false pass on the tests.
refer: 
image

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -430,7 +430,7 @@ function prepare_pullback_cache(fx...; kwargs...)
     y, rvs!! = rule(map((x, dx) -> CoDual(x, fdata(dx)), fx, tangents)...)
 
     # Handle forward pass's primal exceptions
-    __exclude_unsupported_output(y)
+    __exclude_unsupported_output(primal(y))
 
     # Run reverse-pass in order to reset stacks + state.
     rvs!!(zero_rdata(primal(y)))
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -421,6 +421,9 @@ end
 Returns a cache used with [`value_and_pullback!!`](@ref). See that function for more info.
 """
 function prepare_pullback_cache(fx...; kwargs...)
+    # Handle forward pass's primal exceptions before rule construction (zero_tangent is not defined for Ptr)
+    arg_copy = deepcopy.(fx)
+    __exclude_unsupported_output((arg_copy[1])(arg_copy[2:end]...))
 
     # Construct rule and tangents.
     rule = build_rrule(get_interpreter(), Tuple{map(_typeof, fx)...}; kwargs...)
@@ -428,9 +431,6 @@ function prepare_pullback_cache(fx...; kwargs...)
 
     # Run the rule forwards -- this should do a decent chunk of pre-allocation.
     y, rvs!! = rule(map((x, dx) -> CoDual(x, fdata(dx)), fx, tangents)...)
-
-    # Handle forward pass's primal exceptions
-    __exclude_unsupported_output(primal(y))
 
     # Run reverse-pass in order to reset stacks + state.
     rvs!!(zero_rdata(primal(y)))
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -196,9 +196,10 @@ end
 
 """
     __exclude_unsupported_output(y)
+    __exclude_func_with_unsupported_output(fx)
 
 Required for the robust design of [`value_and_pullback`](@ref), [`prepare_pullback_cache`](@ref).  
-Ensures that `y` contains no aliasing, circular references, `Ptr`s or non differentiable datatypes. 
+Ensures that `y` or returned value of `fx::Tuple{f, Targs...}` contains no aliasing, circular references, `Ptr`s or non differentiable datatypes. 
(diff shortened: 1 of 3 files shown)


AstitvaAggarwal:
@yebai requesting reviews.

src/interface.jl
```diff
@@ -421,6 +421,9 @@ end
 Returns a cache used with [`value_and_pullback!!`](@ref). See that function for more info.
 """
 function prepare_pullback_cache(fx...; kwargs...)
+    # Handle forward pass's primal exceptions before rule construction (zero_tangent is not defined for Ptr)
```
yebai: > Handle forward pass's primal exceptions before rule construction (zero_tangent is not defined for Ptr)

I am confused, does `prepare_pullback_cache` support `Ptr`-typed argument in `fx`? If not, let's throw an error before proceeding with the steps below.

yebai:
Two quick things: 

1. Please always format the code for PRs using `JuliaFormatter.format(".")`. 
2. Please avoid pasting screenshots. Instead, you can copy and paste REPL outputs into code blocks, e.g. 

```julia
julia> using Mooncake, Turing

julia> Mooncake.tangent_type(Array{Float64, 1})
Vector{Float64} (alias for Array{Float64, 1})

julia> 
```

src/interface.jl
AstitvaAggarwal: well it should ideally allow it (eg. bitcast PtrFloat32->PtrFloat64), but it fails as `prepare_pullback_cache` tries to construct a `zero_tangent` for the `Ptr` arg.

yebai: Given that it currently fails for `Ptr`-typed arguments, I suggest we check for `Ptr` arguments and throw an argument error early on. 

Then, you can revert to `__exclude_unsupported_output(primal(y))` after calling `y, rvs!! = rule(map((x, dx) -> CoDual(x, fdata(dx)), fx, tangents)...)`.

AstitvaAggarwal: But we also need to check for ptrs in the primal? (Case where function takes in non pointer and outputs a pointer)

yebai: Yes, `Ptr`-typed output should be forbidden too.

AstitvaAggarwal: > Given that it currently fails for `Ptr`-typed arguments, I suggest we check for `Ptr` arguments and throw an argument error early on. 
> 
> Then, you can revert to `__exclude_unsupported_output(primal(y))` after calling `y, rvs!! = rule(map((x, dx) -> CoDual(x, fdata(dx)), fx, tangents)...)`. 

`__exclude_unsupported_output(primal(y))` would have to be called before the rule though as the rule call also fails when an output is a ptr.

AstitvaAggarwal: basically it fails at 2 places - the forward pass for the derived rules or while generating zero_tangent data for the rule's forward pass. to avoid this ive added the changes to `__exclude_unsupported_output(y)` before the rule gen itself, this way we dont need to construct a rule whos forward pass would fail.

src/interface.jl
```diff
@@ -421,6 +421,9 @@ end
...
+    arg_copy = deepcopy.(fx)
+    __exclude_unsupported_output((arg_copy[1])(arg_copy[2:end]...))
```
```suggestion
    # Exclude specific primals; this needs to be handled before rule construction because
    # `zero_tangent` is not defined for some argument types like `Ptr` and will error. 
    _fx = deepcopy(fx)
    _func, _args = _fx[1], _fx[2:end]
    _y = _func(_args...)
    __exclude_unsupported_output(_y)
```

test/interface.jl
```diff
@@ -144,38 +144,46 @@ end
...
 
-        # ---- Unsupported Types ----
-        push!(test_to_fail_cases, Ptr{Float64}(12345))
-        push!(test_to_fail_cases, (1, Ptr{Float64}(12345)))
-        push!(test_to_fail_cases, [Ptr{Float64}(12345)])
-
-        @test_throws(
-            Mooncake.ValueAndPullbackReturnTypeError,
-            Mooncake.__exclude_unsupported_output.(test_to_fail_cases)
+        # ---- include Ptr Unsupported Types ----
```
yebai:
```suggestion
        # ---- Exclude `Ptr` typed input arguments and returned values ----
```

yebai (approved):

yebai:
@AstitvaAggarwal I tweaked the code style for consistency and clarity. Feel free to merge and create a new release once the CI passes.

AstitvaAggarwal:
@JuliaTagBot tag

PR #644, opened 2025-07-04
Add DynamicPPL extension, use `NoCache` for performance

sunxd3:
ref https://github.com/chalk-lab/Mooncake.jl/issues/552


sunxd3:
with `ADTests` models

| Model | Current PR | Release | Ratio (C/R) |
|-------|---------|---------|-------------|
| assume_beta | 4.486810 | 6.528063 | 0.687311 |
| assume_dirichlet | 5.028416 | 6.503771 | 0.773154 |
| assume_lkjcholu | 4.169331 | 4.713440 | 0.884562 |
| assume_mvnormal | 7.584657 | 9.605529 | 0.789614 |
| assume_normal | 4.587607 | 7.555556 | 0.607183 |
| assume_submodel | 3.035702 | 3.936000 | 0.771266 |
| assume_wishart | 7.341956 | 9.142857 | 0.803026 |
| broadcast_macro | 3.552181 | 6.819299 | 0.520901 |
| control_flow | 3.767179 | 7.860805 | 0.479236 |
| demo_assume_dot_observe | 3.822346 | 5.625724 | 0.679441 |
| demo_assume_dot_observe_literal | 3.110979 | 6.891854 | 0.451399 |
| demo_assume_index_observe | 3.540136 | 5.826788 | 0.607562 |
| demo_assume_matrix_observe_matrix_index | 4.456230 | 5.599493 | 0.795827 |
| demo_assume_multivariate_observe | 4.135187 | 7.349003 | 0.562687 |
| demo_assume_multivariate_observe_literal | 4.092533 | 6.768094 | 0.604680 |
| demo_assume_observe_literal | 3.496984 | 6.271389 | 0.557609 |
| demo_assume_submodel_observe_index_literal | 3.079295 | 4.747754 | 0.648579 |
| demo_dot_assume_observe | 4.045807 | 5.916732 | 0.683791 |
| demo_dot_assume_observe_index | 3.737512 | 6.001495 | 0.622764 |
| demo_dot_assume_observe_index_literal | 3.563587 | 6.228103 | 0.572179 |
| demo_dot_assume_observe_matrix_index | 3.861691 | 6.480148 | 0.595926 |
| demo_dot_assume_observe_submodel | 3.258644 | 5.056410 | 0.644458 |
| dot_assume | 3.822570 | 5.309253 | 0.719983 |
| dot_observe | 4.396199 | 8.126444 | 0.540974 |
| dynamic_constraint | 3.199087 | 5.866546 | 0.545310 |
| multiple_constraints_same_var | 7.969536 | 8.591111 | 0.927649 |
| multithreaded | 3.893714 | 4.530427 | 0.859459 |
| n010 | 3.605450 | 3.854293 | 0.935437 |
| n050 | 3.498392 | 3.673095 | 0.952437 |
| n100 | 3.509174 | 3.450000 | 1.017152 |
| n500 | 4.415281 | 7.765084 | 0.568607 |
| observe_bernoulli | 5.329560 | 7.040417 | 0.756995 |
| observe_categorical | 5.303571 | 9.444518 | 0.561550 |
| observe_index | 4.329834 | 7.583224 | 0.570975 |
| observe_literal | 4.903116 | 6.428380 | 0.762730 |
| observe_multivariate | 4.058949 | 8.550705 | 0.474692 |
| observe_submodel | 3.884157 | 5.014414 | 0.774598 |
| pdb_eight_schools_centered | 3.702677 | 5.093955 | 0.726877 |

sunxd3:
As Penny pointed out in https://github.com/TuringLang/DynamicPPL.jl/issues/856, when a model is declared in a function, it becomes a closure.

```julia
julia> function create_closure()
           @model function closure_model(x)
               s ~ InverseGamma(2, 3)
               return x ~ Normal(0, s)
           end
           return closure_model
       end
create_closure (generic function with 1 method)

julia> closure_fn = create_closure()
(::var"#closure_model#17") (generic function with 2 methods)

julia> model = closure_fn([1.0, 2.0])
Model{var"#closure_model#17", (:x,), (), (), Tuple{Vector{Float64}}, Tuple{}, DefaultContext}(var"#closure_model#17"(Core.Box(var"#closure_model#17"(#= circular reference @-2 =#))), (x = [1.0, 2.0],), NamedTuple(), DefaultContext())

julia>  vi = DynamicPPL.VarInfo(Random.default_rng(), model); ldf = DynamicPPL.LogDensityFunction(model, vi, DynamicPPL.DefaultContext()); tangent = zero_tangent(ldf);

julia> tangent.fields.model.fields.f.fields.closure_model.fields.contents.tangent === model_f_tangent
true
```

This is a special case where tangent might refer to itself, and reason for https://github.com/chalk-lab/Mooncake.jl/blob/2f9a1c1dc957a3616152fa581c48c436d88b0567/ext/MooncakeDynamicPPLExt.jl#L88-L96

sunxd3:
The interface is fine, but achieving feature parity with the current code is difficult to do in a clean way. The main reason is that closure detection requires runtime information. We can go for a hybrid approach, which somewhat defeats the purpose of introducing an interface function. Or we can opt for cleanliness and leave some performance gains on the table.

yebai:
> Or we can opt for cleanliness and leave some performance gains on the table.

Yes, that is okay if we do it robustly, excluding invalid cases like closures.

penelopeysm:
I haven't yet looked at the code, but I wanted to point out that if Mooncake has a DynamicPPLExt (and thus a compat entry for DynamicPPL), then breaking changes of DynamicPPL cannot be tested with Mooncake.

That isn't a massive problem in and of itself; we could separate the Mooncake part out from the test suite (in exactly the same way Enzyme is).

However, because breaking changes in DPPL are much more common than breaking changes in Mooncake, I think it may be easier to implement this code inside DynamicPPLMooncakeExt rather than the other way around.

(see: https://github.com/TuringLang/DynamicPPL.jl/issues/740 where this issue was discussed)

ext/MooncakeDynamicPPLExt.jl
```diff
@@ -0,0 +1,117 @@
...
+    if hasfield(typeof(fields.model), :fields)
+        model_fields = propertynames(fields.model.fields)
+        all(f in model_fields for f in (:f, :args, :defaults, :context)) || return false
+    end
+
+    # Verify varinfo tangent structure  
+    if hasfield(typeof(fields.varinfo), :fields)
+        varinfo_fields = propertynames(fields.varinfo.fields)
+        all(f in varinfo_fields for f in (:metadata, :logp, :num_produce)) || return false
+    end
```
penelopeysm:
```suggestion
    is_dppl_varinfo_tangent(fields.varinfo) || return false
    is_dppl_model_tangent(fields.model) || return false
```

I reckon you could cut some duplication here

ext/MooncakeDynamicPPLExt.jl
```diff
@@ -0,0 +1,117 @@
...
+    fields = x.fields
+    # Check for the expected fields in Metadata
+    expected_fields = (:idcs, :vns, :ranges, :vals, :dists, :orders, :flags)
+    all(f in propertynames(fields) for f in expected_fields) || return false
+
+    return true
+end
+
+function Mooncake.set_to_zero!!(x)
+    if x isa Tangent && is_dppl_ldf_tangent(x)
```
penelopeysm:
```suggestion
    if is_dppl_ldf_tangent(x)
```

`x isa Tangent` is already the first thing that `is_dppl_ldf_tangent` checks -- unless you think inserting it explicitly here might help type inference?

penelopeysm (changes requested):
I had a look at the code and I'm generally happy. However, I think it would make sense to move this to DynamicPPL so that we can also test this with the existing Mooncake tests there. If we do that then I can review it there in more detail if needed?

yebai:
Closed in favour of https://github.com/TuringLang/DynamicPPL.jl/pull/975

PR #659, opened 2025-07-21
anonymous function doc fix for @from_rrule

### LINKED ISSUE #558: rrule!! not picked up: Mooncake's `is_primative` is too rigid sometimes leading to false negatives
I am trying and failing to make [my package](https://github.com/thorek1/MacroModelling.jl/) Mooncake compatible. It currently works with Zygote + custom rrules (ChainRulesCore).

I understand this is how I can make Mooncake aware of the rrule:
```Julia
@from_rrule DefaultCtx Tuple{typeof(get_NSSS_and_parameters), ℳ, Vector{S}} where S <: AbstractFloat true
```

but calling it with DifferentiationInterface gives the following error

```julia
vj = DifferentiationInterface.jacobian(x->get_NSSS_and_parameters(m, x)[1], AutoMooncake(;config=nothing), m.parameter_values)

ERROR: No rrule!! available for foreigncall with primal argument types Tuple{Val{(:daxpy_64_, Symbol("@rpath/libblastrampoline.5.dylib"))}, Val{Nothing}, Tuple{Val{Ref{Int64}}, Val{Ref{Float64}}, Val{Ptr{Float64}}, Val{Ref{Int64}}, Val{Ptr{Float64}}, Val{Ref{Int64}}}, Val{0}, Val{:ccall}, Ptr{Int64}, Ptr{Float64}, Ptr{Float64}, Ptr{Int64}, Ptr{Float64}, Ptr{Int64}, Base.RefValue{Int64}, Ptr{Float64}, Base.RefValue{Int64}, Ptr{Float64}, Base.RefValue{Float64}, Base.RefValue{Int64}}. This problem has most likely arisen because there is a ccall somewhere in the function you are trying to differentiate, for which an rrule!! has not been explicitly written.You have three options: write an rrule!! for this foreigncall, write an rrule!! for a Julia function that calls this foreigncall, or re-write your code to avoid this foreigncall entirely. If you believe that this error has arisen for some other reason than the above, or the above does not help you to workaround this problem, please open an issue.
```

the rest of the error message shows that the rrule is not called but the function itself is (attempted to be) differentiated.

What I understand from the documentation is that the type `ℳ` might be a problem. `ℳ` is a large nested mutable struct with, lists, dictionaries, circular buffers, functions, arrays, and scalars among other types therein. See [here](https://github.com/thorek1/MacroModelling.jl/blob/e59ddbf302dbc6ce277cf72e7246bb9d7492e016/src/structures.jl#L268) for details.

I am sorry if I am overlooking something obvious here. Any help is much appreciated

AstitvaAggarwal:
handles/closes #558.

### DIFF
--- a/src/tools_for_rules.jl
+++ b/src/tools_for_rules.jl
@@ -456,7 +456,7 @@ we must specify the respective macro call, `rrule` definition as follows:
     @from_rrule DefaultCtx Tuple{Any,CalculationOptions,typeof(get_NSSS_and_parameters),ℳ,Vector{S}} where S<:Real true
 
     function ChainRulesCore.rrule(
-        Any,
+        func_ir_name::Any,
         opts::CalculationOptions,
         ::typeof(MacroModelling.get_NSSS_and_parameters),
         m::ℳ,
@@ -473,7 +473,7 @@ we must specify the respective macro call, `rrule` definition as follows:
     rule = Mooncake.build_rrule(a, m.parameter_values)
     rule(zero_fcodual(a), zero_fcodual(m.parameter_values))
 
-Here in `@from_rrule` and `ChainRulesCore.rrule`, `Any` is the argument+type information for the lowered IR name of the function call of interest.
+Here in `@from_rrule` and `ChainRulesCore.rrule`, `Any` is the type information for the lowered IR name of the function call of interest.
 You can also use the actual lowered IR name eg. MacroModelling.var"##get_NSSS_and_parameters#1215" (got via `@code_lowered`) for the function call, which is unique for each dispatch of this function.
 We can safely pass `Any`, due to the fact that each dispatch has unique type information for its arguments respectively - This is implicitly satisfied in the `@from_rrule` and `ChainRulesCore.rrule` calls by design.
 This also works when using loose type constraints on arguments as long as the specified function's type information is a subtree of the specific function's dispatch type tree.


yebai:
Closing this as it is a heuristic approach. We need a more principled solution.

AstitvaAggarwal:
@yebai This works for all anonymous functions. In fact the IR function name for the inlined function is unique for all the function's package defined dispatch possibilities (like `build_rrule` has 4-5 function dispatches as mentioned in api docs) but consistent across all repl sessions. Here the `@from_rrule` macro can also be set to loose type specifications only if required.

Otherwise, The inlining can only be handled via the `@noinline` macro + `@from_rrule` macro with `Any` type for all the arguments - for every method inside the anonymous function. we could add either of these ways in the docs as a note maybe?

yebai:
I mentioned in https://github.com/chalk-lab/Mooncake.jl/issues/558 that the issue is not specific to closure. The problem could occur in other cases where Mooncake fails to prevent Julia from using inline functions. 

A principled fix would modify the `@from_rules` / `@zero_adjoint` macros so they always define `is_primative(::Any, ...)`  instead of `is_primative(::ConcreteType, ...)` to become more proactive in preventing inlining. This incurs a small performance penalty, as it prevents more inlining, but that is a price worth paying for correctness.

Issue #661, opened 2025-07-23
Failure of traversing through nested structures


zuhengxu:
As described in https://github.com/TuringLang/NormalizingFlows.jl/pull/53#issuecomment-3111298171

zuhengxu:
I believe the bugs are introduced in `Mooncake v0.4.124` (see https://github.com/TuringLang/NormalizingFlows.jl/pull/53#issuecomment-3116719380)

yebai:
It is likely caused by this change

https://github.com/chalk-lab/Mooncake.jl/pull/594/files#diff-0163b66510af82fcd75ca009591262b191accc7873909a2e7f3b12dfef7976cbR36-R55

@zuhengxu / @MilesCranmer anything obvious that could be incorrect?

MilesCranmer:
"Unreachable reached" usually points to a Julia bug: https://discourse.julialang.org/t/unreachable-reached-at-0x13929e935/38528/6. It is possible it just appeared from that commit but might not be directly related.

In this code: https://github.com/chalk-lab/Mooncake.jl/blob/a26b5c35c55d1e98b9e8c6bfafbbe3dc55784140/src/rrules/new.jl#L36-L57

nothing looks problematic in terms of generated function rules. The only thing being used in the generated function construction is `Val(n)` for `n` the number of fields, and `Val(names)` for `names` the names of those fields. So I don't think that specifically is an issue. Although "Unreachable reached" might be caused by something else and that particular change just caused a particular branch of code to get executed in Julia.

@zuhengxu might be useful to run this on previous versions of Julia and see what the bisect points to there?

yebai:
I confirmed changes to `build_fdata` in #594 introduced this bug. Reverting related changes in https://github.com/chalk-lab/Mooncake.jl/pull/663 fixes example in https://github.com/TuringLang/NormalizingFlows.jl/pull/53#issuecomment-3111298171

MilesCranmer:
I don't think this is fixing the issue though, it's just temporarily concealing a symptom.

The function itself has no visible issues, and also this bug is related to undefined behavior, so it's either caused by something else in the code, or is a Julia bug. It just so happens that a certain combination of Julia code (for this particular Julia version, etc.) brings out the symptom. But the issue is still lurking somewhere

yebai:
[...]

PR #663, opened 2025-07-25
Bugfix in `build_fdata`: replace `Tuple{Vararg{Any,N}}`  with type parameter `Tfdata<:Tuple`

yebai:
This PR fixes https://github.com/chalk-lab/Mooncake.jl/issues/661 

`Tuple{Vararg{Any,N}}` appears to be causing incorrect dispatches in specific `build_fdata` cases and crashing Julia. This issue was discovered while I was iterating through the code. However, it is challenging to provide a minimal, reproducing example. 

Perhaps @MilesCranmer has an idea what is happening. 

Close https://github.com/chalk-lab/Mooncake.jl/issues/661


src/rrules/new.jl
```diff
@@ -33,27 +33,19 @@ function rrule!!(
...
+        F = fdata_field_type(P, n)
+        if n <= length(fdata.parameters)
+            data_expr = Expr(:call, __get_data, P, :x, :fdata, n)
+            return F <: PossiblyUninitTangent ? Expr(:call, F, data_expr) : data_expr
+        else
+            return :($F())
+        end
     end
+    F_out = fdata_type(tangent_type(P))
+    return :($F_out(NamedTuple{$names}($(Expr(:call, tuple, fdata_exprs...)))))
```
yebai, line 48:
```suggestion
    return :(fdata_type(tangent_type(P))(NamedTuple{$names}($(Expr(:call, tuple, fdata_exprs...)))))
```

src/rrules/new.jl
```diff
@@ -33,27 +33,18 @@ function rrule!!(
...
+    fdata_exprs = map(eachindex(names)) do n
+        F = fdata_field_type(P, n)
+        if n <= length(fdata.parameters)
+            data_expr = Expr(:call, __get_data, P, :x, :fdata, n)
+            return F <: PossiblyUninitTangent ? Expr(:call, F, data_expr) : data_expr
+        else
+            return :($F())
+        end
     end
+    return :(fdata_type(tangent_type(P))(NamedTuple{$names}($(Expr(:call, tuple, fdata_exprs...)))))
```
yebai, line 47:
```suggestion
    quote
        F_out = fdata_type(tangent_type(P))
        F_out(NamedTuple{$names}($(Expr(:call, tuple, fdata_exprs...))))
    end
```

src/rrules/new.jl
```diff
@@ -33,26 +33,20 @@ function rrule!!(
...
-                    data = __get_data(P, x, fdata, n)
-                    F <: PossiblyUninitTangent ? F(data) : data
-                else
-                    F()
-                end
-            end
-        )
         F_out = fdata_type(tangent_type(P))
-        return F_out(NamedTuple{$names}(processed_fdata))
+        F_out(NamedTuple{$names}($(Expr(:call, tuple, fdata_exprs...))))
```
yebai, line 49:
```suggestion
        return F_out(NamedTuple{$names}($(Expr(:call, tuple, fdata_exprs...))))
```

yebai:
@MilesCranmer I'd appreciate it if you could find a minimal reproducing example for this bug. Then we can replace the normalising flow integration tests with an MWE.

MilesCranmer:
Sorry which bug do you mean?

yebai:
See the PR description and code changes.

MilesCranmer:
Ah I see. I did actually try yesterday to minify it and run it under rr but I had zero luck. Any effort on my part to minify it would just remove the reproducer effectively

yebai:
It seems like a genuine Julia bug; otherwise, Julia should error if there was something wrong with dispatch. I think it can be quite hard to find an MWE for Julia, though.

We probably want to merge this before identifying the root cause in the Julia compiler code.

yebai:
@MilesCranmer, are you happy with this PR?

yebai:
I played with this a bit further. Here is another "fix" that works: 

```diff
@generated function _build_fdata_cartesian(
    ::Type{P}, x::Tuple, fdata::Tuple{Vararg{Any,N}}, ::Val{nfield}, ::Val{names}
) where {P,N,nfield,names}
+    2*N
    quote
        processed_fdata = Base.Cartesian.@ntuple(
```

This, of course, doesn't make sense since the extra `2*N` statement does nothing. But it fixes the bug. 

Does that ring a bell about what might be wrong?

MilesCranmer:
Super weird. Definitely add that to the JuliaLang issue.

idk it kind of reminds me of this a little bit: https://github.com/JuliaLang/julia/pull/55093#issuecomment-2221037166

MilesCranmer:
Someone should bisect Julia with this bug and see when it shows up

yebai:
> Someone should bisect Julia with this bug and see when it shows up

I did a quick test on `lts` and observed the same segfault. So this is unlikely to be a new bug introduced in recent Julia releases.

I am going to merge this now, as I am confident that the real issue is caused by a Julia bug. In cases like these, it doesn't make sense to find a more minimal regression test for Mooncake than we currently have (i.e. `flux.jl`).

Issue #472, opened 2025-02-10
Utility for stripping tangents


gdalle:
As noted by @mcabbott in #466, Mooncake returns a custom `Tangent` struct that does not natively play nice with Functors, Optimisers and the like. Would it make sense to provide a utility that strips all the `Tangent` types from a pullback result, essentially turning it into a nested `NamedTuple`?
Here's a simple prototype inspired by [`Fluxperimental._moonstrip`](https://github.com/FluxML/Fluxperimental.jl/blob/4e11d268eee0c863e9a7557e9dbd4361ed557b1e/ext/FluxMooncakeExt.jl#L115-L131):

```julia
using DifferentiationInterface
using Mooncake:
    Mooncake, MutableTangent, NoTangent, PossiblyUninitTangent, Tangent, tangent_type

strip_tangents(::NoTangent) = nothing
strip_tangents(t::Union{Tangent,MutableTangent}) = strip_tangents(t.fields)
strip_tangents(t::PossiblyUninitTangent) = strip_tangents(t.tangent)
strip_tangents(t::Union{Tuple,NamedTuple,AbstractArray}) = map(strip_tangents, t)
strip_tangents(t::Base.IEEEFloat) = t

f = Returns(1.0)
backend = AutoMooncake(; config=nothing)

struct MyStruct{A,B}
    a::A
    b::B
end

strip_tangents(gradient(f, backend, 1.0)) == 0.0
strip_tangents(gradient(f, backend, [1.0])) == [0.0]
strip_tangents(gradient(f, backend, (1.0, [1.0]))) == (0.0, [0.0])
strip_tangents(gradient(f, backend, (1, ["hello"]))) == (nothing, [nothing])
strip_tangents(gradient(f, backend, MyStruct(1, 2.0))) == (a=nothing, b=0.0)
```

@willtebbutt what is likely to go horribly wrong if we provide something like that?

Related:

- #467 
- #466

willtebbutt:
I had been thinking about something like this myself. It definitely makes sense to me. The only place that it's likely to run into problems is with possibly uninitialised fields -- i.e. you would probably implicitly rule out structs such as
```julia
mutable struct Foo
    x::Float64
    Foo() = new()
    Foo(x::Float64) = new(x)
end
```
and probably some other edge-cases (these probably aren't a bad thing to rule out anyway).

gdalle:
[...]

PR #664, opened 2025-07-26
add rule for complex logdet

### LINKED ISSUE #662: Complex number support
Is Mooncake supposed to be able to handle complex numbers? I've tried a few examples, and for simple functions it gets the correct result, but as soon as I try something more sophisticated it fails due to a foreign call. Looking at the source code I see that there are rules for a lot of BLAS and LAPACK functions, but always only the real versions. Is this by design? I'm asking here because the docs don't say anything about it. Example:

```julia
using LinearAlgebra
import Mooncake

function test_complex(::Type{T}, simple::Bool) where {T}

    function f(point)
        M = mat(point, T)
        return real(tr(M'M))
    end

    function g(point)
        M = mat(point, T)
        return logdet(M)
    end

    x = randn(T, 2, 2)
    p0 = rvec(x * x')
    h = simple ? f : g
    grad = gradient(h, AutoMooncake(), p0)
    return mat(grad, T)
end

function rvec(M::AbstractMatrix{T}) where {T}
    if T <: Real
        return vec(M)
    else
        return [real(vec(M)); imag(vec(M))]
    end
end

function mat(v::AbstractVector, ::Type{T}) where {T}
    if T <: Real
        n = isqrt(length(v))
        return reshape(v, n, n)
    else
        l2 = div(length(v), 2)
        n = isqrt(l2)
        return reshape(v[1:l2] + im * v[l2+1:end], n, n)
    end
end
```
`test_complex(Float64, true)`, `test_complex(Float64, false)`, and `test_complex(ComplexF64, true)` work, but `test_complex(ComplexF64, false)` fails due to a missing rule for `getrf!`.

araujoms:
Closes #662 

Let's not let perfect be the enemy of good, I don't actually need `logdet(Hermitian(x))`.

### DIFF
--- a/src/rrules/lapack.jl
+++ b/src/rrules/lapack.jl
@@ -1,7 +1,7 @@
-@is_primitive(MinimalCtx, Tuple{typeof(LAPACK.getrf!),AbstractMatrix{<:BlasRealFloat}})
+@is_primitive(MinimalCtx, Tuple{typeof(LAPACK.getrf!),AbstractMatrix{<:BlasFloat}})
 function rrule!!(
     ::CoDual{typeof(LAPACK.getrf!)}, _A::CoDual{<:AbstractMatrix{P}}
-) where {P<:BlasRealFloat}
+) where {P<:BlasFloat}
     A, dA = arrayify(_A)
     A_copy = copy(A)
 
@@ -22,7 +22,7 @@ end
 @is_primitive(
     MinimalCtx,
     Tuple{
-        typeof(Core.kwcall),NamedTuple,typeof(LAPACK.getrf!),AbstractMatrix{<:BlasRealFloat}
+        typeof(Core.kwcall),NamedTuple,typeof(LAPACK.getrf!),AbstractMatrix{<:BlasFloat}
     },
 )
 function rrule!!(
@@ -30,7 +30,7 @@ function rrule!!(
     _kwargs::CoDual{<:NamedTuple},
     ::CoDual{typeof(getrf!)},
     _A::CoDual{<:AbstractMatrix{P}},
-) where {P<:BlasRealFloat}
+) where {P<:BlasFloat}
     check = _kwargs.x.check
     A, dA = arrayify(_A)
     A_copy = copy(A)
--- a/test/rrules/lapack.jl
+++ b/test/rrules/lapack.jl
@@ -1,3 +1,17 @@
 @testset "lapack" begin
     TestUtils.run_rrule!!_test_cases(StableRNG, Val(:lapack))
+
+    @testset "real/complex getrf!" begin
+        d = 3
+        logdet_realmat(x) = logdet(reshape(x, d, d))
+        x0 = vec(randn(d, d)^2)
+        TestUtils.test_rule(StableRNG(1), logdet_realmat, x0; is_primitive=false)
+        logdet_complexmat(x) = real(
+            logdet(reshape(x[1:(d^2)], d, d) + im * reshape(x[(d^2 + 1):(2d^2)], d, d))
+        )
+        z0 = randn(ComplexF64, d, d)
+        y0 = z0'z0
+        x0 = [vec(real(y0)); vec(imag(y0))]
+        TestUtils.test_rule(StableRNG(1), logdet_complexmat, x0; is_primitive=false)
+    end
 end
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.140"
+version = "0.4.141"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"


willtebbutt:
Thanks for this! There should be a block of tests at the bottom of the lapack src file which specifies tests specific to getrf. Would it be possible to also extend them to test the complex version?

(On my phone, so can't easily point to the code directly, apologies!!)

araujoms:
Ah, they were in source file, that's why I couldn't find them! No problem, done.

araujoms:
I thought everything was working because the test for `rrules/lapack` didn't run by default. Now I see that it wasn't actually working. The tests for `SubArray` and `ReshapedArray` fail. I don't know what is wrong; I simply let the corresponding methods accept `BlasFloat`, and if I compute the gradient with the input in the corresponding format I can see that the gradient is correct, albeit in a rather inconvenient format.

But the same happens with a plain numerical matrix, independently of my additions. Compare:
```julia
julia> gradient(x -> real(tr(x'x)), AutoMooncake(), randn(2,2))
2×2 Matrix{Float64}:
  0.434831   3.09532
 -1.09911   -2.19635

julia> gradient(x -> real(tr(x'x)), AutoMooncake(), randn(ComplexF64, 2,2))
2×2 Matrix{Mooncake.Tangent{@NamedTuple{re::Float64, im::Float64}}}:
 Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.847196, im = -0.810308))  …  Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.548862, im = 0.833737))
 Tangent{@NamedTuple{re::Float64, im::Float64}}((re = -0.269572, im = 1.98935))      Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 1.83858, im = 0.926595))
```

yebai:
> that the gradient is correct, albeit in a rather inconvenient format.

The gradient format is due to #472.

yebai:
> The tests for SubArray and ReshapedArray fail.

Cc @willtebbutt

willtebbutt:
The checks which are failing are the "is the gradient numerically correct" checks -- see https://github.com/chalk-lab/Mooncake.jl/blob/05cdce129f2ab14b13c340434f5cd54dab8eb7f0/src/test_utils.jl#L513 . There are a couple of things to check:
1. Is it actually the case that we expect to have exactly the same rule for the real-valued case as the complex-valued case? Or does the general case involve eg. computing the complex conjugate of something (this is purely guess-work on my part as I've not worked through the maths for the complex case)
2. It would be worth running the test cases for just one of the examples that is failing and checking that the approimate equality check which is failing is failing by a large amount. If it's only just failing, then it's possible that the rule is implemented correctly, but for some reason the generated test cases are ill-conditioned (in which case we should look at fixing the test cases)

araujoms:
I did check that the result was correct for the complex case. The gradient of logdet is simply the inverse. Ill conditioning is a possibility, how do I see which matrices were used in the failing tests? Then I could check them manually.

willtebbutt:
Sadly your best bet is probably to include print statements strategically in the code.

araujoms:
I can't make heads or tails of the tests, but it's certainly not an actual error, because the tests succeed for plain matrices. It's just a formatting issue. Look:
```julia
julia> using LinearAlgebra; using DifferentiationInterface; import Mooncake; import Mooncake.Tangent; import Mooncake.NoTangent

julia> f(x) = real(logdet(x));

julia> z0 = randn(ComplexF64, 3, 3); @views y0 = z0[1:2, 1:2]; x0 = Matrix(y0);

julia> inv(x0)'
2×2 adjoint(::Matrix{ComplexF64}) with eltype ComplexF64:
 0.0729042+0.947071im  0.556282-0.362036im
  0.335595-0.15118im   0.460689-0.136424im

julia> gradient(f, AutoMooncake(), x0)
2×2 Matrix{Tangent{@NamedTuple{re::Float64, im::Float64}}}:
 Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.0729042, im = 0.947071))  …  Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.556282, im = -0.362036))
 Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.335595, im = -0.15118))      Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.460689, im = -0.136424))

julia> gradient(f, AutoMooncake(), y0)
Tangent{@NamedTuple{parent::Matrix{Tangent{@NamedTuple{re::Float64, im::Float64}}}, indices::NoTangent, offset1::NoTangent, stride1::NoTangent}}((parent = Tangent{@NamedTuple{re::Float64, im::Float64}}[Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.07290416524300863, im = 0.9470706523876997)) Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.5562823767358989, im = -0.3620364530936287)) Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.0, im = 0.0)); Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.3355953622333809, im = -0.1511804723798711)) Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.46068937497050844, im = -0.13642443117666486)) Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.0, im = 0.0)); Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.0, im = 0.0)) Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.0, im = 0.0)) Tangent{@NamedTuple{re::Float64, im::Float64}}((re = 0.0, im = 0.0))], indices = NoTangent(), offset1 = NoTangent(), stride1 = NoTangent()))
```
When I do it with a plain matrix the result is exactly correct, no complex conjugation missing. When I do it with a `SubArray` the result is also correct, you just need to find it amidst all that noise. My guess is that the tests can't handle this format.

yebai:
https://github.com/chalk-lab/Mooncake.jl/pull/683 will address the gradient format issue and unblock this PR.

willtebbutt:
Just had a look at this. The problem is a bug in `arrayify` for complex types. I need to crack on with other stuff today, but will provide a more careful explanation + fix in a separate PR this evening / tomorrow evening.

willtebbutt:
@araujoms main now contains fixes to arrayify which ought to fix the problems in this PR. Would you be happy to update this PR (or create a new one with the relevant changes from this one on top of main if it's easier), and see if everything is working?

src/rrules/lapack.jl
```diff
@@ -505,6 +503,7 @@ end
 function generate_hand_written_rrule!!_test_cases(rng_ctor, ::Val{:lapack})
     rng = rng_ctor(123)
     Ps = [Float64, Float32]
```
yebai, line 505:
```suggestion
```

araujoms:
The test failures are the same as in main, so I think everything is indeed working now. Specially rrules/lapack, which was failing before.

The tests report a performance regression for `potrf`. That wasn't touched by this PR, so I assume it's a random fluctuation.

src/rrules/lapack.jl
araujoms: I need this for the tests of `trtrs`, `getrs`, `getri`, `potrf`, and `potrs`, which do not accept complex numbers.

test/rrules/lapack.jl
```diff
@@ -1,3 +1,19 @@
...
+        logdet_complexmat(x) = real(
+            logdet(
+                reshape(x[1:(d ^ 2)], d, d) + im * reshape(x[(d ^ 2 + 1):(2d ^ 2)], d, d)
+            ),
+        )
+        z0 = randn(ComplexF64, d, d)
+        y0 = z0'z0
+        x0 = [vec(real(y0)); vec(imag(y0))]
+        TestUtils.test_rule(StableRNG(1), logdet_complexmat, x0; is_primitive=false)
+    end
```
willtebbutt: Would you mind moving these into the `generate_derived_rrule!!_test_cases` method for LAPACK? I would just prefer to keep all tests in the same place.

src/rrules/lapack.jl
```diff
@@ -583,13 +582,23 @@ end
...
+        map_prod([false, true], complexPs) do (check, P)
+            As = blas_matrices(rng, P, 5, 5)
+            return map(As) do A
+                (false, :none, nothing, getrf_wrapper!, A, check)
+            end
+        end...,
+
+        # logdet
+        map(complexPs) do P
+            A = randn(rng, P, 3, 3)
```
willtebbutt: I think you can probably make use of the `Mooncake.invertible_blas_matrices` function here (it's like `blas_matrices`, but ensures that they're invertible). This will ensure that we don't just test on `Matrix{P}`, but also views, reshaped arrays, etc.

araujoms: The problem is that I need to make them positive definite, that's why I'm doing `A'A + I`.

willtebbutt: Ah, good point. Certainly, for the real-valued tests cases it'll need to have positive definite matrices. For the complex-valued test cases, on the other hand, `logdet` should work fine provided that the matrix is only invertible, so we should probably test the complex-valued cases on some matrices which aren't positive definite.

Would you mind splitting this into two collections of test cases? One in which we have real-valued positive definite matrices, and one in which we have complex-valued invertible (and not necessarily positive definite) matrices?

willtebbutt:
Thanks for separating out the testing -- I think this is basically done now. I think you just need to run the formatter to make the "quality" checks happy.

araujoms:
I'm happy with merging, thanks for all the help.

PR #667, opened 2025-07-29
Add `requires_cache` API for optimizing tangent operations without circular reference tracking

sunxd3:
Ref https://github.com/TuringLang/DynamicPPL.jl/pull/975#discussion_r2235727500

### DIFF
--- a/src/Mooncake.jl
+++ b/src/Mooncake.jl
@@ -149,7 +149,7 @@ include("public.jl")
 end
 #! format: on
 
-@public Config, value_and_pullback!!, prepare_pullback_cache
+@public Config, value_and_pullback!!, prepare_pullback_cache, requires_cache
 
 # Public, exported
 export value_and_gradient!!, prepare_gradient_cache
--- a/src/Mooncake.jl
+++ b/src/Mooncake.jl
@@ -149,7 +149,7 @@ include("public.jl")
 end
 #! format: on
 
-@public Config, value_and_pullback!!, prepare_pullback_cache, require_tangent_cache
+@public Config, value_and_pullback!!, prepare_pullback_cache
 
 # Public, exported
 export value_and_gradient!!, prepare_gradient_cache
--- a/src/Mooncake.jl
+++ b/src/Mooncake.jl
@@ -149,7 +149,7 @@ include("public.jl")
 end
 #! format: on
 
-@public Config, value_and_pullback!!, prepare_pullback_cache, requires_cache
+@public Config, value_and_pullback!!, prepare_pullback_cache, require_tangent_cache
 
 # Public, exported
 export value_and_gradient!!, prepare_gradient_cache
--- a/src/tangents.jl
+++ b/src/tangents.jl
@@ -706,7 +706,7 @@ by avoiding cache overhead, but is only safe when the tangent structure is prova
 This function makes decisions about tangent types, not primal values. The key question is:
 "Could the tangent of this primal value contain circular references or shared mutable state?"
 
-The cache prevents infinite loops when traversing tangent structures that might contain:
+The cache prevents infinite loops and incorrect results when traversing tangent structures that might contain:
 - Circular references (A references B, B references A)
 - Aliasing (multiple references to the same mutable object)
 - Uninitialized fields that might later create cycles
--- a/src/tangents.jl
+++ b/src/tangents.jl
@@ -678,16 +678,16 @@ end
 end
 
 """
-    require_tangent_cache(::Type{T}) where {T}
+    require_tangent_cache(::Type{P}) where {P}
 
-Determines whether operations on tangents of type `T` require a cache to handle potential 
+Determines whether operations on tangents of primal type `P` require a cache to handle potential 
 circular references or aliasing. Returns `Val{true}()` if caching is required (the default),
-or `Val{false}()` if tangents of type `T` are guaranteed to be free of circular references,
+or `Val{false}()` if tangents of type `tangent_type(P)` are guaranteed to be free of circular references,
 uninitialized fields that could create circular references, and aliasing.
 
 This function is used internally by operations like `set_to_zero!!`. Returning `Val{false}()` 
 can improve performance by avoiding cache overhead, but is only safe when the memory layout
-of the tangent type is provably tree-like (no back edges). 
+of the tangent type is provably tree-like. 
 
 !!! warning "Advanced Performance Optimization"
     This is an advanced optimization hook. The default behavior (returning `Val{true}()`)
@@ -703,8 +703,8 @@ of the tangent type is provably tree-like (no back edges).
 
 ### Understanding Tangent Caching
 
-This function makes decisions based on the primal type `T` by answering the key question:
-"Could tangents of type `tangent_type(T)` contain circular references?"
+This function makes decisions based on the primal type `P` by answering the key question:
+"Could tangents of type `tangent_type(P)` contain circular references or aliasing?"
 
 The cache prevents infinite loops and incorrect results when traversing tangents that might contain:
 - Circular references (A references B, B references A)
@@ -712,10 +712,10 @@ The cache prevents infinite loops and incorrect results when traversing tangents
 
 ### Safety Requirements for `Val{false}()`
 
-Returning `Val{false}()` is only safe when the tangent type meomry layout is guaranteed to be tree-like,
+Returning `Val{false}()` is only safe when the tangent type memory layout is guaranteed to be tree-like,
 with no possibility of circular references or aliasing.
 
-#### Safe Cases (can return `Val{false}()`):
+#### Safe Cases (non-exhaustive, can return `Val{false}()`):
 
 1. **Pure immutable structures**: Structures with only `Tangent` types (no `MutableTangent`)
    cannot create cycles because immutable objects cannot reference themselves after construction
@@ -727,7 +727,7 @@ with no possibility of circular references or aliasing.
    like `PossiblyUninitTangent{Float64}`. For example, `mutable struct Bar; x::Ref{Float64}; end` 
    produces this safe tangent type because `Ref{Float64}` is concrete
 
-#### Unsafe Cases (must return `Val{true}()`): 
+#### Unsafe Cases (non-exhaustive, must return `Val{true}()`): 
 
 1. **Abstract typed fields**: Any field typed as `Any` or other abstract types can hold
    arbitrary values at runtime, including circular references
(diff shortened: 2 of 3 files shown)


sunxd3:
@mhauru @willtebbutt could you take a quick look and see if this is on the right track?

yebai (commented):
I suggest that we merge `requires_cahe` and `prepare_cache` into a single function `prepare_tangent_cache`. This would also be consistent with the existing `prepare_pullback_cache`. This is clearer and has fewer layers of redirection.

sunxd3:
Conceptually I don't object, but it requires exposure of `NoCache`.

yebai:
~~For clarity, it should return `NoCache()` object instead of `NoCache` so it won't expose `NoCache`.~~

EDIT: happy to keep the current design, but with a clearer name `require_tangent_cache` to replace `require_cache`.

src/Mooncake.jl
```diff
@@ -149,7 +149,7 @@ include("public.jl")
 end
 #! format: on
 
-@public Config, value_and_pullback!!, prepare_pullback_cache
+@public Config, value_and_pullback!!, prepare_pullback_cache, require_tangent_cache
```
yebai: As discussed when we met, no need to export this function. Let's also adopt the simpler design, i.e. `prepare_tangent_cache`. 

```suggestion
@public Config, value_and_pullback!!, prepare_pullback_cache
```

src/tangents.jl
```diff
@@ -677,6 +677,74 @@ end
...
+
+!!! warning "Advanced Performance Optimization - Use With Extreme Caution"
+    This is an advanced optimization. The default behavior (using a cache) is always correct
+    but may be slower. Only override if you have measured a significant performance impact
+    AND can prove your tangent types are safe.
+    
+    **When is it safe to return `Val{false}()`?**
+    Only when the tangent type is guaranteed to have:
+    - No circular references between tangent objects
+    - No uninitialized fields that might later create cycles
```
willtebbutt: I believe that we also need no aliasing. That is, the following also requires a cache (in general):
```julia
x = randn(5)
data_of_interest = (x, x) # thing we want to operate on
```

src/tangents.jl
```diff
@@ -677,6 +677,74 @@ end
...
+       - Objects storing callbacks that might capture the object itself
+    
+    4. **Complex shared structures**: When dealing with intricate data structures where
+       tracking visited nodes is necessary to avoid infinite loops
+    
+    5. **Recursive type definitions**: Types defined recursively often produce tangent
+       structures with circular references
+    
+    6. **Generic/Parametric types**: Even if your type seems simple, its type parameters
+       might have complex tangent structures
```
willtebbutt: I really like that you've included this section, but my feeling is that we ought to tighten it up a bit and make it more precise. I think the best way to do that would be to insert some doctests which provide concrete examples of failure modes. Is this something you would be open to doing?

src/tangents.jl
```diff
@@ -677,6 +677,74 @@ end
...
+"""
+    require_tangent_cache(x)
+
+Returns `Val{true}()` if the tangent type of `x` might contain circular references or 
+aliasing that requires a cache to track during operations like `set_to_zero!!`, 
+`increment!!`, etc. Returns `Val{false}()` otherwise.
+
+Note that this function takes a primal value but makes a decision about properties of
+its corresponding tangent type. The question is not whether the primal has circular
+references, but whether its tangent might.
```
willtebbutt: This is a slightly subtle point, so I wonder whether it ought to be moved further down to live alongside other examples (see comment below), and an example of when this distinction matters added? I can think of one situation in which this distinction matters: if you have a primal which contains circular references, but whose tangent type is `NoTangent`. If there are others that you can think of, please do include them!

src/tangents.jl
```diff
@@ -677,6 +677,74 @@ end
...
+
+The default implementation returns `Val{true}()` for non-bits types to be conservative.
```
willtebbutt:
```suggestion
```
My inclination is to remove this line, because it documents an implementation detail which I believe is quite likely to change. For example, we ought to also (at least) to add a method which returns `Val(false)` for `Vector{BitsType}`.

src/tangents.jl
```diff
@@ -677,6 +677,74 @@ end
...
+Package extensions can overload this for their specific types to assert that their
+tangent types contain no circular references or aliasing.
```
willtebbutt:
```suggestion
Package extensions can overload this for their specific data structures to assert that their
tangents do not require caches.
```
I'm not 100% about my wording, but I think phrasing this in a slightly more generic manner would be good.

willtebbutt (changes requested):
I think this is exactly the right kind of thing. I think the level of detail you've gone into with the docstring is exactly what this function calls for, but my feeling is that we can be more informative and precise.

Roughly speaking, I think it should be re-structured, and split into three sections:
1. general description of the purpose of the function, and a very brief description of the conditions under which it is safe to return `Val(false)` from this function.
2. the warning that you've included, which I think is a really excellent idea. It should strongly advise the reader to look at the extended help section if they are thinking about adding a method of their own.
3. an `## Extended Help` section, which gives a much more detailed description of what's going on, and concrete worked examples (preferably containing doctests).

I left some specific comments before writing this summary, but I think what I've said in this summary best describes my thoughts.

src/tangents.jl
sunxd3: Good point. Initially I wrote about aliasing, but deleted because I figure that, for aliasing, it's a performance improvement rather than correctness concern. Are there cases where aliasing can cause errors?

willtebbutt: It's not so much of an issue at the interface layer, but it can cause problems when testing correctness. For example, the implementation of [`increment_internal!!`](https://github.com/chalk-lab/Mooncake.jl/blob/e8828b53c221c68587e3c2af0523680d796b9d88/src/tangents.jl#L721) needs caching to ensure that we don't "double increment" in the above example.

willtebbutt: I think you're right that it's an optimisation for idempotent operations like `set_to_zero!!`.

src/tangents.jl
```diff
@@ -677,6 +677,185 @@ end
     end
 end
 
+"""
+    require_tangent_cache(x)
+
+Determines whether operations on the tangent of `x` require a cache to handle potential 
+circular references or aliasing. Returns `Val{true}()` if caching is required (the default),
+or `Val{false}()` if the tangent type is guaranteed to be free of circular references,
```
willtebbutt: As you suggest by refering to "tangent type" in this sentence, I think it is possibly true that we have all of the information we need to implement this function in the type of the primal. I wonder whether we should change this function to be implemented in terms of the primal type, rather than the primal value.

src/tangents.jl
```diff
@@ -677,6 +677,185 @@ end
...
+aliasing, and uninitialized fields that could create cycles.
```
willtebbutt:
```suggestion
uninitialized fields that could create circular references, and aliasing.
```
Just a slight tweak to prevent potential misunderstanding. The other ordering leave it open to the interpretation that the probem might be that aliasing can cause circular references, while this ordering makes it clear that aliasing is a problem whether or not it leads to circular references.

I've also suggested changing "cycles" to "circular references" so that there's no risk of thinking that "cycles" are something different from "circular references". Am I correct that this was your intent?

src/tangents.jl
```diff
@@ -677,6 +677,185 @@ end
...
+
+This function is used internally by operations like `set_to_zero!!` and `increment!!` to
+decide whether to track visited objects. Returning `Val{false}()` can improve performance
+by avoiding cache overhead, but is only safe when the tangent structure is provably acyclic.
```
willtebbutt:
```suggestion
by avoiding cache overhead, but is only safe when the tangent structure is provably tree-like.
```
Given that we're considering aliasing here too, I think tree-like is actually the correct thing, rather than acyclic. For example
```julia
x = randn(5)
tangent_of_interest = (x, x)
```
is acyclic, but the memory doesn't have a tree-like structure.

Also, I wonder whether we ought to refer to the "memory layout" of the tangent, rather than the (more generic) "tangent structure"? I'm not sure whether "memory layout" is the correct term here, or whether there's something better to call it that you're aware of.

src/tangents.jl
```diff
@@ -677,6 +677,185 @@ end
...
+    See the Extended Help section for detailed safety requirements and examples.
+
+# Extended help
+
+### Understanding Tangent Caching
+
+This function makes decisions about tangent types, not primal values. The key question is:
+"Could the tangent of this primal value contain circular references or shared mutable state?"
+
+The cache prevents infinite loops when traversing tangent structures that might contain:
```
willtebbutt:
```suggestion
The cache prevents infinite loops and incorrect results when traversing tangent structures that might contain:
```
because aliasing

willtebbutt (commented):
I've got a couple of small thoughts on in bit before the extended help, but I think we ought to discuss the examples in the extended help a bit more carefully.

I think example 2 is correct and good (because it demonstrates clearly what happens when you have a circular reference, but I think we might need to refine 1 and 3, and the safety requirements.

In particular, there are a situations in which I believe that `mutable tangent`s and `PossiblyUninitTangent`s are safe. For example, if you have primal type
```julia
mutable struct Foo
    x::Float64
end
```
which has a tangent type `MutableTangent` with a single `Float64` field (I believe), there is no risk of either aliasing or circular references.

Similarly, if we had primal type
```julia
mutable struct Foo
    x::Vector{Float64}
    Foo() = new()
end
```
there is also no risk of either circular references or aliasing in the tangent, since there is no risk of aliasing inside a `Vector{Float64}` (which is itself a mutable type).

Do you agree with the above, or do you think I might be missing something?

sunxd3:
Thanks @willtebbutt, my intention was to state the sufficient conditions. But upon reflection, it is too confusing and too restrict.

I did a rewrite on multiple parts, most deletion, also incorporated some of your suggestions. Another look?

src/tangents.jl
```diff
@@ -677,6 +677,184 @@ end
     end
 end
 
+"""
+    require_tangent_cache(::Type{T}) where {T}
+
+Determines whether operations on tangents of type `T` require a cache to handle potential 
+circular references or aliasing. Returns `Val{true}()` if caching is required (the default),
+or `Val{false}()` if tangents of type `T` are guaranteed to be free of circular references,
+uninitialized fields that could create circular references, and aliasing.
```
willtebbutt: Ah, sorry. Could you maybe call the type parameter `P` (to indicate "primal") and make it clear that the type passed in is the primal type, and we're asking about whether its tangents can contain circular references? Currently, it's possible for the reader to think that the argument is itself the tangent type, which is not the case.

src/tangents.jl
```diff
@@ -677,6 +677,184 @@ end
...
+The cache prevents infinite loops and incorrect results when traversing tangents that might contain:
+- Circular references (A references B, B references A)
+- Aliasing (multiple references to the same object)
+
+### Safety Requirements for `Val{false}()`
+
+Returning `Val{false}()` is only safe when the tangent type meomry layout is guaranteed to be tree-like,
+with no possibility of circular references or aliasing.
+
+#### Safe Cases (can return `Val{false}()`):
```
willtebbutt: Maybe indicate that this is a non-exhaustive list?

src/tangents.jl
```diff
@@ -677,6 +677,184 @@ end
...
+   cannot create cycles because immutable objects cannot reference themselves after construction
+   
+2. **Concrete-only mutable structs**: `MutableTangent` types where ALL fields have concrete types
+   that cannot hold references (e.g., `mutable struct Foo; x::Float64; end`)
+   
+3. **Concrete `PossiblyUninitTangent`**: When parameterized by concrete non-reference types
+   like `PossiblyUninitTangent{Float64}`. For example, `mutable struct Bar; x::Ref{Float64}; end` 
+   produces this safe tangent type because `Ref{Float64}` is concrete
+
+#### Unsafe Cases (must return `Val{true}()`):
```
willtebbutt: Similarly to the previous section, maybe highlight that this is non-exhaustive?

sunxd3:
@willtebbutt another look?

willtebbutt (approved):
In my view, this is excellent. Please feel free to go ahead and merge + tag a release once CI passes :)

sunxd3:
thanks Will!

PR #653, opened 2025-07-14
Add tangent_type for BigInt and rule for generate_dropout_mask

yebai:
This PR fixes a Lux issue with dropout (https://github.com/chalk-lab/Mooncake.jl/issues/563) and adds a test for https://github.com/chalk-lab/Mooncake.jl/issues/622

Close https://github.com/chalk-lab/Mooncake.jl/issues/563 and add tests.

PR #671, opened 2025-08-03
fix: move luxlib ext to mooncake

avik-pal:
I accidentally created a circular dependency in LuxLib by depending on Mooncake. Trying to fix that by moving the LuxLib patch here. (https://github.com/LuxDL/Lux.jl/issues/1416)

Needs to be followed up with a patch in General to upperbound Mooncake version in LuxLib. Then I will remove the ext from LuxLib


yebai:
Looks good, @avik-pal. Can you fix the formatting and add a test?

AstitvaAggarwal:
This needs tests added in `test\ext\luxlib\luxlib.jl` to be merged.

yebai:
https://github.com/chalk-lab/Mooncake.jl/pull/653 added a test already.

PR #680, opened 2025-08-07
Improve `set_to_zero!!` performance by using other data structures for caching

sunxd3:
@willtebbutt suggests that `IdDict` might be too much overhead and alternative data structures might be better.

This is some experiments along this line of thoughts.


sunxd3:
IdDict{Any,Bool}: 1358.3 ns
Set{UInt}:        757.7 ns (1.79x faster)
Vector{UInt}:     754.3 ns (1.8x faster)
NoCache:          318.8 ns (4.26x faster, unsafe)

on `set_to_zero!!` with `DynamicPPL.TestUtils.DEMO_MODELS[1]` (code [here](https://github.com/chalk-lab/Mooncake.jl/blob/sunxd/improve_set_to_zero/test_idset/benchmark_set_to_zero.jl))

willtebbutt:
This is cool. You're probably already thinking about this, but in the `Vector` case is it clear how much time is spent allocating it and re-sizing it? This would be interesting to know because we ought to be able to pay the price once (when `prepare_gradient_cache` is called, and re-cycle the data structure afterwards).

sunxd3:
> how much time is spent allocating it and re-sizing it

```
  This breaks down to:
  - Membership checking: ~28% (161/577)
  - Resizing/growing: ~5.4% (31/577)
  - Initial allocation: ~1.4% (8/577)
  - Push/append: ~3.5% (20/577)
  - Memory allocation: ~0.9% (5/577)
```
(LLM's break down)
not much, about 10% of the run time, in contrast `in` takes about `30%` of the run time

image

for `Set`

```
Breaking down:
  - Hash operations & lookups: ~21% (120/562)
  - Push/insert: ~18% (100/562)
  - Membership checking: ~7% (40/562)
  - Resizing: ~6% (36/562)
  - Initial allocation: ~0.4% (2/562)
```

image

for the case of this simple DPPL model, hashing is a big overhead.

sunxd3:
On saving the cache vector (say we take this route), is there a reason why we wouldn't just use `NoCache` for all the `set_to_zero!!`s after the first one?

willtebbutt:
Ah, right, yes. It's possible (albeit unlikely) that the user would call `prepare_gradient_cache` with arguments that do not contain any aliasing / circular references, but then call `value_and_gradient!!` with arguments (of the same type) which do contain aliasing / circular references. Since silent correctness issues are by far the worst thing that an AD system can do, I really do think it's best to exclude the possibility of this unlikely situation if it's not prohibitively expensive.

yebai:
> It's possible (albeit unlikely) that the user would call prepare_gradient_cache with arguments that do not contain any aliasing / circular references, but then call value_and_gradient!! with arguments (of the same type) which do contain aliasing / circular references. 

Do you think we can protect against this via an assertion? If so, @sunxd3's suggestion below would work nicely 

> On saving the cache vector (say we take this route), is there a reason why we wouldn't just use NoCache for all the set_to_zero!!s after the first one?

willtebbutt:
The question is: what are you asserting? If you're looking to assert that there is no aliasing / circular references, then you're basically doing the same thing as `set_to_zero!!` is doing. Did you have something else in mind?

yebai:
Assuming that no arguments contain any aliasing/circular references during `prepare_gradient_cache`. Then when calling `set_to_zero!!` later, one can push all `objectid`s into a preallocated `Vector`, then perform `unique` at the end to determine whether specific `objectid` appear multiple times. If so, throw an error. Since `push!` is fast when given preallocated `Vector` and there is no `in` operation, I expect this will be quite close to `NoCache`.

willtebbutt:
Hmmm interesting. I wonder how expensive running `allunique` is vs the current approach -- it would be good to see a benchmark. I find it hard to reason about what is going to be fast for small collections. @sunxd3 's benchmarking indicates that quite a lot of time is spent computing `objectid`, so it might be that all approaches have similar cost in practice.

sunxd3:
running locally with DynamicPPL model with https://turinglang.org/ADTests/. (ReverseDiff is slower on Apple silicon vs x64)

this is benchmarking the gradient computation, not just `set_to_zero!!`, it is consistent with what @willtebbutt and I have seen offline - 2.X time improvement on `set_to_zero!!`.

| Model | Mooncake | Mooncake (optimized) | Speedup (optimization) | ReverseDiffCompiled |
|:--|--:|--:|--:|--:|
| demo_assume_dot_observe | 6.35 | 5.79 | 1.10x | 13.34 |
| demo_assume_dot_observe_literal | 7.70 | 5.79 | 1.33x | 10.48 |
| demo_assume_index_observe | 5.86 | 4.41 | 1.33x | 10.26 |
| demo_assume_matrix_observe_matrix_index | 7.07 | 5.55 | 1.27x | 9.09 |
| demo_assume_multivariate_observe | 6.79 | 5.23 | 1.30x | 11.06 |
| demo_assume_multivariate_observe_literal | 6.70 | 4.91 | 1.36x | 11.18 |
| demo_assume_observe_literal | 7.84 | 5.45 | 1.44x | 11.71 |
| demo_assume_submodel_observe_index_literal | 5.76 | 4.35 | 1.32x | 10.20 |
| demo_dot_assume_observe | 6.50 | 5.19 | 1.25x | 11.40 |
| demo_dot_assume_observe_index | 6.65 | 5.01 | 1.33x | 12.78 |
| demo_dot_assume_observe_index_literal | 6.81 | 5.04 | 1.35x | 12.56 |
| demo_dot_assume_observe_matrix_index | 5.97 | 4.75 | 1.26x | 11.09 |
| demo_dot_assume_observe_submodel | 5.02 | 3.93 | 1.28x | 8.54 |

<details>
  <summary>benchmark script</summary>

to benchmark between two versions of Mooncake, need manually `develop Mooncake`; run the script in environment `ADTest`

```julia
const DYNAMICPPL_MODELS = [
    "demo_assume_dot_observe",
    "demo_assume_dot_observe_literal",
    "demo_assume_index_observe",
    "demo_assume_matrix_observe_matrix_index",
    "demo_assume_multivariate_observe",
    "demo_assume_multivariate_observe_literal",
    "demo_assume_observe_literal",
    "demo_assume_submodel_observe_index_literal",
    "demo_dot_assume_observe",
    "demo_dot_assume_observe_index",
    "demo_dot_assume_observe_index_literal",
    "demo_dot_assume_observe_matrix_index",
    "demo_dot_assume_observe_submodel"
]

const ADTYPES = ["Mooncake", "ReverseDiffCompiled"]

# Create output directory
mkpath("benchmark_results")

# Run benchmarks for each combination
for adtype in ADTYPES
    results = String[]
    for model in DYNAMICPPL_MODELS
        cmd = `julia --project=. main.jl --run $model $adtype`
        result = try
            output = read(cmd, String)
            output = strip(output)
            println("Result: $output")
            "$model: $output"
        catch e
            println("Failed with error")
            "$model: error"
        end
        
        push!(results, result)
    end
    
    # Save results to file
    filename = "benchmark_results/dynamicppl_$(lowercase(adtype))_results.txt"
    open(filename, "w") do f
        println(f, "Benchmark Results for DynamicPPL Models with $adtype")
        println(f, "=" ^ 60)
        for result in results
            println(f, result)
        end
    end
    
    println("\nResults saved to: $filename")
end
```

</details>

willtebbutt:
This is a massive improvement. This is with the vector implementation?

sunxd3:
a current summary: for models like https://github.com/TuringLang/ADTests/blob/main/models/demo_assume_dot_observe.jl and https://github.com/TuringLang/DynamicPPL.jl/blob/ea6b6de5db56d408e4befbba1d0f211a74840f71/src/test_utils/models.jl#L189-L198, the overhead of caching is about 45% of `set_to_zero!!`, while `set_to_zero!!` is about 40% of computing gradient ([source](https://github.com/chalk-lab/Mooncake.jl/issues/552#issuecomment-3131998657)).

sunxd3:
> This is a massive improvement. This is with the vector implementation?

yes (but without the `NoTangent` for `Memory{NonDifferentiableType}` optimization)

src/tangents.jl
```diff
@@ -813,7 +813,8 @@ same tangent twice and producing incorrect results.
 """
 require_tangent_cache(::Type{P}) where {P} = Val{!isbitstype(P)}()
```
yebai, line 814: Maybe rename this to `require_SetToZeroCache` for consistency? 
```suggestion
require_SetToZeroCache(::Type{P}) where {P} = Val{!isbitstype(P)}()
```

sunxd3:
@willtebbutt I don't have a clear idea how to fix the tests, could you give it a quick look and let me know how?

src/tangents.jl
willtebbutt: I don't think this is needed -- `require_tangent_cache` is a more general piece of functionality which applies to a range of functions (`increment!!`, `_dot`, etc). We're not currently making use of it in this PR, but we ought to do so in a follow-up PR.

willtebbutt:
Fix should just be changing https://github.com/chalk-lab/Mooncake.jl/blob/feffb6a1b1b60cf8851bbc7338e34d0729d0f8da/src/test_utils.jl#L883 to pass in a `Vector{UInt}` rather than an `IdDict`. This is an annoying (but necessary, I think) piece of code duplication.

sunxd3:
I adjusted the membership test function (was `prod` now just use `in`) and benchmarked again

| Model | Mooncake | Mooncake (optimized) | Speedup (optimization) | ReverseDiffCompiled |
|:--|--:|--:|--:|--:|
| demo_assume_dot_observe | 6.35 | 5.90 | 1.08x | 13.34 |
| demo_assume_dot_observe_literal | 7.70 | 5.95 | 1.29x | 10.48 |
| demo_assume_index_observe | 5.86 | 4.70 | 1.25x | 10.26 |
| demo_assume_matrix_observe_matrix_index | 7.07 | 5.37 | 1.32x | 9.09 |
| demo_assume_multivariate_observe | 6.79 | 5.36 | 1.27x | 11.06 |
| demo_assume_multivariate_observe_literal | 6.70 | 5.34 | 1.26x | 11.18 |
| demo_assume_observe_literal | 7.84 | 5.20 | 1.51x | 11.71 |
| demo_assume_submodel_observe_index_literal | 5.76 | 4.44 | 1.30x | 10.20 |
| demo_dot_assume_observe | 6.50 | 5.16 | 1.26x | 11.40 |
| demo_dot_assume_observe_index | 6.65 | 5.14 | 1.29x | 12.78 |
| demo_dot_assume_observe_index_literal | 6.81 | 5.01 | 1.36x | 12.56 |
| demo_dot_assume_observe_matrix_index | 5.97 | 5.10 | 1.17x | 11.09 |
| demo_dot_assume_observe_submodel | 5.02 | 4.18 | 1.20x | 8.54 |

compare to https://github.com/chalk-lab/Mooncake.jl/pull/680#issuecomment-3167417420, pretty much can be considered the same, modulo benchmark noise.

src/tangents.jl
```diff
@@ -867,29 +868,32 @@ end
...
 end
-function set_to_zero_internal!!(c::IncCache, x::MutableTangent)
-    haskey(c, x) && return x
-    setindex!(c, false, x)
+function set_to_zero_internal!!(c::SetToZeroCache, x::MutableTangent)
+    if c isa Vector{UInt}
+        oid = objectid(x)
+        oid in c && return x
+        push!(c, oid)
+    end
```
willtebbutt: IIUC, this 5 lines of code is repeated everywhere. Since it's not just coincidental copying, could you add a function to do it, and replace all occurences of these 5 lines with a function call?

sunxd3: Done now, what do you think?

willtebbutt: Perfect.

willtebbutt:
Please feel free to merge @sunxd3

PR #686, opened 2025-08-12
Fix Turing integration test for DynamicPPL 0.37

penelopeysm:
I think this should fix it. BTW, would it be more technically correct to call it a DynamicPPL integration test (not Turing)?


yebai:
> BTW, would it be more technically correct to call it a DynamicPPL integration test (not Turing)?

Can you fix that too?

PR #693, opened 2025-08-13
Fix changelog version

penelopeysm:


yebai:
Nice catch -- thanks @penelopeysm

PR #699, opened 2025-08-15
minimal print io output

### LINKED ISSUE #684: Very noisy test suite output
I'm not exactly sure when it happened, but it appears to be the case that the `avoiding_non_differentiable_code` test suite is printing an awful lot to the terminal. It would be great if this could be suppressed somehow. I had to do this inside the standard testing functionality (see the `test_rule` function in #389 -- for some reason I can't link to it directly) to make doctests work sensibly, so I wonder if similar bits of Julia functionality can be used to sort this problem out.

AstitvaAggarwal:
closes #684 . this loop https://github.com/LilithHafner/Chairmarks.jl/blob/52678207dfb4b4ca7ba33fa4a224d13c1acde4c2/src/benchmarking.jl#L114 in `Chairmarks.jl` calls the function being benchmarked multiple times. for `print, show, println` and other such function that directly  `write` to chosen io stream cause loud outputs while being benchmarked. Not a forward mode bug/problem.

In both, handwritten or derived rule case : `(..., print, "Testing print")` or `(..., (x) -> print(x), "Testing print")`, we will face this issue. Its just that in current tests [`benchmark_hand_written_rrules`](https://github.com/chalk-lab/Mooncake.jl/blob/727ef49941611436f4b1d6e41cac2083a1fb8a0c/bench/run_benchmarks.jl#L301) has to be run for all primitive rule test sets while [`benchmark_derived_rrules!!`](https://github.com/chalk-lab/Mooncake.jl/blob/727ef49941611436f4b1d6e41cac2083a1fb8a0c/bench/run_benchmarks.jl#L322) is limited to the `:test_resources` derived rule test set.

### DIFF
--- a/src/rrules/avoiding_non_differentiable_code.jl
+++ b/src/rrules/avoiding_non_differentiable_code.jl
@@ -138,7 +138,6 @@ function generate_hand_written_rrule!!_test_cases(
             ("hello" => "hi",),
             1,
         ),
-        (false, :none, nothing, show, "Testing show"),
 
         # non-kwargs sprint rule test
         (false, :stability, nothing, sprint, show, "Testing sprint"),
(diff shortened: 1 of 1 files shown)


yebai (approved):
Good investigation, @AstitvaAggarwal. The explanation can be clearer. Presumably, `(..., (x) -> print(x), "Testing print")`'s output got redirected, but `(..., print, "Testing print")`'s didn't. Can you give a more crisp explanation for why? 

Feel free to merge and tag a release.

Issue #704, opened 2025-08-17
CI errors in logging related tests


yebai:
I noticed a few CI errors arising from logging-related tests, see eg,

https://github.com/chalk-lab/Mooncake.jl/actions/runs/17025944433/job/48261320366?pr=703

This was fine on the `main` branch, so not sure why they are broken on PRs with no code changes, e.g. #703

PR #705, opened 2025-08-26
Fix 660

### LINKED ISSUE #660: `prep` fails with wrong type, `outer::UnionAll`
I'd like to switch to this AD. It works in most of my cases, however, I just hit a wall.
Simplified example drastically, and still do not see why it fails.

The AD cannot get the type right when pulling through flow.

the line
```julia
prep = prepare_gradient(loss, ADbackend, θ) # breaks
```

leads to
```julia
ERROR: TypeError: in typeassert, expected
Mooncake.CoDual{@NamedTuple{outer::Type{Trivial}, inner::Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}, @NamedTuple{outer::Mooncake.NoFData, inner::Mooncake.FData{@NamedTuple{weight::Matrix{Float32}, bias::Vector{Float32}, σ::Mooncake.NoFData}}}}, got a value of type 
Mooncake.CoDual{@NamedTuple{outer::UnionAll, inner::Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}, @NamedTuple{outer::Mooncake.NoFData, inner::Mooncake.FData{@NamedTuple{weight::Matrix{Float32}, bias::Vector{Float32}, σ::Mooncake.NoFData}}}}
Stacktrace:
```
where `outer::Type{Trivial}` is not correctly inferred.

## Full example

EDIT: simplified a bit

```julia
using Bijectors
using NormalizingFlows.Optimisers
import Flux: Dense, Chain, relu
using Mooncake: Mooncake

using Random
rng = MersenneTwister(42)


struct Trivial{T} <: Bijectors.Bijector
    params::AbstractVector{T}
end

transform(b::Trivial, x) = with_logabsdet_jacobian(b, x)[1]
logabsdetjac(b::Trivial, x) = with_logabsdet_jacobian(b, x)[2]

Bijectors.with_logabsdet_jacobian(b::Trivial, x::Real) = x, 0.0
Bijectors.with_logabsdet_jacobian(b::Trivial, x::AbstractVector) = x, 0.0
Bijectors.with_logabsdet_jacobian(ib::Inverse{<:Trivial}, y::Real) = y, 0.0
Bijectors.with_logabsdet_jacobian(ib::Inverse{<:Trivial}, y::AbstractArray) = y, 0.0

const conditioner = Dense(1, 7)

flow = let
    q₀ = Distributions.Product([Uniform(0, 1), Uniform(0, 1)])
    mask = Bijectors.PartitionMask(2, [1], [2])
    ts = Coupling(Trivial ∘ conditioner, mask)
    transformed(q₀, ts)
end

θ, re = Optimisers.destructure(flow)
loss(θ) = -logpdf(re(θ), rand(Float32, 2))

loss(θ) # works
ADbackend = AutoMooncake(; config = nothing)
prep = prepare_gradient(loss, ADbackend, θ) # breaks
```

willtebbutt:
Resolves #660 . The issue was using `Mooncake._typeof` rather than `Base.typeof` in this particular instance. They generally give the same result, but `Mooncake._typeof` can sometimes give a more specific type than `Base.typeof`, speifically when asking for the type of a type.

Anyway, I've added a regression tests and checked manually against Markus' MWE (thanks for minimising as much as you did @mhauru ).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.150"
+version = "0.4.151"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/src/interpreter/reverse_mode.jl
+++ b/src/interpreter/reverse_mode.jl
@@ -1746,7 +1746,7 @@ DynamicDerivedRule(debug_mode::Bool) = DynamicDerivedRule(Dict{Any,Any}(), debug
 _copy(x::P) where {P<:DynamicDerivedRule} = P(Dict{Any,Any}(), x.debug_mode)
 
 function (dynamic_rule::DynamicDerivedRule)(args::Vararg{Any,N}) where {N}
-    sig = Tuple{map(_typeof ∘ primal, args)...}
+    sig = Tuple{map(typeof ∘ primal, args)...}
     rule = get(dynamic_rule.cache, sig, nothing)
     if rule === nothing
         interp = get_interpreter(ReverseMode)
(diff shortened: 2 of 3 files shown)


willtebbutt:
The fix fixes our failing test case, but seems to cause issues in other situations. I'll have to revisit this later.

willtebbutt:
I believe I'm correct in saying that the failures in the `avoiding_non_differentiable_code` tests are related to #704 .

willtebbutt:
Good idea. I'll do that and merge.

PR #715, opened 2025-08-27
Tidy up 

willtebbutt:
Copilot didn't to the best job in #710 , so I've tidied up a bit. Specificially:
- `error(...)` throws, so you don't need to write `throw(error(...))`
- I've bumped the patch version so we can make a release
- Copilot wrote a bunch of redundant tests. It's enough to check that we hit the errors and roughly the string that we expected is produced when an error occurs. Since the implementation is so simple, there's really no reason to test multiple types etc.


yebai (approved):

willtebbutt:
Thanks for the review. I'll merge when CI finishes.

PR #716, opened 2025-08-27
fix CI errors.

### LINKED ISSUE #704: CI errors in logging related tests

AstitvaAggarwal:
closes #704. `Mooncake.zero_rdata_from_type` needs new overloads for every new super type whose subtype's  object are encountered in `Mooncake.rrule!!` eg. Subtypes of `Base.CoreLogging.AbstractLogger`.

### DIFF
--- a/test/fwds_rvs_data.jl
+++ b/test/fwds_rvs_data.jl
@@ -1,8 +1,5 @@
 module FwdsRvsDataTestResources
 
-using Test: TestLogger
-using Logging: ConsoleLogger
-
 struct Foo{A} end
 struct Bar{A,B,C}
     a::A
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,5 +1,8 @@
 include("front_matter.jl")
 
+using Test: TestLogger
+using Logging: ConsoleLogger
+
 @testset "Mooncake.jl" begin
     if test_group == "quality"
         Aqua.test_all(Mooncake)


AstitvaAggarwal:
Im done here @yebai.

src/rrules/avoiding_non_differentiable_code.jl
```diff
@@ -88,6 +88,9 @@ import Base.CoreLogging as CoreLogging
     }
 )
 
+# zero_rdata_from_type must handle new object types.
+zero_rdata_from_type(::Type{T}) where {T<:Base.CoreLogging.AbstractLogger} = NoRData()
```
yebai, line 92: I suggest we add a unit test for this, so it is easier to debug in the future.

yebai (commented):
Good work, @AstitvaAggarwal -- I left a minor comment below.

src/rrules/avoiding_non_differentiable_code.jl
yebai: @willtebbutt, is there a reason why we don't have a generic fallback: 

```julia
zero_rdata_from_type(::Type{T}) = rdata(set_to_zero!!(tangent_type(T)()))
```

This would work for all cases where the `tangent_type` is defined to be `NoTangent()`.

src/rrules/avoiding_non_differentiable_code.jl
```diff
@@ -88,6 +88,10 @@ import Base.CoreLogging as CoreLogging
     }
 )
 
+# Mooncake.zero_rdata_from_type() must handle new object types.
+# Normally it should not be defined for AbstractTypes as subtypes may have different rdata, but for Logging all Logger subtypes have rdata as NoRData().
```
yebai:
```suggestion
# New `zero_rdata_from_type` overloads are required for `Logger` subtypes, which violates Mooncake’s usual assumption for default `zero_rdata_from_type` implementation. 
```

yebai (approved):
Feel free to merge after fixing the comment below and addressing the formatting issues.

src/rrules/avoiding_non_differentiable_code.jl
willtebbutt: We already have this. You're asking here to set the value of the tangent _type_ to zero, which doesn't make sense. You can only apply `set_to_zero!!` to instances of a type.

willtebbutt: > We already have this.

By which I mean that if `rdata_type(tangent_type(P))` is `NoRData`, `zero_rdata_from_type` will just return `NoRData()`.

willtebbutt:
Unless I'm mistaken, I don't think that this method is doing anything new here. There's a method of `tangent_type` for `AbstractLogger` which returns `NoTangent`. `rdata_type(NoTangent) == NoRData`, so the default method of `zero_rdata_from_type` will yield `NoRData()`. Might be worth having a look to see whether this is the case.

AstitvaAggarwal:
```
julia> Mooncake.zero_rdata_from_type(Test.TestLogger)
NoRData()

julia> Mooncake.zero_rdata_from_type(AbstractLogger)
ERROR: ArgumentError: type does not have a definite number of fields
Stacktrace:
 [1] fieldcount
   @ .\reflection.jl:895 [inlined]
 [2] fieldnames
   @ .\reflection.jl:167 [inlined]
 [3] #s11#142
   @ C:\Users\astit\OneDrive\Desktop\GSOC\Julia\contri\Mooncake.jl\src\fwds_rvs_data.jl:693 [inlined]
 [4] var"#s11#142"(P::Any, ::Any, ::Any)
   @ Mooncake .\none:0
 [5] (::Core.GeneratedFunctionStub)(::UInt64, ::LineNumberNode, ::Any, ::Vararg{Any})
   @ Core .\boot.jl:602
 [6] top-level scope
   @ REPL[54]:1

julia> rdata_type(Mooncake.tangent_type(AbstractLogger))
NoRData

julia> rdata_type(Mooncake.tangent_type(Test.TestLogger))
NoRData
```
@willtebbutt  This is without the current overload on `zero_rdata_from_type()`. As this method checks for [any](https://github.com/chalk-lab/Mooncake.jl/blob/14ce26cef43495840d137fcf757f79046d145ef6/src/fwds_rvs_data.jl#L693) DataType's fields, it fails for the supertype `AbstractLogger` wrapper struct.

yebai:
https://github.com/chalk-lab/Mooncake.jl/pull/736 should address this more generally.

PR #730, opened 2025-08-28
Minor bugfix in `rdata_type`

yebai:


willtebbutt:
It looks like this bug was hit because we never tested this code path. Could you declare add a unit test? (declare a new primitive type in the test suite, and check that this error is thrown)

yebai:
@AstitvaAggarwal, can you help add a unit test for this? Feel free to ask questions if anything is unclear.

yebai:
Thanks @AstitvaAggarwal.

PR #756, opened 2025-09-02
Specialise require_tangent_cache for Arrays + add test cases

### LINKED ISSUE #523: Allocation-free gradient?
Hi there @willtebbutt,
Do you think there is a way to get allocation-free in-place gradients with Mooncake?

```julia
using Chairmarks
using Mooncake: Mooncake

f(x) = sum(abs2, x)
x = rand(10)
g = zeros(10)
prep = prepare_gradient(f, backend, x)
@b gradient!($f, $g, $prep, $backend, $x)  # 48.285 ns (2 allocs: 320 bytes)
@profview for _ in 1:1_000_000; gradient!(f, g, prep, backend, x); en
```

Profiling shows that `set_to_zero!!` allocates an `IdDict{Any,Bool}`. I don't really understand what it is used for, presumably to avoid setting something to zero twice? If that is correct, then it is useless whenever the struct in question has no aliasing?

I'm thinking perhaps we could add a parameter to the Mooncake config that chooses between naive zeroing (zero everything even when it has already been seen) and clever zeroing (remember the objects that have been seen).
However this trick won't work for incrementing, because unlike zeroing it is not idempotent.
An alternative might be to store these `IdDict`s in the rule/cache object itself? But that doesn't solve the issue of `setindex!`, which is type-unstable due to the `Any`s.

As a motivation, as soon as we get alloc-free gradients with Mooncake, the refactor I've done in https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/743 will give us alloc-free Hessians (dense and sparse) with FiniteDiff-over-Mooncake.

willtebbutt:
Resolves #523

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.156"
+version = "0.4.157"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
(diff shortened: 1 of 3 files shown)


willtebbutt:
I'll merge when CI passes.

PR #759, opened 2025-09-03
Fix inlining avoidance

willtebbutt:
~~Don't review this yet please -- I just want to run CI.~~

I'm pleased to say that this is now ready for review. I've checked, and I'm pleased to say that (as expected) it resolves https://github.com/chalk-lab/Mooncake.jl/issues/558 . I believe that the unit test added in `test/interpreter/abstract_interpretation.jl` is fairly representative of what was going on in that issue.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -317,6 +317,19 @@ function _copy_to_output!!(dst::P, src::P) where {P}
     end
 end
 
+# fallback for invalid type combinations
+function _copy_to_output!!(dst::T, src::P) where {T,P}
+    throw(
+        ArgumentError(
+            "Mooncake.jl does not currently have a method " *
+            "`_copy_to_output!!` to handle this type combination: " *
+            "dst passed is of type $T, while src is a $P. " *
+            "This often happens when differentiating over " *
+            "non-differentiable types (e.g. integers or booleans).",
+        ),
+    )
+end
+
 """
     _copy_output(x::T)
 
--- a/src/interpreter/abstract_interpretation.jl
+++ b/src/interpreter/abstract_interpretation.jl
@@ -72,8 +72,8 @@ function Base.show(io::IO, mime::MIME"text/plain", mc::MooncakeInterpreter)
 end
 Base.show(io::IO, mc::MooncakeInterpreter) = _show_interp(io, MIME"text/plain"(), mc)
 
-function _show_interp(io::IO, ::MIME"text/plain", ::MooncakeInterpreter)
-    return print(io, "MooncakeInterpreter()")
+function _show_interp(io::IO, ::MIME"text/plain", ::MooncakeInterpreter{C,M}) where {C,M}
+    return print(io, "MooncakeInterpreter($M)")
 end
 
 MooncakeInterpreter(M::Type{<:Mode}) = MooncakeInterpreter(DefaultCtx, M)
--- a/src/interpreter/reverse_mode.jl
+++ b/src/interpreter/reverse_mode.jl
@@ -1042,7 +1042,7 @@ end
 
 function Base.showerror(io::IO, err::MooncakeRuleCompilationError)
     msg =
-        "MooncakeRuleCompilationError: an error occured while Mooncake was compiling a " *
+        "MooncakeRuleCompilationError: an error occurred while Mooncake was compiling a " *
         "rule to differentiate something. If the `caused by` error " *
         "message below does not make it clear to you how the problem can be fixed, " *
         "please open an issue at github.com/chalk-lab/Mooncake.jl describing your " *
--- a/src/rrules/low_level_maths.jl
+++ b/src/rrules/low_level_maths.jl
@@ -74,6 +74,7 @@
 @from_chainrules MinimalCtx Tuple{typeof(atan),P,P} where {P<:IEEEFloat}
 @from_chainrules MinimalCtx Tuple{typeof(max),P,P} where {P<:IEEEFloat}
 @from_chainrules MinimalCtx Tuple{typeof(min),P,P} where {P<:IEEEFloat}
+@from_chainrules MinimalCtx Tuple{typeof(mod),P,P} where {P<:IEEEFloat}
 
 @is_primitive MinimalCtx Tuple{typeof(mod2pi),IEEEFloat}
 function frule!!(::Dual{typeof(mod2pi)}, x::Dual{P}) where {P<:IEEEFloat}
@@ -283,6 +284,8 @@ function hand_written_rule_test_cases(rng_ctor, ::Val{:low_level_maths})
                 (deg2rad, P(185.4)),
                 (rad2deg, P(0.45)),
                 (mod2pi, P(0.1)),
+                (mod, P(7.5), P(2.3)),
+                (mod, P(10.2), P(3.1)),
                 (^, P(4.0), P(5.0)),
                 (atan, P(4.3), P(0.23)),
                 (hypot, P(4.0), P(5.0)),


yebai:
I'll leave this to @gdalle now that he is a co-author of Mooncake.

src/interpreter/contexts.jl
```diff
@@ -1,20 +1,20 @@
...
-
 """
     struct DefaultCtx end
 
 Context for all usually used AD primitives. Anything which is a primitive in a MinimalCtx is
 a primitive in the DefaultCtx automatically. If you are adding a rule for the sake of
 performance, it should be a primitive in the DefaultCtx, but not the MinimalCtx.
 """
-struct DefaultCtx end
+abstract type DefaultCtx end
```
yebai: Converting an existing type `DefaultCtx` into an abstract type and swapping the subtype relationship between `DefaultCtx` and `MinimalCtx` can be considerably confusing to people who are already familiar with the context system. Maybe introduce a new type `AbstractCtx` and deprecate `DefaultCtx`?  

```suggestion
abstract type AbstractCtx end
```

src/interpreter/contexts.jl
```diff
@@ -1,20 +1,20 @@
...
+
+"""
+    struct MinimalCtx end
+
+Functions should only be primitives in this context if not making them so would cause AD to
+fail. In particular, do not add primitives to this context if you are writing them for
+performance only -- instead, make these primitives in the DefaultCtx.
+"""
+abstract type MinimalCtx <: DefaultCtx end
```
yebai: See above. 

```suggestion
abstract type MinimalCtx <: AbstractCtx end
```

src/interpreter/contexts.jl
```diff
@@ -109,10 +99,105 @@ end
...
+    world::UInt
+end
+
+const _IS_PRIMITIVE_CACHE = IdDict{Any,Bool}()
+
+"""
+    is_primitive(ctx::Type, mode::Type{<:Mode}, sig::Type{<:Tuple}, world::UInt)
+
+Returns a `Bool` specifying whether the methods specified by `sig` are considered primitives
+in the context of contexts of type `ctx` in mode `mode` at world age `world`.
```
yebai:
```suggestion
in the context of context `ctx` in mode `mode` at world age `world`.
```

src/interpreter/contexts.jl
```diff
@@ -109,10 +99,105 @@ end
...
             ::Type{$(esc(Tctx))}, ::Type{<:$(Tmode)}, ::Type{<:$(esc(sig))}
         )
             return true
         end
     end
 end
+
+struct PrimitiveCacheKey{tt}
+    world::UInt
+end
```
yebai: This struct doesn't seem to be used anywhere. 
```suggestion
```

src/interpreter/contexts.jl
```diff
@@ -109,10 +99,105 @@ end
...
+
+const _IS_PRIMITIVE_CACHE = IdDict{Any,Bool}()
+
+"""
+    is_primitive(ctx::Type, mode::Type{<:Mode}, sig::Type{<:Tuple}, world::UInt)
```
yebai, line 114: The argument `world` is new in this PR. Perhaps provide a brief explanation for the benefit of the future audience, why this is needed?

src/interpreter/contexts.jl
willtebbutt: Good catch.

src/interpreter/contexts.jl
willtebbutt: Hmmm. I don't think I would be in favour of this change, because changing `DefaultCtx` to an `abstract type` is meant to be an implementation detail, rather than a change to the behaviour of `DefaultCtx` (ie. from the perspective of which rules are considered primitives by `is_primitive`, nothing should have changed).

I would be happy to extend the docstring to explain this -- would that help?

src/interpreter/contexts.jl
willtebbutt: Good point, will do.

willtebbutt: Docstring extended.

src/interpreter/contexts.jl
yebai: I see your point. The main concern is that `MinimalCtx` becomes a subtype of `DefaultCtx`. Intuitively, DefaultCtx feels more minimal, since subtypes usually inherit their parent type’s behaviour. Even for someone familiar with Mooncake’s design, like me, this creates a bit of mental overhead each time I think about it.

src/interpreter/contexts.jl
```diff
@@ -109,10 +99,126 @@ end
...
+```jldoctest
+julia> using Mooncake: is_primitive, DefaultCtx, ReverseMode
+
+julia> is_primitive(DefaultCtx, ReverseMode, Tuple{typeof(sin), Float64}, Base.get_world_counter())
+true
+```
+
+`world` is needed as rules which Mooncake derives are associated to a particular Julia world
+age. As a result, anything declared a primitive after the construction of a rule ought not
+to be considered a primitive by that rule. To see how this works, consider the following:
```
yebai: I'd clarify further by explaining how to override an existing rule from a world age where primitives are different. Feel free to edit the wording as you see fit. 

```suggestion
to be considered a primitive by that rule. One can explicitly derive a new rule via `build_rrule` after new functions are declared as primitive, should it be needed in cases where a rule has been previously derived. To see how this works, consider the following:
```

sunxd3 (commented):
In general I see no problem.

(Maybe for future) it would be good to add some documentation in `/docs` on the difference between `maybe_primitive` and `is_primitive`. The docstrings make it pretty clear, but still good to have a synthesized writing -- especially when the abstract interpretation is probably the most confusing part of the whole package. This is non-blocking obviously.

src/interpreter/contexts.jl
willtebbutt: That's fair enough -- I also find it a bit confusing to be honest. Sadly, I think it probably is necessary in order to produce a correct implementation of `maybe_primitive`. I'll add a note to the implementation to remind the reader that it's an implementation detail.

willtebbutt: Okay, docstring extended + a comment added to the implementation of `is_primitive`. Do you think this suffices? It doesn't make my head hurt any less, but it's at least now signposted as being a bit tricky.

src/interpreter/contexts.jl
willtebbutt: This is a good point -- I've merged something similar to this into the docstring (it has some proper cross references to the functions you might use to build a rule).

src/interpreter/contexts.jl
yebai: Thanks @willtebbutt! 

I agree it’s a slight annoyance to change the current design in this PR.
That said, I think there’s a straightforward solution and that it would be worth it for the sake of clarity.

Specifically, this would make the code easier to follow:
- Remove the subtyping MinimalCtx <: DefaultCtx and treat both context types equally.
- Update is_primative and maybe_primative for DefaultCtx so they automatically call the corresponding functions internally, rather than relying on dispatch semantics. For example, 

```julia
funciton is_primative(ctx::DefaultCtx, args...; kwargs...) 
  ...
  # check whether something is considered as primitive by `MinimalCtx`. 
  is_primative(ctx::MinimalCtx, args..., kwargs...)
  ...
end
```

willtebbutt: Hmm. I agree that this would provide a correct implementation.

I've been thinking about this some more, and I think I have a good way to think about this. All `MinimalCtx` primitives are `DefaultCtx` primitives, but not all `DefaultCtx` primitives are `MinimalCtx` primitives. This is analogous to all `Float64`s being `Real`s, but not all `Real`s being `Float64`. Does this description make the subtyping relationship clearer?

yebai: I’ve thought about this example myself—it makes sense if we keep the particular case in mind, but it can feel a bit counterintuitive in a broader context. Since I’m onboarding new RA members, I’d suggest prioritising code readability.

willtebbutt: > but it can feel a bit counterintuitive in a broader context

What do you mean by the broader context here?

> Since I’m onboarding new RA members, I’d suggest prioritising code readability.

I'm always pro readability, I'm just pushing back on changing this implementation detail because my view on it has changed over the weekend. Previously, I thought it distracted from the core ideas, but I now find that this implementation clarifies the core ideas, and prevents me from getting confused (per the above analogy with other subtyping relationships).

gdalle:
Sorry I don't have the bandwidth to review this

sunxd3 (approved):
I think the changes from my last review looks solid.

sunxd3:
But maybe wait for another review to reduce risk

src/test_utils.jl
```diff
@@ -928,6 +928,7 @@ function test_rule(
     debug_mode::Bool=false,
     unsafe_perturb::Bool=false,
     print_results=true,
+    world::UInt=Base.get_world_counter(),
```
Technici4n: I would maybe remove this, and instead get the world age from the interpreter? (see other comment)

src/test_utils.jl
```diff
@@ -990,15 +991,15 @@ function test_rule(
             @testset "Caching" begin
                 if test_fwd
                     C_fwd = Mooncake.context_type(fwd_interp)
-                    if !Mooncake.is_primitive(C_fwd, ForwardMode, sig)
+                    if !Mooncake.is_primitive(C_fwd, ForwardMode, sig, world)
```
Technici4n: Use `fdw_interp.world` as the world?

src/test_utils.jl
```diff
@@ -990,15 +991,15 @@ function test_rule(
...
                         cache_key = (sig, false, :forward)
                         k = Mooncake.ClosureCacheKey(fwd_interp.world, cache_key)
                         @test haskey(fwd_interp.oc_cache, k)
                     end
                 end
                 if test_rvs
                     C_rvs = Mooncake.context_type(rvs_interp)
-                    if !Mooncake.is_primitive(C_rvs, ReverseMode, sig)
+                    if !Mooncake.is_primitive(C_rvs, ReverseMode, sig, world)
```
Technici4n: Use `rvs_interp.world` as the world?

Technici4n (commented):
Looks good, just a couple quick remarks

src/test_utils.jl
yebai: I think these are intentional: they are meant to test whether primitive rules correctly hold in a later world age than Mooncake's interpreter. Users can always pass `rvs_interp/fwd_interp.word` here if needed.

Technici4n: The interpreter is obtained below with `get_interpreter`, which internally gets the current world age. So it's weird to make it half-customizable.

yebai: Good point. `get_interpreter` does use the [current world age internally](https://github.com/chalk-lab/Mooncake.jl/blob/12cc097e0f0893eeb18d72ad4b3373dd80ea3f13/src/interpreter/abstract_interpretation.jl#L253).

Technici4n (approved):

PR #761, opened 2025-09-09
Relax some blas dispatches to `AbstractVecOrMat`

### LINKED ISSUE #622: Missing `rrule!!` for Lux.jl matrix operations
Simple training of an MLP generates this traceback which points to MooncakeLuxLibExt.jl, specifically `LuxLib.Impl.fused_dense`, which hits this error:

```julia
    nested task error: No rrule!! available for foreigncall with primal argument types Tuple{Val{(:sgemm_64_, Symbol("@rpath/libblastrampoline.5.dylib"))}, Val{Nothing}, Tuple{Val{Ref{UInt8}}, Val{Ref{UInt8}}, Val{Ref{Int64}}, Val{Ref{Int64}}, Val{Ref{Int64}}, Val{Ref{Float32}}, Val{Ptr{Float32}}, Val{Ref{Int64}}, Val{Ptr{Float32}}, Val{Ref{Int64}}, Val{Ref{Float32}}, Val{Ptr{Float32}}, Val{Ref{Int64}}, Val{Int64}, Val{Int64}}, Val{0}, Val{:ccall}, Ptr{UInt8}, Ptr{UInt8}, Ptr{Int64}, Ptr{Int64}, Ptr{Int64}, Ptr{Float32}, Ptr{Float32}, Ptr{Int64}, Ptr{Float32}, Ptr{Int64}, Ptr{Float32}, Ptr{Float32}, Ptr{Int64}, Int64, Int64, Int64, Int64, Base.RefValue{Int64}, MemoryRef{Float32}, Base.RefValue{Float32}, Base.RefValue{Int64}, MemoryRef{Float32}, Base.RefValue{Int64}, SubArray{Float32, 1, Vector{Float32}, Tuple{UnitRange{Int64}}, true}, Base.RefValue{Float32}, Base.RefValue{Int64}, Base.RefValue{Int64}, Base.RefValue{Int64}, Base.RefValue{UInt8}, Base.RefValue{UInt8}}. This problem has most likely arisen because there is a ccall somewhere in the function you are trying to differentiate, for which an rrule!! has not been explicitly written.You have three options: write an rrule!! for this foreigncall, write an rrule!! for a Julia function that calls this foreigncall, or re-write your code to avoid this foreigncall entirely. If you believe that this error has arisen for some other reason than the above, or the above does not help you to workaround this problem, please open an issue.
    Stacktrace:
      [1] rrule!!(::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Mooncake.CoDual{…}, ::Vararg{…})
        @ Mooncake ~/.julia/packages/Mooncake/JFoRx/src/rrules/foreigncall.jl:12
      [2] gemm!
        @ ~/.julia/juliaup/julia-1.11.5+0.aarch64.apple.darwin14/share/julia/stdlib/v1.11/LinearAlgebra/src/blas.jl:1633 [inlined]
      [3] (::Tuple{…})(none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…}, none::Mooncake.CoDual{…})
      [4] DerivedRule
        @ ~/.julia/packages/Mooncake/JFoRx/src/interpreter/s2s_reverse_mode_ad.jl:966 [inlined]
```

cc @yebai @avik-pal

RomeoV:
Tries to fix #622, and perhaps [#1392.](https://github.com/LuxDL/Lux.jl/issues/1392).
We check [blas.jl from LinearAlgebra](https://github.com/JuliaLang/LinearAlgebra.jl/blob/2c3fe9b7e0ca4e2c7bf506bd16ae5900f04a8023/src/blas.jl#LL1633C1-L1637C51) to check which of the operations and parameters need `AbstractVecOrMat`:

``` julia
$ rg -B8 AbstractVecOrMat LinearAlgebra/src/blas.jl | rg "function"
        function gemv!(trans::AbstractChar, alpha::Union{($elty), Bool},
        function gemmt!(uplo::AbstractChar, transA::AbstractChar, transB::AbstractChar,
        function gemm!(transA::AbstractChar, transB::AbstractChar,
        function syrk!(uplo::AbstractChar, trans::AbstractChar,
function syrk(uplo::AbstractChar, trans::AbstractChar, alpha::Number, A::AbstractVecOrMat)
        function herk!(uplo::AbstractChar, trans::AbstractChar,
        function herk(uplo::AbstractChar, trans::AbstractChar, α::$relty, A::AbstractVecOrMat{$elty})
        function syr2k!(uplo::AbstractChar, trans::AbstractChar,
function syr2k(uplo::AbstractChar, trans::AbstractChar, alpha::Number, A::AbstractVecOrMat, B::AbstractVecOrMat)
        function her2k!(uplo::AbstractChar, trans::AbstractChar, alpha::($elty1),
        function her2k(uplo::AbstractChar, trans::AbstractChar, alpha::($elty1), A::AbstractVecOrMat{$elty1}, B::AbstractVecOrMat{$elty1})
```
We can also check which of these are implemented in Mooncake:

``` bash
$ rg "BLAS\.((gemv)|(gemmt)|(gemm)|(syrk)|(herk)|(syr2k)|(her2k))" Mooncake/src/rrules/blas.jl -o | uniq 
BLAS.gemv
BLAS.gemm
BLAS.syrk
BLAS.syr2k
BLAS.syrk
BLAS.gemv
BLAS.gemm
BLAS.syrk
BLAS.gemm
```

| Function | taken care of |
|:---------|:--------------|
| `gemv`   | ✔️             |
| `gemm`   |             |
| `syrk`   |             |

### DIFF
--- a/src/rrules/blas.jl
+++ b/src/rrules/blas.jl
@@ -352,7 +352,7 @@ end
 )
 
 @inline function frule!!(
-    d::Dual{typeof(BLAS.gemv!)},
+    ::Dual{typeof(BLAS.gemv!)},
     tA::Dual{Char},
     alpha::Dual{P},
     A_dA::Dual{<:AbstractVector{P}},
@@ -360,10 +360,17 @@ end
     beta::Dual{P},
     y_dy::Dual{<:AbstractVector{P}},
 ) where {P<:BlasFloat}
-    Avec, dAvec = arrayify(A_dA)
-    return frule!!(
-        d, tA, alpha, Dual(reshape(Avec, :, 1), reshape(dAvec, :, 1)), x_dx, beta, y_dy
+    A, dA = arrayify(A_dA)
+    x, dx = arrayify(x_dx)
+    y, dy = arrayify(y_dy)
+    α, dα = numberify(alpha)
+    β, dβ = numberify(beta)
+
+    _gemv!_frule_core!(
+        primal(tA), α, dα, reshape(A, :, 1), reshape(dA, :, 1), x, dx, β, dβ, y, dy
     )
+
+    return y_dy
 end
 
 @inline function frule!!(
@@ -381,10 +388,28 @@ end
     α, dα = numberify(alpha)
     β, dβ = numberify(beta)
 
+    _gemv!_frule_core!(primal(tA), α, dα, A, dA, x, dx, β, dβ, y, dy)
+
+    return y_dy
+end
+
+@inline function _gemv!_frule_core!(
+    tA::Char,
+    α::P,
+    dα::P,
+    A::AbstractMatrix{P},
+    dA::AbstractMatrix{P},
+    x::AbstractVector{P},
+    dx::AbstractVector{P},
+    β::P,
+    dβ::P,
+    y::AbstractVector{P},
+    dy::AbstractVector{P},
+) where {P<:BlasFloat}
     # Derivative computation.
-    BLAS.gemv!(primal(tA), dα, A, x, β, dy)
-    BLAS.gemv!(primal(tA), α, dA, x, one(P), dy)
-    BLAS.gemv!(primal(tA), α, A, dx, one(P), dy)
+    BLAS.gemv!(tA, dα, A, x, β, dy)
+    BLAS.gemv!(tA, α, dA, x, one(P), dy)
+    BLAS.gemv!(tA, α, A, dx, one(P), dy)
 
     # Strong zero is essential here, in case `y` has undefined element values.
     if !iszero(dβ)
@@ -395,13 +420,12 @@ end
     end
 
     # Primal computation.
-    BLAS.gemv!(primal(tA), α, A, x, β, y)
-
-    return y_dy
+    BLAS.gemv!(tA, α, A, x, β, y)
+    return nothing
 end
 
 @inline function rrule!!(
-    _d::CoDual{typeof(BLAS.gemv!)},
+    ::CoDual{typeof(BLAS.gemv!)},
     _tA::CoDual{Char},
     _alpha::CoDual{P},
     _A::CoDual{<:AbstractVector{P}},
@@ -409,10 +433,20 @@ end
     _beta::CoDual{P},
     _y::CoDual{<:AbstractVector{P}},
 ) where {P<:BlasFloat}
-    Avec, dAvec = arrayify(_A)
-    return rrule!!(
-        _d, _tA, _alpha, CoDual(reshape(Avec, :, 1), reshape(dAvec, :, 1)), _x, _beta, _y
+
+    # Pull out primals and tangents (the latter only where necessary).
+    trans = _tA.x
+    alpha = _alpha.x
+    A, dA = arrayify(_A)
+    x, dx = arrayify(_x)
+    beta = _beta.x
+    y, dy = arrayify(_y)
+
+    pb = _gemv!_rrule_core!(
+        trans, alpha, reshape(A, :, 1), reshape(dA, :, 1), x, dx, beta, y, dy
     )
+
+    return _y, pb
 end
 
 @inline function rrule!!(
@@ -432,6 +466,23 @@ end
     x, dx = arrayify(_x)
     beta = _beta.x
     y, dy = arrayify(_y)
+
+    pb = _gemv!_rrule_core!(trans, alpha, A, dA, x, dx, beta, y, dy)
+
+    return _y, pb
+end
+
+@inline function _gemv!_rrule_core!(
+    trans::Char,
+    alpha::P,
+    A::AbstractMatrix{P},
+    dA::AbstractMatrix{P},
+    x::AbstractVector{P},
+    dx::AbstractVector{P},
+    beta::P,
+    y::AbstractVector{P},
+    dy::AbstractVector{P},
+) where {P<:BlasFloat}
 
     # Take copies before adding.
     y_copy = copy(y)
@@ -477,7 +528,7 @@ end
         )
     end
 
-    return _y, gemv!_pb!!
+    return gemv!_pb!!
 end
 
 @is_primitive(
--- a/src/rrules/blas.jl
+++ b/src/rrules/blas.jl
@@ -1237,11 +1237,11 @@ function positive_definite_blas_matrices(rng::AbstractRNG, P::Type{<:BlasFloat},
     end
 end
 
-function blas_vectors(rng::AbstractRNG, P::Type{<:BlasFloat}, p::Int)
+function blas_vectors(rng::AbstractRNG, P::Type{<:BlasFloat}, p::Int; only_contiguous=false)
     xs = Any[
         randn(rng, P, p),
         view(randn(rng, P, p + 5), 3:(p + 2)),
-        view(randn(rng, P, 3p, 3), 1:2:(2p), 2),
+        (only_contiguous ? copy : identity)(view(randn(rng, P, 3p, 3), 1:2:(2p), 2)),
         reshape(view(randn(rng, P, 1, p + 5), 1:1, 1:p), p),
     ]
     @assert all(x -> length(x) == p, xs)
--- a/src/rrules/blas.jl
+++ b/src/rrules/blas.jl
@@ -1292,7 +1292,7 @@ function blas_vectors(rng::AbstractRNG, P::Type{<:BlasFloat}, p::Int; only_conti
     xs = Any[
         randn(rng, P, p),
         view(randn(rng, P, p + 5), 3:(p + 2)),
-        (only_contiguous ? copy : identity)(view(randn(rng, P, 3p, 3), 1:2:(2p), 2)),
+        (only_contiguous ? collect : identity)(view(randn(rng, P, 3p, 3), 1:2:(2p), 2)),
         reshape(view(randn(rng, P, 1, p + 5), 1:1, 1:p), p),
     ]
     @assert all(x -> length(x) == p, xs)
--- a/src/rrules/blas.jl
+++ b/src/rrules/blas.jl
@@ -1288,9 +1288,12 @@ function generate_hand_written_rrule!!_test_cases(rng_ctor, ::Val{:blas})
         ) do (tA, M, N, P, α, β)
             P <: BlasRealFloat && (imag(α) > 0 || imag(β) > 0) && return []
 
-            As = blas_matrices(rng, P, tA == 'N' ? M : N, tA == 'N' ? N : M)
-            xs = blas_vectors(rng, P, N)
-            ys = blas_vectors(rng, P, M)
+            As = [
+                blas_matrices(rng, P, tA == 'N' ? M : N, tA == 'N' ? N : M)
+                blas_vectors(rng, P, M; only_contiguous=true)
+            ]
+            xs = [blas_vectors(rng, P, N); blas_vectors(rng, P, tA == 'N' ? 1 : M)]
+            ys = [blas_vectors(rng, P, M); blas_vectors(rng, P, tA == 'N' ? M : 1)]
             flags = (false, :stability, (lb=1e-3, ub=10.0))
             return map(As, xs, ys) do A, x, y
                 (flags..., BLAS.gemv!, tA, P(α), A, x, P(β), y)
--- a/src/rrules/blas.jl
+++ b/src/rrules/blas.jl
@@ -347,9 +347,24 @@ end
 @is_primitive(
     MinimalCtx,
     Tuple{
-        typeof(BLAS.gemv!),Char,P,AbstractMatrix{P},AbstractVector{P},P,AbstractVector{P}
+        typeof(BLAS.gemv!),Char,P,AbstractVecOrMat{P},AbstractVector{P},P,AbstractVector{P}
     } where {P<:BlasFloat},
 )
+
+@inline function frule!!(
+    d::Dual{typeof(BLAS.gemv!)},
+    tA::Dual{Char},
+    alpha::Dual{P},
+    A_dA::Dual{<:AbstractVector{P}},
+    x_dx::Dual{<:AbstractVector{P}},
+    beta::Dual{P},
+    y_dy::Dual{<:AbstractVector{P}},
+) where {P<:BlasFloat}
+    Avec, dAvec = arrayify(A_dA)
+    return frule!!(
+        d, tA, alpha, Dual(reshape(Avec, :, 1), reshape(dAvec, :, 1)), x_dx, beta, y_dy
+    )
+end
 
 @inline function frule!!(
     ::Dual{typeof(BLAS.gemv!)},
@@ -383,6 +398,21 @@ end
     BLAS.gemv!(primal(tA), α, A, x, β, y)
 
     return y_dy
+end
+
+@inline function rrule!!(
+    _d::CoDual{typeof(BLAS.gemv!)},
+    _tA::CoDual{Char},
+    _alpha::CoDual{P},
+    _A::CoDual{<:AbstractVector{P}},
+    _x::CoDual{<:AbstractVector{P}},
+    _beta::CoDual{P},
+    _y::CoDual{<:AbstractVector{P}},
+) where {P<:BlasFloat}
+    Avec, dAvec = arrayify(_A)
+    return rrule!!(
+        _d, _tA, _alpha, CoDual(reshape(Avec, :, 1), reshape(dAvec, :, 1)), _x, _beta, _y
+    )
 end
 
 @inline function rrule!!(
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.170"
+version = "0.4.171"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"


willtebbutt:
Thanks for this @RomeoV -- this broadly looks like the right thing to me.

To ensure that the behaviour is correct for the `AbstractVector`s that these methods now admit as arguments, would you mind extending the unit testing to include test cases which contain `AbstractVector`s?

For example, we generate the test cases for `gemv!` [here](https://github.com/chalk-lab/Mooncake.jl/blob/682a83dc15cb8850c7811dc73222fd069c03a491/src/rrules/blas.jl#L1171-L1179) -- currently `As` only includes matrices, but it ought now to contain some vectors as well. Could you maybe concatenate the current values of `As`, `xs` and `ys` with appropriately-sized vectors generated by `blas_vectors`? (we'll need to do something analogous for the other functions as well, but this is probably a good starting point)

RomeoV:
I added some tests [here](https://github.com/chalk-lab/Mooncake.jl/commit/8bf83a438f41d7e51aa32502ee2a0cb20fdaedae). Do you agree with that format? Then I can also add some for the other blas functions.
Trying to run the tests locally now...

RomeoV:
One issue I have is that `blas_vectors` creates non-contiguous vectors, which seem to be unsupported for `gemv!`:
```
Correctness: Error During Test at /home/romeo/.julia/dev/Mooncake/src/test_utils.jl:925
  Got exception outside of a @test
  matrix does not have contiguous columns
  Stacktrace:
    [1] error(s::String)
      @ Base ./error.jl:35
    [2] _chkstride1(ok::Bool)
      @ LinearAlgebra ~/.julia/juliaup/julia-1.11.6+0.x64.linux.gnu/share/julia/stdlib/v1.11/LinearAlgebra/src/LinearAlgebra.jl:281
    [3] _chkstride1
      @ ~/.julia/juliaup/julia-1.11.6+0.x64.linux.gnu/share/julia/stdlib/v1.11/LinearAlgebra/src/LinearAlgebra.jl:282 [inlined]
    [4] chkstride1
      @ ~/.julia/juliaup/julia-1.11.6+0.x64.linux.gnu/share/julia/stdlib/v1.11/LinearAlgebra/src/LinearAlgebra.jl:280 [inlined]
    [5] gemv!(trans::Char, alpha::Float32, A::SubArray{Float32, 1, Matrix{Float32}, Tuple{StepRange{Int64, Int64}, Int64}, true}, X::SubArray{Float32, 1, Matrix{Float32}, Tuple{StepRange{Int64, Int64}, Int64}, true}, beta::Float32, Y::SubArray{Float32, 1, Matrix{Float32}, Tuple{StepRange{Int64, Int64}, Int64}, true})
      @ LinearAlgebra.BLAS ~/.julia/juliaup/julia-1.11.6+0.x64.linux.gnu/share/julia/stdlib/v1.11/LinearAlgebra/src/blas.jl:658
    [6] test_frule_correctness(::AbstractRNG, ::Any, ::Vararg{Any}; frule::Function, unsafe_perturb::Bool)
      @ Mooncake.TestUtils ~/.julia/dev/Mooncake/src/test_utils.jl:443
```

EDIT: I'll enforce stride 1 for all A vectors for now.

willtebbutt:
Ahh, interesting. I'm guessing we're going to need stride-1 vectors in a few places. Maybe add a function called something like `stride_1_blas_vectors`, which is basically the same as `blas_vectors`, but only returns vectors with stride 1? It would definitely be good to continue testing on a range of types of `AbstractVector`, so that we're sure we've covered them.

RomeoV:
Writing tests for the other variants is getting a bit out of control... For `gemm!`, in theory we'd need tests for every combination `mat x mat`, `mat x vec`, `vec x mat`, `vec x vec` with the correct transposes etc...
I can do that but it won't be pretty...

willtebbutt:
Good point. I think it's probably enough to ensure that each argument that it permitted to be a vector has at least seen a vector -- is there a way to avoid considering every possible combination, but to still have this property? I fully agree that it feels excessive to test every possible combination -- these tests already take way longer than I would like haha.

willtebbutt:
Thanks for digging into this further. I've just been looking through the code that you've had to add in order to make everything `gemv!` and `gemm!` work correctly. Do you know if it is the case that, any time we see an `AbstractVector` in one of these functions with widened signatures, we want it to behave like an `AbstractMatrix` with exactly one column?

If so, could we just insert some canonicalising code at the top of these functions which reshapes any incoming vectors to be a single-column matrix, and leave the rest of the bodies of the functions otherwise unchanged?

RomeoV:
Yes I agree that it would be nice to simplify this. Vectors are not always column matrices: For instance, we can have $v^T M$, where $v^T$ becomes a row matrix. However, I think we can just check the tranposed flag for the vector typed inputs, which should tell us whether to resize as `(:, 1)` or `(1, :)`. Will update the code soon.

yebai:
A gentle reminder @RomeoV.

RomeoV:
Thanks @yebai. Had a busy two weeks but will pick this back up next week.

RomeoV:
Talked with @willtebbutt earlier and just force-pushed a new version just working on `gemv!`, both for `frule!!` and `rrule!!` by defining a new dispatch for the `AbstractVector` inputs that converts to column matrices. I have convinced myself that this is correct for all tranposed flags, since we're just taking the memory address in the end and working in place in the end.

I have added tests for the `rrule!!` and ran them locally, but I am not sure where the `frule!!` tests go. Any pointers? Let me know if this approach is fine with you, then I will also look into some of the other functions again.

RomeoV:
Just rebased as well.

RomeoV:
Tests pass except for code coverage. Where to add the `frule` tests?

EDIT: Looks like the `frule` tests are the same as the `rrule` tests... so then I'm not sure why we fail for coverage...

yebai:
> Tests pass except for code coverage. 

It is okay to ignore that -- random failures from code coverage are observed in other PRs, they are not always reliable.

src/rrules/blas.jl
```diff
@@ -1211,7 +1241,10 @@ function generate_hand_written_rrule!!_test_cases(rng_ctor, ::Val{:blas})
         ) do (tA, M, N, P, α, β)
             P <: BlasRealFloat && (imag(α) > 0 || imag(β) > 0) && return []
 
-            As = blas_matrices(rng, P, tA == 'N' ? M : N, tA == 'N' ? N : M)
+            As = [
+                blas_matrices(rng, P, tA == 'N' ? M : N, tA == 'N' ? N : M)
+                blas_vectors(rng, P, M; only_contiguous=true)
+            ]
             xs = blas_vectors(rng, P, N)
             ys = blas_vectors(rng, P, M)
```
willtebbutt: On this particular occassion, I believe that coverage is quite possibly telling us the truth! I believe you've hit some slightly strange semantics of `map`. Unless I'm mistaken, `As` is a longer `Vector` than either `xs` or `ys`. The behaviour of multi-argument `map` in this case is to only iterate until the shortest collection has been finished, so it's quite possible that all if the vector test cases aren't actually being tested. I find this behaviour rather odd, but I'm quite sure it's by design rather than by accident.

`Mooncake._map` is an internal function which has the same behaviour as `map` when its arguments all have the same length, but which errors if they are different lengths. Perhaps replace `map(As, xs, ys)` below with `Mooncake._map(As, xs, ys)`, and see if that catches the problem?

RomeoV: Great catch! Fixed in the latest commit.

RomeoV:
Ok, it seems that the double `arrayify` that happens first in
```julia
function (..., ::Dual{<:AbstractVector}, ...)
    Avec, dAvec = arrayify(A_dA)
    return frule!!(
        d, tA, alpha, Dual(reshape(Avec, :, 1), reshape(dAvec, :, 1)), x_dx, beta, y_dy
    )
end
```
and then again
```julia
function (..., ::Dual{<:AbstractMatrix}, ...)
    Avec, dAvec = arrayify(A_dA)
end
```
doesn't work. 
I'm getting errors such as
```
  caused by: Encountered unexpected array type in `Mooncake.arrayify`. This error is likely due to a call to a BLAS or LAPACK function with an array type that Mooncake has not been told about. A new method of `Mooncake.arrayify` is needed. Please open an issue at https://github.com/chalk-lab/Mooncake.jl/issues . It should contain this error message and the associated stack trace.
  
  Array type: Base.ReshapedArray{Float32, 2, SubArray{Float32, 2, Matrix{Float32}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}
  
  FData type: Base.ReshapedArray{Float32, 2, SubArray{Float32, 2, Matrix{Float32}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}.
  Stacktrace:
    [1] error(s::String)
      @ Base ./error.jl:35
    [2] arrayify(x::Base.ReshapedArray{Float32, 2, SubArray{Float32, 2, Matrix{Float32}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}, dx::Base.ReshapedArray{Float32, 2, SubArray{Float32, 2, Matrix{Float32}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}})
      @ Mooncake ~/.julia/dev/Mooncake/src/rrules/blas.jl:66
    [3] arrayify(x::Mooncake.Dual{Base.ReshapedArray{Float32, 2, SubArray{Float32, 2, Matrix{Float32}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}, Base.ReshapedArray{Float32, 2, SubArray{Float32, 2, Matrix{Float32}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}})
      @ Mooncake ~/.julia/dev/Mooncake/src/rrules/blas.jl:32
    [4] frule!!
      @ ~/.julia/dev/Mooncake/src/rrules/blas.jl:324 [inlined]
    [5] frule!!(d::Mooncake.Dual{typeof(LinearAlgebra.BLAS.gemv!), NoTangent}, tA::Mooncake.Dual{Char, NoTangent}, alpha::Mooncake.Dual{Float32, Float32}, A_dA::Mooncake.Dual{Base.ReshapedArray{Float32, 1, SubArray{Float32, 2, Matrix{Float32}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}, Tangent{@NamedTuple{parent::Tangent{@NamedTuple{parent::Matrix{Float32}, indices::NoTangent, offset1::NoTangent, stride1::NoTangent}}, dims::NoTangent, mi::NoTangent}}}, x_dx::Mooncake.Dual{Base.ReshapedArray{Float32, 1, SubArray{Float32, 2, Matrix{Float32}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}, Tangent{@NamedTuple{parent::Tangent{@NamedTuple{parent::Matrix{Float32}, indices::NoTangent, offset1::NoTangent, stride1::NoTangent}}, dims::NoTangent, mi::NoTangent}}}, beta::Mooncake.Dual{Float32, Float32}, y_dy::Mooncake.Dual{Base.ReshapedArray{Float32, 1, SubArray{Float32, 2, Matrix{Float32}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}, Tangent{@NamedTuple{parent::Tangent{@NamedTuple{parent::Matrix{Float32}, indices::NoTangent, offset1::NoTangent, stride1::NoTangent}}, dims::NoTangent, mi::NoTangent}}})
      @ Mooncake ~/.julia/dev/Mooncake/src/rrules/blas.jl:310
```

Any ideas how we can solve this more cleanly?


EDIT: I replaced the dispatch approach with a simple if-condition now that seems to do the trick.
EDIT2: Seems like this is giving type stability issues, so I'll try to back to the original dispatch approach.

RomeoV:
Ok this latest approach, pulling out a shared "core" function, circumvents the dispatch issues and all tests pass. The one test that failed was due to a network issue in the CI, unrelated to the PR. Perhaps someone can retrigger it.

src/rrules/blas.jl
```diff
@@ -1153,11 +1234,11 @@ function positive_definite_blas_matrices(rng::AbstractRNG, P::Type{<:BlasFloat},
     end
 end
 
-function blas_vectors(rng::AbstractRNG, P::Type{<:BlasFloat}, p::Int)
+function blas_vectors(rng::AbstractRNG, P::Type{<:BlasFloat}, p::Int; only_contiguous=false)
     xs = Any[
         randn(rng, P, p),
         view(randn(rng, P, p + 5), 3:(p + 2)),
-        view(randn(rng, P, 3p, 3), 1:2:(2p), 2),
+        (only_contiguous ? copy : identity)(view(randn(rng, P, 3p, 3), 1:2:(2p), 2)),
```
willtebbutt: Should we use `collect` rather than `copy` here? My understanding is that `collect` promises to return an `Array`, while `copy` doesn't make any promises around what type it will return (albeit it plainly does happen to return an `Array` at preset)

willtebbutt:
This looks like it's basically there now, and provides a nice approach for other BLAS rules. Other than my comment above:
1. it would be good to hit the `Update branch` button before we merge this,
2. could you please bump the patch version so that we can tag a release?

src/rrules/blas.jl
RomeoV: Done

RomeoV:
Ready for review @sumiya11

RomeoV:
Having worked this out, when I find some more time I'll also do the same for the `gemm!` dispatch, which may then fix the Lux issue.

RomeoV:
I am still struggling to implement `gemm!`, not because it's so challenging, but because I am still struggling to efficiently test my changes. I am already commenting out all the other test cases in `src/rrules/blas.jl` and have even commented out the JET test for now, but when something fails I get an endless stream of errors which I can't interrupt without crashing Julia and then setting up everything again which takes several minutes. Conversely, I am also not sure how to just call the `rrule!!` directly on the data that I want to test on. So in the end my development loop here has been so slow that I can't continue to spend the time working on this.

willtebbutt:
Ah, apologies for not making it more clear how to achieve a quick development cycle! I would recommend the following.

To run the rule manually at the command line, you might do something like the following:
```julia
julia> using LinearAlgebra, Mooncake, Random

julia> using Mooncake: rrule!!, zero_rdata, zero_fcodual, primal

julia> fargs = (BLAS.gemm!, 'N', 'N', 1.1, randn(3, 2), randn(2, 1), 0.5, randn(3, 1));

julia> fargs[1](fargs[2:end]...)
3×1 Matrix{Float64}:
  3.0675765934204846
  0.3241193013431162
 -0.10724921573866608

julia> # Run the rule
       y, rvs = rrule!!(map(zero_fcodual, fargs)...)
(Mooncake.CoDual{Matrix{Float64}, Matrix{Float64}}([3.097836282610327; 1.434872548283475; 0.36413373015673733;;], [0.0; 0.0; 0.0;;]), Mooncake.var"#gemm!_pb!!#344"{Float64, Base.RefValue{Matrix{Float64}}, Matrix{Float64}, Matrix{Float64}, Matrix{Float64}, Matrix{Float64}, Matrix{Float64}, Matrix{Float64}, Matrix{Float64}, Float64, Float64, Char, Char}(Base.RefValue{Matrix{Float64}}([1.4126922025789126; 0.8205107411804217; 0.23693699035842655;;]), [3.0877497195470465; 1.064621465970022; 0.20700608152493621;;], [0.0; 0.0; 0.0;;], [3.097836282610327; 1.434872548283475; 0.36413373015673733;;], [0.0; 0.0;;], [1.6096030714358025; 0.01575435958162854;;], [0.0 0.0; 0.0 0.0; 0.0 0.0], [0.8794851649178947 -0.18596885135024463; 0.5096859991842587 0.007527530712906747; 0.14416956239782056 0.3098329633805674], 0.5, 1.1, 'N', 'N'))

julia> # Run the reverse pass
       rvs(zero_rdata(primal(y)))
(Mooncake.NoRData(), Mooncake.NoRData(), Mooncake.NoRData(), 0.0, Mooncake.NoRData(), Mooncake.NoRData(), 0.0, Mooncake.NoRData())
```
If you want to run the test suite on a single collection of arguments, you can do something like
```julia
julia> Mooncake.TestUtils.test_rule(Xoshiro(123), fargs...);
Test Summary:                                                                                                            | Pass  Total  Time
Tuple{typeof(LinearAlgebra.BLAS.gemm!), Char, Char, Float64, Matrix{Float64}, Matrix{Float64}, Float64, Matrix{Float64}} |   34     34  0.0s
```
When debugging, it might be helpful to set the `interface_only` kwarg to `test_rule` equal to `true`, in order to avoid running correctness tests.

Does the above help at all? If it does, I'll make a docs PR explaining it in order to make everyone's life easier.

PR #762, opened 2025-09-09
Revise documentation for internal `_copy` utility

yebai:
The PR cleans up docs on `_copy` and adds an overview of `_copy`'s current behaviour for different types

Replacing and closes https://github.com/chalk-lab/Mooncake.jl/pull/760

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -429,14 +429,17 @@ end
 
 Utility for copying AD-related data structures (e.g. rules, caches, and other internals).
 
-# Semantics
+# Examples of current behaviour
 
-`_copy` specifies how different types are duplicated within the AD system:
+Currently, `_copy` has the following behaviours for specific types:
 
-- Immutable AD types (e.g. `CoDual`, `Dual`) → usually return the same object
-- Mutable containers → new instances with copied contents
-- Composite types → recursively applies `_copy` to fields
-- Other types → falls back to `Base.copy`
+- `CoDual`, `Dual` types → no copying needed  
+- `Stack` types → create a new empty instance  
+- Composite types (e.g. `Tuple`) → recursively copy elements  
+- Rule types (e.g. `DerivedRule`) → construct new instances with copied captures and caches  
+- Misty closure reverse data → construct a new Misty closure with deep copy of captured data  
+- Tangent types (`PossiblyUninitTangent`) → copy conditionally based on initialisation state  
+- Forward/reverse data types (i.e., `FData`, `RData`) → recursively copy wrapped data  
 """
 
 # Generic fallback that works with any type
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -422,71 +422,29 @@ end
 """
     _copy(x)
 
-*Note:* this is not part of the public Mooncake.jl interface, and may change without warning.
+!!! warning
+    This is an internal utility, not part of the public `Mooncake.jl` API, and may change
+    without notice. Its behaviour can vary across data types, so the description below
+    should be treated as guidance, not a strict contract.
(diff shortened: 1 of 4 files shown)


yebai:
@gdalle, I'd appreciate it if you could take a look at this.

src/utils.jl
```diff
@@ -422,71 +422,35 @@ end
...
+- Tangent types (`PossiblyUninitTangent`) → copy conditionally based on initialisation state  
+- Forward/reverse data types (e.g. `FData`, `RData`, `LazyZeroRData`) → recursively copy wrapped data  
+- `RRuleZeroWrapper` → recursively copy the wrapped rule into a new instance
+- `DerivedRule` → construct new instances with copied captures and caches  
+- `LazyFRule`, `LazyDerivedRule` → construct new lazy rules with the same method instance and debug mode  
+- `DynamicFRule`, `DynamicDerivedRule` → construct new dynamic rules with an empty cache and the same debug mode  
+"""
 
-# Fallback to Base.copy
+# Generic fallback that works with any type
```
mhauru:
```suggestion
# Generic fallback to Base.copy
```
Proposing this change because many types don't implement `Base.copy`.

src/utils.jl
```diff
@@ -422,71 +422,35 @@ end
...
-*Note:* this is not part of the public Mooncake.jl interface, and may change without warning.
+!!! warning
+    This is an internal utility, not part of the public `Mooncake.jl` API, and may change
+    without notice. Its behaviour can vary across data types, so the description below
+    should be treated as guidance, not a strict contract.
 
-Internal protocol for creating copies of AD-related data structures. This function is used 
-throughout the automatic differentiation system to create appropriate copies of rules, 
-caches, and other internal data structures when building new AD contexts.
+Utility for copying AD-related data structures (e.g. rules, caches, and other internals).
```
mhauru, line 430: Would it be useful to say something about where `_copy` gets used? Or something else that would help answer the question of, if I was to implement my own type `Foo`, what should I make `_copy(::Foo)` do?

src/utils.jl
```diff
@@ -422,71 +422,35 @@ end
...
+- `CoDual`, `Dual` types → no copying needed  
+- `Stack` types → create a new empty instance  
+- Composite types (e.g. `Tuple`, `NamedTuple`) → recursively copy elements  
+- Misty closure → construct a new Misty closure with deep copy of captured data  
+- Tangent types (`PossiblyUninitTangent`) → copy conditionally based on initialisation state  
+- Forward/reverse data types (e.g. `FData`, `RData`, `LazyZeroRData`) → recursively copy wrapped data  
+- `RRuleZeroWrapper` → recursively copy the wrapped rule into a new instance
+- `DerivedRule` → construct new instances with copied captures and caches  
+- `LazyFRule`, `LazyDerivedRule` → construct new lazy rules with the same method instance and debug mode  
+- `DynamicFRule`, `DynamicDerivedRule` → construct new dynamic rules with an empty cache and the same debug mode
```
mhauru, line 445: Might it make sense to document these in the same place where they are implemented? Some of these implementations must live in other files, and could be added as docstrings to the methods that implement them. Less risk of the documentation falling out of sync with the implementation.

src/utils.jl
yebai: It is a bit awkward to say any more about that for now -- the `_copy` function is a bit scattered and thus requires some tidy up down the road.

src/utils.jl
yebai: That is already done as comments, the point of repeating them here is to have an overview.

Issue #763, opened 2025-09-12
No rrule for foreigncall geev


kshyatt:
Hello,

I'm trying to write an `rrule` for generalized eigensolving for (possibly complex-valued) matrices. I have the pullback written but hit this:

```
Mooncake.MissingForeigncallRuleError("No rrule!! available for foreigncall with primal argument types Tuple{Val{(:dgeev_64_, Symbol(\"@rpath/libblastrampoline.5.dylib\"))}, Val{Nothing}, Tuple{Val{Ref{UInt8}}, Val{Ref{UInt8}}, Val{Ref{Int64}}, Val{Ptr{Float64}}, Val{Ref{Int64}}, Val{Ptr{Float64}}, Val{Ptr{Float64}}, Val{Ptr{Float64}}, Val{Ref{Int64}}, Val{Ptr{Float64}}, Val{Ref{Int64}}, Val{Ptr{Float64}}, Val{Ref{Int64}}, Val{Ptr{Int64}}, Val{Int64}, Val{Int64}}, Val{0}, Val{:ccall}, Ptr{UInt8}, Ptr{UInt8}, Ptr{Int64}, Ptr{Float64}, Ptr{Int64}, Ptr{Float64}, Ptr{Float64}, Ptr{Float64}, Ptr{Int64}, Ptr{Float64}, Ptr{Int64}, Ptr{Float64}, Ptr{Int64}, Ptr{Int64}, Int64, Int64, Int64, Int64, Base.RefValue{Int64}, Base.RefValue{Int64}, MemoryRef{Float64}, Base.RefValue{Int64}, MemoryRef{ComplexF64}, Base.RefValue{Int64}, MemoryRef{Float64}, Base.RefValue{Int64}, MemoryRef{Float64}, Base.RefValue{Int64}, Base.RefValue{UInt8}, Base.RefValue{UInt8}, DataType, Base.ReinterpretArray{Float64, 1, ComplexF64, Vector{ComplexF64}, false}, DataType, Base.ReinterpretArray{Float64, 1, ComplexF64, Vector{ComplexF64}, false}}.
[...]

PR #767, opened 2025-09-12
Add arrayify for Diagonals

kshyatt:
This wasn't yet supported but is nice to have for some BLAS methods.

### DIFF
--- a/src/rrules/blas.jl
+++ b/src/rrules/blas.jl
@@ -1101,7 +1101,13 @@ function blas_matrices(rng::AbstractRNG, P::Type{<:BlasFloat}, p::Int, q::Int)
         view(randn(rng, P, 3p, 3, 2q), (p + 1):(2p), 2, 1:2:(2q)),
         reshape(view(randn(rng, P, p * q + 5), 1:(p * q)), p, q),
     ]
-    append!(Xs, map(Diagonal, blas_vectors(rng, P, p)))
+    @assert all(X -> size(X) == (p, q), Xs)
+    @assert all(Base.Fix2(isa, AbstractMatrix{P}), Xs)
+    return Xs
+end
+
+function special_matrices(rng::AbstractRNG, P::Type{<:BlasFloat}, p::Int, q::Int)
+    Xs = map(Diagonal, blas_vectors(rng, P, p))
     @assert all(X -> size(X) == (isa(X, Diagonal) ? (p, p) : (p, q)), Xs)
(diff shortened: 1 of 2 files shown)


willtebbutt:
Thanks for this @kshyatt . Would you mind pointing me towards the BLAS functionality that uses `Diagonal` matrices directly? I wasn't aware that this is a thing!

kshyatt:
If your wrapper uses diagonals and wants to pass them into BLAS by
referencing the diag field (which is very common in higher level wrappers
for eig or svd) you need this arrayify method for the derivative

willtebbutt:
Ohh cool, okay. I don't have a problem with an extra method of `arrayify`.

Would you mind adding some instances of `Diagonal{P}` matrices to the `xs` vector found [here in the BLAS tests](https://github.com/chalk-lab/Mooncake.jl/blob/682a83dc15cb8850c7811dc73222fd069c03a491/test/rrules/blas.jl#L13)? Maybe you could generate them using the `Mooncake.blas_vectors` function to ensure that the implementation copes nicely with `diag` fields other than `Vector{P}` (it looks to me like your implementation ought to be fine).

Assuming these tests pass, I think this will be good to go with a patch bump.

edit: oh, also, it looks like the formatter is complaining. Running JuliaFormatter in the root of the repo should resolve this.

kshyatt:
Can do, I wanted to get the maintainer take before going all in on these.
Thanks for the suggestions!

willtebbutt:
Fab, thanks.

> I wanted to get the maintainer take before going all in on these.

Should I take it that you have other methods of `arrayify` in mind, or are you referring to the rules discussed in https://github.com/chalk-lab/Mooncake.jl/issues/763 ?

kshyatt:
I meant before writing tests for the Diagonal one. I can certainly add some
support for other linear algebra wrapper types like Symmetric in a future
PR if you like

willtebbutt:
Nono, just the `Diagonal` one is great. I'm quite wary of adding methods to `arrayify` in general, because I think they're quite hard to get right, and I'm not sure they make sense for most matrix types. I can't see that any harm will come from the `Diagonal` case though, so have no issue with it.

kshyatt:
Perfect! This is why I thought it would be wise to ask :)

kshyatt:
Oof, looks like this broke the lapack tests. @willtebbutt maybe I can make a new function, `special_matrices` or something, and pass that to the tests where appropriate?

kshyatt:
Go for it!

willtebbutt:
Great. Thanks for the contribution :)

PR #769, opened 2025-09-15
Improve potrf performance

willtebbutt:
Tests have been a bit flakey recently, and it appears to be to do with potrf. This implementation approximately halves the number of allocations, and more than halves the total amount of memory allocated, so I'm hoping that it stabilises the tests. Also, a faster potrf is just nice to have.


yebai:
@willtebbutt, you can ask @sunxd3 to review PRs from now on to avoid delays.

src/rrules/lapack.jl
```diff
@@ -408,15 +411,19 @@ function rrule!!(
...
             L = LowerTriangular(A)
-            B = L' \ (E' .* (dA2'L)) / L
-            dA .= 0.5 * __sym(B) .* E .+ triu!(dA2, 1)
+            tmp = dA2'L
+            tmp .*= E'
+            B = rdiv!(ldiv!(L', tmp), L)
+            dA .= __sym_lower!(B) .* E ./ 2 .+ triu!(dA2, 1)
         else
-            E = UpperTriangular(2 * ones(N, N) - Diagonal(ones(N)))
+            E = UpperTriangular(__E(N))
```
sunxd3: intuitively, I sense that we might be able to reduce allocation further? (`__E`) still allocate.

I might be wrong, also this is obviously non blocking

src/rrules/lapack.jl
```diff
@@ -427,6 +434,28 @@ function rrule!!(
...
+
+function __sym_upper!(X::Matrix)
+    @inbounds for q in 1:size(X, 2), p in 1:(q - 1)
+        X[p, q] = (X[p, q] + X[q, p]) / 2
+    end
+    return X
+end
+
+@inline function __E(N::Int)
+    E = fill(2.0, (N, N))
```
sunxd3: I assume not, but would `2.0` cause some potential floating type trouble?

willtebbutt: This is a very good point. The code definitely runs on `Float32`, but it must be the case that we're doing a bunch of computations in `Float64` in that case. I'll have a dig, and might fix this.

src/rrules/lapack.jl
willtebbutt: I think you're right that there's probably more that we can do to improve performance here. Since this is such a simple mask, it feels to me like we ought to be able to avoid allocating it. Something I'd be keen to tackle in a follow-up PR at some point :)

willtebbutt:
Thanks for the review @sunxd3 . I'll merge when CI passes.

PR #770, opened 2025-09-16
Move JuliaFormatter to a separate environment

willtebbutt:
There are currently a couple of different versions of (I think) JuliaSyntax.jl being used by test deps of Julia, which is causing some trouble. By separating JuliaFormatter out into its own environment, we circumvent this issue. Additionally:
1. Aqua.jl tests now happen in the "basic" test set. They're cheap, so this should be fine.
2. I've been able to remove the JuliaInterpreter test dependency, which is nice because it was previously purely there purely to control versions of other packages, rather than to be used.


willtebbutt:
Great. Thanks for checking re version bump -- there are no changes to src or ext, so no need to bump :)

PR #771, opened 2025-09-17
Move FunctionWrappers to extension

willtebbutt:
Should be merged after #770 .

Note: FunctionWrappers code is not currently an extension because, at the time of originally implementing it, we had some issues with generated functions and world ages. These were resolved quite some time ago now, so there's no problem with doing this now.

edit: note that this PR now also modifies the nnlib extension tests to make use of a manually-initialised RNG. These correctness tests have been a little flakey for a while, which is usually a sign of non-determinism being introduced via different random seeds on different runs of the test suite.


willtebbutt:
thanks @sunxd3 -- I'll bump the patch version and merge.

PR #776, opened 2025-09-18
Allow users to pass output_tangent kwarg to test_rule

kshyatt:
Allows you to specify the output tangent for reproducibility and correctness


yebai:
@kshyatt, can you provide a brief example of how this would be used?

kshyatt:
@yebai I'm currently using this to test my complex matrix eig/svd methods -- it protects against the FiniteDifferences check exploding due to a discontinuity in the output tangent manifold. You can remove some gauge freedom in the eigenvector derivative (let's say), pass that back as part of `output_tangent`, and check your reverse rule in a "safe" region without the discontinuity.

PR #778, opened 2025-09-20
Fix Formatter

willtebbutt:
I messed up the formatter in https://github.com/chalk-lab/Mooncake.jl/pull/770 -- it was pointing to the `test` folder only, rather than the entire package. As a result, a couple of recent PRs were merged without proper formatting.

This PR
1. fixes the formatter, making sure that it checks code in all directories in the package
2. fixes formatting of PRs that were merged while the formatter was not doing the right thing
3. bumps the patch version so that we can merge the updated code.


sunxd3 (dismissed):
👍

PR #780, opened 2025-09-24
Add output_tangent to kwargs list

kshyatt:
Missed this when I added the option


willtebbutt:
Ah, thanks for catching this.

Could you please update this branch so that it's up to date with main, and bump the patch version?

kshyatt:
Done

willtebbutt:
Thanks! -- will merge when CI passes. There's obviously no way that your changes could have made CI fail, but I'm trying to run them to completion whenever possible at the minute because I _think_ we fixed some flakiness over the last week, and the only real way to know for sure is to run the tests lots and see if they fall over 😂 .

PR #782, opened 2025-09-25
Remove DiffRules Dep

willtebbutt:
While working on resolving the problems in #471 , I realised that it would be rather straightforward to remove our DiffRules.jl dep. This has several advantages (see the note at the top of the file for info on them), hence this PR.

PR #785, opened 2025-09-26
More dependency reduction

willtebbutt:
~~Please do not review -- need to see if CI still passes.~~

This seems to work nicely. Unlike #782 , this won't actually change what code gets loaded, because ChainRules depends on GPUArraysCore. This is therefore essentially just some rather trivial house keeping.

edit: this now also deals with some non-determinism in the benchmark suite where (I think) certain functions were getting run different amount of time on each run and, as a result, sometimes causing errors. We shall see.

edit2: the hand-written rule benchmarks now run quite a bit faster, because I've reduced the amount of time spent on each benchmark case. It looks like the test suite passed on this run, suggesting that doing this hasn't massively increased the variability in how long things take to run.

### DIFF
--- a/bench/run_benchmarks.jl
+++ b/bench/run_benchmarks.jl
@@ -176,7 +176,7 @@ function generate_inter_framework_tests()
 end
 
 function benchmark_rules!!(
-    test_case_data, default_ratios, include_other_frameworks::Bool, seconds=nothing,
+    test_case_data, default_ratios, include_other_frameworks::Bool, seconds=nothing
 )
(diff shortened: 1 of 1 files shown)


yebai:
To clarify, does this fix the underministisism of `gp_lml`'s performance?

willtebbutt:
> To clarify, does this fix the underministisism of gp_lml's performance?

This will not. This should (hopefully) just stop the benchmark suite occassionally erroring on `potrf!` calls.

PR #796, opened 2025-10-02
Add a test mode for forward-on-reverse

kshyatt:
This mode is helpful for testing rules where the choice of gauge for outputs (e.g. for eigenvectors) is meaningful. Finite differences will pick an arbitrary gauge which leads to "false" correctness failures in the rule tests. This PR introduces a new mode that runs the reverse mode first (the output tangent allows a particular choice of gauge), checks it with FD, then uses that to check the forward rule.


kshyatt:
OK, so I have the tests for the x's passing, which I think is correct and not testing that `a === a` 😅 . But my tests for the tangents of `y` is not working even for a rule I'm pretty confident in -- not sure what I've done wrong here but probably something!

src/test_utils.jl
```diff
@@ -608,6 +608,83 @@ function test_rrule_correctness(
...
+    x̄ = map(tangent, x_x̄)
+
+    # Run primal, and ensure that we still have access to mutated inputs afterwards.
+    x_primal = _deepcopy(x)
+    y_primal = x_primal[1](x_primal[2:end]...)
+
+    # Construct tangent to inputs, and normalise to be of unit length.
+    ẋ_unnormalised = map(_x -> randn_tangent(rng, _x), x)
+    nrm = sqrt(sum(x -> _dot(x, x), ẋ_unnormalised))
+    ẋ = map(_x -> _scale(1 / nrm, _x), ẋ_unnormalised)
```
willtebbutt:
```suggestion
    ẋ_unnormalised = randn_tangent(rng, x)
    nrm = sqrt(_dot(ẋ_unnormalised, ẋ_unnormalised))
    ẋ = _scale(1 / nrm, ẋ_unnormalised)
```
Unless I'm mistaken, we should just be able to apply these operations to entire tuples. Should save some overhead, and also ensures correctness if there's aliasing between arguments.

willtebbutt: (It might be that I don't do this elsewhere, but probably should...)

kshyatt: This came directly from `test_rrule_correctness` so should probably be changed there as well

kshyatt: Updated here and at the source

kshyatt: This seems to have broken tests rip

yebai:
@sunxd3, can you help review this?

src/test_utils.jl
```diff
@@ -608,6 +608,50 @@ function test_rrule_correctness(
...
+    x_primal = _deepcopy(x)
+    y_primal = x_primal[1](x_primal[2:end]...)
+    # Use AD to compute Frechet derivative at ẋ.
+    x_ẋ_rule = map((x, ẋ) -> dual_type(_typeof(x))(_deepcopy(x), ẋ), x, ẋ)
+    y_ẏ_rule = frule(x_ẋ_rule...)
+    ẋ_ad = map(tangent, x_ẋ_rule)
+    ẏ_ad = tangent(y_ẏ_rule)
+
+    # now use ẏ_ad as output_tangent for rrule again
+    test_rrule_correctness(
```
willtebbutt, line 650: I'm a little confused by why we're running reverse-mode again here. Would you mind running be through the logic of what's going on here again @kshyatt ?

src/test_utils.jl
```diff
@@ -608,6 +608,50 @@ function test_rrule_correctness(
...
+)
+    # Generate random tangents for anything that is not already a CoDual.
+    x_ẋ = map(x -> x isa CoDual ? Dual(primal(x), tangent(x)) : randn_dual(rng, x), x)
+    # test reverse rule first, we will recapitulate some work this way
+    # but it's much less duplicated code
+    test_rrule_correctness(rng, x_ẋ...; rrule, unsafe_perturb, output_tangent, atol, rtol)
+
+    # now we are certain the reverse rule works -- use it to compute
+    # the tangents to compare with, rather than finite differences
+    @nospecialize rng x_x̄
```
```suggestion
    @nospecialize rng x_ẋ
```

src/test_utils.jl
```diff
@@ -608,6 +608,50 @@ function test_rrule_correctness(
     @test any(isapprox_results)
 end
 
+# Test the reverse rule using FD, then, if that succeeds
+# test the forwrad rule using the reverse result, *not*
```
```suggestion
# test the forward rule using the reverse result, *not*
```

src/test_utils.jl
```diff
@@ -887,6 +932,10 @@ signature associated to `x` corresponds to a primitive, a hand-written rule will
 - `unsafe_perturb::Bool=false`: value passed as the third argument to `_add_to_primal`.
     Should usually be left `false` -- consult the docstring for `_add_to_primal` for more
     info on when you might wish to set it to `true`.
+- `test_rewind=false`: rather than testing the forwards rule using finite differences,
+    test the reverse rule first, then run the forwards rule to generate new output tangents
+    and use these to test the reverse rule. This protects against a forwards pass mapping
+    to some arbitrary gauge for matrix transformations.
 - `output_tangent=nothing`: final output tangent to initialize reverse mode with for testing
     the correctnes of reverse rules.
```
```suggestion
    the correctness of reverse rules.
```

src/test_utils.jl
sunxd3: I have similar question here. It feels like to me that the forward rule is not tested?

kshyatt: The idea is to generate the derivatives from the forward pass then use them as the outputs to the reverse pass (to ensure you stay in the correct gauge) -- but it's very possible I did this wrong, I find the whole testing framework very hard to make sense of to be honest

yebai: @AstitvaAggarwal, can you help?

yebai:
@kshyatt, would you be open to combining this PR with https://github.com/chalk-lab/Mooncake.jl/pull/833?

kshyatt:
Definitely!

yebai:
Closed in favour of  #808

PR #797, opened 2025-10-02
NaN handling for some rrules

### LINKED ISSUE #794: Unused quantity causes NaNs in perfectly boring derivative
In this function, `b` is not actually used:
```julia
function f(x)
    a = x
    b = sqrt(abs(x-1.0))
    return a
end
```
Obviously, the derivative of `f` is really just 1.  Nonetheless, if we evaluate at an unfortunate spot for `b`, then Mooncake's derivative of `f` results in `NaN`:
```julia
julia> DifferentiationInterface.derivative(f, AutoMooncake(), 1.0)
NaN
```
Though the value of `b` at 1.0 is 0.0, its derivative is `NaN`.  That part is actually reasonable, but I did not expect it to affect the derivative of `a`.

Of course, if we comment out that line, we get the expected result:
```julia
julia> function g(x)
           a = x
           # b = sqrt(x-1)  # This is the only change from f
           return a
       end
g (generic function with 1 method)

julia> DifferentiationInterface.derivative(g, AutoMooncake(), 1.0)
1.0
```
But this is just an MWE; in my real code I can't just delete the offending line.

AstitvaAggarwal:
closes #794. 
This PR extends the NaN-handling logic to additional functions (cbrt, sqrt, and the log variants), which can also trigger the same issue described in #794.

Currently, correctness tests for these special cases will fail because the mathematically correct gradients would be NaN. However, since we want to avoid polluting upstream gradients in the computational graph, we return 0 instead in these cases.

Suggestions and feedback on whether and how to add tests for this special-case behaviour are welcome.
@willtebbutt @yebai

This issue primarily arises in Reverse Mode AD, therefore this PR addresses the same only for Reverse Mode AD.
In Forward Mode, each tangent is tagged uniquely so it's not a direct issue, unless in the case where final gradients used in downstream applications are sensitive to `Inf/NaN`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.171"
+version = "0.4.173"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"


src/rrules/builtins.jl
```diff
@@ -1250,6 +1250,7 @@ end
 
 function generate_derived_rrule!!_test_cases(rng_ctor, ::Val{:builtins})
     test_cases = Any[
+        (false, :none, nothing, _apply_iterate_equivalent, Base.iterate, *, 5.0, 4.0),
```
willtebbutt: Is this not just a repeat of the test case below?

src/rrules/low_level_maths.jl
```diff
@@ -105,11 +99,93 @@ function rrule!!(::CoDual{typeof(log)}, b::CoDual{P}, x::CoDual{P}) where {P<:IE
...
+    function log_adjoint(dy::P)
+        return NoRData(), ifelse(iszero(primal(x)), 0.0, dy / primal(x))
+    end
+    return zero_fcodual(y), log_adjoint
+end
+
+@is_primitive MinimalCtx Tuple{typeof(sqrt),IEEEFloat}
+function frule!!(::Dual{typeof(sqrt)}, x::Dual{P}) where {P<:IEEEFloat}
+    _x, dx = extract(x)
+    return Dual(sqrt(_x), dx / (2 * sqrt(_x)))
```
willtebbutt: `sqrt(_x)` is called twice here -- it would be good to only compute it once, and use it twice.

willtebbutt (commented):
Would you mind double checking that there's a `Float32` test case for any modified rules? I notice there is `0.0` being returned in various places, which _should_ causes errors in the `Float32` setting. Notably, the tests for `sqrt_llvm` and `sqrt_llvm_fast` seem to pass (I think that they ought to fail under the current implementation).

Additionally, would you mind adding a two-argument version of `hypot` and removing the `@from_chainrule` for the two-argument `hypot`? This way we can ensure consistent behaviour.

src/rrules/builtins.jl
AstitvaAggarwal: my bad i forgot to clean this up

AstitvaAggarwal:
@willtebbutt yeah in the rrules for `log` typecasting for `base` was required. Also ive set the rrule for `hypot` to take in account `Vararg` inputs as it has many dispatches (1/2/3... arguments).

yebai:
@moble, I'd encourage you to review this as well, as it addresses your issue in https://github.com/chalk-lab/Mooncake.jl/issues/794.

src/rrules/low_level_maths.jl
```diff
@@ -105,11 +99,94 @@ function rrule!!(::CoDual{typeof(log)}, b::CoDual{P}, x::CoDual{P}) where {P<:IE
     y = log(primal(b), primal(x))
     function log_adjoint(dy::P)
         log_b = log(primal(b))
-        return NoRData(), -dy * y / (log_b * primal(b)), dy / (primal(x) * log_b)
+        return NoRData(),
+        ifelse(iszero(primal(b)) || isnan(log_b), P(0), -dy * y / (log_b * primal(b))),
```
sunxd3: edge case (`b` can be 1): maybe add `iszero(log_b)`?

sunxd3 (commented):
two high level things:
1. I am not sure what is the "accepted" behavior of this. Is there a way to allow `NaN` to propagate for debugging fail-fast?
2. Maybe make it clear that this is only for reverse mode in the description?

moble:
> @moble, I'd encourage you to review this as well, as it addresses your issue in #794.

This PR is doing something quite different from @willtebbutt 's suggestion in that issue: he suggested to test for when the "argument to the reverse pass is zero", *not* when the derivative is zero.  So, by that logic, all of these tests are testing the wrong quantity.  That is, if `dy` is nonzero, I would still usually expect to get an infinite derivative; but when `dy` is zero, I really do want a zero return to not pollute the return when a derivative is not used (as in #794).

That said, while several other AD backends return Inf at that value, Enzyme returns 0.0 — so maybe there's some argument for it?

I've put together a little code to look at various backends acting on `sqrt` and the `f` I pointed out in that issue, with a nonzero or zero `dy`.  Note that for `f`, the pullback of `dy` should just be `dy`.

```julia
using ForwardDiff: ForwardDiff
using Enzyme: Enzyme
using Zygote: Zygote
using ReverseDiff: ReverseDiff

function f(x)
    a = x
    b = sqrt(abs(x-1.0))
    return a
end

for backend in (AutoForwardDiff(), AutoEnzyme(), AutoZygote(), AutoReverseDiff())
    @info "$backend" pullback(sqrt, backend, 0.0, (1.0,)) pullback(sqrt, backend, 0.0, (0.0,)) pullback(f, backend, 1.0, (1.0,)) pullback(f, backend, 1.0, (0.0,))
end
```

The results are

```julia
┌ Info: AutoForwardDiff()
│   pullback(sqrt, backend, 0.0, (1.0,)) = (Inf,)
│   pullback(sqrt, backend, 0.0, (0.0,)) = (NaN,)
│   pullback(f, backend, 1.0, (1.0,)) = (1.0,)
└   pullback(f, backend, 1.0, (0.0,)) = (0.0,)
┌ Info: AutoEnzyme()
│   pullback(sqrt, backend, 0.0, (1.0,)) = (0.0,)
│   pullback(sqrt, backend, 0.0, (0.0,)) = (0.0,)
│   pullback(f, backend, 1.0, (1.0,)) = (1.0,)
└   pullback(f, backend, 1.0, (0.0,)) = (0.0,)
┌ Info: AutoZygote()
│   pullback(sqrt, backend, 0.0, (1.0,)) = (Inf,)
│   pullback(sqrt, backend, 0.0, (0.0,)) = (NaN,)
│   pullback(f, backend, 1.0, (1.0,)) = (1.0,)
└   pullback(f, backend, 1.0, (0.0,)) = (0.0,)
┌ Info: AutoReverseDiff()
│   pullback(sqrt, backend, 0.0, (1.0,)) = (Inf,)
│   pullback(sqrt, backend, 0.0, (0.0,)) = (NaN,)
│   pullback(f, backend, 1.0, (1.0,)) = (NaN,)
└   pullback(f, backend, 1.0, (0.0,)) = (NaN,)
```

Notably, `ReverseDiff` also fails on `f(1.0)`, but all the others get it right.

AstitvaAggarwal:
@moble thank you for your review. I have a few points around the issue. AD libraries can handle Inf or NaN gradients in two following ways:

1) Return as is `Inf/NaN`. This is mathematically correct but limits usage for ML systems, optimisers or solvers that depend on gradient updates (gradient explosion, NaN sudden stopping of training). These will always result in gradient poisoning and are only suitable for functions being differentiated in safe domains.

2) Return 0. This is a more Robust/safe way of handling the situation for applications using the AD library as the invalid gradient update for parameters is simply set to 0 and is ignored. This ensures that the following iterations or larger function's gradient is not Inf/NaN. This is however "mathematically incorrect".

3) The simplest example is using the function :

```
function f(x)
    a = x
    b = sqrt(abs(x-1.0))
    return a
end
```

Here in the CFG when computing the gradient, if we dont set the gradient of b as 0, it will pollute the final function f's gradient as follows:

df/dx=df*(da/dx) + df*(db/dx)
df/dx=something + NaN/Inf.

Therefore it is imperative to set db/dx as 0 when encountering Inf/NaN.

4) In the PR I specifically test for the edge cases where for some common function's the gradient of the leaf node in the program CFG can end up with Inf/NaN values and return them as 0. Therefore the tests must only pass if for those specific functions the gradient is 0 at the edge case inputs.

5) for f above, ReverseDiff is unable to handle it, While all the rest are.
```
julia> DifferentiationInterface.derivative(f, AutoMooncake(), 1.0)
1.0

julia> DifferentiationInterface.derivative(f, AutoEnzyme(), 1.0)
1.0

julia> DifferentiationInterface.derivative(f, AutoZygote(), 1.0)
1.0

julia> DifferentiationInterface.derivative(f, AutoForwardDiff(), 1.0)
1.0

julia> DifferentiationInterface.derivative(f, AutoReverseDiff(), 1.0)
NaN

```

```
julia> DifferentiationInterface.derivative(sqrt, AutoMooncake(), 0.0)
0.0

julia> DifferentiationInterface.derivative(sqrt, AutoEnzyme(), 0.0)
0.0

julia> DifferentiationInterface.derivative(sqrt, AutoZygote(), 0.0)
Inf

julia> DifferentiationInterface.derivative(sqrt, AutoForwardDiff(), 0.0)
Inf


```

6) As suggested by @sunxd3, I agree there is a possibility of an option where NaN propagation may be allowed. This would allow for getting safe + mathematically correct results. I feel like this is only possible if I set a wrapper around NaN and Inf and allow arithmetic operations to treat the wrapper as zero while the user gets something like NAN_VAL/INF_VAL. But maybe this is too much of a stretch as this PR allows for handling of Inf/NaN gradients exactly how `Enzyme` does.

AstitvaAggarwal:
@moble Thank you for your review. I have a few points, clarifications regarding the PR. AD libraries can handle `Inf` or `NaN` gradients in essentially **two ways**:

1. **Return `Inf/NaN` as-is.**
   This is mathematically correct but can limit usage in ML systems, optimizers, or solvers that depend on gradient updates. Gradient explosion or sudden NaNs can halt training. This approach is suitable only when differentiating functions in safe domains.

Note: 
`Inf`: 
1) propagate as `Inf` (potentially destabilizing optimizer), or
2) multiply with 0 → produce `NaN` → then poison.

`NaN`: immediate poisoning; almost always invalidates the gradient.

2. **Return 0.**
   This is a more robust and safe approach for practical applications. Invalid gradient updates are simply set to 0 and ignored, ensuring subsequent iterations or larger function gradients are not polluted by `Inf/NaN`. The tradeoff is that this is "mathematically incorrect", but it prevents optimizer failures.

---

**Example:**

```julia
function f(x)
    a = x
    b = sqrt(abs(x-1.0))
    return a
end
```

* In the computational graph (CFG), if we do **not** set the gradient of `b` to 0, it can pollute the final gradient of `f`:

```
df/dx = df*(da/dx) + df*(db/dx)
      = something + NaN/Inf
```

* Therefore, it is imperative to set `db/dx = 0` whenever `Inf`/`NaN` is encountered.

---

In this PR, I specifically test edge cases where leaf-node gradients in the CFG can end up as `Inf`/`NaN` and therefore must return them as 0. Tests pass only **if the gradient at these edge-case inputs is 0; they fail if it is `Inf/NaN`.**

For `f` above, **ReverseDiff** fails, while all other backends handle it correctly:

```julia
julia> DifferentiationInterface.derivative(f, AutoMooncake(), 1.0)
1.0

julia> DifferentiationInterface.derivative(f, AutoEnzyme(), 1.0)
1.0

julia> DifferentiationInterface.derivative(f, AutoZygote(), 1.0)
1.0

julia> DifferentiationInterface.derivative(f, AutoForwardDiff(), 1.0)
1.0

julia> DifferentiationInterface.derivative(f, AutoReverseDiff(), 1.0)
NaN
```

For `sqrt(x)` at 0, Mooncake and Enzyme return 0 (**safer for optimization**), while Zygote, ForwardDiff, and ReverseDiff return `Inf`, which may break gradient-based training (explained below) :

```julia
julia> DifferentiationInterface.derivative(sqrt, AutoMooncake(), 0.0)
0.0

julia> DifferentiationInterface.derivative(sqrt, AutoEnzyme(), 0.0)
0.0

julia> DifferentiationInterface.derivative(sqrt, AutoZygote(), 0.0)
Inf

julia> DifferentiationInterface.derivative(sqrt, AutoForwardDiff(), 0.0)
Inf

julia> DifferentiationInterface.derivative(sqrt, AutoReverseDiff(), 0.0)
Inf
```

In suppose a standard gradient descent step, returning `0` instead of `Inf/NaN` ensures that other parameters or parts of the model with valid gradients can still update normally.

For instance, in a multivariate function `g(x, y) = sqrt(y) + x^2`, if `y=0` and `grad(sqrt(y))` is `Inf`: 
```
julia> function g(x, y)
           a = x^2       # normal active branch
           b = sqrt(y)   # risky active branch if y = 0
           return a + b
       end

julia> grad = Zygote.gradient(g, 1.0, 0.0)

julia> println("dg/dx = ", grad[1],", dg/dy = ", grad[2])
dg/dx = 2.0, dg/dy = Inf

julia> cache = Mooncake.prepare_gradient_cache(g, (1.0, 0.0)...)

julia> Mooncake.value_and_gradient!!(cache, g, 1.0, 0.0)
(1.0, (Mooncake.NoTangent(), 2.0, 0.0))
```

Here in Mooncake, by returning `0` for the problematic `y` gradient, the optimizer ignores that step for `y` but can still update `x` safely, allowing training to progress rather than halting entirely.

---

Currently, the PR handles `Inf`/`NaN` gradients **similar to Enzyme**, prioritizing **robust backpropagation for ML, optimizer or solver use cases**.

Finally, as suggested by @sunxd3, one could imagine an option allowing `NaN` propagation while still protecting numerical stability. This might involve **wrapping `NaN` and `Inf`** so arithmetic treats them as 0 internally while exposing the original `NaN`/`Inf` to the user. However, this may be beyond the scope of this PR. (Maybe some kinda pruning in `BBcode IR` can also solve this...)

---

In case I am missing anything please let me know.

sunxd3:
I have not problem other than https://github.com/chalk-lab/Mooncake.jl/pull/797#discussion_r2409716119

yebai:
Cc @willtebbutt

AstitvaAggarwal:
GTM?

yebai:
Let's wait till Monday in case @moble or @willtebbutt wants to chime in.

PR #798, opened 2025-10-03
Some Housekeeping

willtebbutt:
1. adds `dual_ir` to the dev docs, alongisde `primal_ir`, `fwd_ir`, and `rvs_ir`.
2. renames the test case functions, since they're now used for both forward-mode and reverse-mode.


willtebbutt:
Thanks @sunxd3 . I'll let CI run again to completion then merge, having now bumped the patch version.

PR #802, opened 2025-10-08
Temporarily pin CI to 1.11

willtebbutt:
Since we're not quite ready for 1.12 yet, I figure we ought to run CI on 1.11 in the meantime.

### DIFF
--- a/bench/run_benchmarks.jl
+++ b/bench/run_benchmarks.jl
@@ -325,7 +325,7 @@ function benchmark_hand_written_rrules!!(rng_ctor)
         tags = fill(nothing, length(test_cases))
         return map(x -> x[4:end], test_cases), memory, ranges, tags
     end
-    return benchmark_rules!!(test_case_data, (lb=1e-3, ub=50.0), false, 0.02)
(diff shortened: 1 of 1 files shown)


willtebbutt:
@yebai is there any way to pin the Turing docs CI pipeline to use 1.11 rather than release?

yebai:
@shravanngoswamii

shravanngoswamii:
> @yebai is there any way to pin the Turing docs CI pipeline to use 1.11 rather than release?

@willtebbutt Just pass `julia-version` to DocsDocumenter in [documentation](https://github.com/chalk-lab/Mooncake.jl/blob/4169e2dd68224454ea060b7ee05d9a05926ec588/.github/workflows/documentation.yml#L18C1-L22C53) GHA like this:
```
uses: TuringLang/actions/DocsDocumenter@main
with:
  julia-version: '1.11'
```

See: https://github.com/TuringLang/actions/blob/main/README.md#docsdocumenter

willtebbutt:
I'm going to go ahead and merge this as I would rather like to have CI running again so that I can push forward some other PRs in the morning (when their CI will hopefully have passed), and I'm quite sure this is uncontraversial. If anyone feels otherwise, I will be more than happy to revert in the morning.

Issue #672, opened 2025-08-03
Type Checking in Forward Mode


willtebbutt:
I can see where you're coming from with this, but I think it's better done as part of debug mode (to check for internal bugs) and as part of `prepare_derivative_cache` (to check for user errors).

I think this needs doing, but I'm going to open a separate issue to implement it in order to get this PR merged.

_Originally posted by @willtebbutt in https://github.com/chalk-lab/Mooncake.jl/pull/389#discussion_r2250047102_

yebai:
> verify that the user has provided the correct tangent type when passing in Duals in forward mode, nor do we have a functional debug mode in forward mode

Though I am supportive of merging #309, I think the above two features are very, very important.

sumiya11:
Hi @willtebbutt, do you think type checking for Forward mode should be implemented similarly to how it is done in Reverse mode? 

For example, by adding a method `(rule::DebugFRule)(x::Vararg{Dual,N}) where {N}` that would check the consistency of `x`, compute `y = rule.rule(x)`, and check the consistency of `y`. This method can live just below:

https://github.com/chalk-lab/Mooncake.jl/blob/3320db7f2e6764dadfa6fe12e28b268e8a88f996/src/debug_mode.jl#L2

sumiya11:
Related note: would it be fine to test `DebugFRule` similarly to how debug_mode is tested for Reverse mode? 

By looking at `test/debug_mode.jl`, I can infer some examples that look like they should error in debug mode in Forward mode:

```julia
using Mooncake: zero_dual, build_frule, Dual

# 1. Size mismatch
rule = build_frule(zero_dual(size), randn(10));
rule(zero_dual(size), Dual(randn(11), randn(10)))

# 2. Type mismatch
rule = build_frule(zero_dual(identity), Any[1.]);
rule(zero_dual(identity), Dual(Any[1.], Any[1.f16]))

# 3. Other examples?
```

yebai:
@willtebbutt, can you spell out what the checks you would like are so others can pick up? Or, would you rather implement this yourself?

Technici4n:
From a quick read, there is indeed expanding `DebugFRule` to perform similar checks as `DebugRRule`.

I think there is also another layer of validation to be done:
[...]

Issue #777, opened 2025-09-19
Nested forward stack overflow


Technici4n:
Given the following code:
```jl
using DifferentiationInterface
import ForwardDiff
import Mooncake

function mooncake_derivative(f, x)
    rule = Mooncake.prepare_derivative_cache(f, x)
    Mooncake.value_and_derivative!!(
        rule,
        Mooncake.zero_dual(f),
        Mooncake.Dual(x, one(x))) |> Mooncake.tangent
end
```

I am able to compute a second derivative using Mooncake over ForwardDiff:
```jl
der_mc = mooncake_derivative(
    x -> x * DifferentiationInterface.derivative(
        y -> x + y^2,
        AutoForwardDiff(),
        x),
    1.0)
```

But Mooncake over Mooncake gives a stack overflow:
```jl
der_mc = mooncake_derivative(
    x -> x * mooncake_derivative(
        y -> x + y^2,
        x),
    1.0)
```
the exception is:
```
ERROR: LoadError: StackOverflowError:
Stacktrace:
      [1] macro expansion
        @ ~\.julia\packages\Mooncake\fW8wv\src\tangents.jl:469 [inlined]
      [2] macro expansion
        @ .\none:0 [inlined]
      [3] tangent_type(::Type{Core.MethodInstance})
        @ Mooncake .\none:0
--- the above 3 lines are repeated 79982 more times ---
 [239950] macro expansion
        @ ~\.julia\packages\Mooncake\fW8wv\src\tangents.jl:469 [inlined]
 [239951] macro expansion
        @ .\none:0 [inlined]
```

yebai:
Likely related to https://github.com/chalk-lab/Mooncake.jl/issues/632#issuecomment-3192451600 and https://github.com/chalk-lab/Mooncake.jl/pull/747#issuecomment-3245384872

willtebbutt:
This should now be resolved on main / version 0.4.164 (which should be released soon). I'm going to close this for the time being, but please do re-open if this hasn't resolved the problem!

gdalle:
Not really resolved in the sense that nested forward mode still doesn't work, but the error is now different, so I'd suggest keeping this open

Technici4n:
Indeed, the new error is this:
```
[...]

Issue #632, opened 2025-06-29
`DI.hessian`: Tuple field type cannot be `Union{}` 


timholy:
This example arose from using [GAMSFiles](https://github.com/HolyLab/GAMSFiles.jl) to parse a bunch of files used as a test suite for optimization problems (specifically `convex3_10_1.gms`). I've set this up so you don't need GAMSFiles to run this demo, although there are commented-out references to that package. (In my local checkout of that package I defined a `@zero_adjoint` for `GAMSFiles.checkaxes`, and that fixed one problem, woohoo. But this is different.)

Unless I'm missing something, the only operations in this objective are `+`, `-`, and `*`. This is with Mooncake v0.4.129 and DI v0.7.1.

```julia
using Mooncake
using DifferentiationInterface

module MooncakeDemo
    # using GAMSFiles
    function objective(x)
    #   begin
        #   (GAMSFiles.checkaxes)(x, (Base.OneTo(10),))
          objvar = (((((((((((((((((((((((((((x[1] * x[1] + x[10] * x[10]) * (x[1] * x[1] + x[10] * x[10]) - 4 * x[1]) + 3) + (x[2] * x[2] + x[10] * x[10]) * (x[2] * x[2] + x[10] * x[10])) - 4 * x[2]) + 3) + (x[3] * x[3] + x[10] * x[10]) * (x[3] * x[3] + x[10] * x[10])) - 4 * x[3]) + 3) + (x[4] * x[4] + x[10] * x[10]) * (x[4] * x[4] + x[10] * x[10])) - 4 * x[4]) + 3) + (x[5] * x[5] + x[10] * x[10]) * (x[5] * x[5] + x[10] * x[10])) - 4 * x[5]) + 3) + (x[6] * x[6] + x[10] * x[10]) * (x[6] * x[6] + x[10] * x[10])) - 4 * x[6]) + 3) + (x[7] * x[7] + x[10] * x[10]) * (x[7] * x[7] + x[10] * x[10])) - 4 * x[7]) + 3) + (x[8] * x[8] + x[10] * x[10]) * (x[8] * x[8] + x[10] * x[10])) - 4 * x[8]) + 3) + (x[9] * x[9] + x[10] * x[10]) * (x[9] * x[9] + x[10] * x[10])) - 4 * x[9]) + 3) - 0
          return objvar
    #   end
  end
end
x0 = [0.0; fill(1.0, 9)]
f = MooncakeDemo.objective
f(x0)  # works

[...]

PR #808, opened 2025-10-17
Add Type Checking in Forward Mode

sunxd3:
start to address https://github.com/chalk-lab/Mooncake.jl/issues/672

mostly conceptually copy-paste

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -61,7 +61,7 @@ jobs:
             version: 'lts'
             arch: x64
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: julia-actions/setup-julia@v2
         with:
           version: ${{ matrix.version }}
@@ -116,7 +116,7 @@ jobs:
         arch:
           - x64
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: julia-actions/setup-julia@v2
         with:
           version: ${{ matrix.version }}
@@ -149,7 +149,7 @@ jobs:
           - 'hand_written'
           - 'derived'
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: julia-actions/setup-julia@v2
         with:
           version: '1.11'
@@ -168,7 +168,7 @@ jobs:
     strategy:
       fail-fast: false
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: julia-actions/setup-julia@v2
         with:
           version: '1.11'


yebai:
Thanks, @sunxd3 — I’ll take a look at this soon. In the meantime, could you check whether the debugging mode helps us identify the cause of https://github.com/chalk-lab/Mooncake.jl/issues/777 / https://github.com/chalk-lab/Mooncake.jl/issues/632#issuecomment-3330837960

src/debug_mode.jl
```diff
@@ -1,5 +1,91 @@
-# TODO: make it non-trivial. See https://github.com/chalk-lab/Mooncake.jl/issues/672
-DebugFRule(rule) = rule
+"""
+    DebugFRule(rule)
+
+Construct a callable equivalent to `rule` but with additional type checking for forward-mode
+AD. Checks:
+- Each `Dual` argument has tangent type matching `tangent_type(typeof(primal))`
```
yebai, line 6: Double check: `tangent_type(typeof(primal))` is for reverse mode only? 

```suggestion
- Each `Dual` argument has tangent type matching `tangent_type(typeof(primal))`
```

src/debug_mode.jl
```diff
@@ -1,5 +1,91 @@
...
+using Mooncake
+
+f(x) = sum(x)
+rule = build_frule(zero_dual(f), randn(10); debug_mode=true)
+
+# Valid input - passes
+result = rule(zero_dual(f), Dual(randn(10), randn(10)))
+
+# Invalid input - throws ErrorException wrapping the underlying validation error
+rule(zero_dual(f), Dual(randn(10), randn(11)))  # Size mismatch
```
yebai: > rule(zero_dual(f), Dual(randn(10), randn(11))) # Size mismatch

I don't think this is correct -- both `frule` and `rrule` only dispatch on argument types, _NOT_ their values.

sunxd3: this will trap https://github.com/chalk-lab/Mooncake.jl/blob/e06720509d58751cbda65dd23d5d0e28d2cb0e3a/src/fwds_rvs_data.jl#L330-L333

but I think the example here is a bit confusing. the functions are not to be called individually anyway, so I removed them.

src/debug_mode.jl
sunxd3: my thinking is that forward mode still uses tangent type, so it should enforce
https://github.com/chalk-lab/Mooncake.jl/blob/e06720509d58751cbda65dd23d5d0e28d2cb0e3a/docs/src/developer_documentation/forwards_mode_design.md#L45-L57

Technici4n: Both forward and reverse use the `tangent_type`, yeah. (Note that in reverse mode the tangent is split into fdata and rdata as well).

src/debug_mode.jl
```diff
@@ -1,5 +1,78 @@
...
+
+# Recursively copy the wrapped rule
+_copy(x::P) where {P<:DebugFRule} = P(_copy(x.rule))
+
+"""
+    (rule::DebugFRule)(x::Vararg{Dual,N}) where {N}
+
+Apply pre- and post-condition type checking. See [`DebugFRule`](@ref).
+"""
+@noinline function (rule::DebugFRule)(x::Vararg{Dual,N}) where {N}
```
Technici4n: Did you also look into providing an equivalent of `verify_args`?

src/debug_mode.jl
```diff
@@ -1,5 +1,78 @@
...
+    end
+end
+
+@noinline function verify_dual_value(d::Dual)
+    p, t = primal(d), tangent(d)
+
+    # Fast path: type-level check
+    T_expected = tangent_type(typeof(p))
+    T_actual = typeof(t)
+    if !(T_actual <: T_expected)
```
Technici4n: Would it not be better to write `Dual{TP, TT}` in the function signature and check that `tangent_type(TP) == TT`?

src/debug_mode.jl
```diff
@@ -1,5 +1,78 @@
...
+        throw(
+            InvalidFDataException(
+                "Dual tangent type mismatch: primal $(typeof(p)) requires tangent type " *
+                "$T_expected, but got $T_actual",
+            ),
+        )
+    end
+
+    # Slow path: deep structural validation
```
Technici4n, line 125: Do you want to add a comment that we check `fdata` and `rdata` separately such that the logic can be shared with reverse mode?

test/debug_mode.jl
```diff
@@ -41,4 +41,87 @@
     rule_with_bad_pb_length(x::CoDual{Float64}) = x, dy -> (5, 5.0) # returns the wrong type
     y, pb!! = Mooncake.DebugRRule(rule_with_bad_pb_length)(zero_fcodual(5.0))
     @test_throws ErrorException pb!!(1.0)
+
+    # Forward mode debug tests
+    @testset "forward debug mode" begin
```
Technici4n, line 47: Maybe move to top-level and rename the other testset to `"reverse debug mode"`?

src/debug_mode.jl
```diff
@@ -1,5 +1,78 @@
...
+Apply pre- and post-condition type checking. See [`DebugFRule`](@ref).
+"""
+@noinline function (rule::DebugFRule)(x::Vararg{Dual,N}) where {N}
+    verify_dual_inputs(rule.rule, x)
+    y = rule.rule(x...)
+    verify_dual_output(x, y)
+    return y::Dual
+end
+
+@noinline function verify_dual_inputs(rule, @nospecialize(x::Tuple))
```
Technici4n: are you sure you want to specialize on the `rule`, which is currently not used?

Technici4n: note: it will be used if something like `verify_args` is added

src/debug_mode.jl
sunxd3: I haven't put thoughts on it, maybe another PR?

sunxd3:
thanks Bruno

src/debug_mode.jl
Technici4n: Sure yeah

yebai: Please feel free to add a TODO item "TODO: need to implement `verify_args`"

Technici4n:
I wonder why this segfaults on 1.10 😓

sunxd3:
https://github.com/chalk-lab/Mooncake.jl/blob/e4b657614318a32ba3a7a9e85cf41901ef51b79a/test/debug_mode.jl#L84-L89 triggered the segfault.

I think they related to https://github.com/JuliaLang/julia/issues/51016. Essentially, when `DebugFRule` receives a `Dual{Float64, Float32}` but the inner rule was compiled for `Dual{Float64, Float64}`, the compiler segfaults while trying to specialize the call.

Not certain if `Base.inferencebarrier` is the best solution here, but it seems to prevent the crash.

Technici4n:
Is this maybe caused by the lack of `verify_args`?

sunxd3:
I think adding `verify_args` alone wouldn't solve it -- it won't prevent the Julia compiler from trying to do the type inference. 

I added `verify_args` similar to the reverse_mode one, but also had to wrap the whole debug rule function in generated block so that we can properly intercept the error.

src/debug_mode.jl
```diff
@@ -25,17 +25,65 @@ _copy(x::P) where {P<:DebugFRule} = P(_copy(x.rule))
...
 Apply pre- and post-condition type checking. See [`DebugFRule`](@ref).
 """
-@noinline function (rule::DebugFRule)(x::Vararg{Dual,N}) where {N}
-    verify_dual_inputs(x)
-    # On Julia 1.10, calling an OpaqueClosure with mismatched types causes a segfault
-    # during codegen (JuliaLang/julia#51016). Use inferencebarrier to prevent specializing.
-    @static if VERSION < v"1.11-"
-        y = Base.inferencebarrier(rule.rule)(x...)
-    else
+@static if VERSION < v"1.11-"
```
Technici4n, line 28: Since this is not needed for reverse debug mode, I am a bit skeptical that it's required? 😄

sunxd3: it's tricky, reverse debug mode was not testing the case where the segfault appear for the forward debug mode
https://github.com/chalk-lab/Mooncake.jl/blob/bcfa006fff53bfc2e057e0122dc54c98356f3f97/test/debug_mode.jl#L3-L9

but e.g., same segfault will happen with 
```julia
julia +lts --project -e '
      using Mooncake
      using Mooncake: CoDual, zero_fcodual, build_rrule

      f = x -> 5x
      rule = build_rrule(f, 5.0; debug_mode=true)

      rule(zero_fcodual(f), CoDual(5.0, 1.0))
      '
```

Technici4n: Ah I see. Funny how simple PRs turn into a rabbit hole sometimes... 😅

sunxd3:
Ideally, the coding style between forward and backward debug mode (tests too) would match better, but I don't want to make more style only changes now

yebai:
@penelopeysm, can you help take a look?

src/debug_mode.jl
```diff
@@ -1,5 +1,135 @@
...
+
+        return quote
+            verify_dual_inputs(x)
+            y = rule.rule(x...)
+            verify_dual_output(x, y)
+            return y
+        end
+    end
+else
+    @noinline function (rule::DebugFRule)(x::Vararg{Dual,N}) where {N}
```
Technici4n, line 80: Nit: inline?

sunxd3: I am following `DebugRRule` (line 220) here and I think `noinline` makes more sense to me just from that we want this function to fail and correctly attribute to this definition. Frankly, I am not sure whether the inline hint is doing anything.

Technici4n: Sure, either is fine by me

yebai:
thanks @sunxd3 and @Technici4n -- please feel free to merge and make new releases!

Issue #810, opened 2025-10-21
`MethodError` for `CuArray` in reverse mode


kshyatt:
MWE:
```julia
using CUDA, Mooncake, Random

A = CUDA.rand(Float64, 16, 16)
B = CUDA.rand(Float64, 16, 16)
Mooncake.TestUtils.test_rule(Random.default_rng(), *, A, B; mode=Mooncake.ReverseMode, is_primitive=false)
```

When I run this I see:

```julia
MethodError: no method matching rrule!!(::Mooncake.CoDual{typeof(Mooncake.lgetfield), Mooncake.NoFData}, ::Mooncake.CoDual{CuArray{Float64, 2, CUDA.DeviceMemory}, CuArray{Float64, 2, CUDA.DeviceMemory}}, ::Mooncake.CoDual{Val{:dims}, Mooncake.NoFData})
  The function `rrule!!` exists, but no method is defined for this combination of argument types.
  
  Closest candidates are:
    rrule!!(::Mooncake.CoDual{typeof(Mooncake.lgetfield)}, !Matched::Mooncake.CoDual{Task, Mooncake.TaskTangent}, ::Mooncake.CoDual{Val{f}}) where f
     @ Mooncake ~/.julia/packages/Mooncake/cIVTD/src/rrules/tasks.jl:86
    rrule!!(!Matched::Mooncake.CoDual{typeof(setfield!)}, ::Mooncake.CoDual, ::Mooncake.CoDual, !Matched::Mooncake.CoDual)
     @ Mooncake ~/.julia/packages/Mooncake/cIVTD/src/rrules/builtins.jl:892
    rrule!!(!Matched::Mooncake.CoDual{typeof(typeassert)}, ::Mooncake.CoDual, ::Mooncake.CoDual)
     @ Mooncake ~/.julia/packages/Mooncake/cIVTD/src/rrules/builtins.jl:939
```

This seems pretty deep in the internals so I wasn't sure how to approach it. 

I also see:

```julia
Correctness: Error During Test at /home/kshyatt/.julia/packages/Mooncake/cIVTD/src/test_utils.jl:946
  Got exception outside of a @test
  MethodError: no method matching _add_to_primal_internal(::IdDict{Any, Any}, ::CuArray{Float64, 2, CUDA.DeviceMemory}, ::CuArray{Float32, 2, CUDA.DeviceMemory}, ::Bool)
[...]

Issue #278, opened 2024-10-01
GPU Rules


willtebbutt:
Places to look for rules and tests for GPU stuff:
- Zygote.jl tests
- Tracker.jl tests
- Lux.jl integration test
- LuxLib.jl unit tests
- NNlib.jl unit tests
- Flux.jl unit tests

@yebai anything else that you can think of that we should be looking at?

kshyatt:
A couple more rules we need for CUDA arrays:
- [ ] `sort` and friends
- [x] `broadcast`, `mapreduce`
- [ ] LinearAlgebra methods (CUBLAS)
    - [x] ~~LinearAlgebra.mul! / generic_matmatmul!~~ Fixed by #1044
- [x] Whatever is blocking `adjoint` in forward mode (`_new_`?)

Will add more as I find them

Technici4n:
For cuBLAS I wonder if there would be a way to share the rules with the standard BLAS 😅

kshyatt:
There may be but there are some subtleties (which I am happy to discuss)

yebai:
@kshyatt, as discussed, it is helpful to make the above list more comprehensive and specify what is missing to provide reasonable GPU support for linear algebra. 

It's important to be realistic about one's bandwidth and aim for a very high standard within those stated scopes.

yebai:
cc @lkdvos

yebai:
> [!CAUTION]
>  Produced with help from Copilot. 

## What Mooncake currently has (`ext/MooncakeCUDAExt.jl`)

| Rule | Coverage |
|---|---|
| Tangent type system | `CuArray` is its own tangent; `CuPtr`, `CuRefValue`, `CuContext` handled |
| `CuArray(undef, dims...)` | Construction of uninitialised arrays |
| `lgetfield` on `CuArray` | Field access (`.data`, `.dims`) |
| `adjoint`/`transpose` of `CuArray` | Via `_new_` rules |
| `sum(::CuFloatArray)` | Reduction to scalar — **reals only, no function arg** |
| `mul!(C,A,B)` for `CuMaybeComplexArray` | 3-arg GEMM via CUBLAS |
| `@from_rrule` NNlib ops | conv, pool, softmax, bias_act!, etc. |

---

## What's missing (vs. Zygote's CUDA test suite)

### 1. GPU broadcasting — the root gap, blocks most tests

[...]

PR #811, opened 2025-10-21
Make tangent of CuArray have the same type

kshyatt:
The current method causes the tangent to have element type `Float32` no matter the element type of the primal array

Fix https://github.com/chalk-lab/Mooncake.jl/issues/810

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -214,7 +214,7 @@ function __exclude_func_with_unsupported_output(fx)
     _fx = deepcopy(fx)
     _func, _args = _fx[1], _fx[2:end]
     _y = _func(_args...)
-    __exclude_unsupported_output(_y)
+    return __exclude_unsupported_output(_y)
 end
 
 """
@@ -427,6 +427,8 @@ end
     prepare_pullback_cache(f, x...)
 
 Returns a cache used with [`value_and_pullback!!`](@ref). See that function for more info.
+
+The API guarantees that tangents are initialized at zero before the first autodiff pass.
 """
 @unstable function prepare_pullback_cache(fx...; kwargs...)
 
@@ -450,7 +452,7 @@ Returns a cache used with [`value_and_pullback!!`](@ref). See that function for 
 end
 
 """
-    value_and_pullback!!(cache::Cache, ȳ, f, x...)
+    value_and_pullback!!(cache::Cache, ȳ, f, x...; args_to_zero=(true, ...))
 
 !!! info
     If `f(x...)` returns a scalar, you should use [`value_and_gradient!!`](@ref), not this
@@ -480,6 +482,10 @@ will return both to their original state as part of the process of computing the
     around over multiple calls to this function with the same `cache`, you should take a
     copy (using `copy` or `deepcopy`) of them before calling again.
 
+The keyword argument `args_to_zero` is a tuple of boolean values specifying which cotangents should be reset to zero before differentiation.
+It contains one boolean for each element of `(f, x...)`.
+It is used for performance optimizations if you can guarantee that the initial cotangent allocated in `cache` (created by `zero_tangent`) never needs to be zeroed out again.
+
 # Example Usage
 ```jldoctest
 f(x, y) = sum(x .* y)
@@ -493,8 +499,14 @@ Mooncake.value_and_pullback!!(cache, 1.0, f, x, y)
 (4.0, (NoTangent(), [1.0, 1.0], [2.0, 2.0]))
 ```
 """
-function value_and_pullback!!(cache::Cache, ȳ, f::F, x::Vararg{Any,N}) where {F,N}
-    tangents = tuple_map(set_to_zero!!, cache.tangents)
+function value_and_pullback!!(
+    cache::Cache,
+    ȳ,
+    f::F,
+    x::Vararg{Any,N};
+    args_to_zero::NTuple=ntuple(Returns(true), Val(N + 1)),
+) where {F,N}
+    tangents = tuple_map(set_to_zero_maybe!!, cache.tangents, args_to_zero)
     coduals = tuple_map(CoDual, (f, x...), tangents)
     return __value_and_pullback!!(cache.rule, ȳ, coduals...; y_cache=cache.y_cache)
 end
@@ -503,6 +515,8 @@ end
     prepare_gradient_cache(f, x...)
 
 Returns a cache used with [`value_and_gradient!!`](@ref). See that function for more info.
+
+The API guarantees that tangents are initialized at zero before the first autodiff pass.
 """
 @unstable function prepare_gradient_cache(fx...; kwargs...)
     rule = build_rrule(fx...; kwargs...)
@@ -514,7 +528,7 @@ Returns a cache used with [`value_and_gradient!!`](@ref). See that function for 
 end
 
 """
-    value_and_gradient!!(cache::Cache, f, x...)
+    value_and_gradient!!(cache::Cache, f, x...; args_to_zero=(true, ...))
 
 Computes a 2-tuple. The first element is `f(x...)`, and the second is a tuple containing the
 gradient of `f` w.r.t. each argument. The first element is the gradient w.r.t any
@@ -538,6 +552,10 @@ will return both to their original state as part of the process of computing the
     around over multiple calls to this function with the same `cache`, you should take a
     copy (using `copy` or `deepcopy`) of them before calling again.
 
+The keyword argument `args_to_zero` is a tuple of boolean values specifying which cotangents should be reset to zero before differentiation.
+It contains one boolean for each element of `(f, x...)`.
+It is used for performance optimizations if you can guarantee that the initial cotangent allocated in `cache` (created by `zero_tangent`) never needs to be zeroed out again.
+
 # Example Usage
 ```jldoctest
 f(x, y) = sum(x .* y)
@@ -551,8 +569,14 @@ value_and_gradient!!(cache, f, x, y)
 (4.0, (NoTangent(), [1.0, 1.0], [2.0, 2.0]))
 ```
 """
-function value_and_gradient!!(cache::Cache, f::F, x::Vararg{Any,N}) where {F,N}
-    coduals = tuple_map(CoDual, (f, x...), tuple_map(set_to_zero!!, cache.tangents))
+function value_and_gradient!!(
+    cache::Cache,
+    f::F,
+    x::Vararg{Any,N};
+    args_to_zero::NTuple=ntuple(Returns(true), Val(N + 1)),
+) where {F,N}
+    tangents = tuple_map(set_to_zero_maybe!!, cache.tangents, args_to_zero)
+    coduals = tuple_map(CoDual, (f, x...), tangents)
     return __value_and_gradient!!(cache.rule, coduals...)
 end
 
--- a/src/tangents.jl
+++ b/src/tangents.jl
@@ -910,6 +910,19 @@ set_to_zero!!(x, ::Val{true}) = set_to_zero_internal!!(Vector{UInt}(), x)
 set_to_zero!!(x, ::Val{false}) = set_to_zero_internal!!(NoCache(), x)
 
 """
+    set_to_zero_maybe!!(x, doit::Bool)
+
+If `doit` is `true`, return `set_to_zero!!(x)`, otherwise return `x`.
+"""
+function set_to_zero_maybe!!(x, doit::Bool)
+    if doit
+        return set_to_zero!!(x)
+    else
+        return x
+    end
+end
+
+"""
     set_to_zero_internal!!(c::SetToZeroCache, x)
 
 Implementation for [`Mooncake.set_to_zero!!`](@ref). Use `c` to ensure that circular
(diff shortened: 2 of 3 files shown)


Technici4n:
Would the other uses of `CuFloatArray` not need adapting as well? (`CuFloatArray` should probably be removed entirely in favor of `CuArray{T}`)

kshyatt:
They may well but this was the first one I came across

yebai:
@Technici4n, would you like to review this formally?

ext/MooncakeCUDAExt.jl
```diff
@@ -33,19 +33,49 @@ import Mooncake.TestUtils:
     populate_address_map_internal, AddressMap, __increment_should_allocate
 
 const CuFloatArray = CuArray{<:IEEEFloat}
+const CuComplexArray = CuArray{<:Complex{<:IEEEFloat}}
 
 # Tell Mooncake.jl how to handle CuArrays.
 
+Mooncake.tangent_type(::Type{<:CuPtr}) = Mooncake.NoTangent
+Mooncake.tangent_type(::Type{<:CuStream}) = Mooncake.NoTangent
+Mooncake.tangent_type(::Type{<:DeviceMemory}) = Mooncake.NoTangent
```
Technici4n: What is the logic here? The tangent type of a `Ptr` is a `Ptr`, so I am a bit worried that e.g. giving `NoTangent` to a `CuPtr` is incorrect.

ext/MooncakeCUDAExt.jl
```diff
@@ -33,19 +33,49 @@ import Mooncake.TestUtils:
     populate_address_map_internal, AddressMap, __increment_should_allocate
 
 const CuFloatArray = CuArray{<:IEEEFloat}
+const CuComplexArray = CuArray{<:Complex{<:IEEEFloat}}
```
Technici4n, line 38: Could we make all the functions generic in `CuArray{<:T}`?

Technici4n: Maybe the existing `Array` functions in Mooncake (some seem to be in `src/rrules/array_legacy.jl`) can provide inspiration?

yebai: We test `CuArray{<:Complex{<:IEEEFloat}}`  and similarly `Array{<:IEEEFloat}` extensively, so this is a deliberate decision. The general rule of thumb is that we only allow array types with extensive test coverage.

ext/MooncakeCUDAExt.jl
kshyatt: Most likely me not quite understanding how the internals of Mooncake work 😓

ext/MooncakeCUDAExt.jl
Technici4n: Ah yes I see. I got confused between `Float32` and `IEEEFloat` somehow. Still, is there any testing of `CuArray` for something else than `Float32`? I couldn't find any.

Would it also not make sense to support any `CuArray{T}`, similar to how "any" `Array{T}` seems to be currently supported? (Even if relatively untested) I suppose it could also be left for another PR

(Note from CuArray:
```jl
# CuArray only supports element types that are allocated inline (`Base.allocatedinline`).
# These come in three forms:
# 1. plain bitstypes (`Int`, `(Float32, Float64)`, plain immutable structs, etc).
#    these are simply stored contiguously in memory.
# 2. structs of unions (`struct Foo; x::Union{Int, Float32}; end`)
#    these are stored with a selector at the end (handled by Julia).
# 3. bitstype unions (`Union{Int, Float32}`, etc)
#    these are stored contiguously and require a selector array (handled by us)
# As well as "mutable singleton" types like `Symbol` that use pointer-identity
```
)

yebai: > Still, is there any testing of CuArray for something else than Float32? I couldn't find any.

IIRC, GPU support is not complete. See, eg, #278 

> Even if relatively untested) I suppose it could also be left for another PR

I'd suggest we relax the type constraint for `CuArray{T}` in the same PR that adds tests.

ext/MooncakeCUDAExt.jl
yebai: In general, `NoTangent` should _NOT_ be used for types that contain "active data" (ie, their instance could contribute to gradients). 

So, let's remove these `tangent_type` definitions.

ext/MooncakeCUDAExt.jl
```diff
@@ -33,19 +33,49 @@ import Mooncake.TestUtils:
...
+
 Mooncake.@foldable tangent_type(::Type{P}) where {P<:CuFloatArray} = P
 function zero_tangent_internal(x::CuFloatArray, dict::MaybeCache)
     haskey(dict, x) && return dict[x]::tangent_type(typeof(x))
     t = zero(x)
     dict[x] = t
     return t
 end
-function randn_tangent_internal(rng::AbstractRNG, x::CuFloatArray, dict::MaybeCache)
+function randn_tangent_internal(
```
yebai: Looks good to me!

ext/MooncakeCUDAExt.jl
```diff
@@ -33,19 +33,46 @@ import Mooncake.TestUtils:
...
 const CuFloatArray = CuArray{<:IEEEFloat}
+const CuComplexArray = CuArray{<:Complex{<:IEEEFloat}}
 
 # Tell Mooncake.jl how to handle CuArrays.
 
+Mooncake.@foldable tangent_type(::Type{<:CuArray{P,N,M}}) where {P,N,M} = CuArray{
+    tangent_type(P),N,M
+}
+
+#Mooncake.@unstable tangent_type(::Type{<:CuArray{P,N} where {P}}) where {N} = CuArray
```
yebai:
```suggestion
```

ext/MooncakeCUDAExt.jl
```diff
@@ -33,19 +33,45 @@ import Mooncake.TestUtils:
     populate_address_map_internal, AddressMap, __increment_should_allocate
 
 const CuFloatArray = CuArray{<:IEEEFloat}
+const CuComplexArray = CuArray{<:Complex{<:IEEEFloat}}
 
 # Tell Mooncake.jl how to handle CuArrays.
 
+Mooncake.@foldable tangent_type(::Type{<:CuArray{P,N,M}}) where {P,N,M} = CuArray{
```
yebai: As discussed in https://github.com/chalk-lab/Mooncake.jl/pull/811/files#r2451846833, let's restrict to `Union{Complex{<:IEEEFloat}, IEEEFloat}` for now and relax it when tests for new types can be added. 

```suggestion
Mooncake.@foldable tangent_type(::Type{<:CuArray{P,N,M}}) where {P<:Union{Complex{<:IEEEFloat}, IEEEFloat},N,M} = CuArray{
```

ext/MooncakeCUDAExt.jl
```diff
@@ -3,7 +3,7 @@ module MooncakeCUDAExt
 using LinearAlgebra, Random, Mooncake
 
 using Base: IEEEFloat
-using CUDA: CuArray, cu
+using CUDA: CuPtr, CuArray, cu, DeviceMemory, CuStream
```
yebai:
```suggestion
using CUDA: CuArray, cu
```

Technici4n:
> there are automated tests: see test/ext/cuda/cuda.jl. It would be good to extend that test to Float64, ComplexF32 and ComplexF64 at the very least I think.

Don't forget 😉

kshyatt:
Don’t merge yet I’m still fixing texts

yebai:
@kshyatt, a gentle reminder on this.

kshyatt:
I'm back on this!

kshyatt:
OK, so nearly everything is working but for an FData related thing I'm unsure about

kshyatt:
Ok, I think this is now working after I fixed the `rrule` for `Complex` arrays!

ext/MooncakeCUDAExt.jl
```diff
@@ -3,13 +3,15 @@ module MooncakeCUDAExt
 using LinearAlgebra, Random, Mooncake
 
 using Base: IEEEFloat
-using CUDA: CuArray, cu
+using CUDA: CuArray, cu, CuDeviceArray, @cuda, threadIdx, blockIdx, blockDim, i32
```
Technici4n: Which parts of this are still used?

ext/MooncakeCUDAExt.jl
```diff
@@ -33,35 +35,107 @@ import Mooncake.TestUtils:
...
+const CuComplexArray = CuArray{<:Complex{<:IEEEFloat}}
 
 # Tell Mooncake.jl how to handle CuArrays.
 
-Mooncake.@foldable tangent_type(::Type{P}) where {P<:CuFloatArray} = P
+Mooncake.@foldable tangent_type(::Type{<:CuArray{P,N,M}}) where {P<:Union{Complex{<:IEEEFloat},IEEEFloat},N,M} = CuArray{
+    tangent_type(P),N,M
+}
+
+Mooncake.@foldable fdata_type(::Type{CuArray{P,N,M}}) where {T<:IEEEFloat,P<:Complex{T},N,M} = CuArray{
```
Technici4n: Are you sure that this one is needed?

ext/MooncakeCUDAExt.jl
```diff
@@ -33,35 +35,107 @@ import Mooncake.TestUtils:
...
 set_to_zero_internal!!(::Mooncake.SetToZeroCache, x::CuFloatArray) = x .= 0
+function set_to_zero_internal!!(
+    ::Mooncake.SetToZeroCache, x::CuArray{Mooncake.Tangent{@NamedTuple{re::T,im::T}},N,M}
+) where {T<:IEEEFloat,N,M}
+    x_ = reinterpret(Complex{T}, x)
+    x_ .= zero(Complex{T})
+    return x
+end
+
+function _add_to_primal_internal(
```
Technici4n, line 127: nitpick: move below other overload?

ext/MooncakeCUDAExt.jl
```diff
@@ -78,32 +152,69 @@ function _diff_internal(c::MaybeCache, x::P, y::P) where {P<:CuFloatArray}
...
+end
 function populate_address_map_internal(m::AddressMap, p::CuArray, t::CuArray)
     k = pointer_from_objref(p)
     v = pointer_from_objref(t)
     haskey(m, k) && (@assert m[k] == v)
     m[k] = v
     return m
 end
-function Mooncake.__verify_fdata_value(::IdDict{Any,Nothing}, p::CuArray, f::CuArray)
+function Mooncake._verify_fdata_value(::IdDict{Any,Nothing}, p::CuArray, f::CuArray)
```
Technici4n: why not the double _ version?

ext/MooncakeCUDAExt.jl
```diff
@@ -78,32 +152,69 @@ function _diff_internal(c::MaybeCache, x::P, y::P) where {P<:CuFloatArray}
...
+Mooncake.@foldable tangent_type(::Type{CuArray{P,N,M}}, ::Type{NoRData}) where {T<:IEEEFloat,P<:Mooncake.Tangent{@NamedTuple{re::T,im::T}},N,M} = CuArray{
+    P,N,M
+}
+tangent(p::CuFloatArray, ::NoRData) = p
+function tangent(
+    p::CuArray{P,N,M}, ::NoRData
+) where {T<:IEEEFloat,P<:Mooncake.Tangent{@NamedTuple{re::T,im::T}},N,M}
+    p
+end
+tangent(p::CuArray{P,N,M}, ::NoRData) where {T<:IEEEFloat,P<:Complex{T},N,M} = tangent.(p)
```
Technici4n: Is this one needed as well?

ext/MooncakeCUDAExt.jl
```diff
@@ -120,5 +231,12 @@ function rrule!!(
     _dims = map(primal, dims)
     return CoDual(P(undef, _dims), P(undef, _dims)), NoPullback(p, init, dims...)
 end
+function rrule!!(
+    p::CoDual{Type{P}}, init::CoDual{UndefInitializer}, dims::CoDual{Int}...
+) where {P<:CuComplexArray}
+    _dims = map(primal, dims)
+    return CoDual(P(undef, _dims), tangent_type(P)(undef, _dims)),
```
Technici4n: Maybe wrap in `return (...)`?

ext/MooncakeCUDAExt.jl
kshyatt: The double one was not getting called 🤷 Unclear to me why

Technici4n: Hmmm, given the definition of `_verify_fdata_value` that seems strange 😅
```jl
function _verify_fdata_value(c::IdDict{Any,Nothing}, p, f)::Nothing
    verify_fdata_type(_typeof(p), typeof(f))
    return __verify_fdata_value(c, p, f)
end
```

kshyatt: I agree!!! I can try removing it/going back to double underscores and see what happens. It's hard to figure out what's going on in the internals here.

kshyatt: Back to two underscores

Technici4n (approved):
LGTM!

PR #812, opened 2025-10-22
Try adding fwd/rvs rules for matrix inv and det

kshyatt:
First attempt at adding some internal rules. These are pretty simple and drawn from https://people.maths.ox.ac.uk/gilesm/files/NA-08-01.pdf

### DIFF
--- a/src/rrules/linear_algebra.jl
+++ b/src/rrules/linear_algebra.jl
@@ -1,6 +1,4 @@
 @is_primitive MinimalCtx Tuple{typeof(exp),Matrix{<:IEEEFloat}}
-@is_primitive MinimalCtx Tuple{typeof(inv),Matrix{<:IEEEFloat}}
-@is_primitive MinimalCtx Tuple{typeof(det),Matrix{<:IEEEFloat}}
 
 struct ExpPullback{P}
     pb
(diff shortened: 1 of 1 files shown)


Technici4n:
I wonder if rules should be added for the `lu` factorization and then `inv!(lu)` and `det(lu)`, compared to directly writing them against the "higher-level" wrappers. It's probably easier to write them against the higher-level wrappers for now. I just realized that Mooncake seems to have rules for `LAPACK.getrf!` (general triangular factorization i.e. LU) 👀

kshyatt:
It's possible, although the rule for `inv` is so simple I might keep that as is... up to the maintainers!

yebai (approved):
Many thanks @kshyatt! 

Can you also bump the version to make a minor release?

yebai:
> I wonder if rules should be added for the lu factorization and then inv!(lu) and det(lu), compared to directly writing them against the "higher-level" wrappers. It's probably easier to write them against the higher-level wrappers for now. I just realized that Mooncake seems to have rules for LAPACK.getrf! (general triangular factorization i.e. LU) 👀

As a rule of thumb, we prefer to write rules at the lowest possible level (e.g., foreign calls), since the number of rules we need to write and test is smaller.

src/rrules/linear_algebra.jl
```diff
@@ -12,6 +12,46 @@ function (pb::ExpPullback)(::NoRData)
     return NoRData(), NoRData()
 end
```
Technici4n, line 14: Coverage reports 0% diff. Indeed it looks like `@is_primitive` was not defined, so the rules don't run. Somehow Mooncake is still able to differentiate however?

kshyatt: Which is weird because when I tried using `inv` with a matrix earlier, it freaked out because of no foreign call rule! Sorry for missing the macro, I'll add it and retest

Technici4n:
I am quite confused:
- Apparently the linear algebra test did not do its job because it didn't catch the missing `@is_primitive` even though it is supposed to. Is something going wrong with the test utils?
- Apparently Mooncake already handles matrix `inv` and `det` automatically (otherwise how was the test passing?), likely via `LAPACK.getrf!` and `LAPACK.getri!`, and so I am not sure that a custom rule is even necessary?

src/rrules/linear_algebra.jl
Technici4n: > it freaked out because of no foreign call rule!

Oh hmmm. Do you still have the stack trace by any chance?

kshyatt:
Lmao well this is very confusing! I'm happy to delete my custom rules and add `inv` and `det` to the derived rules tests instead

yebai:
Thanks @kshyatt and @Technici4n!

src/rrules/linear_algebra.jl
```diff
@@ -35,4 +35,15 @@ function hand_written_rule_test_cases(rng_ctor, ::Val{:linear_algebra})
...
 
-derived_rule_test_cases(rng_ctor, ::Val{:linear_algebra}) = Any[], Any[]
+function derived_rule_test_cases(rng_ctor, ::Val{:linear_algebra})
+    rng = rng_ctor(123)
+    Ps = [Float64, Float32]
+    test_cases = vcat(
+        map_prod([3, 7], Ps) do (N, P)
+            return (false, :none, nothing, inv, randn(rng, P, N, N))
+            return (false, :none, nothing, det, randn(rng, P, N, N))
```
Technici4n, line 44: This never runs because of the `return` above

Technici4n: That also explains why the `hand_written_rule_test_cases` did not trigger any error... Nothing was running heh.

kshyatt: RIP lol

Technici4n: Sending a fix...

PR #817, opened 2025-10-24
[don't merge] Fix `sig_argcount_mismatch` typeassert error

sunxd3:

### DIFF
--- a/src/interpreter/forward_mode.jl
+++ b/src/interpreter/forward_mode.jl
@@ -100,11 +100,13 @@ end
     rest = Base.tail(remaining)
 
     # Unwrap Tuple containers that only hold Duals (length 1 or more)
-    if first_arg isa Tuple && !isa(first_arg, Dual) && length(first_arg) > 0 &&
+    if first_arg isa Tuple &&
+        !isa(first_arg, Dual) &&
+        length(first_arg) > 0 &&
         all(x -> x isa Dual, first_arg)
         return _flatten_tuple_duals_impl((first_arg..., rest...), acc)
-    # Expand Duals whose primal is a Tuple so downstream sees element-wise Duals
-    elseif first_arg isa Dual{P,T} where {P<:Tuple, T<:Tuple}
+        # Expand Duals whose primal is a Tuple so downstream sees element-wise Duals
+    elseif first_arg isa Dual{P,T} where {P<:Tuple,T<:Tuple}
         p_tuple = primal(first_arg)
         t_tuple = tangent(first_arg)
         expanded = ntuple(i -> Dual(p_tuple[i], t_tuple[i]), length(p_tuple))
@@ -478,7 +480,7 @@ function frule_type(
     # For varargs, the OpaqueClosure uses Vararg{Any} for the last parameter
     dual_arg_types = map(dual_type, arg_types)
     if isva
-        dual_args_type = Tuple{dual_arg_types[1:(end-1)]..., Vararg{Any}}
+        dual_args_type = Tuple{dual_arg_types[1:(end - 1)]...,Vararg{Any}}
     else
         dual_args_type = Tuple{dual_arg_types...}
     end
--- a/src/interpreter/reverse_mode.jl
+++ b/src/interpreter/reverse_mode.jl
@@ -1013,16 +1013,18 @@ end
     rest = Base.tail(remaining)
 
     # Case 1: Tuple containing only CoDuals - unwrap all elements
-    if first_arg isa Tuple && !isa(first_arg, CoDual) && length(first_arg) > 0 &&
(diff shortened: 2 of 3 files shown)


sunxd3:
To take stock.

The code in this branch aims to bring the vararg handling inline with the 1.2.1 new OC behavior. The core change is about passing `isva` flag and fix surrounding issues. (Essentially, this is what need to be fixed following https://github.com/JuliaLang/julia/pull/59772.) I couldn't quite make it work.

<details>

<summary>conceptual trace to demonstrate why missing `isva` is problematic</summary>

### The Test Case
```julia
sig_argcount_mismatch(x) = vcat(x[1], x[2:2], x[3:3], x[4:4])
# Called with: ones(4)
```

### Step-by-Step Execution Flow

1. User calls:
`sig_argcount_mismatch(ones(4))`

2. This expands to:
```julia
vcat(1.0, [1.0], [1.0], [1.0])
# 4 arguments of different types: Float64, Vector{Float64}, Vector{Float64}, Vector{Float64}
```

4. Julia lowers vcat to:
```julia
Base.__cat_offset!(A, shape, catdims, offsets, 1.0, [1.0], [1.0], [1.0])
# __cat_offset! is defined as:
# __cat_offset!(A, shape, catdims, offsets, x, X...) # varargs
```

6. Inside __cat_offset!, Julia uses Core._apply_iterate which Mooncake replaces with _apply_iterate_equivalent:
```julia
_apply_iterate_equivalent(iterate, some_function, args...)
# Where args... are the varargs being iterated
```

7.   When Mooncake differentiates `__cat_offset!`, it needs to create an `OpaqueClosure` for the AD rule:
`__cat_offset!` has argtypes:
```julia
[
  Tuple{},                    # captures/env (slot 1)
  typeof(__cat_offset!),      # slot 2
  Vector{Float64},            # A (slot 3)
  Tuple{Int64},               # shape (slot 4)
  Tuple{Bool},                # catdims (slot 5)
  Tuple{Int64},               # offsets (slot 6)
  Float64,                    # x (slot 7)
  Tuple{Vector{Float64}, ...} # X... - THE VARARGS (slot 8)
]
```

8. What Julia's OpaqueClosure constructor does when `isva=false`:
```julia
# From compute_oc_signature with isva=false:
sig = Tuple{
  Dual{typeof(__cat_offset!), NoTangent},
  Dual{Vector{Float64}, Vector{Float64}},
  Dual{Tuple{Int64}, NoTangent},
  Dual{Tuple{Bool}, NoTangent},
  Dual{Tuple{Int64}, NoTangent},
  Dual{Float64, Float64},
  Dual{Tuple{Vector{Float64}, ...}, Tuple{Vector{Float64}, ...}}  ← Last arg stays as Tuple!
}

What Julia's OpaqueClosure constructor does when isva=true:
# From compute_oc_signature with isva=true:
# It POPS the last arg and unpacks it:
lastarg = Dual{Tuple{Vector{Float64}, ...}, Tuple{Vector{Float64}, ...}}
# Then appends as Vararg{Any}:
sig = Tuple{
  Dual{typeof(__cat_offset!), NoTangent},
  Dual{Vector{Float64}, Vector{Float64}},
  Dual{Tuple{Int64}, NoTangent},
  Dual{Tuple{Bool}, NoTangent},
  Dual{Tuple{Int64}, NoTangent},
  Dual{Float64, Float64},
  Vararg{Any}  ← Transformed to Vararg{Any}!
}
```

9. When Mooncake tries to call the OpaqueClosure:
```julia
The generated AD code passes arguments as:
oc(dual_f, dual_A, dual_shape, dual_catdims, dual_offsets, dual_x, dual_X...)
# The last part: dual_X... gets splatted

But the OpaqueClosure was created with isva=false, so it expects:
oc(dual_f, dual_A, dual_shape, dual_catdims, dual_offsets, dual_x, dual_X_as_single_tuple)
# Expects the varargs as ONE argument
```

</details>

What I tried:
1. pass `isva` flag into `optimized_misty_closure`
2. handle `Tuple{CoDual}` and `CoDual{Tuple}` at the interface, which emerges from nested oc call and vararg grouping by some Mooncake functions (this is really superficial fix and probably very wrong)

yebai:
Cc @serenity4 in case anything is obvious to you

serenity4:
> Cc @serenity4 in case anything is obvious to you

Not at first glance, seems like this requires further debugging. A MWE would be helpful so we can identify whether this is a bug in Base or in Mooncake.

sunxd3:
https://github.com/chalk-lab/Mooncake.jl/blob/sunxd/draft_fix_sig_argcount_mismatch/test_sig_argcount_mismatch.jl would  trigger the error, not sure this is MWE, but a start at least.

PR #827, opened 2025-10-30
Support differentiable data in SimpleVector

Technici4n:
In 1.12, Julia uses `svec` when calling `_apply_iterate`. As such `SimpleVector` can now contain differentiable data. This PR fixes the `_svec_ref` rule accordingly, and adds an `svec` rule.

Initially PRed in https://github.com/chalk-lab/Mooncake.jl/pull/824, but since this change does not require Julia 1.12 it is being sent to `main` directly.


yebai:
@Technici4n, there doesn't seem to be any performance regression, so this is good to go.

sunxd3:
should we make a release?

Technici4n:
@sunxd3 I suppose we have to wait for 1.12 support for a new release (which is what motivated this change anyway).

Issue #826, opened 2025-10-29
Higher-order derivatives


Technici4n:
I am opening this issue to track support for higher-order derivatives (second order and more). In principle all 4 combinations (forward and reverse over forward and reverse) should be supported. Currently none seems to be.

Some preliminary questions would be:
- How to represent duals and coduals for higher-order differentiation? I tried to ask Mooncake what the tangent type of a `Dual` should be, but this seems suboptimal:
```jl
julia> Mooncake.Dual{Mooncake.Dual(1.0, 2.0),tangent_type(typeof(Mooncake.Dual(1.0, 2.0)))}
Mooncake.Dual{Mooncake.Dual{Float64, Float64}(1.0, 2.0), Tangent{@NamedTuple{primal::Float64, tangent::Float64}}}
```
In principle these nested duals should usually not be interacted with so an ugly representation might be OK.

- How to avoid perturbation confusion? Is this something that compiler-based AD systems need to be careful with as well?

yebai:
@sunxd3 is working on testing and fixing forward over reverse, which ought to be the lowest hanging fruit.

gdalle:
> How to represent duals and coduals for higher-order differentiation? I tried to ask Mooncake what the tangent type of a Dual should be, but this seems suboptimal:

This is probably fine

> How to avoid perturbation confusion? Is this something that compiler-based AD systems need to be careful with as well?

I think we might end up needing tags as in ForwardDiff. When you look at Mooncake forward mode, it's ForwardDiff with a mustache. But they have those in [JAX too](https://docs.jax.dev/en/latest/autodidax2_part1.html#jvp-interpreter)

yebai:
Fixed by https://github.com/chalk-lab/Mooncake.jl/pull/878

PR #833, opened 2025-11-02
Fix Hessian

sunxd3:
Ref: https://github.com/chalk-lab/Mooncake.jl/issues/826

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -98,6 +98,12 @@ end
 function Base.:+(x::Dual{P,T}, y::Dual{P,T}) where {P<:IEEEFloat,T<:IEEEFloat}
     return Dual(primal(x) + primal(y), tangent(x) + tangent(y))
 end
+function Base.:+(x::Dual{P,T}, y::Integer) where {P<:IEEEFloat,T<:IEEEFloat}
+    return Dual(primal(x) + y, tangent(x))
+end
+function Base.:+(x::Integer, y::Dual{P,T}) where {P<:IEEEFloat,T<:IEEEFloat}
+    return Dual(x + primal(y), tangent(y))
+end
 
 # Subtraction
 Base.:-(x::Dual{P,T}) where {P<:IEEEFloat,T<:IEEEFloat} = Dual(-primal(x), -tangent(x))
@@ -109,6 +115,12 @@ function Base.:-(x::P, y::Dual{P,T}) where {P<:IEEEFloat,T<:IEEEFloat}
 end
 function Base.:-(x::Dual{P,T}, y::Dual{P,T}) where {P<:IEEEFloat,T<:IEEEFloat}
     return Dual(primal(x) - primal(y), tangent(x) - tangent(y))
+end
+function Base.:-(x::Dual{P,T}, y::Integer) where {P<:IEEEFloat,T<:IEEEFloat}
+    return Dual(primal(x) - y, tangent(x))
+end
+function Base.:-(x::Integer, y::Dual{P,T}) where {P<:IEEEFloat,T<:IEEEFloat}
+    return Dual(x - primal(y), -tangent(y))
 end
 
 # Multiplication (product rule)
--- a/test/ext/differentiation_interface/differentiation_interface.jl
+++ b/test/ext/differentiation_interface/differentiation_interface.jl
@@ -42,4 +42,149 @@ test_differentiation(
         @test DI.hessian(x -> x[1]^4, backend, [2.0]) ≈ [48.0]
         @test DI.hessian(x -> x[1]^6, backend, [2.0]) ≈ [480.0]
     end
+
+    @testset "https://github.com/chalk-lab/Mooncake.jl/issues/632" begin
+        function gams_objective(x)
+            return (
+                (
+                    (
+                        (
+                            (
+                                (
+                                    (
+                                        (
+                                            (
+                                                (
+                                                    (
+                                                        (
+                                                            (
+                                                                (
+                                                                    (
+                                                                        (
+                                                                            (
+                                                                                (
+                                                                                    (
+                                                                                        (
+                                                                                            (
+                                                                                                (
+                                                                                                    (
+                                                                                                        (
+                                                                                                            (
+                                                                                                                (
+                                                                                                                    (
+                                                                                                                        x[1] *
+                                                                                                                        x[1] +
+                                                                                                                        x[10] *
+                                                                                                                        x[10]
+                                                                                                                    ) *
+                                                                                                                    (
+                                                                                                                        x[1] *
+                                                                                                                        x[1] +
+                                                                                                                        x[10] *
+                                                                                                                        x[10]
+                                                                                                                    ) -
+                                                                                                                    4 *
+                                                                                                                    x[1]
+                                                                                                                ) +
+                                                                                                                3
+                                                                                                            ) +
+                                                                                                            (
+                                                                                                                x[2] *
+                                                                                                                x[2] +
+                                                                                                                x[10] *
+                                                                                                                x[10]
+                                                                                                            ) *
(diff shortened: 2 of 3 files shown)


sunxd3:
The tests
https://github.com/chalk-lab/Mooncake.jl/blob/b1984f00d76bca2031c8b24fe8b3eb28ff0313cf/test/ext/differentiation_interface/differentiation_interface.jl#L15-L26
are passing, which is good. 

Next things to do:
 - [ ] add more robust tests (probably shouldn't be put in DI tests)
 - [ ] fix `basic` test error
 - [ ] better documentation of the issue and fix

sunxd3:
this is still very drafty, but I think it's somewhat ready for a quick look.

@Technici4n sorry for tagging, but maybe you are the best arbiter right now. if you do find some time to look over, please don't hold back

sunxd3:
A more complicated example would blow up the compilation.

I try to push a little bit, but I couldn't find the exact boundary of making a more complex function work and marking too many things non-differentiable

src/rrules/memory.jl
```diff
@@ -605,6 +605,16 @@ function rrule!!(
     return CoDual(x, dx), NoPullback((NoRData(), NoRData(), NoRData()))
 end
 
+# Core.memorynew is called inside rrule!! for Memory allocation. Forward-over-reverse
+# (Hessian) needs this frule when forward mode differentiates through reverse mode code.
+@inline function frule!!(
```
yebai: Let's also add the rrule implementation for completness.

test/ext/differentiation_interface/differentiation_interface.jl
```diff
@@ -3,10 +3,48 @@ Pkg.activate(@__DIR__)
...
     [AutoMooncake(; config=nothing), AutoMooncake(; config=Mooncake.Config())];
     excluded=SECOND_ORDER,
     logging=true,
 )
+
+# Test Hessian computation using forward-over-reverse with DITest scenarios.
+# Using linalg=false to select loop-based test functions instead of the default
+# versions that use broadcasting and linear algebra ops (vec, transpose).
+# Broadcasting in forward-over-reverse mode causes compilation hangs due to
+# complex nested types that overwhelm Julia's type inference.
```
yebai: Would be great to have a minimal example demonstrating this and understanding why, eg, Julia compiler normally gives up after a certain threshhold is reached.

src/rrules/memory.jl
sunxd3: turns out we probably don't need this

sunxd3:
Can't say this is ready, but it seems to pass DITest's Hessian test

yebai:
@gdalle might want to take a look given that this PR interacts deeply with DI. No pressure.

gdalle:
Thanks for the ping, I wasn't aware of this PR.
I don't have a lot of bandwidth at the moment but I don't think it is a good idea to have Mooncake depend on DI, given that DI already depends on Mooncake. This would make things rather confusing for users. @sunxd3 can you explain why the fixes necessarily involve DI functions? Why isn't second-order autodiff fixable within Mooncake itself?

gdalle:
Also, why would it be necessary to define arithmetic on Dual numbers? Isn't that what frules are for?

ext/MooncakeDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,147 @@
+module MooncakeDifferentiationInterfaceExt
```
yebai, line 1: Let's move this file to the DI repo.

ext/MooncakeDifferentiationInterfaceExt.jl
```diff
@@ -0,0 +1,147 @@
...
+
+using Mooncake:
+    Mooncake, @is_primitive, MinimalCtx, ForwardMode, Dual, primal, tangent, NoTangent
+import DifferentiationInterface as DI
+
+# Mark shuffled_gradient as forward-mode primitive to avoid expensive type inference hang.
+# This prevents build_frule from trying to derive rules for the complex gradient closure.
```
yebai, line 8: Can you explain clearly with a small example, why Mooncake/Julia struggle with `shuffled_gradient`.

src/dual.jl
```diff
@@ -55,3 +55,128 @@ verify_dual_type(x::Dual) = tangent_type(typeof(primal(x))) == typeof(tangent(x)
 function Dual(x::Type{P}, dx::NoTangent) where {P}
     return Dual{@isdefined(P) ? Type{P} : typeof(x),NoTangent}(x, dx)
 end
+
+# Dual of numeric types is self-tangent
```
yebai, line 59: Can you explain when `tangent_type(::Type{Dual})` will be called during forward-over-reverse? It is not self-evident why this is needed.

yebai (commented):
Thank you, @sunxd3 — I’ve added a few points of clarification below.

ext/MooncakeDifferentiationInterfaceExt.jl
gdalle: Until @sunxd3 explains why it is necessary, I'm not convinced this file should be anywhere.

sunxd3:
Thanks @gdalle for taking a look and the acute observations! In short, ~current code is more like reverse-over-forward (conceptually, but not implemented as such), but I am not sure if it's standard at all.~ current code use reverse mode to compute hvp that mathematically equivalent to reverse-over-forward, but computed using one pass of reverse mode only.

This is very much a Hail Mary—I had a ton of trouble trying to nail down the correct set of frules that would let forward mode differentiate through reverse mode without blowing up compilation. That wasn't productive, so I thought: maybe reverse-over-forward could work? I'm not convinced what's here is correct and the right approach, which is why I haven't written it up in detail yet.

The (old) title is misleading for the current implementation (I started by trying to work out forward-over-reverse, but that's not what the current code does). And the **DI extension is not necessary**—I was hacking the interface to pretend to do forward-over-reverse for testing purposes.

And I should say I am very very far from being fluent in Differentiable programming, so this is mixed with LLM's ideas.

---

The idea is making `Dual` self-tangent, then compute hvp using a single reverse-mode pass. **Dual arithmetic** is needed because the forward pass of reverse mode operates on Dual inputs. 

To see how it works, we pass `Dual.(x, v)` into reverse mode—the forward pass uses Dual arithmetic (this is why all the arithmetic operations on Duals are needed), and the backward pass uses rrules for Duals. The result is $\text{Dual}.(\nabla f(x), Hv)$—the tangent component gives the Hvp.

With $f(x) = x_1^2 + x_1 x_2$:

**Forward pass**:

$$
\begin{aligned}
t_1 &= \text{Dual}(x_1, v_1)^2 = \text{Dual}(x_1^2,\; 2x_1 v_1) \\
t_2 &= \text{Dual}(x_1, v_1) \cdot \text{Dual}(x_2, v_2) = \text{Dual}(x_1 x_2,\; x_1 v_2 + v_1 x_2)
\end{aligned}
$$

**Backward pass**

$$
\begin{aligned}
\bar{x}_1 &= 2 \cdot \text{Dual}(x_1, v_1) \cdot \text{Dual}(1, 0) + \text{Dual}(x_2, v_2) \cdot \text{Dual}(1, 0) \\
&= \text{Dual}(2x_1 + x_2,\; 2v_1 + v_2) \\
\bar{x}_2 &= \text{Dual}(x_1, v_1) \cdot \text{Dual}(1, 0) = \text{Dual}(x_1,\; v_1)
\end{aligned}
$$

So $Hv = \text{tangent}.([\bar{x}_1, \bar{x}_2])$

---

Let me know if this make sense and worth pursuing at all.

sunxd3:
This is not the way to go and commit/conversation histories are getting too confusing, I started https://github.com/chalk-lab/Mooncake.jl/pull/878 instead.

PR #839, opened 2025-11-06
Add rules for LAPACK.lacpy! and BLAS.trsv!

Technici4n:
This should fix a lot of failures on Julia 1.12 as it seems that these functions are now being used (e.g. `cholesky` now calls `lacpy!`).

I can target `main` instead if preferred.


Technici4n:
We need a rule for `BLAS.trsv!` as well... tomorrow hopefully 😉

Technici4n:
@yebai ready now, seems to fix a bunch of tests! Regarding the target, it's easier for me if this goes straight into the 1.12 porting branch.

yebai:
Thanks, Bruno -- it looks good to me at a high level. I've requested a more thorough review from @sunxd3.

Technici4n:
You don't trust the test cases? 😁

yebai:
It's probably okay to merge, but let's see whether @sunxd3 can offer his perspective.

PR #840, opened 2025-11-06
Try to use count_allocs instead of @allocations

Technici4n:
Maybe this works? Unfortunately the issue cannot be reproduced locally so I have to rely on github to test it...

### DIFF
--- a/src/test_utils.jl
+++ b/src/test_utils.jl
@@ -820,7 +820,7 @@ function test_rrule_performance(
 
         # Test allocations in primal.
         f(x...)
-        @test (@allocations f(x...)) == 0
+        @test count_allocs(f, x...) == 0
 
         # Test allocations in round-trip.
         f_f̄_fwds = to_fwds(f_f̄)


yebai:
The [rules](https://github.com/chalk-lab/Mooncake.jl/tree/main/src/rrules/performance_patches.jl) for `sum`, [rules](https://github.com/chalk-lab/Mooncake.jl/tree/main/src/rrules/random.jl) for `randn` and `randexp` appear to have a genuine allocation on Julia 1.12. However, it is not exactly whether it is from the `Base.sum` or Mooncake rules. 

See, also, https://github.com/chalk-lab/Mooncake.jl/pull/714#issuecomment-3499883712

yebai:
I don’t think this is needed anymore — I’ve added the changes directly in [#714](https://github.com/chalk-lab/Mooncake.jl/pull/714/files/0b97be5cc2b5aedeb3af9b02f67e554826c64b24..75299ef0e8b1e5fc947295e3b6f32468e13f0c0c)￼.

PR #844, opened 2025-11-10
Skip certain stability and performance tests for 1.12

yebai:
These errors are from Julia Base, not Mooncake, so skipping them.


Technici4n:
Are you sure that it's a problem of the stdlib itself and not of how we verify allocations? You are probably right though, it looks related to the scoped values. It might be worth opening an issue on Julia.

yebai:
I might not have time to produce an MWE; perhaps someone else can pick this up? 

Cc @serenity4

PR #845, opened 2025-11-10
Fix for TwicePrecision allocs

penelopeysm:
Tested locally and this should fix it (without breaking any of the other failures), but I'd let CI run to check that it didn't mess _something else_ up too.

I think it should also fix the array tests because those have the same underlying problem of `f` being a type, although I didn't check those myself.


yebai:
Thanks @penelopeysm for the quick fix!

PR #846, opened 2025-11-11
fix compat bound for main

penelopeysm:
Notice that 0.4.174 was never released: https://github.com/JuliaRegistries/General/pull/141069 https://github.com/chalk-lab/Mooncake.jl/commit/1c0a6d03ed07ae15640c55a1d096ffdcd420f814#commitcomment-168672266

The reason is because this came out at a time after 1.12 was released, but the compat bound was not adjusted accordingly. If I'm not wrong, Automerge always tests the lowest + highest Julia versions declared in the compat. Since Mooncake failed to build on 1.12, it never got merged into the General registry.

Merging this and re-triggering the 0.4.174 release should make it go through.

As an alternative solution you could merge #714 and just release the entire thing as 0.4.174.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -72,7 +72,7 @@ SpecialFunctions = "2"
 StableRNGs = "1"
 Static = "1.1.1"
 Test = "1"
-julia = "~1.10.8, 1.11.6"
+julia = "~1.10.8, ~1.11.6"
 
 [extras]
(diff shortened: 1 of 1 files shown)


penelopeysm:
plan is to release #714 as 0.4.174 instead

PR #850, opened 2025-11-13
Allow selective zeroing of cotangents

gdalle:
This is a very low-tech way for DI to bypass `set_to_zero!!` when it can prove that tangents don't need to be reset. As explained on [this page](https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/dev/dev/math/), the restrictive assumptions made by DI about what is done to each argument make it possible to avoid zeroing for all context cotangents.
@willtebbutt this is what we discussed at JuliaCon Paris, you're right that Mooncake can't make this kind of promises but DI can, and the present PR allows me to exploit them.

The tuple-based API is not necessarily the best one, but I think it's close.

Related:

- https://github.com/chalk-lab/Mooncake.jl/issues/552

### DIFF
--- a/src/tangents.jl
+++ b/src/tangents.jl
@@ -77,7 +77,7 @@ Gets the `i`th field of data in `t`.
 
 Has the same semantics that `getfield!` would have if the data in the `fields` field of `t`
 were actually fields of `t`. This is the moral equivalent of `getfield` for
-`MutableTangent`.
+[`MutableTangent`](@ref).
 """
 @unstable @inline get_tangent_field(t::PossiblyMutableTangent, i::Int) = val(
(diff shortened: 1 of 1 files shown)


gdalle:
@penelopeysm I'm curious whether this brings sizeable improvements to Turing too?
You can test it by using this branch of Mooncake and the following branch of DI: https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/916

penelopeysm:
Yes, a little bit faster (maybe 30%)! https://turinglang.org/ADTests/pr/

gdalle:
Superseded by #851

PR #851, opened 2025-11-14
 Allow selective zeroing of cotangents, take 2

gdalle:
This is a very low-tech way for DI to bypass `set_to_zero!!` when it can prove that tangents don't need to be reset. As explained on [this page](https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/dev/dev/math/), the restrictive assumptions made by DI about what is done to each argument make it possible to avoid zeroing for all context cotangents.
@willtebbutt this is what we discussed at JuliaCon Paris, you're right that Mooncake can't make this kind of promises but DI can, and the present PR allows me to exploit them.

The tuple-based API is not necessarily the best one, but I think it's close.

Related:

- https://github.com/chalk-lab/Mooncake.jl/issues/552

Supersedes #850 with signed commits

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -504,7 +504,7 @@ function value_and_pullback!!(
     ȳ,
     f::F,
     x::Vararg{Any,N};
-    args_to_zero::NTuple=ntuple(_ -> true, Val(length(x) + 1)),
+    args_to_zero::NTuple=ntuple(Returns(true), Val(N + 1)),
 ) where {F,N}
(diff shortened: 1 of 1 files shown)


gdalle:
I don't understand why the code coverage isn't 100%?

gdalle:
Another worry I have is on type inference for the kwarg as `Val(length(x)+1)`. It should work but I better test it locally to make sure

Technici4n:
Does `Val(N+1)` not work?

gdalle:
There we go, I think the codecov failure was due to my dirty squashing

PR #853, opened 2025-11-14
Start AMDGPU extension

kshyatt:
We should be able to base a lot of this off the existing CUDA work

### DIFF
--- a/ext/MooncakeAMDGPUExt.jl
+++ b/ext/MooncakeAMDGPUExt.jl
@@ -21,7 +21,6 @@ import Mooncake:
     tangent_to_primal_internal!!,
     primal_to_tangent_internal!!,
     _add_to_primal_internal,
-    _diff_internal,
     _dot_internal,
     _scale_internal,
     TestUtils,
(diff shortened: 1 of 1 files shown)


ext/MooncakeAMDGPUExt.jl
```diff
@@ -0,0 +1,239 @@
...
+    to_cr_tangent,
+    increment_and_get_rdata!,
+    MaybeCache,
+    IncCache,
+    NoRData
+
+import Mooncake.TestUtils:
+    populate_address_map_internal, AddressMap, __increment_should_allocate
+
+const ROCFloatArray = ROCArray{<:IEEEFloat}
```
yebai, line 39: Is it possible that we can do a union type to help share code between `CuArray` and `ROCArray`? 

```julia
GPUFloatArray = Union{ROCArray{<:IEEEFloat},CuArray{<:IEEEFloat}}
GPUComplexArray = Union{ROCArray{<:Complex{<:IEEEFloat}},CuArray{<:Complex{<:IEEEFloat}}}
```

kshyatt: For some methods probably. Tbh I wonder if we couldn't use `GPUArrays.jl` for this?

yebai: We want to support only GPUArray types that we can test, leaving only `CuArray`, `ROCArray` and maybe `MtlArray `.

yebai: I tried to add a rule for kron for abstract array types, but eventually gave up, as it makes robust rule testing impossible, in case it’s of interest.

https://github.com/chalk-lab/Mooncake.jl/pull/891

This doesn’t mean we can’t write rules for abstract types. However, the fact that their subtypes are not known when the rule is defined makes such rules difficult to test, and avoiding this has been one of Mooncake’s greatest strengths.

cc @Technici4n and @sunxd3

Technici4n: I think we should make sure to share the code between cuda and amd where possible. Worst case we limit the methods and testing to only those two.

yebai:
I'd suggest keeping it as an extension in the AMDGPU package -- it is outside Mooncake's team's expertise to review, as no one on the team is familiar with AMDGPU.

Issue #403, opened 2024-12-01
Stateful Hand-Written Rules


willtebbutt:
At present, if an `rrule!!` needs to allocate storage for some value it overwrites on the forwards pass, it must do so each time the rule is called. This is distinct from derived rules, which have state which is preserved between calls.

There's no particular reason for things to be this way. This kind of functionality would be helpful, because it would make it possible to trade off some memory in exchange for runtime performance. Take, for example, `BLAS.gemm!(transA, transB, alpha, A, B, beta, C)` -- this rule has to allocate memory to hold whatever value `C` has upon entry, in order to restore it on the reverse-pass. Currently, it must allocate this memory each time the rule is called, but if we permit rules to be stateful we can just use the same heuristic that we use for the derived rules, and avoid de-allocating this memory -- subsequent calls to a rule would be fast.

This change could probably be made non-breaking (`rrule!!`s would remain unchanged). We would just add a function `build_primitive_rrule` or something, which returns `rrule!!` by default, but which one can over-ride to return e.g. a custom callable struct instead.

I first realised that this is probably going to be required while addressing https://github.com/compintell/Mooncake.jl/discussions/394 . The primal is type-stable and non-allocating, but Mooncake has a great deal of allocations. Now that a couple of performance bugs have been fixed (the generated IR was not type stable due to these bugs), the removal of the remaining allocations will require solving this issue.

RoyCCWang:
While I was going through [The Rule Interface Round 2 of the current documentation](https://compintell.github.io/Mooncake.jl/stable/understanding_mooncake/rule_system/#The-Rule-Interface-(Round-2)), I  wrote the following code that should be equivalent to the coded `rrule!!` example in that section of the documentation; i.e., `rrule_mod!!` below should behave the same as the `rrule!` from the example.

```
using LinearAlgebra
import Mooncake as MN

T = Float64

# This is from the example, the primal function.
function eval_model(p::Tuple{T, Vector{T}}) where T <: AbstractFloat
    a, b = p[1], p[2]
    return a + sum(b)
end

# This is from the example.
function rrule!!(
    ::MN.CoDual{typeof(eval_model)},
    x::MN.CoDual{
        Tuple{T, Vector{T}},
    },
    ) where T <: AbstractFloat

    dx_fdata = x.dx
    function df_adjoint(dy::T)

        dx_fdata[2] .+= dy
        dx_1_rdata = dy
[...]

PR #860, opened 2025-11-19
Conversions between primals and tangents

### LINKED ISSUE #393: More Friendly Types at the Interface Level
This issue https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/642 makes me wonder whether we need a systematic approach to translating between primal types and tangent types at the interface level.

For example, while users probably want to represent the tangent of a, `SArray` with another `SArray`, rather than a `Tangent`, Mooncake requires that users provide a `Tangent`.

I think we can probably define sensible translation functionality between primals and tangents which makes some choices around how to handle non-differentiable fields, but which works quite generically. This function would be something like
```julia
translate_to_tangent(t::IEEEFloat) = t
translate_to_tangent(t::VariousIntegerTypes) = NoTangent()
translate_to_tangent(t::Array{<:IEEEFloat}) = t
translate_to_tangent(t::Array) = map(translate_to_tangent, t)
function translate_to_tangent(t::P) where {P}
    isprimitivetype(P) && throw(error("need a translation rule"))
    return # recursively transform into tangent_type(P)
end
```
This would have the effect of, for example, dropping any non-differentiable fields.

On the way back, we could do a similar thing, but would need to pick a placeholder value for any non-differentiable fields. Not all types have a well-defined zero value (e.g. `String`s and `Symbol`s), so it might just make sense to make the conversion the other way require that you pass in the primal value, and we just copy its fields. For example
```julia
translate_to_primal(::P, t::P) where {P<:IEEEFloat} = t
translate_to_primal(p::P, t) where {P<:VariousIntegerTypes} = p
translate_to_primal(p::Array{P}, t::Array{P}) where {P<:IEEEFloat} = p
translate_to_primal(p::Array, t::Array) = map(translate_to_primal, p, t)
function translate_to_primal(p::P, t) where {P}
    # same idea as translate_to_tangent, but in the other direction
end
```
We could then e.g. shove this in `value_and_pullback` or whatever, so that users get "nice" types. We would want to ensure that there is a sensible way to opt-out of this translation, of course.

Technici4n:
This PR introduces functions `primal_to_tangent!!` and `tangent_to_primal!!` to convert between primal types and their corresponding tangent types. This allows tangents to be expressed using the more "friendly" primal types. The interface functions for reverse-mode AD accept a `friendly_tangents` boolean kwarg (defaulting to `false`) to automatically perform the conversion. Fixes #393.

The `_diff` is removed since it is superseded by `primal_to_tangent!!` and `tangent_to_primal!!`. In the future we could remove `add_to_primal` as well.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.193"
+version = "0.4.194"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"


gdalle:
Can you add a flag to [`Mooncake.Config`](https://github.com/chalk-lab/Mooncake.jl/blob/main/src/config.jl)? We don't want to enable it by default but it would be nice to experiment inside DI. Something like:

```julia
@kwdef struct Config
    debug_mode::Bool = false
    silence_debug_messages::Bool = false
    translate_tangents::Bool = false
end
```
Although I guess there's a question of who does the conversion: Mooncake or DI?

Technici4n:
A good reason to let Mooncake optionally perform the conversion as part of `value_and_pullback!!`, would be that it can cache the storage for the converted tangent.

gdalle:
Good point. Another reason is that only Mooncake can offer guarantees as to the type of tangents it returns, so the best DI can do is follow the lead without making additional promises

src/tangents.jl
```diff
@@ -1334,6 +1334,61 @@ for T in [Symbol, Int, Val]
     @eval increment_field!!(::NoTangent, ::NoTangent, f::Union{$T}) = NoTangent()
 end
 
+# TODO: x instead of primal?
+# TODO: a primal shouldn't be modified in-place? e.g. a vector should not be overwritten?
```
gdalle: Agreed, it should be `translate_to_primal!!(tangent, primal)` where the `tangent` can get modified (because our assumption is that we won't need it i `Tangent` form anymore) but the `primal` cannot.

Technici4n:
Alright, updated with a reasonable implementation. What is also nice is that the testing is completely automated and should be quite robust. Now we still need to figure out how exactly this should be wired up for users.

@gdalle in the end it was more convenient to mutate the primal and not the tangent. Part of the reason is that there is already functionality to copy primal types inside of Mooncake (although it's probably a little broken, see https://github.com/chalk-lab/Mooncake.jl/issues/865).

Other things to do:
- What is the best name for this new function? `translate_to_primal!!` is a bit long.
- Implementing the method for the other tangent types.
- Fixing the implementation of `_diff_internal`, which is currently broken for mutable types that have an early field unset and a later field set 😄

gdalle:
> Now we still need to figure out how exactly this should be wired up for users.

I would put a toggle inside the `config` object and add a toggle-dependent conversion step at the end of every user-facing function (`value_and_pullback!!`, `value_and_gradient!!`, `value_and_derivative!`). What do you think? 

> What is the best name for this new function? `translate_to_primal!!` is a bit long.

I like it, and I don't think we can do better with only two words.

yebai:
Minor comment: `tangent_to_primal!!` is better than `translate_to_primal!!`, making it clear what we are translating from.

Technici4n:
OK, I added `primal_to_tangent!!` now, as well as a flag to `Mooncake.Config`. I think that it would actually be preferable to convert at the DI level given that it already has related infrastructure (`prep.dy_righttype`), if @gdalle you would be willing to give it a try while I continue working on the Mooncake side.

gdalle:
Thanks! But I'm a bit skeptical about converting at the DI level if the flag is in Mooncake.Config, because that means the flag does nothing in the package per se, and its semantics are defined externally. If we want conversion to be in DI, I'd put the flag in ADTypes?

Technici4n:
> If we want conversion to be in DI, I'd put the flag in ADTypes?

Probably yes. The config is a bit weird anyway because none of it is handled at the Mooncake level anyway. :P

gdalle:
So the only reason why it should be done in DI is because we have the memory preallocated for the conversion?

Technici4n:
If you put it that way... 😓 The answer is that I don't know yet. Need to fix all the custom tangent types first.

yebai:
@Technici4n Looking at this more closely, these operations are similar to `arrayify` and related helpers, which exist to simplify writing rules for array types. Since Julia has many array types, unifying them within rules significantly reduces implementation complexity.

In that spirit, the `tangent_to_primal!!` functions could also be helpful as rule-writing helpers.

*PS:* An alternative naming for `tangent_to_primal!!` / `primal_to_tangent!!` is `primalify!!` / `tangentify!!`, following the same spirit as `arrayify` / `numberify` / `viewify`: 

https://github.com/chalk-lab/Mooncake.jl/blob/bd1e5fe42778140dec3ce64277684ae1be10d788/src/rrules/blas.jl#L24-L100

Technici4n:
They look similar on the surface but are different in the details: `arrayify` will typically just return a view, e.g. by reinterpreting the `Tangent`s as complex numbers, so the "arrayified" tangent is still not of the same type as the primal. Whereas `tangent_to_primal!!` will convert to the exact primal type that was used. So I think they serve different purposes.

yebai:
The differences are relatively minor: `arrayify` was initially intended for writing BLAS rules and uses views to support in-place tangent accumulation. 

I think `primalify!!` / `tangentify!!` can be more general and applicable to a broader range of use cases. One example is they may be helpful in writing rules involving abstract types too, ie, unifying different input types within rules to reduces implementation complexity.

Technici4n:
@yebai I agree that converting between primal and tangent would be helpful when writing rules, but unfortunately I don't see how we could make both fit under the same interface. Do you have some ideas?

To expand over the problem, consider for example a primal `x::Vector{ComplexF64}`. Its tangent is a `t::Vector{Tangent{@NamedTuple{re::Float64, im::Float64}}}`.
- `tangent_to_primal!!(x, t)` will _copy_ the tangent into a pre-allocated vector of the primal type.
- `arrayify(x, t)` will produce a _view_ of type `Base.ReinterpretArray{ComplexF64, 1, Tangent{@NamedTuple{re::Float64, im::Float64}}, Vector{Tangent{@NamedTuple{re::Float64, im::Float64}}}, false}`.

I think that primal<->tangent conversions should really be allocation-free and give back the exact primal type, since this is what we will want users of Mooncake to receive by default. It would be good to reuse the same machinery to make writing rules easier, but I don't see how that can be done.

yebai:
I’m only pointing out the connection between `arrayify` and `tangent_to_primal` here. Further exploration of whether this can help with writing rules for generic types can be continued in separate PRs.

It’s worth noting that there are already mechanisms to preallocate memory in manual rules (see #403).

Technici4n:
Ah that is interesting, I didn't know about `build_primitive_rrule`. #403 also offers a nice path to call back into AD (e.g. for implicit function rules). For "lazy" users pre-allocating an array and then converting it to a primal type would work 😄

However for a BLAS rule involving complex matrices it would still be preferable to reinterpret the input array I think. Copying into another array, even if preallocated, is a potential waste of cycles since at the end of the day we will be making another BLAS call that will pass a pointer to the start of the data anyway.

PS: Looks like I am quite close to done with this PR, I just need to fix DispatchDoctor, CUDA and the docs :)

Technici4n:
@gdalle is this PR now suitable for your needs in DI? For reverse-mode you should be able to just pass `friendly_tangents` from the `Config`; for forward-mode you'd have to call the conversion methods yourself since Mooncake doesn't have a cache for forward mode yet.

src/tangents/tangents.jl
```diff
@@ -1349,6 +1248,241 @@ for T in [Symbol, Int, Val]
...
+    # There doesn't seem to be a nice way to modify a SimpleVector in-place,
+    # so we just create a new one.
+    x′ = svec(map(x, tx) do xn, txn
+        tangent_to_primal_internal!!(xn, txn, c)
+    end...)
+    c[x] = x′
+    return x′
+end
+function primal_to_tangent_internal!!(tx, x::SimpleVector, c::MaybeCache)
+    haskey(c, x) && c[x]::Vector{Any}
```
sunxd3:
```suggestion
    haskey(c, x) && return c[x]::Vector{Any}
```

docs/src/developer_documentation/custom_tangent_type.md
```diff
@@ -448,23 +449,21 @@ function Mooncake._add_to_primal_internal(c::Mooncake.MaybeCache, p::A{T}, t::Ta
...
+    haskey(c, p) && return c[p]::A{T}
+    c[p] = p
+    p.x = Mooncake.tangent_to_primal_internal!!(p.x, t.x, c)
+    p.a = Mooncake.tangent_to_primal_internal!!(p.a, t.a, c)
+    return p
+end
+
+function Mooncake.primal_to_tangent_internal!!(t, p::A{T}, c::Mooncake.MaybeCache) where {T}
+    t isa Mooncake.NoTangent && return Mooncake.NoTangent()
+    haskey(c, p) && return c[p]::TangentForA{Tx}
```
sunxd3: `Tx` is undefined here

sunxd3 (approved):
Looks great in general. One bug, I think.

Maybe some extra doc [here](https://github.com/chalk-lab/Mooncake.jl/blob/tech/friendly-types/docs/src/tutorial.md#differentiationinterfacejl-api)?

Also, do we still need these: https://github.com/chalk-lab/Mooncake.jl/blob/33192859009e3d0ae2160ddef42dc2bc24e6af4b/ext/MooncakeCUDAExt.jl#L178-L193 ?

yebai:
> for forward-mode you'd have to call the conversion methods yourself since Mooncake doesn't have a cache for forward mode yet.

Does it make sense to define a caching mechanism for forward differentiation so that the Mooncake interface is consistent? If this adds too much additional work, please feel free to open an issue and address it in a separate PR.

Technici4n:
Thanks a lot for the review @sunxd3, very nice catches! Will go through them soon.

> Does it make sense to define a caching mechanism for forward differentiation so that the Mooncake interface is consistent?

Absolutely, yes. I will open a follow-up issue / PR once this one is done, as it can be done after this.

Technici4n:
I was also thinking that it might be nice to make `friendly_tangents=true` by default, maybe in Mooncake 0.5?

sunxd3:
@Technici4n Thanks a lot for working on this!

gdalle:
@Technici4n I'm still on (sorely needed) vacation, I'll try the integration when I get back. But you can also have a crack at it yourself if you like, the DI bindings are fairly simple

Technici4n:
> @Technici4n I'm still on (sorely needed) vacation, I'll try the integration when I get back. But you can also have a crack at it yourself if you like, the DI bindings are fairly simple

No worries, enjoy! I really don't want to take away from your vacation. :)

Technici4n:
@sunxd3 All addressed! I wanted to properly test `custom_tangent_types.md` with a cycle to hit the `Tx` that did not exist, and I had to fix a bunch of edge cases in the cache handling of various methods, so if you could a look at the changes that would be great. :)

docs/src/developer_documentation/custom_tangent_type.md
```diff
@@ -385,15 +392,21 @@ end
...
 function Mooncake.zero_tangent_internal(p::A{T}, dict::Mooncake.MaybeCache) where {T}
     Tx = Mooncake.tangent_type(T)
     Tx == Mooncake.NoTangent && return Mooncake.NoTangent()
     if haskey(dict, p)
         return dict[p]::TangentForA{Tx}
     end
-    x_t = Mooncake.zero_tangent_internal(p.x, dict)::Tx
-    t = TangentForA{Tx}(x_t)
+    t = TangentForA{Tx}()
     dict[p] = t
```
Technici4n, line 402: Note: we always have to put a new "empty" tangent in the cache before recursing to prevent infinite recursion. This was not done correctly before.

docs/src/developer_documentation/custom_tangent_type.md
```diff
@@ -737,7 +739,14 @@ end
...
+We run the check for both a non-cyclic case to check our method implementations,
+as well as a cyclic case to make sure that our interactions with the caches are correct.
 
 ```@example custom_tangent_type
+# Non-cyclic A
 Mooncake.TestUtils.test_data(Random.default_rng(), A(1.0, A(2.0, A(3.0))))
+# Cyclic A
+cyclic_a = A(1.0, A(2.0))
+cyclic_a.a.a = cyclic_a
+Mooncake.TestUtils.test_data(Random.default_rng(), cyclic_a)
```
sunxd3, line 759: Do you think whether it's better to split this into a new code block? Currently only the result for line 751 will show.

docs/src/developer_documentation/custom_tangent_type.md
```diff
@@ -385,15 +392,21 @@ end
 Implement the essential tangent manipulation functions:
 
 ```@example custom_tangent_type
+# Once we add this method, the TangentForA constructor that takes a NamedTuple can be removed
+Base.delete_method(
+    # (we need a concrete type to find the method hence Float64)
+    only(methods(TangentForA{Float64}, Tuple{@NamedTuple{x::Tx, a::Union{Mooncake.NoTangent, TangentForA{Tx}}} where Tx}))
+)
+
```
sunxd3: Is this necessary for the test logic?

sunxd3 (commented):
Thanks Bruno. Tiny things from my end, very solid improvements

docs/src/developer_documentation/custom_tangent_type.md
sunxd3: (I am slightly worried if this is TMI for a user facing tutorial. "someone" might think this is necessary and copy without thinking? not certain)

or maybe one of these https://documenter.juliadocs.org/stable/showcase/#Admonitions ?

Technici4n: It's not necessary, it's just that the method is unnecessary once you overload `zero_tangent_internal`. At the same time I wouldn't want to move `zero_tangent_internal` too high on that docs page. 😅

My assumption was that someone familiar with Julia would not copy `delete_method` like that. It's also a very technical docs page anyway.

sunxd3: 👍 make sense, I think we can keep it then. may I be nitpicky and say let's add something along the line of "this is not necessary for correctness ..."?

Technici4n: I tweaked the wording a bit, what do you think?

docs/src/developer_documentation/custom_tangent_type.md
```diff
@@ -415,6 +409,15 @@ function Mooncake.zero_tangent_internal(p::A{T}, dict::Mooncake.MaybeCache) wher
     return t
 end
 
+# Once we add a zero_tangent_internal method,
+# the TangentForA constructor that takes a NamedTuple can be removed.
+# NOTE: We use delete_method on this docs page because we defined the method above.
+# NOTE: In your package, you should instead remove the original method definition.
```
yebai:
```suggestion
# Once `zero_tangent_internal` is added, the NamedTuple-based `TangentForA`
# constructor is obsolete; we use `Base.delete_method` here to undo
# a method defined earlier in this documentation.
# 
# NOTE: In a real package, do not call `delete_method`; remove the method
# definition from the source instead.
```

docs/src/developer_documentation/custom_tangent_type.md
sunxd3: I like it a lot, thanks

Issue #818, opened 2025-10-25
Bitcast Error on Modulo Operation


jmurphy6895:
A recent change to Mooncake seems to have broken differentiation through a function containing modulo. Here is an MWE

```
using Mooncake
using DifferentiationInterface 
f_mod(x) = mod(x, 360.0)
result, deriv = value_and_derivative(f_mod, AutoMooncake(;config=nothing), 450.0)
```

Interestingly, this doesn't occur in rem2pi (although the error does seem to be popping up for deeper nested functions with rem2pi as well)

```
f_mod(x) = rem2pi(x, RoundDown)
result, deriv = value_and_derivative(f_mod, AutoMooncake(;config=nothing), 7.5)
```

These functions were working in the past and the library I'm using them in hasn't changed so I'm not sure what the cause is

yebai:
For posterity, here is the error message. 

> ERROR: ArgumentError: It is not permissible to bitcast to a differentiable type during AD, as this risks dropping tangents, and therefore risks silently giving the wrong answer. If this call to bitcast appears as part of the implementation of a differentiable function, you should write a rule for this function, or modify its implementation to avoid the bitcast.

bitcast is considered unsafe for AD, so it is forbidden in Mooncake (see, eg, [here](https://chalk-lab.github.io/Mooncake.jl/stable/utilities/debugging_and_mwes/#Example)). What may have happened here is that Julia Base updated its implementation and introduced a bitcast in some underlying code (ie, `mod → Base.rem_internal → Mooncake.IntrinsicsWrappers.bitcast`). If that is the case, then a corresponding Mooncake rule will be required.

PR #863, opened 2025-11-21
feat: import rules for modulo from chainrules.jl

jmurphy6895:
A small PR so that the modulo operation rule gets pulled from ChainRules.jl and doesn't throw a bitcast error. Resolves https://github.com/chalk-lab/Mooncake.jl/issues/818


Technici4n:
Looks good. Can you bump the version in `Project.toml` as well?

Technici4n (approved):

PR #866, opened 2025-11-21
Document limitations of differentiating CUDA kernels

### LINKED ISSUE #861: Document that support for custom kernels is not planned
It came up on slack recently that support for custom cuda kernels is explicitly a non-goal of Mooncake. 

I think it's good to have non-goals and I respect this design decision. Currently, however, the only "public" evidence of this non-goal seems to be the fact that related issues such as #648  are closed as not planned.

It would be good to document this more visibly (perhaps in the readme?). Such a disclaimer may make it easier for folks to choose between the many AD libraries in Julia.

lassepe:
Add section on CUDA kernel differentiation limitations. Resolves #861

### DIFF
--- a/docs/src/known_limitations.md
+++ b/docs/src/known_limitations.md
@@ -70,9 +70,9 @@ If you encounter a practical situation in which it is very important that this e
 
 ## Differentiating CUDA Kernels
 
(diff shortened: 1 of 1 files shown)


yebai:
thanks @lassepe!

PR #874, opened 2025-12-02
fix DynamicPPL integration tests

### LINKED ISSUE #873: DynamicPPL test errors
The new DynamicPPL release seem to break integration test. See, https://github.com/chalk-lab/Mooncake.jl/actions/runs/19859589835/job/56905976820

penelopeysm:
Closes #873 (probably... but let CI check!)
 
- changes SimpleVarInfo to VarInfo (in DPPL 0.39 onwards, LDF does not store a varinfo so the type of varinfo used to construct it is meaningless)
- remove call to `rand_prior_true`; instantiating a varinfo from the model will sample from the prior anyway so don't think there's much purpose in the specialised call
- enable the LDA model in the tests (which apparently used to not work with SimpleVarInfo)

### DIFF
--- a/test/integration_testing/dynamicppl/dynamicppl.jl
+++ b/test/integration_testing/dynamicppl/dynamicppl.jl
@@ -97,7 +97,7 @@ end
             (false, "broadcast_demo", broadcast_demo(rand(LogNormal(1.5, 0.5), 1_000))),
             (false, "large model", make_large_model()),
             (
-                false,
+                true,
                 "CollapsedLDA",
                 LatentDirichletAllocationVectorizedCollapsedManual(
                     data.D, data.K, data.V, data.α, data.η, data.w, data.doc


penelopeysm:
so, on the bright side there aren't any interface errors, i.e. i think my job is done as far as DPPL is concerned

but the newly enabled LDA model has a correctness error, which is a Mooncake issue

penelopeysm:
@sunxd3 ^ should we disable the test first (to make ci pass), merge this and then worry about the LDA model separately?

yebai:
@penelopeysm, FiniteDiff likely will struggle with the simplex distribution so the Mooncake result might be correct here. If so, feel free to pass `interface_only` to test_rule for this specific test.

penelopeysm:
Ok! I suspected that it was not the best for finite differences too, so that's reassuring to hear it from someone else too.

test/integration_testing/dynamicppl/dynamicppl.jl
```diff
@@ -80,42 +80,36 @@ end
...
+    @testset "$(typeof(model))" for (interface_only, name, model) in vcat(
         Any[
-            (false, "simple_model", simple_model(), nothing),
-            (false, "demo", demo(), nothing),
+            (false, "simple_model", simple_model()),
+            (false, "demo", demo()),
+            (false, "broadcast_demo", broadcast_demo(rand(LogNormal(1.5, 0.5), 1_000))),
+            (false, "large model", make_large_model()),
             (
                 false,
```
yebai:
```suggestion
                true,
```

yebai:
thanks, @penelopeysm!

Issue #870, opened 2025-12-01
Time to first gradient


kbarros:
First of all, awesome project!

I am wondering about "time to first gradient". On my machine, it's currently about 33 seconds for a simple function (Mooncake v0.4.180). Is this something that might improve over time? 

```julia
import DifferentiationInterface as DI
import Mooncake
using LinearAlgebra

backend_mooncake = DI.AutoMooncake()

x = randn(1_000)
f(x) = sum(x -> x^2, x)
@time DI.gradient(f, backend_mooncake, x) # ~ 33s

# Redefine f interactively
f(x) = sum(x -> x^3, x)
@time DI.gradient(f, backend_mooncake, x) # ~ 6.8s
```

This might be improved with #868.

yebai:
Thanks @kbarros -- this is likely beyond the expertise of the core Mooncake team, which is mainly machine learning researchers. We welcome community efforts to improve this. 

Related PR: https://github.com/chalk-lab/Mooncake.jl/pull/593

PR #877, opened 2025-12-05
Improve compile times -- Don't use `_methods_by_ftype` for primitive detection

### LINKED ISSUE #868: Bottleneck in `_methods_by_ftype`
The following MWE popped up on Slack (credit goes to Christopher Fisher):

```julia
using Mooncake
using SequentialSamplingModels
using Turing
using StatProfilerHTML

# Generate some data with known parameters
dist = LBA(ν=[3.0, 2.0], A = .8, k = .2, τ = .3)
data = rand(dist, 100)

# Specify LBA model
@model function model(data; min_rt = minimum(data.rt))
    # Priors
    ν ~ MvNormal(zeros(2), I * 2)
    A ~ truncated(Normal(.8, .4), 0.0, Inf)
    k ~ truncated(Normal(.2, .2), 0.0, Inf)
    τ ~ Uniform(0.0, min_rt)

    # Likelihood
    data ~ LBA(; ν, A, k, τ )
end

# Estimate parameters
chain = sample(model(data), NUTS(1000, .85; adtype = AutoMooncake()), MCMCThreads(), 1000, 4)
```

with the following profiling results

[statprof_lba_single_thread.zip](https://github.com/user-attachments/files/23815730/statprof_lba_single_thread.zip)

Image

MasonProtter:
Closes #868. This also goes part-way towards addressing #870.

Benchmark:
```julia
import DifferentiationInterface as DI
import Mooncake
using LinearAlgebra

# Some simple scalar-valued function
f(x) = sum(x -> x^2, x)

# Problem size
N = 1_000
x = randn(N)

backend_mooncake = DI.AutoMooncake()
@time DI.prepare_gradient(f, backend_mooncake, x);

# Redefine f interactively
f(x) = sum(x -> x^3, x)
@time DI.gradient(f, backend_mooncake, x);

f(x) = sum(x -> x^4, x)
@time DI.gradient(f, backend_mooncake, x);
```
Before this PR:
```julia
 38.599238 seconds (89.85 M allocations: 4.272 GiB, 1.13% gc time, 94.45% compilation time: <1% of which was recompilation)
  5.638958 seconds (5.39 M allocations: 265.009 MiB, 0.34% gc time, 63.89% compilation time: <1% of which was recompilation)
 22.847218 seconds (45.38 M allocations: 2.158 GiB, 0.78% gc time, 72.58% compilation time: <1% of which was recompilation)
```
After this PR:
```julia
27.035642 seconds (72.93 M allocations: 3.465 GiB, 1.36% gc time, 99.16% compilation time: <1% of which was recompilation)
 0.793844 seconds (2.18 M allocations: 104.561 MiB, 86.42% compilation time: <1% of which was recompilation)
 4.708629 seconds (11.59 M allocations: 559.929 MiB, 0.83% gc time, 92.41% compilation time: <1% of which was recompilation)
```

For this benchmark, that means I see a ~30% reduction in the time-to-first-gradient,  and a ~500-700% reduction in the time it takes to compile similar gradients that touch a lot of the same functions as the previous gradient. 

This *should* work without regressing on e.g. https://github.com/chalk-lab/Mooncake.jl/issues/558
__________

This was achieved via two coupled changes:

1. Instead of introducing a `maybe_is_primitive` that tries to indirectly check if there's any possible methods which might match `_is_primitive`, I take advantage of `Core.Compiler.find_method_matches` to just give me a list of all possible matching methods at a call-site, and then I can iterate over those matches and check if any is a primitive using regular `is_primitive`
2. `is_primitive` relied on a very slow method table lookup via `Base._methods_by_ftype`, and tried to compensate for this slowness by caching the result based on the world in a global dictionary. The problem with this is that the cache is extremely pessimistic. It stores a single world age, rather than e.g. a range of applicable worlds, so it means the cache needs to be totally recomputed for a ton of functions every time the worldage advances, even if no methods are actually invalidated. Instead, I threw out the caching and simply `invoke` the `_is_primitive` function in the requested world, which is more straightforward and faster.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.182"
+version = "0.4.183"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"


Technici4n:
This looks great, thanks a lot! How does this deal with the union types that can appear in the IR?

MasonProtter:
That's what `find_matching_methods` does. The `applicable` list gives the list of possible matching method specializations that can occur when there's a finite union in the IR.

I use that function to iterate over the possible methods, and if any of them are primitives, then i stop the callsite from inlining

MasonProtter:
Also, I should note that here's the benchmark speed I see on the last commit before https://github.com/chalk-lab/Mooncake.jl/pull/759 (i.e. benchmarking on commit `832b3cb`):
```julia
 28.417532 seconds (72.73 M allocations: 3.456 GiB, 1.14% gc time, 99.24% compilation time: <1% of which was recompilation)
  0.683032 seconds (2.14 M allocations: 102.898 MiB, 85.13% compilation time: <1% of which was recompilation)
  3.862572 seconds (11.24 M allocations: 543.919 MiB, 1.77% gc time, 91.75% compilation time: <1% of which was recompilation)
```
so this PR recovers most of the (compile-time) performance lost in #759, but without compromising on accuracy with union callsites.

src/interpreter/contexts.jl
```diff
@@ -157,115 +152,37 @@ function is_primitive(
...
-    # return the answer from the cache.
-
-    tt = Tuple{typeof(_is_primitive),Type{ctx},Type{mode},Type{sig}}
-    return get!(_IS_PRIMITIVE_CACHE_MinimalCtx, (world, tt)) do
-        return !isempty(Base._methods_by_ftype(tt, -1, world))
+    try
+        Base.invoke_in_world(world, _is_primitive, ctx, mode, sig)::Bool
+    catch e
+        ;
+        # Sometimes there are ambiguous `_is_pimitive` declarations, especially ones where
```
sunxd3:
```suggestion
        # Sometimes there are ambiguous `_is_primitive` declarations, especially ones where
```

src/interpreter/contexts.jl
```diff
@@ -157,115 +152,37 @@ function is_primitive(
...
-    # ctx-mode-signature triple in world age `world`. If we have looked this up before,
-    # return the answer from the cache.
-
-    tt = Tuple{typeof(_is_primitive),Type{ctx},Type{mode},Type{sig}}
-    return get!(_IS_PRIMITIVE_CACHE_MinimalCtx, (world, tt)) do
-        return !isempty(Base._methods_by_ftype(tt, -1, world))
+    try
+        Base.invoke_in_world(world, _is_primitive, ctx, mode, sig)::Bool
+    catch e
+        ;
```
sunxd3:
```suggestion
```

src/interpreter/contexts.jl
```diff
@@ -157,115 +152,37 @@ function is_primitive(
...
-    # of this ctx-mode-signature triple in world age `world`. If we have looked this up
-    # before, return the answer from the cache.
-    tt = Tuple{typeof(_is_primitive),Type{ctx},Type{mode},Type{<:sig}}
-    return get!(_MAYBE_PRIMITIVE_CACHE_MinimalCtx, (world, tt)) do
-        return !isempty(Base._methods_by_ftype(tt, -1, world))
+    # `DefaultCtx` _or_ `MinimalCtx`.
+    try
+        Base.invoke_in_world(world, _is_primitive, ctx, mode, sig)::Bool
+    catch e
+        ;
```
sunxd3:
```suggestion
```

sunxd3 (approved):
minor nits. this is great, thanks!

sunxd3:
another tiny thing, `abstract_call_gf_by_type` calls `find_matching_methods`, maybe we can reuse, but I don't think it will give much gain

src/interpreter/contexts.jl
MasonProtter: Oops, sorry that was added by the formatter, it used to be `catch e;`

MasonProtter:
> another tiny thing, `abstract_call_gf_by_type` calls `find_matching_methods`, maybe we can reuse, but I don't think it will give much gain

Yeah, I thought a bit about this, but I figured the only way to avoid the reuse was to copy the whole implementation from `Core.Compiler` over here, but then that would cause some maintenance burden I figured you wouldn't want.

If you like, I could try it out though and report how it changes the benchmark? I also suspect it won't make a huge difference, but who knows.

sunxd3:
not necessary, but I would be curious to see, only if you end up doing it of course

PR #878, opened 2025-12-06
Enable second-order differentiation via forward-over-reverse

sunxd3:
Ref: https://github.com/chalk-lab/Mooncake.jl/issues/826

This PR aims to enable forward-over-reverse for second order derivatives. Many of the necessary plumbings are already implemented before this PR. This PR smooths out the remaining edges.

## Fixes:

1. World age fix for MistyClosure: Forward mode now uses MistyClosure's original world age instead of current world, fixing failures when `zero_tangent(pb_mc)` was called.
2. `Union{}` type issue (this is somewhat 1.10 specific) `Union{}` can showup when type inference fails, the approach here is to replace them with `Any`
3. Add some necessary rules to avoid hanging when testing some particular functions
    1. `Core.memorynew`
    2. `LazyDerivedRule`: avoid diff through `get_interpreter` (`jl_get_world_counter`)
    3. `jl_alloc_array_1d/2d/3d`

## Pointers to some possible future work
- `second_derivative` DITest fails on 1.11 CI (pass locally on laptop) https://github.com/chalk-lab/Mooncake.jl/pull/878#issuecomment-3699183272
- reverse mode primitives are inlined https://github.com/chalk-lab/Mooncake.jl/pull/878#issuecomment-3666924752
- potential new interface functions for hessian https://github.com/chalk-lab/Mooncake.jl/pull/878#issuecomment-3696441923
- use of `DI.inner_preparation_behavior` https://github.com/chalk-lab/Mooncake.jl/pull/878#issuecomment-3696647362

Closes https://github.com/chalk-lab/Mooncake.jl/issues/632

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.192"
+version = "0.4.193"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"


sunxd3:
@yebai @Technici4n this is ready for a review.

Integration with DI is not complete, I have some guess of reasons, but not certain yet. The PR is probably big enough.

~One thing I don't know how to resolve: examples like `x -> sum(x .* x)` will hang on 1.10. Function like these uses `DerivedFRule` and `frule_type` on 1.10 hangs. A hot fix is something like https://github.com/chalk-lab/Mooncake.jl/pull/878/commits/03eb3c036952d456731ae2960b088c2bef492022. it works, but creates allocations.~ 
`x -> sum(x .* x)` is working fine on 1.10 in fact, I mistake "taking a bit long" as "hangs"

yebai:
> One thing I don't know how to resolve: examples like x -> sum(x .* x) will hang on 1.10. Function like these uses DerivedFRule and frule_type on 1.10 hangs.

This might be an actual Julia compiler bug. Can you try to create a MWE so we can report and patch over it? 

> Integration with DI is not complete, I have some guess of reasons, but not certain yet. The PR is probably big enough.

Probably okay to keep fixing DI related issues in this PR if they are not major. Also happy to start a seprate PR otherwise.

test/ext/differentiation_interface/differentiation_interface.jl
```diff
@@ -5,8 +5,17 @@ Pkg.develop(; path=joinpath(@__DIR__, "..", "..", ".."))
...
+# Test first-order differentiation (reverse mode)
 test_differentiation(
     [AutoMooncake(; config=nothing), AutoMooncake(; config=Mooncake.Config())];
     excluded=SECOND_ORDER,
     logging=true,
 )
+
+# Test second-order differentiation (forward-over-reverse for Hessian)
+test_differentiation(
+    [SecondOrder(AutoMooncakeForward(; config=nothing), AutoMooncake(; config=nothing))],
```
gdalle: Note that forward over forward would be a good test too

gdalle:
Very cool that you found a way to do this without touching DI internals @sunxd3, congrats and thanks for the effort!!!
Let me know when this is in a good enough state, I'll put it through some more extensive DI testing

sunxd3:
@yebai 

> This might be an actual Julia compiler bug. 

It might be, but whatever it is, it's fixed in 1.11. Curious what it is, though.
The tests pass, but is taking quite a long while (https://github.com/chalk-lab/Mooncake.jl/actions/runs/20024598254/job/57418981172?pr=878). 

My guess is that type inference failed, produce `Union{}`, but handled by the `Union{}` fix, so the test didn't fail. see https://github.com/chalk-lab/Mooncake.jl/pull/878#issuecomment-3632457694

I would prefer to stop here and fix DI integration in another PR.

sunxd3:
Thanks @gdalle, unfortunately, the DI test [doesn't work yet](https://github.com/chalk-lab/Mooncake.jl/actions/runs/20024598254/job/57418981330?pr=878).

I wanted to keep it simple, and use [Mooncake internal functions](https://github.com/chalk-lab/Mooncake.jl/blob/03eb3c036952d456731ae2960b088c2bef492022/test/interpreter/forward_over_reverse.jl#L1-L34) to make minimal f-o-r work.

I might start DI integration in another PR, I don't think we would need extension, but def need more investigation.

gdalle:
I agree, 2nd order should work inside of Mooncake instead of requiring DI hacks. I just meant that DI has more possible tests to run than what is currently inside your integration tests

src/rrules/low_level_maths.jl
```diff
@@ -71,6 +71,30 @@
 @from_chainrules MinimalCtx Tuple{typeof(deg2rad),IEEEFloat}
 @from_chainrules MinimalCtx Tuple{typeof(rad2deg),IEEEFloat}
 @from_chainrules MinimalCtx Tuple{typeof(^),P,P} where {P<:IEEEFloat}
+
+@is_primitive MinimalCtx Tuple{
```
Technici4n: Why is this needed? The methods look like they should be differentiable by Mooncake (it's just `x*x` or similar).

src/rrules/misc.jl
```diff
@@ -39,6 +39,26 @@
     @zero_derivative MinimalCtx Tuple{typeof(Base.dataids),Memory}
 end
 
+# Forward rule for push! - needed for forward-over-reverse on code that accumulates pullbacks
+# Only primitive in ForwardMode - reverse mode uses derived rule
```
Technici4n: Why would reverse mode not need the primitive if forward mode needs it?

src/interpreter/forward_mode.jl
```diff
@@ -27,8 +55,8 @@ function build_frule(
     @nospecialize sig_or_mi
 
     # To avoid segfaults, ensure that we bail out if the interpreter's world age is greater
-    # than the current world age.
-    if Base.get_world_counter() > interp.world
+    # than the current world age. Exception: MistyClosure uses its own world age.
```
Technici4n: We could maybe add a kwarg to disable the check? Special-casing MC here is a bit strange.

src/rrules/foreigncall.jl
```diff
@@ -256,6 +256,34 @@ end
 
 @zero_derivative MinimalCtx Tuple{typeof(Base.has_free_typevars),Any}
 
+# jl_genericmemory_owner is used by Base.dataids to determine memory aliasing.
+# The result is just a pointer for aliasing detection, so it has zero derivative.
```
Technici4n: Why not use `@zero_derivative`?

src/interpreter/reverse_mode.jl
```diff
@@ -1893,6 +1893,76 @@ end
     return rule.rule(args...)
 end
 
+# Forward-mode primitive for _build_rule! on LazyDerivedRule.
+# This avoids differentiating through get_interpreter which has a ccall to jl_get_world_counter.
+# The tangent propagation happens through the fwds_oc MistyClosure call, not the rule building.
+# Only primitive in ForwardMode - reverse mode uses derived rule.
```
Technici4n: Is said derived rule working? If not it's likely better to define a rule that throws rather than letting Mooncake try to differentiate through it and fail.

yebai:
>> This might be an actual Julia compiler bug.

>It might be, but whatever it is, it's fixed in 1.11. Curious what it is, though.
The tests pass, but is taking quite a long while (https://github.com/chalk-lab/Mooncake.jl/actions/runs/20024598254/job/57418981172?pr=878). My guess is that type inference failed, produce Union{}, but handled by the Union{} fix, so the test didn't fail.

@sunxd3 That’s possible. For the sake of thoroughness, let’s get to the bottom of this. For example, if your guess above is correct, we ought to see Union{} returned by frule_type for the MWE that’s causing issues.

src/rrules/foreigncall.jl
sunxd3: My understanding is that [`@zero_derivative`](https://github.com/chalk-lab/Mooncake.jl/blob/09c496a3d7a98817f3305520523501ccd45c92a0/src/tools_for_rules.jl#L177-L180) needs the function name to the first argument. `_foreigncall_` is special case because because the first argument is `_foreigncall_` not the ccall name.

src/rrules/low_level_maths.jl
sunxd3: Good question, it turns out we don't. It was added as a fix, but I think the rule for `_build_rule!` superseded this, so it's not needed.

src/rrules/misc.jl
sunxd3: also removed

src/interpreter/reverse_mode.jl
sunxd3: added a rrule saying we don't support reverse-o-r yet

src/rrules/foreigncall.jl
sunxd3: https://github.com/chalk-lab/Mooncake.jl/actions/runs/20058347212/job/57528828789?pr=878 interesting only fails on 1.11 (I removed the rule temporarily to see where it traps)

test/rules/high_order_derivative_patches.jl
```diff
@@ -0,0 +1,118 @@
...
+        f(x) = sum([x[1] * x[1], x[2] * x[2]])
+        x = [2.0, 3.0]
+        grad, hess_col = _hessian_column(f, x, 1)
+        @test grad ≈ [4.0, 6.0] rtol = 1e-10
+        @test hess_col ≈ [2.0, 0.0] rtol = 1e-10
+    end
+
+    @testset "broadcast sum of squares" begin
+        # Tests broadcast operations: x .* x uses broadcasting
+        f(x) = sum(x .* x)
```
yebai, line 89: @sunxd3 does this work on Julia 1.0? According to your comment in https://github.com/chalk-lab/Mooncake.jl/pull/878#issuecomment-3626404003, this will hang?

sunxd3: It works, won't hang.

sunxd3: I find the reason for Union{}, going to write it down later

yebai: Pleaase update your comment https://github.com/chalk-lab/Mooncake.jl/pull/878#issuecomment-3626404003 to avoid confusing others.

sunxd3:
On the `Tuple field type cannot be Union{}` error:

With the following code example,
```julia
using Core: ReturnNode
using Core.Compiler
const CC = Core.Compiler
using Mooncake: build_rrule, lookup_ir, get_interpreter, ReverseMode

f(x) = sum(x .* x)
x = [1.0, 2.0]

tt = Tuple{typeof(f),typeof(x)}
interp = get_interpreter(ReverseMode)
julia_ir, _ = lookup_ir(interp, tt)

rule = build_rrule(f, x)
interp = get_interpreter(ReverseMode)
mooncake_ir, _ = lookup_ir(interp, rule.fwds_oc)

for i in 1:length(mooncake_ir.stmts)
    mooncake_ir.stmts[i][:type] === Union{} || continue
    stmt = @static if VERSION >= v"1.11-"
        mooncake_ir.stmts[i][:stmt]
    else
        mooncake_ir.stmts[i][:inst]
    end
    typ = mooncake_ir.stmts[i][:type]
    println("SSA $i :: $typ = $(typeof(stmt).name.name) - $stmt")
end
```

<details>
<summary>On Julia 1.10, this returns (click to expand)</summary>

```julia
SSA 124 :: Union{} = GotoIfNot - goto %44 if not %123
SSA 127 :: Union{} = GotoNode - goto %47
SSA 129 :: Union{} = GotoIfNot - goto %46 if not %128
SSA 132 :: Union{} = GotoNode - goto %47
SSA 134 :: Union{} = GotoNode - goto %47
SSA 144 :: Union{} = Expr - (Mooncake.rrule!!)(...)
SSA 477 :: Union{} = Expr - invoke Base.throw_boundserror(...)
SSA 504 :: Union{} = Expr - invoke Base.throw_boundserror(...)
SSA 544 :: Union{} = Expr - invoke Base.throw_boundserror(...)
SSA 600 :: Union{} = Expr - invoke MistyClosure(throwdm)(...)
SSA 602 :: Union{} = PhiNode - φ ()                          # ← ORPHANED!
SSA 604 :: Union{} = Expr - invoke Mooncake._build_rule!(...)
SSA 606 :: Union{} = PhiNode - φ ()                          # ← ORPHANED!
SSA 607 :: Union{} = Expr - (getfield)(%606, 1)              # ← USES Union{} phi!
SSA 608 :: Union{} = Expr - (getfield)(%606, 2)              # ← USES Union{} phi!
SSA 609 :: Union{} = Expr - (tuple)(%608)
SSA 610 :: Union{} = Expr - (push!)(%9, %609)
```

</details>

<details>
<summary>On Julia 1.12, this returns (click to expand)</summary>

```julia
SSA 45 :: Union{} = Expr - invoke Base.throw_boundserror(...)
SSA 46 :: Union{} = ReturnNode - unreachable     # ← CLEAN!
SSA 79 :: Union{} = Expr - invoke Base.throw_boundserror(...)
SSA 80 :: Union{} = ReturnNode - unreachable     # ← CLEAN!
# ... (pattern continues: throw call → unreachable, no orphaned phi nodes)
SSA 969 :: Union{} = Expr - invoke MistyClosure(throwdm)(...)
SSA 970 :: Union{} = ReturnNode - unreachable    # ← CLEAN!
SSA 971 :: Union{} = Expr - invoke Mooncake._build_rule!(...)
SSA 972 :: Union{} = ReturnNode - unreachable    # ← CLEAN!
# ... (61 total Union{} statements, all are Expr or ReturnNode)
```

</details>

## The Culprit

The orphaned `φ ()` (empty PhiNode) has `Union{}` type. Operations that use it also get `Union{}` type:

```
SSA 606 :: Union{} = PhiNode - φ ()              # ← orphaned empty phi
SSA 607 :: Union{} = Expr - (getfield)(%606, 1)  # ← uses the Union{} phi
```

When Mooncake processes `getfield(%606, 1)`, it builds:
```julia
sig_types = [type_of(%606), type_of(1)]  # = [Union{}, Int]
sig = Tuple{Union{}, Int}                 # ← ERROR on Julia 1.10+!
```

I am still trying to nail down which julia compiler PRs fix these. The fix should be applied to `IncrementalCompact`, mostly likely https://github.com/JuliaLang/julia/pull/52614 and https://github.com/JuliaLang/julia/pull/52863.

### Why Other `Union{}` Statements Are Fine

Mooncake handles other `Union{}`-typed statements gracefully:

- **Expr (throw calls)**: Legitimate "never returns" calls - Mooncake has rules for these
- **ReturnNode - unreachable**: Mooncake skips these (`isdefined(stmt, :val) || return nothing`)

Only the orphaned `PhiNode` and operations using it cause problems, because Mooncake tries to build type signatures containing `Union{}`.

The fix currently in this PR is fine IMO. I would consider this a julia compiler bug, but it was fixed in 1.11 release.

test/rules/high_order_derivative_patches.jl
sunxd3: done

src/rrules/foreigncall.jl
sunxd3: I did a bit more digging. This turns out to be a pretty interesting hidden issue.

The question to ask here is why https://github.com/chalk-lab/Mooncake.jl/blob/b7413d5ca1eded31815b308ee65213bc39f07504/src/rrules/misc.jl#L37-L41 didn't prevent the error.

The answer is that the call is inlined away, and that's why [`jl_genericmemory_owner`](https://github.com/JuliaLang/julia/blob/f2b3dbda30a314d042178ec70ae2115735176f7e/base/abstractarray.jl#L1550-L1564) is exposed (1.12 doesn't use `jl_genericmemory_owner` anymore, so this test doesn't error on 1.12).

To verify, with `julia +1.11`
```julia
using Mooncake
using Mooncake: CC, get_interpreter, BugPatchInterpreter, lookup_ir, optimise_ir!, ForwardMode

f(x::Vector{Float64}) = Base.dataids(x.ref.mem)

interp = get_interpreter(ForwardMode)

# IR from lookup_ir - dataids is NOT inlined (MooncakeInterpreter prevents it)
ir, _ = lookup_ir(interp, Tuple{typeof(f), Vector{Float64}})

# IR after optimise_ir! - dataids IS inlined, exposing jl_genericmemory_owner foreigncall
ir_opt = optimise_ir!(CC.copy(ir); do_inline=true)
```

gives

```julia
julia> ir  # dataids is NOT inlined
1 ─ %1 = Base.getfield(_2, :ref)::MemoryRef{Float64}
│   %2 = Base.getfield(%1, :mem)::Memory{Float64}
│   %3 = invoke Base.dataids(%2::Memory{Float64})::Tuple{UInt64}   # <-- stays as invoke
└──      return %3

julia> ir_opt  # dataids IS inlined, exposing foreigncall
1 ─ %1  = Base.getfield(_2, :ref)::MemoryRef{Float64}
│   %2  = Base.getfield(%1, :mem)::Memory{Float64}
│   %3  = foreigncall(:jl_genericmemory_owner, ...)::Any           # <-- inlined!
│   %4  = (%3 isa Memory{Float64})::Bool
└──       goto #3 if not %4
2 ─ %6  = π (%3, Memory{Float64})
└──       goto #4
3 ─       nothing
4 ┄ %9  = φ (#2 => %6, #3 => %2)::Memory{Float64}
│   %10 = Base.getfield(%9, :ptr)::Ptr{Nothing}
│   %11 = Base.bitcast(Ptr{Float64}, %10)::Ptr{Float64}
│   %12 = Core.bitcast(Core.UInt, %11)::UInt64
│   %13 = Core.tuple(%12)::Tuple{UInt64}
└──       return %13
```

## Why This Only Affects Forward-over-Reverse

When building rules, `lookup_ir` uses `MooncakeInterpreter` which prevents inlining primitives. But then `optimise_ir!` uses `BugPatchInterpreter` which has no such protection—it inlines everything including `rrule!!(dataids, ...)` down to the raw foreigncall.

In pure reverse mode, `rule(...)` executes this optimized IR at runtime. The foreigncall just runs and returns a value.

In forward-over-reverse, `build_frule` calls `lookup_ir(MistyClosure)` which returns the already-optimized IR containing the foreigncall. Now we try to differentiate through this IR, which requires an `frule!!` for the foreigncall—but none exists.

sunxd3:
Repost https://github.com/chalk-lab/Mooncake.jl/pull/878#discussion_r2611300394

A proper fix might be adding a new interpreter that combine both interpreters.

sunxd3:
Context: https://github.com/chalk-lab/Mooncake.jl/pull/878#discussion_r2611300394 and https://github.com/chalk-lab/Mooncake.jl/pull/878/commits/ed10d3887c529dd1a7e6cef362288377a7ec8473 (reverted, but shows what was attempted).

In forward-over-reverse, we build a forward rule over the reverse rule. Primitives in reverse mode can get inlined when building the forward rule, exposing internal calls that lack an `frule!!`. One example: `dataids` is a reverse-mode primitive, but inlining it exposes `jl_genericmemory_owner`.

I tried making `BugPatchInterpreter` also able to wrap `MooncakeInterpreter` (not just `NativeInterpreter`) during `optimise_ir!`. This lets `abstract_call_gf_by_type` check if calls are Mooncake primitives and wrap them in `NoInlineCallInfo`, which `inlining_policy` then uses to block inlining.

Two problems emerged:

1. Cascading rule requirements. Blocking primitive inlining means more function calls survive to codegen. This exposed `getfield` calls on `Dual{P, P}` (self-tangent types where `tangent_type(P) == P`). The existing `frule!!(getfield, ::Dual{P, <:StandardTangentType}, ...)` doesn't cover these, so I added `frule!!(getfield, ::Dual{P, P}, ...)`. But this created method ambiguity for types that are both self-tangent and subtypes of `StandardTangentType`. Restricting to `AbstractArray` fixed that case but felt like whack-a-mole — other types could surface similar issues.

2. Allocation regressions. Blocking inlining caused massive allocations across all Julia versions. Simple functions like `bar(x,y) = x*y` showed 2-7 allocations; loop-heavy code like `test_naive_mat_mul!` showed much more.

The interpreter approach is more fundamental but requires deeper work: properly handling the tangent type dispatch and understanding the allocation mechanism. For now, I added an `frule!!/rrule!!` for `jl_genericmemory_owner` directly in the test file as a workaround.

sunxd3:
@yebai @Technici4n I believe this is ready for a final look. I'll leave proper inline handling and DI integration in future PRs.

MasonProtter:
Sorry if this is bad timing, but I'd like to just note that I have a new version of my compile-time rule generation system that's almost ready for a PR, and one of the things it does is make higher-order AD effortless without any custom rules. It should also work with all combinations of Forward and Reverse (I'll be able to say this more definitively in the next couple days once reverse is working properly)

The idea is that since the creation of rules happens inside a generated function generator at compile time, when you do higher order derivatives, the outer derivative never sees the code which creates the rule in the inner derivative, so the only thing it ever needs to actually differentiate is the code *produced* by the inner rule.

gdalle:
That sounds very very cool!!!

gdalle:
And now that you mention it, it does seem like the right approach. I suspect the reason @sunxd3 is struggling is that we currently need to differentiate through the IR generator itself, which kinda sucks

Technici4n:
@MasonProtter that sounds awesome, I would indeed be curious to see what your suggestion looks like!

sunxd3:
It's perfect timing! Guillaume is right on the struggle. Current approach might work okay, after some painful cherry-picking and whack-a-mole. Your idea sounds more elegant. Looking forward to review!

yebai:
I don’t think there’s a “magical” solution once you start looking at the autograd compiler details. One example is that Mooncake can derive rules at runtime to support dynamic dispatch (i.e., [`LazyDerivedRule`](https://github.com/chalk-lab/Mooncake.jl/blob/5daa968fb1e01a8bea563cf68f2d75c889410a41/src/interpreter/reverse_mode.jl#L1807)), which is what’s encountered when making the tests in this PR pass.

This PR only adds a single rule for the autograd IR compiler, `_build_rule!`; other rules are for foreign calls, etc.

Beyond that, it also fixes issues that apply more generally, regardless of whether the compiled rule is used or not.

We should still review and merge this PR and remove the rule for `_build_rule!` if it becomes obsolete.

yebai:
> Primitives in reverse mode can get inlined when building the forward rule, exposing internal calls that lack an frule!!. One example: dataids is a reverse-mode primitive, but inlining it exposes jl_genericmemory_owner.

@sunxd3, note that further compiler optimisation runs after reverse-mode autograd builds a rule (see below). This is where things could go wrong, since Mooncake no longer prevents inlining because higher-order/nested autograd was not considered initially. 

https://github.com/chalk-lab/Mooncake.jl/blob/5daa968fb1e01a8bea563cf68f2d75c889410a41/src/interpreter/reverse_mode.jl#L1219-L1223

Technici4n:
For such an important feature I think that we should wait a bit and see what the best solution is. I don't see a need to rush this in when an alternative implementation is apparently around the corner.

sunxd3:
> > Primitives in reverse mode can get inlined when building the forward rule, exposing internal calls that lack an frule!!. One example: dataids is a reverse-mode primitive, but inlining it exposes jl_genericmemory_owner.
> 
> @sunxd3, note that further compiler optimisation runs after reverse-mode autograd builds a rule (see below). This is where things could go wrong, since Mooncake no longer prevents inlining because higher-order/nested autograd was not considered initially.
> 
> https://github.com/chalk-lab/Mooncake.jl/blob/5daa968fb1e01a8bea563cf68f2d75c889410a41/src/interpreter/reverse_mode.jl#L1219-L1223

thanks, this makes sense

MasonProtter:
Okay, I've posted a bit of a write-up on my progress with the compile-time rule generation here: https://github.com/chalk-lab/Mooncake.jl/pull/900

I think there's still a lot to figure out there, so it's probably best to not let it block this work. Also, at least some of this will definitely be needed even if I totally figure out the compile-time stuff, because we'll want to preserve `is_primitive` calls and whatnot, which my mechanism erases.

sunxd3:
With current branch,

<details><summary>the following code</summary>
<p>

```julia
using Pkg
Pkg.activate(@__DIR__)

using Mooncake: Mooncake

f_simple(x) = sum(x .^ 2)

# Test input
x = [1.0, 2.0]

# Backends
reverse_backend = AutoMooncake(; config=nothing)
forward_backend = AutoMooncakeForward(; config=nothing)
second_order_backend = SecondOrder(forward_backend, reverse_backend)

v = [1.0, 0.0]
hvp_result = hvp(f_simple, second_order_backend, x, (v,))

H = hessian(f_simple, second_order_backend, x)
```

</p>
</details> 

will produce

<details><summary>error message</summary>
<p>

```julia
julia> hvp_result = hvp(f_simple, second_order_backend, x, (v,))
ERROR: No frule!! available for foreigncall with primal argument types Tuple{Val{:jl_get_world_counter}, Val{UInt64}, Tuple{}, Val{0}, Val{:ccall}}. This problem has most likely arisen because there is a ccall somewhere in the function you are trying to differentiate, for which an frule!! has not been explicitly written.You have three options: write an frule!! for this foreigncall, write an frule!! for a Julia function that calls this foreigncall, or re-write your code to avoid this foreigncall entirely. If you believe that this error has arisen for some other reason than the above, or the above does not help you to workaround this problem, please open an issue.
Stacktrace:
  [1] throw_missing_foreigncall_rule_error(::Symbol, ::Mooncake.Dual{Val{…}, Mooncake.NoTangent}, ::Vararg{Any})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/rrules/foreigncall.jl:19
  [2] frule!!(::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/rrules/foreigncall.jl:12
  [3] get_world_counter
    @ ./runtime_internals.jl:1269 [inlined]
  [4] get_interpreter
    @ ~/TuringLang/Mooncake.jl/src/interpreter/abstract_interpretation.jl:284 [inlined]
  [5] (::Tuple{…})(_2::Mooncake.Dual{…}, _3::Mooncake.Dual{…})
  [6] (::MistyClosures.MistyClosure{…})(::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ MistyClosures ~/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
  [7] DerivedFRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:120 [inlined]
  [8] LazyFRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:475 [inlined]
  [9] #build_rrule#233
    @ ~/TuringLang/Mooncake.jl/src/interpreter/reverse_mode.jl:1070 [inlined]
 [10] (::Tuple{…})(_2::Mooncake.Dual{…}, _3::Mooncake.Dual{…}, _4::Mooncake.Dual{…}, _5::Mooncake.Dual)
 [11] (::MistyClosures.MistyClosure{…})(::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ MistyClosures ~/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
 [12] DerivedFRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:120 [inlined]
 [13] LazyFRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:475 [inlined]
 [14] tuple_splat
    @ ~/TuringLang/Mooncake.jl/src/utils.jl:54 [inlined]
 [15] (::Tuple{…})(_2::Mooncake.Dual{…}, _3::Mooncake.Dual{…}, _4::Mooncake.Dual{…})
 [16] (::MistyClosures.MistyClosure{…})(::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ MistyClosures ~/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
 [17] (::Mooncake.DerivedFRule{…})(::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:120
 [18] (::Mooncake.DynamicFRule{…})(::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:523
 [19] _apply_iterate_equivalent
    @ ~/TuringLang/Mooncake.jl/src/rrules/builtins.jl:744 [inlined]
 [20] (::Tuple{…})(_2::Mooncake.Dual{…}, _3::Mooncake.Dual{…}, _4::Mooncake.Dual{…}, _5::Mooncake.Dual{…})
 [21] (::MistyClosures.MistyClosure{…})(::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ MistyClosures ~/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
 [22] (::Mooncake.DerivedFRule{…})(::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:120
 [23] (::Mooncake.DynamicFRule{…})(::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ Mooncake ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:523
 [24] #prepare_gradient_cache#495
    @ ~/TuringLang/Mooncake.jl/src/interface.jl:535 [inlined]
 [25] (::NTuple{…})(_2::Mooncake.Dual{…}, _3::Mooncake.Dual{…}, _4::Mooncake.Dual{…}, _5::Mooncake.Dual)
 [26] (::MistyClosures.MistyClosure{…})(::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ MistyClosures ~/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
 [27] DerivedFRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:120 [inlined]
 [28] LazyFRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:475 [inlined]
 [29] prepare_gradient_nokwarg
    @ ~/.julia/packages/DifferentiationInterface/6H4dc/ext/DifferentiationInterfaceMooncakeExt/onearg.jl:124 [inlined]
 [30] (::Tuple{…})(_2::Mooncake.Dual{…}, _3::Mooncake.Dual{…}, _4::Mooncake.Dual{…}, _5::Mooncake.Dual{…}, _6::Mooncake.Dual{…}, _7::Mooncake.Dual{…})
 [31] (::MistyClosures.MistyClosure{…})(::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ MistyClosures ~/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
 [32] DerivedFRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:120 [inlined]
 [33] LazyFRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:475 [inlined]
 [34] gradient
    @ ~/.julia/packages/DifferentiationInterface/6H4dc/src/first_order/gradient.jl:62 [inlined]
 [35] (::Tuple{…})(_2::Mooncake.Dual{…}, _3::Mooncake.Dual{…}, _4::Mooncake.Dual{…}, _5::Mooncake.Dual{…}, _6::Mooncake.Dual{…})
 [36] (::MistyClosures.MistyClosure{…})(::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ MistyClosures ~/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
 [37] DerivedFRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:120 [inlined]
 [38] LazyFRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:475 [inlined]
 [39] shuffled_gradient
    @ ~/.julia/packages/DifferentiationInterface/6H4dc/src/first_order/gradient.jl:160 [inlined]
 [40] (::Tuple{…})(_2::Mooncake.Dual{…}, _3::Mooncake.Dual{…}, _4::Mooncake.Dual{…}, _5::Mooncake.Dual{…}, _6::Mooncake.Dual{…}, _7::Mooncake.Dual{…})
 [41] (::MistyClosures.MistyClosure{…})(::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…}, ::Mooncake.Dual{…})
    @ MistyClosures ~/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
 [42] DerivedFRule
    @ ~/TuringLang/Mooncake.jl/src/interpreter/forward_mode.jl:120 [inlined]
 [43] value_and_derivative!!
    @ ~/TuringLang/Mooncake.jl/src/interface.jl:610 [inlined]
 [44] (::DifferentiationInterfaceMooncakeExt.var"#20#21"{…})(dx::Vector{…})
    @ DifferentiationInterfaceMooncakeExt ~/.julia/packages/DifferentiationInterface/6H4dc/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl:43
 [45] map
    @ ./tuple.jl:358 [inlined]
 [46] value_and_pushforward
    @ ~/.julia/packages/DifferentiationInterface/6H4dc/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl:40 [inlined]
 [47] pushforward
    @ ~/.julia/packages/DifferentiationInterface/6H4dc/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl:67 [inlined]
 [48] hvp(::typeof(f_simple), ::DifferentiationInterface.ForwardOverAnythingHVPPrep{…}, ::SecondOrder{…}, ::Vector{…}, ::Tuple{…})
    @ DifferentiationInterface ~/.julia/packages/DifferentiationInterface/6H4dc/src/second_order/hvp.jl:331
 [49] hvp(::typeof(f_simple), ::SecondOrder{AutoMooncakeForward{…}, AutoMooncake{…}}, ::Vector{Float64}, ::Tuple{Vector{…}})
    @ DifferentiationInterface ~/.julia/packages/DifferentiationInterface/6H4dc/src/second_order/hvp.jl:75
 [50] top-level scope
    @ ~/TuringLang/Mooncake.jl/dev/dev.jl:26
``` 

</p>
</details> 

From the error, it looks like the forward mode is diffing through the prep. This led me trying to overload (this is not a proposed final solution, just for discussion)

```julia
DI.inner_preparation_behavior(::AutoMooncakeForward) = DI.PrepareInnerSimple()
```

```julia
test_differentiation(
    [second_order_backend];
    excluded=[FIRST_ORDER; :second_derivative],
    logging=true,
)
```

all passes (in another words, `hvp` and `hessian` word reasonably well). 

For me, this feels like good news. Right at this moment, it's not obvious to me how good and bad the overload above is.

yebai:
Perhaps the best approach is for Mooncake provide its own `hessian` operator, similar to `ForwardDiff` ([code](https://github.com/JuliaDiff/ForwardDiff.jl/blob/e1eb522db95731630ef0e55df880517900efae98/src/hessian.jl#L6)). This would involve adding the functions `value_and_hessian!!` (returns value + Hessian, and fills a gradient buffer) and `prepare_hessian_cache`, analogous to [`value_and_gradient!!`](https://github.com/chalk-lab/Mooncake.jl/blob/0b55297d1418502eac42b69a6b3a252264288e5b/src/interface.jl#L146-L168) and [`prepare_gradient_cache`](https://github.com/chalk-lab/Mooncake.jl/blob/0b55297d1418502eac42b69a6b3a252264288e5b/src/interface.jl#L527-L533).

sunxd3:
> This would involve adding the functions `value_and_hessian!!` (returns value + Hessian, and fills a gradient buffer) and `prepare_hessian_cache`, analogous to [`value_and_gradient!!`](https://github.com/chalk-lab/Mooncake.jl/blob/0b55297d1418502eac42b69a6b3a252264288e5b/src/interface.jl#L146-L168) and [`prepare_gradient_cache`](https://github.com/chalk-lab/Mooncake.jl/blob/0b55297d1418502eac42b69a6b3a252264288e5b/src/interface.jl#L527-L533).

This certainly means Mooncake can have more leeway for optimizations.

I just made a commit with some frules that, at least on my local machine, passes the DITest second order tests.

test/ext/differentiation_interface/differentiation_interface.jl
```diff
@@ -7,7 +7,5 @@ using Mooncake: Mooncake
 
 # Test first-order differentiation (reverse mode)
 test_differentiation(
-    [AutoMooncake(; config=nothing), AutoMooncake(; config=Mooncake.Config())];
-    excluded=SECOND_ORDER,
-    logging=true,
+    [AutoMooncake(; config=nothing), AutoMooncake(; config=Mooncake.Config())]; logging=true
```
yebai: ~~Is this testing reverse-over-reverse for high order deriatives?~~

EDIT: The commit is outdated.

yebai:
I suggest we move the rules for `_build_rule!` and `jl_genericmemory_owner` into a new file named `rules/high_order_derivative_patches.jl`. This keeps their utility clear. We can also remove this new patch file later if no longered needed.

gdalle:
> From the error, it looks like the forward mode is diffing through the prep. This led me trying to overload (this is not a proposed final solution, just for discussion)
>
>```julia
>DI.inner_preparation_behavior(::AutoMooncakeForward) = DI.PrepareInnerSimple()
>```

This seems correct to me. I introduced the `inner_preparation_behavior` trait in DI to handle second order with operator overloading backends like ForwardDiff, where the inner differentiation happens on `Dual` numbers, and must be prepared accordingly. Hence, we have

```julia
DI.inner_preparation_behavior(::AutoForwardDiff) = DI.PrepareInnerOverload()
```

In the case of Mooncake, we're dealing with source transformation, so `DI.PrepareInnerSimple()` is more appropriate, good job for finding that.
The default setting is `DI.DontPrepareInner()` because redoing preparation at each call is the easiest and most natural option, but here it requires differentiating through Mooncake compilation.

sunxd3:
The DI second-order test failure on 1.11 is a bit confusing to me:
* looking at https://github.com/chalk-lab/Mooncake.jl/actions/runs/20573058900/job/59109194638, it seems to run just fine then suddenly exploded
* I run the test locally and it works just fine

sorry to tag @gdalle have you seen this before, it does seem to me OpaqueClosure can be a bit flaky on 1.11?

gdalle:
Is it just hitting a 60-minute time limit on the CI job? If you take a look at DI's GitHub Actions log, you'll see that they can go above one hour, depending on what you put in them

src/interpreter/forward_mode.jl
```diff
@@ -16,19 +44,27 @@ end
...
+    debug_mode=false,
+    silence_debug_messages=true,
+    skip_world_age_check=false,
 ) where {C}
     @nospecialize sig_or_mi
 
     # To avoid segfaults, ensure that we bail out if the interpreter's world age is greater
     # than the current world age.
-    if Base.get_world_counter() > interp.world
+    if !skip_world_age_check && Base.get_world_counter() > interp.world
```
yebai, line 67: Might this be the cause of the DI crash, since MistyClosure skips this check?

sunxd3:
> Is it just hitting a 60-minute time limit on the CI job?

I was curious about this, [according to Github](https://docs.github.com/en/actions/reference/limits#existing-system-limits) the time limit is 6 hours. I don't think Mooncake has a time limit set. But this prompted me to break the second-order DITest out. Unfortunately, [it still failed ](https://github.com/chalk-lab/Mooncake.jl/actions/runs/20594485899/job/59146164602) at around 57 mins.  (Since then, I limit the DI-second-order to only run the `:hessian` test group as a check.)

Without digging deep (which I plan to do, potentially in a another optimization PR), I suspect the current implementation is quite taxing on the system, so the segfault might be legitimate OOM.

src/interpreter/forward_mode.jl
sunxd3: This is a good heuristic. I don't see worldage complaints in the error message, but they could be buried.

<details><summary>end of CI log</summary>
<p>

```julia
◒ Testing correctness    Time: 0:49:13
          backend: SecondOrder(AutoMooncakeForward(), AutoMooncake()) - 1/1
    scenario_type: second_derivative - 3/3
         scenario: 1/5
   operator_place: out
   function_place: out
         function: num_to_mat
       input_type: Float64
       input_size: ()
      output_type: Matrix{Float64}
      output_size: (2, 2)
      nb_tangents: nothing
      nb_contexts: 0












[2247] signal 11 (1): Segmentation fault
in expression starting at /home/runner/work/Mooncake.jl/Mooncake.jl/test/ext/differentiation_interface_second_order/differentiation_interface_second_order.jl:9
typekeyvalue_hash at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/jltypes.c:1662 [inlined]
lookup_typevalue at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/jltypes.c:1088
jl_inst_arg_tuple_type at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/jltypes.c:2319
arg_type_tuple at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/gf.c:2280 [inlined]
jl_lookup_generic_ at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/gf.c:3077 [inlined]
ijl_apply_generic at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/gf.c:3124
tuple_splat at /home/runner/work/Mooncake.jl/Mooncake.jl/src/utils.jl:54 [inlined]
opaque closure at ./<missing>:0
MistyClosure at /home/runner/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
unknown function (ip: 0x7fb795855273)
jl_apply at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/julia.h:2157 [inlined]
do_apply at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/builtins.c:831
DerivedFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:120
unknown function (ip: 0x7fb795855193)
DynamicFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:523
unknown function (ip: 0x7fb795852483)
_apply_iterate_equivalent at /home/runner/work/Mooncake.jl/Mooncake.jl/src/rules/builtins.jl:744 [inlined]
opaque closure at ./<missing>:0
MistyClosure at /home/runner/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
unknown function (ip: 0x7fb795828283)
jl_apply at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/julia.h:2157 [inlined]
do_apply at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/builtins.c:831
DerivedFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:120
unknown function (ip: 0x7fb79582818a)
DynamicFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:523
typejoin_union_tuple at ./promotion.jl:227 [inlined]
promote_typejoin_union at ./promotion.jl:193 [inlined]
_collect at ./array.jl:768 [inlined]
opaque closure at ./<missing>:0
MistyClosure at /home/runner/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
DerivedFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:120 [inlined]
_build_rule! at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:481
LazyFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:475 [inlined]
collect_similar at ./array.jl:720 [inlined]
map at ./abstractarray.jl:3371 [inlined]
_value_and_pushforward_via_pullback at /home/runner/.julia/packages/DifferentiationInterface/6H4dc/src/first_order/pushforward.jl:364 [inlined]
opaque closure at ./<missing>:0
MistyClosure at /home/runner/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
DerivedFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:120 [inlined]
_build_rule! at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:481
LazyFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:475 [inlined]
value_and_pushforward at /home/runner/.julia/packages/DifferentiationInterface/6H4dc/src/first_order/pushforward.jl:403 [inlined]
pushforward at /home/runner/.julia/packages/DifferentiationInterface/6H4dc/src/first_order/pushforward.jl:430 [inlined]
derivative at /home/runner/.julia/packages/DifferentiationInterface/6H4dc/src/first_order/derivative.jl:200 [inlined]
opaque closure at ./<missing>:0
MistyClosure at /home/runner/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
DerivedFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:120
unknown function (ip: 0x7fb795ae546a)
DynamicFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:523
unknown function (ip: 0x7fb795ab59da)
derivative at /home/runner/.julia/packages/DifferentiationInterface/6H4dc/src/first_order/derivative.jl:102 [inlined]
opaque closure at ./<missing>:0
MistyClosure at /home/runner/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
DerivedFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:120 [inlined]
_build_rule! at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:481
LazyFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:475 [inlined]
shuffled_derivative at /home/runner/.julia/packages/DifferentiationInterface/6H4dc/src/first_order/derivative.jl:284 [inlined]
opaque closure at ./<missing>:0
MistyClosure at /home/runner/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
DerivedFRule at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interpreter/forward_mode.jl:120 [inlined]
value_and_derivative!! at /home/runner/work/Mooncake.jl/Mooncake.jl/src/interface.jl:610 [inlined]
#13 at /home/runner/.julia/packages/DifferentiationInterface/6H4dc/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl:43
map at ./tuple.jl:355 [inlined]
value_and_pushforward at /home/runner/.julia/packages/DifferentiationInterface/6H4dc/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl:40 [inlined]
value_and_derivative at /home/runner/.julia/packages/DifferentiationInterface/6H4dc/src/first_order/derivative.jl:171 [inlined]
value_derivative_and_second_derivative at /home/runner/.julia/packages/DifferentiationInterface/6H4dc/src/second_order/second_derivative.jl:145
unknown function (ip: 0x7fb795b4ebcb)
value_derivative_and_second_derivative at /home/runner/.julia/packages/DifferentiationInterface/6H4dc/src/second_order/second_derivative.jl:71
#238 at /home/runner/.julia/packages/DifferentiationInterfaceTest/4ejkD/src/tests/correctness_eval.jl:351
#70 at ./tuple.jl:692 [inlined]
BottomRF at ./reduce.jl:86 [inlined]
afoldl at ./operators.jl:553 [inlined]
_foldl_impl at ./reduce.jl:68 [inlined]
foldl_impl at ./reduce.jl:48 [inlined]
mapfoldl_impl at ./reduce.jl:44 [inlined]
#mapfoldl#337 at ./reduce.jl:175 [inlined]
mapfoldl at ./reduce.jl:175 [inlined]
#foldl#338 at ./reduce.jl:198 [inlined]
foldl at ./reduce.jl:198 [inlined]
foreach at ./tuple.jl:692
unknown function (ip: 0x7fb795b4e766)
#test_correctness#236 at /home/runner/.julia/packages/DifferentiationInterfaceTest/4ejkD/src/tests/correctness_eval.jl:350
test_correctness at /home/runner/.julia/packages/DifferentiationInterfaceTest/4ejkD/src/tests/correctness_eval.jl:324
unknown function (ip: 0x7fb795cde7b6)
macro expansion at /home/runner/.julia/packages/DifferentiationInterfaceTest/4ejkD/src/test_differentiation.jl:157 [inlined]
macro expansion at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/usr/share/julia/stdlib/v1.11/Test/src/Test.jl:1709 [inlined]
macro expansion at /home/runner/.julia/packages/DifferentiationInterfaceTest/4ejkD/src/test_differentiation.jl:157 [inlined]
macro expansion at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/usr/share/julia/stdlib/v1.11/Test/src/Test.jl:1798 [inlined]
macro expansion at /home/runner/.julia/packages/DifferentiationInterfaceTest/4ejkD/src/test_differentiation.jl:133 [inlined]
macro expansion at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/usr/share/julia/stdlib/v1.11/Test/src/Test.jl:1798 [inlined]
macro expansion at /home/runner/.julia/packages/DifferentiationInterfaceTest/4ejkD/src/test_differentiation.jl:131 [inlined]
macro expansion at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/usr/share/julia/stdlib/v1.11/Test/src/Test.jl:1798 [inlined]
macro expansion at /home/runner/.julia/packages/DifferentiationInterfaceTest/4ejkD/src/test_differentiation.jl:128 [inlined]
macro expansion at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/usr/share/julia/stdlib/v1.11/Test/src/Test.jl:1709 [inlined]
#test_differentiation#729 at /home/runner/.julia/packages/DifferentiationInterfaceTest/4ejkD/src/test_differentiation.jl:128
test_differentiation at /home/runner/.julia/packages/DifferentiationInterfaceTest/4ejkD/src/test_differentiation.jl:94 [inlined]
test_differentiation at /home/runner/.julia/packages/DifferentiationInterfaceTest/4ejkD/src/test_differentiation.jl:69
unknown function (ip: 0x7fb7df31f306)
jl_apply at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/julia.h:2157 [inlined]
do_call at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:126
eval_value at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:223
eval_stmt_value at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:174 [inlined]
eval_body at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:666
jl_interpret_toplevel_thunk at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:824
jl_toplevel_eval_flex at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/toplevel.c:943
jl_toplevel_eval_flex at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/toplevel.c:886
ijl_toplevel_eval_in at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/toplevel.c:994
eval at ./boot.jl:430 [inlined]
include_string at ./loading.jl:2775
_include at ./loading.jl:2835
include at ./sysimg.jl:38
unknown function (ip: 0x7fb803900092)
jl_apply at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/julia.h:2157 [inlined]
do_call at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:126
eval_value at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:223
eval_stmt_value at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:174 [inlined]
eval_body at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:666
jl_interpret_toplevel_thunk at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:824
jl_toplevel_eval_flex at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/toplevel.c:943
jl_toplevel_eval_flex at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/toplevel.c:886
ijl_toplevel_eval_in at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/toplevel.c:994
eval at ./boot.jl:430 [inlined]
include_string at ./loading.jl:2775
_include at ./loading.jl:2835
include at ./sysimg.jl:38
unknown function (ip: 0x7fb803900092)
jl_apply at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/julia.h:2157 [inlined]
do_call at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:126
eval_value at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:223
eval_stmt_value at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:174 [inlined]
eval_body at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:666
jl_interpret_toplevel_thunk at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/interpreter.c:824
jl_toplevel_eval_flex at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/toplevel.c:943
jl_toplevel_eval_flex at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/toplevel.c:886
ijl_toplevel_eval_in at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/toplevel.c:994
eval at ./boot.jl:430 [inlined]
exec_options at ./client.jl:289
_start at ./client.jl:524
jfptr__start_73646.1 at /opt/hostedtoolcache/julia/1.11.8/x64/lib/julia/sys.so (unknown line)
jl_apply at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/julia.h:2157 [inlined]
true_main at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/jlapi.c:900
jl_repl_entrypoint at /cache/build/builder-amdci4-7/julialang/julia-release-1-dot-11/src/jlapi.c:1059
main at julia (unknown line)
unknown function (ip: 0x7fb80a82a1c9)
__libc_start_main at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
unknown function (ip: 0x4010b8)
Allocations: 5899917079 (Pool: 5899586236; Big: 330843); GC: 840
/home/runner/work/_temp/9047e814-cf84-42f5-b851-8873ee652cbb.sh: line 5:  2247 Segmentation fault      (core dumped) julia --code-coverage=user --eval 'include("test/run_extra.jl")'
```

</p>
</details>

src/rules/misty_closures.jl
```diff
@@ -17,7 +17,18 @@ struct MistyClosureTangent
...
 
-_dual_mc(p::MistyClosure) = build_frule(get_interpreter(ForwardMode), p)
+# Build a forward-mode rule for a MistyClosure using its original world age.
+# We skip the world age check because:
+# 1. MistyClosure captures its own IR at creation time (see lookup_ir in ir_utils.jl)
+# 2. build_frule -> generate_dual_ir -> lookup_ir(interp, mc) returns mc.ir[] directly,
+#    bypassing method table lookups that would require a current-world interpreter
+# 3. Any nested non-primitive calls use LazyFRule/DynamicFRule which obtain a current-world
+#    interpreter via get_interpreter() at runtime
+function _dual_mc(p::MistyClosure)
```
yebai, line 34: What would happen if we use the current world age to build a new rule, instead of mistry closure's world age? 

Let's add an explanation in the code.

test/rules/high_order_derivative_patches.jl
```diff
@@ -0,0 +1,118 @@
+function _compute_grad(rule, f, x::Vector{Float64}, x_fdata::Vector{Float64})
```
yebai, line 1: Let's rename this to `test/rules/high_order_derivative_patches.jl` and run it as a sepraste test group (instead of part of `basic`)) for consistency with the rest of files inside `src` folder.

yebai (commented):
Thanks @sunxd3 -- I am happy with the PR.

src/rules/misty_closures.jl
sunxd3: I added some explanations in the [lastest commit](https://github.com/chalk-lab/Mooncake.jl/pull/878/changes/b10b372909946c82faf4bab749e37a9a7557cfdd)

sunxd3:
@copilot Can you bump the patch version?

PR #881, opened 2025-12-08
Add LogExpFunctions rules

### LINKED ISSUE #879: Missing rules for LogExpFunctions
... results in loss of numerical stability for large inputs on `logistic`:

```julia
import DifferentiationInterface as DI
import ForwardDiff, Mooncake
using LogExpFunctions

f(x) = LogExpFunctions.logistic(x[1])
prep_fd = DI.prepare_gradient(f, DI.AutoForwardDiff(), [1.0]);
prep_mcf = DI.prepare_gradient(f, DI.AutoMooncakeForward(), [1.0]);
prep_mcr = DI.prepare_gradient(f, DI.AutoMooncake(), [1.0]);

for x in [-5000.0, -2000.0, -1000.0, -100.0, 0.0, 100.0, 1000.0, 2000.0, 5000.0]
    grad_fd = DI.gradient(f, prep_fd, DI.AutoForwardDiff(), [x]);
    grad_mcf = DI.gradient(f, prep_mcf, DI.AutoMooncakeForward(), [x]);
    grad_mcr = DI.gradient(f, prep_mcr, DI.AutoMooncake(), [x]);
    @show x, grad_fd, grad_mcf, grad_mcr
end

#=
(x, grad_fd, grad_mcf, grad_mcr) = (-5000.0, [0.0], [0.0], [0.0])
(x, grad_fd, grad_mcf, grad_mcr) = (-2000.0, [0.0], [0.0], [0.0])
(x, grad_fd, grad_mcf, grad_mcr) = (-1000.0, [0.0], [0.0], [0.0])
(x, grad_fd, grad_mcf, grad_mcr) = (-100.0, [3.720075976020836e-44], [3.720075976020836e-44], [3.720075976020836e-44])
(x, grad_fd, grad_mcf, grad_mcr) = (0.0, [0.25], [0.25], [0.25])
(x, grad_fd, grad_mcf, grad_mcr) = (100.0, [0.0], [0.0], [0.0])
(x, grad_fd, grad_mcf, grad_mcr) = (1000.0, [0.0], [0.0], [NaN])
(x, grad_fd, grad_mcf, grad_mcr) = (2000.0, [0.0], [0.0], [NaN])
(x, grad_fd, grad_mcf, grad_mcr) = (5000.0, [0.0], [0.0], [NaN])
=#
```

(I don't know if other functions have similar problems, but I assume that it wouldn't hurt to just import every rule that's available.)

I think this should be quite easy to fix by importing rules from ChainRules (defined [here](https://github.com/JuliaStats/LogExpFunctions.jl/blob/master/ext/LogExpFunctionsChainRulesCoreExt.jl)), for example the above is fixed by

```julia
Mooncake.@from_chainrules Mooncake.MinimalCtx Tuple{typeof(LogExpFunctions.logistic),P} where {P<:Base.IEEEFloat}
```

and I don't mind doing a PR, but question is whether you want these in an extension or to take on LogExpFunctions as a hard dep?

penelopeysm:
Closes #879. I found a couple of tests that would fail previously due to Mooncake returning NaN's.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.184"
+version = "0.4.185"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"


Technici4n:
Should this be limited to the functions that can cause a NaN? Or maybe all of them do?

In any case it would be good to explain why this is done, e.g. `# While Mooncake is able to differentiate through these functions, we import rules from ChainRules for improved numerical stability` or similar.

Would it also be possible to test the tricky edge cases? (E.g. logistic function for very large x, which will give a NaN without the custom rule)

penelopeysm:
> Should this be limited to the functions that can cause a NaN? Or maybe all of them do?

Yeah, I was thinking about this too. I could only get logistic and logaddexp to error so far. In fact I also got logsumexp to error whenever it's given an array with both elements equal, which I think is because of this https://github.com/JuliaStats/LogExpFunctions.jl/blob/77c5cf030b58b14f118f237b6b518005499a7f40/src/logsumexp.jl#L91-L102 I haven't yet checked whether it's Mooncake or FiniteDifferences that is the problem here.

Going to mark as draft again while I check that.

> Would it also be possible to test the tricky edge cases? (E.g. logistic function for very large x, which will give a NaN without the custom rule)

GitHub's diff doesn't show it well because I also added the is_primitive flag, but there are two more lines of tests.

penelopeysm:
Yeah, I think logsumexp needs a custom rule 😢 

```julia
import DifferentiationInterface as DI
import ForwardDiff, Mooncake, ReverseDiff
using LogExpFunctions: logsumexp
for adtype in [DI.AutoForwardDiff(), DI.AutoReverseDiff(), DI.AutoMooncakeForward(), DI.AutoMooncake()]
    x = [1.0, 1.0]
    prep = DI.prepare_gradient(logsumexp, adtype, x)
    grad = DI.gradient(logsumexp, prep, adtype, x);
    @show adtype, grad
end
```

```
julia> for adtype in [DI.AutoForwardDiff(), DI.AutoReverseDiff(), DI.AutoMooncakeForward(), DI.AutoMooncake()]
           x = [1.0, 1.0]
           prep = DI.prepare_gradient(logsumexp, adtype, x)
           grad = DI.gradient(logsumexp, prep, adtype, x);
           @show adtype, grad
       end
(adtype, grad) = (AutoForwardDiff(), [0.5, 0.5])
(adtype, grad) = (AutoReverseDiff(), [1.0, 0.0])
(adtype, grad) = (AutoMooncakeForward(), [1.0, 0.0])
(adtype, grad) = (AutoMooncake(), [1.0, 0.0])
```

Technici4n:
Ouch, I am pretty sure that the issue lies in logsumexp itself... Mooncake is computing the correct "selection derivative" of that function 😓 Of course this doesn't match what one would expect from math.

penelopeysm:
@Technici4n am I right in saying that rules for both `Core.kwcall(::NamedTuple, ::typeof(logsumexp), ::AbstractArray)` as well as `logsumexp(::AbstractArray)` need to be implemented? or can we get away with only doing one?

Technici4n:
Ideally you want to implement a rule for a low-level function that is called by the various high-level functions. But I am not sure that this will be possible here :/

We can target the two `logsumexp` calls as you suggest but then `logsumexp!` will still not be supported. One way is to provide rules for the 3 relevant `_logsumexp_onepass_op` methods, but that sounds quite tedious as well.

penelopeysm:
Yeah, implementing rules for `logsumexp` is tedious :(. I don't like depending on underscored methods since they are technically internal, but it might be easier to just target the method of `_logsumexp_onepass_op` that has the branching.

edit: no, we can't do that because `_logsumexp_onepass_op(x, y)` returns `(max(x, y), exp(abs(y - x)))` which is not differentiable at the point we need

penelopeysm:
Forward-mode is a bit slower now, but reverse-mode about 40% faster

```julia
import DifferentiationInterface as DI
import Mooncake
using LogExpFunctions
using Random: Xoshiro
using Chairmarks: @be, median
f, x = logsumexp, randn(Xoshiro(468), 5, 5)
display(median(@be $f($x)))
for adtype in [DI.AutoMooncakeForward(), DI.AutoMooncake()]
    prep = DI.prepare_gradient(f, adtype, x)
    display(median(@be DI.gradient($f, $prep, $adtype, $x)))
end

#= main
121.795 ns
4.535 μs (52 allocs: 6.906 KiB)
547.170 ns (2 allocs: 272 bytes)
=#

#= This PR
121.277 ns
6.865 μs (52 allocs: 6.906 KiB)
291.670 ns (2 allocs: 272 bytes)
=# 
```

penelopeysm:
Sorry to keep pushing, but just for the record, I benchmarked performance for most of the chain rules defined in LogExpFunctionsChainRulesCoreExt, and found that for the most part they are the same as Mooncake's derived rules. So I removed all the imported rules except for the `logistic` and `logaddexp` ones which do provide improved numerical stability.

<details><summary>Benchmarking code (comparison was done on main v0.4.183 vs `b9c10c21e`)</summary>

import DifferentiationInterface as DI
import Mooncake
using LogExpFunctions
using Chairmarks: @be, median

function benchmark(fn, args)
    @info fn
    f(args) = fn(args...)
    ptime = median(@be $f($args))
    print("primal "); display(ptime)
    for adtype in [DI.AutoMooncakeForward(), DI.AutoMooncake()]
        mode = adtype isa DI.AutoMooncakeForward ? "forward" : "reverse"
        prep = DI.prepare_gradient(f, adtype, args)
        grad = DI.gradient(f, prep, adtype, args)
        gtime = median(@be DI.gradient($f, $prep, $adtype, $args))
        print("$mode (res=$grad) "); display(gtime)
    end
end

benchmark(xlogx, [1.5])
benchmark(xlogy, [1.5, 2.5])
benchmark(xlog1py, [1.5, 2.5])
benchmark(xexpx, [1.5])
benchmark(xexpy, [1.5, 2.5])
benchmark(logistic, [1.5])
benchmark(logit, [0.7])
benchmark(logcosh, [1.5])
benchmark(logabssinh, [1.5])
benchmark(log1psq, [1.5])
benchmark(log1pexp, [1.5])
benchmark(log1mexp, [-1.5])
benchmark(log2mexp, [-1.5])
benchmark(logexpm1, [1.5])
benchmark(log1pmx, [1.5])
benchmark(logmxp1, [1.5])
benchmark(logaddexp, [1.5, 2.5])
benchmark(logsubexp, [2.5, 1.5])
benchmark(cloglog, [0.5])
benchmark(cexpexp, [1.5])

</details>

ext/MooncakeLogExpFunctionsExt.jl
```diff
@@ -0,0 +1,126 @@
...
+    logsumexp!(primal(out), primal(x))
+    y, dy = extract(out)
+    dx = tangent(x)
+    function logsumexp!_pb!!(::NoRData)
+        dx .+= dy .* exp.(primal(x) .- y)
+        copyto!(y, old_out)
+        return NoRData(), NoRData(), NoRData()
+    end
+    return out, logsumexp!_pb!!
+end
```
penelopeysm, line 126: please do look at the mutating ones closely since i'm less confident of these 😄 they do give the right result but i could not get finite differences to work nicely with it which makes me feel anxious

test/integration_testing/logexpfunctions/logexpfunctions.jl
```diff
@@ -8,43 +8,61 @@ using Mooncake.TestUtils: test_rule
...
+                    rand(sr(6), Float64, 5),
+                    randn(sr(7), Float32, 5, 4),
+                ),
+                (:none, nothing, false, softmax, randn(sr(7), P, 10)),
+                (:allocs, nothing, false, cloglog, P(0.5)),
+                (:allocs, nothing, false, cexpexp, -P(0.3)),
+                (:allocs, nothing, false, loglogistic, P(0.5)),
+                (:allocs, nothing, false, logitexp, -P(0.3)),
+                (:allocs, nothing, false, log1mlogistic, -P(0.9)),
+                (:allocs, nothing, false, logit1mexp, -P(0.6)),
```
penelopeysm: To be clear, the new test cases are: `logistic(1000.0)` (line 20), `logaddexp(1.5, 1.5)` (line 32), and all the logsumexp ones with non-random inputs (lines 38, 40, 42, 44).

penelopeysm:
(I am done now I think, so will stop adding to this and wait for review)

ext/MooncakeLogExpFunctionsExt.jl
```diff
@@ -0,0 +1,126 @@
...
+    dy = sum(tangent(x) .* (exp.(primal(x) .- y)); primal(kwargs)...)
+    return Dual(y, dy)
+end
+function frule!!(
+    ::Dual{typeof(logsumexp)}, x::Dual{<:AbstractArray{P}}
+) where {P<:IEEEFloat}
+    y = logsumexp(primal(x))
+    dy = zero(P)
+    xp, dx = extract(x)
+    # same as dy = dot(dx, exp.(xp .- y)) but unrolled to avoid allocations
```
Technici4n: Would something like `sum(tangent(x) .* (exp.(primal(x) .- y)))` not work?

ext/MooncakeLogExpFunctionsExt.jl
```diff
@@ -0,0 +1,126 @@
...
+function rrule!!(
+    ::CoDual{typeof(Core.kwcall)},
+    kwargs::CoDual{<:NamedTuple{(:dims,),<:Tuple{Colon}}},
+    ::CoDual{typeof(logsumexp)},
+    x::CoDual{<:AbstractArray{P}},
+) where {P<:IEEEFloat}
+    y = logsumexp(primal(x); primal(kwargs)...)
+    dx = tangent(x)
+    function logsumexp_pb!!(dy::P)
+        dx .+= dy * exp.(primal(x) .- y)
```
Technici4n: you probably want `dy .*` to avoid allocations?

ext/MooncakeLogExpFunctionsExt.jl
```diff
@@ -0,0 +1,126 @@
...
+    # all other dim arguments - i.e. ints or tuples thereof
+    kwargs::CoDual{<:NamedTuple{(:dims,),<:Tuple{Any}}},
+    ::CoDual{typeof(logsumexp)},
+    x::CoDual{<:AbstractArray{P}},
+) where {P<:IEEEFloat}
+    y = logsumexp(primal(x); primal(kwargs)...)
+    dy = zero(y)
+    dx = tangent(x)
+    function logsumexp_pb!!(::NoRData)
+        dx .+= dy .* exp.(primal(x) .- y)
```
Technici4n, line 83: If you use `.*` above the function bodies seem identical.

ext/MooncakeLogExpFunctionsExt.jl
```diff
@@ -0,0 +1,126 @@
...
+    out::CoDual{<:AbstractArray{P}},
+    x::CoDual{<:AbstractArray{P}},
+) where {P<:IEEEFloat}
+    old_out = copy(primal(out))
+    logsumexp!(primal(out), primal(x))
+    y, dy = extract(out)
+    dx = tangent(x)
+    function logsumexp!_pb!!(::NoRData)
+        dx .+= dy .* exp.(primal(x) .- y)
+        copyto!(y, old_out)
```
Technici4n, line 121: I think that you need to zero out `dy` here: if you see the function as going from `(out0, x0) -> (out1 = logsumexp(x0), x1 = x0)`, then the pullback is `(dout0 = 0, dx0 = dx1 + <the actual VJP>)`,

ext/MooncakeLogExpFunctionsExt.jl
Technici4n: Any reason why it didn't work?

ext/MooncakeLogExpFunctionsExt.jl
penelopeysm: yes, thanks!

ext/MooncakeLogExpFunctionsExt.jl
penelopeysm: Not exactly because here `dy` is captured whereas above `dy` is the argument to the pullback :(
Unless you are suggesting factorising that out into a separate function, but for one line i think the benefit is minimal

ext/MooncakeLogExpFunctionsExt.jl
penelopeysm: unfortunately no, benchmarking the gradient for `logsumexp(randn(5, 5))` is

- manual loop: 6.917 μs (52 allocs: 6.906 KiB)
- `sum(tangent(x) .* (exp.(primal(x) .- y))`: 8.653 μs (102 allocs: 13.547 KiB)
- `reduce(+, tangent(x) .* (exp.(primal(x) .- y)))`: 7.431 μs (102 allocs: 13.547 KiB)
- `reduce(+, Iterators.map(t -> t[1] * exp(t[2] - y), zip(dx, xp)))`: 7.708 μs (52 allocs: 6.906 KiB)

(Also tried various combinations of mapreduce which were even worse)

ext/MooncakeLogExpFunctionsExt.jl
penelopeysm: because `dy` wasn't zeroed :)

ext/MooncakeLogExpFunctionsExt.jl
penelopeysm: Thank you, that makes a lot of sense intuitively too - I think the docs could do with some examples of these - might write it up if I find some time...

ext/MooncakeLogExpFunctionsExt.jl
```diff
@@ -0,0 +1,129 @@
...
+) where {P<:IEEEFloat}
+    y = logsumexp(primal(x); primal(kwargs)...)
+    dy = sum(tangent(x) .* (exp.(primal(x) .- y)); primal(kwargs)...)
+    return Dual(y, dy)
+end
+function frule!!(
+    ::Dual{typeof(logsumexp)}, x::Dual{<:AbstractArray{P}}
+) where {P<:IEEEFloat}
+    y = logsumexp(primal(x))
+    dy = zero(P)
```
sunxd3, line 50:
```suggestion
```
seem overwritten by line 53?

penelopeysm: Oh, darn, I committed the wrong version. Sorry.

Technici4n (approved):
Looks good!

penelopeysm:
Thanks 😊 @sunxd3, are you happy too?

sunxd3:
very happy

PR #883, opened 2025-12-11
Add debug mode check for aliased fdata/tangent between input and output

sunxd3:
## Summary

Adds a check in debug mode to detect when an `rrule!!` or `frule!!` returns an output whose tangent/fdata is aliased (`===`) with an input's tangent/fdata. This is a subtle bug that can cause silent incorrect gradients.

## Motivation

In reverse mode, fdata is incremented in-place during the reverse pass. If the output fdata is the same object as an input fdata, modifications during gradient accumulation will incorrectly affect the input, leading to wrong results without any error.

This bug is particularly insidious because:
- It doesn't crash
- It silently produces incorrect gradients
- It can be very hard to debug without this check

## Changes

- Added `verify_tangent_not_aliased` function in `src/debug_mode.jl`
- Integrated the check into both `DebugRRule` (reverse mode) and `DebugFRule` (forward mode)
- Only checks mutable tangents (arrays, `MutableTangent`) since immutable values can't be mutated
- Added tests for both modes

## Note

This check only verifies top-level aliasing. Nested aliasing (e.g., output tangent contains an array that is also in an input tangent) is not checked, as it's more context-dependent.

### DIFF
--- a/src/debug_mode.jl
+++ b/src/debug_mode.jl
@@ -105,6 +105,7 @@ end
     try
         y isa Dual || error("frule!! must return a Dual, got $(typeof(y))")
         verify_dual_value(y)
+        verify_tangent_not_aliased(x, y)
     catch e
         error("Error in outputs of rule with input types $(_typeof(x))")
     end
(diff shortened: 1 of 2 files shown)


sunxd3:
Closing this PR: the aliasing check is too strict and causes false positives on legitimate code paths like `Base.unalias` or `identity`.

Issue #862, opened 2025-11-21
Flaky blas and GPU related CI failures


yebai:
See, eg, 

- [blas_3](https://github.com/chalk-lab/Mooncake.jl/actions/runs/19496944718/job/55801351720#step:6:257)
- [GPU](https://buildkite.com/julialang/mooncake-dot-jl/builds/2257/steps/canvas?jid=019a9c46-ca6e-4633-87d4-6f97c142f06b#019a9c46-ca6e-4633-87d4-6f97c142f06b/153-367)

These are likely due to unlucky random seeds.

PR #884, opened 2025-12-11
Add basic arrayify for CuArray

kshyatt:
I'll add a test soon

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.185"
+version = "0.4.186"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"


yebai:
thanks @kshyatt -- can you help take a look at GPU failures on buildkite? These are obsrved on main branch too.

kshyatt:
@yebai we've seen this on CUDA.jl as well, it's related to a bad LLVM version I think. Retrying the build often resolves it

yebai:
Cc @AstitvaAggarwal who is looking into it to address https://github.com/chalk-lab/Mooncake.jl/issues/862. Is there a way to detect LLVM issue and skip tests?

Technici4n:
Oh, interesting. Is this enough to fix BLAS with CUDA? That would be very nice. I was under the impression that the primal would take a completely different path for CuArrays and so the BLAS rules wouldn't apply.

kshyatt:
We might need to write some custom rules for CUBLAS, as its API isn't 1:1 with BLAS

kshyatt:
I'm testing this implicitly over here (https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/93) though not directly with CUBLAS

Technici4n:
Don't forget to bump the version:)

kshyatt:
Done!

yebai:
thanks @kshyatt!

PR #885, opened 2025-12-11
Add rules for `nextfloat` and `prevfloat`

MasonProtter:
`nextfloat` and `prevfloat` are used a lot in e.g. SciML solvers, but Mooncake trips up on them because these functions do some internal bitcasting: https://github.com/JuliaLang/julia/blob/v1.12.3/base/float.jl#L819

With this PR:
```julia
julia> using Mooncake, DifferentiationInterface

julia> derivative(AutoMooncake(), 1.0) do x
           nextfloat(x) + 1
       end
1.0

julia> derivative(AutoMooncakeForward(), 1.0) do x
           prevfloat(x) + 1
       end
1.0
```
without this PR:
```julia
julia> derivative(AutoMooncakeForward(), 1.0) do x
           nextfloat(x) + 1
       end
ERROR: ArgumentError: It is not permissible to bitcast to a differentiable type during AD, as this risks dropping tangents, and therefore risks silently giving the wrong answer. If this call to bitcast appears as part of the implementation of a differentiable function, you should write a rule for this function, or modify its implementation to avoid the bitcast.
Stacktrace:
  [1] frule!!(f::Mooncake.Dual{…}, t::Mooncake.Dual{…}, x::Mooncake.Dual{…})
    @ Mooncake.IntrinsicsWrappers ~/Nextcloud/Julia/Mooncake/src/rrules/builtins.jl:250
  [2] reinterpret
    @ ./essentials.jl:733 [inlined]
  [3] nextfloat
    @ ./float.jl:852 [inlined]
  [4] nextfloat
    @ ./float.jl:863 [inlined]
  [5] #44
    @ ./REPL[22]:2 [inlined]
  [6] (::Tuple{})(_2::Mooncake.Dual{var"#44#45", Mooncake.NoTangent}, _3::Mooncake.Dual{Float64, Float64})
  [7] (::MistyClosures.MistyClosure{Core.OpaqueClosure{…}})(::Mooncake.Dual{var"#44#45", Mooncake.NoTangent}, ::Mooncake.Dual{Float64, Float64})
    @ MistyClosures ~/.julia/packages/MistyClosures/2vtLL/src/MistyClosures.jl:22
  [8] DerivedFRule
    @ ~/Nextcloud/Julia/Mooncake/src/interpreter/forward_mode.jl:84 [inlined]
  [9] value_and_derivative!!
    @ ~/Nextcloud/Julia/Mooncake/src/interface.jl:610 [inlined]
 [10] #20
    @ ~/.julia/packages/DifferentiationInterface/4n6vR/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl:43 [inlined]
 [11] map
    @ ./tuple.jl:358 [inlined]
 [12] value_and_pushforward
    @ ~/.julia/packages/DifferentiationInterface/4n6vR/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl:40 [inlined]
 [13] pushforward
    @ ~/.julia/packages/DifferentiationInterface/4n6vR/ext/DifferentiationInterfaceMooncakeExt/forward_onearg.jl:67 [inlined]
 [14] derivative(::var"#44#45", ::DifferentiationInterface.PushforwardDerivativePrep{…}, ::AutoMooncakeForward{…}, ::Float64)
    @ DifferentiationInterface ~/.julia/packages/DifferentiationInterface/4n6vR/src/first_order/derivative.jl:200
 [15] derivative(::var"#44#45", ::AutoMooncakeForward{Nothing}, ::Float64)
    @ DifferentiationInterface ~/.julia/packages/DifferentiationInterface/4n6vR/src/first_order/derivative.jl:102
 [16] top-level scope
    @ REPL[22]:1
```

### DIFF
--- a/src/rrules/low_level_maths.jl
+++ b/src/rrules/low_level_maths.jl
@@ -228,7 +228,22 @@ function rrule!!(::CoDual{typeof(Base.eps)}, x::CoDual{P}) where {P<:IEEEFloat}
     return zero_fcodual(y), eps_pb!!
 end
 
+#=
+## nextfloat / prevfloat
+
+The idea is that `nextfloat(x) = x + ϵ`, where `ϵ` is the smallest possible number such that
+`x + ϵ` is not equal to `x`.
+
+Conceptually, this should just be considered as you adding a (tiny) constant to `x`, so
+the derivative is just `1`.  Technically, the size of  `ϵ` does depend on `x`, but more or
+less by construction, the derivative should be too small to represent using whatever floating
+point system you are using.
+
+Therefore, we take `nextfloat` and `prevfloat` to just be the identity function so far as
+derivatives are concerned
+=#
 @is_primitive MinimalCtx Tuple{typeof(nextfloat),<:IEEEFloat}
+@is_primitive MinimalCtx Tuple{typeof(prevfloat),<:IEEEFloat}
 function frule!!(::Dual{typeof(nextfloat)}, x::Dual{<:IEEEFloat})
     return Dual(nextfloat(primal(x)), tangent(x))
 end
@@ -237,8 +252,6 @@ function rrule!!(::CoDual{typeof(nextfloat)}, x::CoDual{P}) where {P<:IEEEFloat}
     nextfloat_pb!!(dy::P) = (NoRData(), dy)
     return zero_fcodual(y), nextfloat_pb!!
 end
-
-@is_primitive MinimalCtx Tuple{typeof(prevfloat),<:IEEEFloat}
 function frule!!(::Dual{typeof(prevfloat)}, x::Dual{<:IEEEFloat})
     return Dual(prevfloat(primal(x)), tangent(x))
 end


Technici4n:
I wonder if these rules would be incorrect very close to max float. Are you sure that this is the right thing to do? These operations don't look very differentiable ^^

MasonProtter:
I would argue that this is the only sensible meaning for these functions. The idea is that `nextfloat(x) = x + ϵ`, where `ϵ` is the smallest possible number such that `x + ϵ` is not equal to `x`. 

Conceptually, this should just be considered as you adding a (tiny) constant to `x`, so the derivative is just `1`.  Technically, the size of  `ϵ` does depend on `x`, but more or less by construction, the derivative should be too small to represent using whatever floating point system you are using.

MasonProtter:
As an additional reference, this is the same as what ForwardDiff.jl does: 
https://github.com/JuliaDiff/ForwardDiff.jl/blob/a7991dd592263aeb7197afc6e5c9b3bd7e950503/src/dual.jl#L318-L324

Technici4n:
Thanks!

Issue #526, opened 2025-03-16
Perfomance rule for `kron`.


yebai:
`kron` is the most underperforming benchmark function and requires a performance rule. The following Zygote rule can be adapted for Mooncake 

https://github.com/FluxML/Zygote.jl/blob/c0bfc26de90db3778e1f3953f04d489e3413127b/src/lib/array.jl#L358-L374

penelopeysm:
Naive rules are slower than letting Mooncake differentiate through but maybe there's a way to optimise this, I haven't tried very hard.

Edit: Ooh, hold on, using `dot` and `@view` made it a lot better. Gradient on `x -> sum(kron(x, x))` for `x = randn(5, 5)` went down from 9 µs with no rule to 3 µs with this rule.

```julia
@is_primitive DefaultCtx Tuple{typeof(kron),AbstractMatrix{T},AbstractMatrix{T}} where {T<:Base.IEEEFloat}
function Mooncake.frule!!(
    ::Dual{typeof(kron)},
    x1::Dual{<:AbstractMatrix{<:T}},
    x2::Dual{<:AbstractMatrix{<:T}}) where {T<:Base.IEEEFloat}
    px1, dx1 = extract(x1)
    px2, dx2 = extract(x2)
    y = kron(px1, px2)
    dy = kron(dx1, px2) .+ kron(px1, dx2)
    return Dual(y, dy)
end
function Mooncake.rrule!!(
    ::CoDual{typeof(kron)},
    x1::CoDual{<:AbstractMatrix{<:T}},
    x2::CoDual{<:AbstractMatrix{<:T}}) where {T<:Base.IEEEFloat}
    px1, dx1 = extract(x1)
    px2, dx2 = extract(x2)
    y = kron(px1, px2)
    dy = zero(y)
    function kron_pb!!(::NoRData)
        M, N = size(dx1)
        P, Q = size(dx2)
        for m in 1:M, n in 1:N
            dx1[m, n] += dot((@view dy[(m-1)*P+1:m*P, (n-1)*Q+1:n*Q]), px2)
        end
        for p in 1:P, q in 1:Q
            dx2[p, q] += dot((@view dy[p:P:end, q:Q:end]), px1)
        end
        return NoRData(), NoRData(), NoRData()
    end
    return CoDual(y, dy), kron_pb!!
end

using Random
Mooncake.TestUtils.test_rule(Xoshiro(468), kron, randn(2, 2), randn(2, 2); is_primitive=true)
```

Issue #156, opened 2024-05-15
Document Loop Optimisation Opportunities


willtebbutt:

willtebbutt:
I still need to add more concrete info about what loop optimisations are possible, but here's a summary of the state of affairs currently:

- `map`, `broadcast`, `mapreduce`, and any other higher-order functions I’ve forgotten about, all lower to loops in the CFG. Tapir doesn’t have rules for them, so Tapir sees these loops.
- Loops in Tapir are reasonably performant. For example, they are completely type-stable and are (usually) allocation-free. So if you’re doing a bit of work at each iteration (e.g. `sin(cos(exp(x[n])))` ), the time spent managing “overhead” associated to looping (e.g. logging stuff on the forwards pass at each iteration which you need on the reverse-pass) is small in comparison to the time spent doing the work that you care about (e.g. computing sin(cos(exp(x))) and doing AD on each operation in it etc)
- If you’re doing a very small amount of work at each iteration of a loop, then your computation is (currently) dominated by “overhead”. `sum` is an extreme case of this, because adding two `Float64`s together at each iteration is about the cheapest differentiable operation that you could imagine doing. Moreover, the current way that we handle looping in Tapir “gets in the way” of vectorisation (on the forwards-pass and reverse-pass).
- The loop optimisations that I will discuss in the issue will largely target this overhead. They will therefore improve the performance of every single example in this table, but the largest improvements will be seen for `kron` and `sum`. I imagine they’ll be especially great on `sum` as they should be able to “get out of the way” of vectorisation (i.e. things should vectorise nicely).

That we just rely on everything boiling down to the same kind of looping structure in the CFG is a great advantage of this approach -- basically everything CPU-based that’s performant gets reduced to a loop in the CFG (specifically, a thing called a “Natural Loop” in compiler optimisation terminology). There are well-established optimisation strategies for loops, so we don’t need to implement separate rules for all the different higher-order functions to get good performance, nor do we need to tell people to steer clear of writing for or while loops.
Rather, we just optimise these so-called “natural loop” structures which appear in the CFG, and then everything (or, rather, most things) will (should) be fast.

(The situation in which this strategy breaks down is if people use `@goto` to produce certain kinds of “weird” looping structures. Such structures will only ever be as performant as they are currently. Frankly, it’s not bad, but we should probably discourage people from using `@goto` , which is definitely something that I can live with)

willtebbutt:
[...]

PR #886, opened 2025-12-11
Add rule for `LinearAlgebra._kron!`

penelopeysm:
## Matrix–matrix

This PR implements a rule for `LinearAlgebra._kron!(::Matrix, ::Matrix, ::Matrix)`. Here are benchmarks for matrix-matrix kron with two 25x25 inputs, `f1` uses the true `LinearAlgebra.kron` whereas `f2` uses a redefinition of it to force Mooncake to derive a rule.

```julia
using DifferentiationInterface, Mooncake, Chairmarks, LinearAlgebra, Random

Mooncake.TestUtils.test_rule(Xoshiro(468), LinearAlgebra._kron!, zeros(20, 20), rand(4, 4), rand(5, 5); is_primitive=true)

function _mykron!(C, A::AbstractMatrix, B::AbstractMatrix)
    m = firstindex(C)
    @inbounds for j in axes(A,2), l in axes(B,2), i in axes(A,1)
        Aij = A[i,j]
        for k in axes(B,1)
            C[m] = Aij*B[k,l]
            m += 1
        end
    end
    return C
end;
function mykron(A::AbstractVecOrMat{S}, B::AbstractVecOrMat{S}) where {S}
    C = Matrix{S}(undef, LinearAlgebra._kronsize(A, B)...)
    return _mykron!(C, A, B)
end;
f1(x) = sum(kron(x, x));
f2(x) = sum(mykron(x, x));
x = rand(Xoshiro(468), 25, 25);

for backend in [AutoMooncakeForward(), AutoMooncake()]
    prep1 = prepare_gradient(f1, backend, x);
    grad1 = gradient(f1, prep1, backend, x);
    @info "$backend LinearAlgebra.kron"
    display(@be gradient(f1, prep1, backend, x))
    prep2 = prepare_gradient(f2, backend, x);
    grad2 = gradient(f2, prep2, backend, x);
    @info "$backend mykron"
    display(@be gradient(f2, prep2, backend, x))
    @assert isapprox(grad1, grad2)
end
```

```
[ Info: AutoMooncakeForward() LinearAlgebra.kron
Benchmark: 1 sample with 1 evaluation
        641.764 ms (5632 allocs: 3.668 GiB, 23.55% gc time, without a warmup)
[ Info: AutoMooncakeForward() mykron
Benchmark: 1 sample with 1 evaluation
        769.250 ms (5632 allocs: 3.668 GiB, 23.23% gc time, without a warmup)

[ Info: AutoMooncake() LinearAlgebra.kron
Benchmark: 33 samples with 1 evaluation
 min    1.720 ms (13 allocs: 9.005 MiB)
 median 2.341 ms (13 allocs: 9.005 MiB)
 mean   3.065 ms (13 allocs: 9.005 MiB, 19.58% gc time)
 max    10.296 ms (13 allocs: 9.005 MiB, 79.64% gc time)
[ Info: AutoMooncake() mykron
Benchmark: 15 samples with 1 evaluation
 min    4.765 ms (10 allocs: 6.005 MiB)
 median 5.255 ms (10 allocs: 6.005 MiB)
 mean   6.732 ms (10 allocs: 6.005 MiB, 8.45% gc time)
 max    22.176 ms (10 allocs: 6.005 MiB, 76.31% gc time)
```

## Other combinations

Instead of repeating the code for the vector–matrix and matrix–vector combinations, here we use an overlay so that the matrix–matrix rule gets used for all of them. This gives about a 2x improvement over the derived rule. The code is given here so you can run it, but the results aren't attached here.

I benchmarked the overlay vs. a handwritten rule (not shown here) and there wasn't much appreciable difference (the overlay has 2 more allocs than the non-overlay, timings are the same to within benchmark noise).

<details><summary>Benchmark code for vector–matrix and matrix–vector</summary>

```julia
function _mykron!(C, A::AbstractMatrix, b::AbstractVector)
    m = firstindex(C)
    @inbounds for j in axes(A,2), i in axes(A,1)
        Aij = A[i,j]
        for k in eachindex(b)
            C[m] = Aij*b[k]
            m += 1
        end
    end
    return C
end;
function _mykron!(C, a::AbstractVector, B::AbstractMatrix)
    m = firstindex(C)
    @inbounds for l in axes(B,2), i in eachindex(a)
        ai = a[i]
        for k in axes(B,1)
            C[m] = ai*B[k,l]
            m += 1
        end
    end
    return C
end;
f3(x) = sum(kron(x, x[1,:]));
f4(x) = sum(mykron(x, x[1,:]));
f5(x) = sum(kron(x[1,:], x));
f6(x) = sum(mykron(x[1,:], x));

for backend in [AutoMooncakeForward(), AutoMooncake()]
    prep3 = prepare_gradient(f3, backend, x);
    grad3 = gradient(f3, prep3, backend, x);
    @info "$backend LinearAlgebra.kron M/V"
    display(@be gradient(f3, prep3, backend, x))
    prep4 = prepare_gradient(f4, backend, x);
    grad4 = gradient(f4, prep4, backend, x);
    @info "$backend mykron M/V"
    display(@be gradient(f4, prep4, backend, x))
    @assert isapprox(grad3, grad4)
    prep5 = prepare_gradient(f5, backend, x);
    grad5 = gradient(f5, prep5, backend, x);
    @info "$backend LinearAlgebra.kron V/M"
    display(@be gradient(f5, prep5, backend, x))
    prep6 = prepare_gradient(f6, backend, x);
    grad6 = gradient(f6, prep6, backend, x);
    @info "$backend mykron V/M"
    display(@be gradient(f6, prep6, backend, x))
    @assert isapprox(grad3, grad6)
end
```

</details>

## Should `kron` itself have a rule?

As can be seen from the benchmarks, using the rule above for `_kron!` does lead to a nice performance boost for `kron`, but in reverse mode there are still extra allocations compared to Mooncake's derived rule. The only way of getting rid of these is to write a rule for `kron` itself as well, and that brings the gradient time down by another 50%, from 2.3 ms to 1.6 ms. The speedup scales with matrix size (i.e. it's around 50% at all sizes). That is actually quite a substantial improvement so I've elected to leave this rule in.

Closes https://github.com/chalk-lab/Mooncake.jl/issues/526

### DIFF
--- a/src/Mooncake.jl
+++ b/src/Mooncake.jl
@@ -151,12 +151,8 @@ else
     include(joinpath("rrules", "array_legacy.jl"))
 end
 
-end # DispatchDoctor block
-
 # Including this in DispatchDoctor causes precompilation error. 
-include(joinpath("rrules", "performance_patches.jl"))
-
-@stable default_mode = "disable" default_union_limit = 2 begin
+@unstable include(joinpath("rrules", "performance_patches.jl"))
 
 include("interface.jl")
 include("config.jl")


Technici4n:
Did you look into why the rule generated by Mooncake is slow?

penelopeysm:
I think it's because `kron!` is implemented with nested loops, and https://github.com/chalk-lab/Mooncake.jl/issues/156 describes the overhead associated with these.

Technici4n:
I see. How relevant is it to special case `kron` in the meanwhile, vs waiting for a more general fix (although that will probably require significant effort)?

penelopeysm:
That's possibly subjective and certainly hard to answer since I don't know how widely kron is used :-) they are in the CI benchmarks but that obviously does not on its own mean that it is widely used.
I am definitely quite wary of dumping another 4 sets of rules in here just for this, and am pretty happy to leave this PR here until there's some consensus on whether it's worth adding a new rule. (assuming it even gets to a mergeable state)

penelopeysm:
(I came across this earlier, which is relevant: https://github.com/chalk-lab/Mooncake.jl/issues/249#issuecomment-2633182716)

penelopeysm:
> But I just want to see the effects on the CI benchmark.

Okay, so for `kron_sum` which has arguments of `randn(20, 20), randn(40, 40)`, that gives about 4-5x improvement (depending on which PR you look at since benchmarks are a bit flaky). Doesn't work for `kron_view_sum` because the rule is restricted to `Matrix` and not `AbstractMatrix`.

yebai:
Happy to support having rules for kron on the grounds that it is a standard linear algebra operation. It’s unlikely that the loop optimisation will be implemented, as the gain would be minimal (only benefiting extremely cheap loops) compared with the substantial added complexity.

src/test_resources.jl
```diff
@@ -881,7 +881,7 @@ function generate_test_functions()
         ),
         (
             false,
-            :allocs,
+            :none,
             nothing,
             LinearAlgebra._kron!,
             randn(25, 25),
```
penelopeysm, line 887: so apparently Mooncake's derived rule is allocation-free; but I don't know how to recover this property, because there will always be allocs coming from copying the old primal (in fact I checked and removing the line to `copy(pout)` does indeed make it allocation-free, but obviously incorrect). Hence I had to relax this. Suggestions welcome

yebai: See, https://github.com/chalk-lab/Mooncake.jl/issues/403

yebai:
See, https://github.com/MilesCranmer/DispatchDoctor.jl/issues/101 for the DispatchDoctor issue.

src/rrules/performance_patches.jl
```diff
@@ -46,23 +46,130 @@ function rrule!!(
...
+        for p in axes(px2, 1), q in axes(px2, 2)
+            dx2[p, q] += dot((@view dout[p:P:end, q:Q:end]), px1)
+        end
+        copyto!(pout, old_pout)
+        fill!(dout, zero(T))
+        return NoRData(), NoRData(), NoRData(), NoRData()
+    end
+    return out, _kron!_pb!!
+end
+@mooncake_overlay function LinearAlgebra._kron!(
```
MilesCranmer, line 104:
```suggestion
@unstable @mooncake_overlay function LinearAlgebra._kron!(
```

src/rrules/performance_patches.jl
```diff
@@ -46,23 +46,130 @@ function rrule!!(
...
+    out::Matrix{T}, a::Vector{T}, b::Matrix{T}
+) where {T<:IEEEFloat}
+    return LinearAlgebra._kron!(out, reshape(a, :, 1), b)::Matrix{T}
+end
+@mooncake_overlay function LinearAlgebra._kron!(
```
MilesCranmer, line 109:
```suggestion
@unstable @mooncake_overlay function LinearAlgebra._kron!(
```

src/rrules/performance_patches.jl
MilesCranmer, line 1: And add

```julia
using DispatchDoctor: @unstable
```

to the top of the file.

MilesCranmer (commented):
Here's how to fix the DispatchDoctor.jl issue - given in comments below. It seems the `@mooncake_overlay` is simply not yet compatible with the `@stable` macro when enabled in warn/error mode for the integration tests.

yebai:
I am involved in this PR so better to have a second perspective.

src/Mooncake.jl
```diff
@@ -150,8 +150,14 @@ include(joinpath("rrules", "twice_precision.jl"))
...
     include(joinpath("rrules", "array_legacy.jl"))
 end
+
+end # DispatchDoctor block
+
+# Including this in DispatchDoctor causes precompilation error. 
 include(joinpath("rrules", "performance_patches.jl"))
 
+@stable default_mode = "disable" default_union_limit = 2 begin
+
```
MilesCranmer, line 156:
```suggestion

# Including this in DispatchDoctor causes precompilation error. 
@unstable include(joinpath("rrules", "performance_patches.jl"))

```

Then the macro settings don't need to be duplicated. Also the same pattern is already done for the lines

```julia
@unstable include("test_utils.jl")
@unstable include("test_resources.jl")
```

and everything under `interpreter/`

src/rrules/blas.jl
```diff
@@ -60,6 +62,12 @@ function arrayify(x::Base.ReinterpretArray{T}, dx::TangentOrFData) where {T<:Bla
     _, _dx = arrayify(x.parent, _fields(dx).parent)
     return x, reinterpret(T, _dx)
 end
```
sunxd3, line 64: looking at the diff, should we update these `BlasFloat` to `IEEEFloat` too?

yebai: `BlasFloat` includes complex numbers. Therefore, using `Union{IEEEFloat, BlasFloat}` is more accurate. It would be better to replace occurrences of `BlasFloat` with `Union{IEEEFloat, BlasFloat}` (or `Union{Float16, BlasFloat}` where appropriate) systematically in a separate PR.

yebai: `BlasFloat` includes complex numbers. Therefore, using `Union{IEEEFloat, BlasFloat}` is more accurate. It would be better to replace occurrences of `BlasFloat` with `Union{IEEEFloat, BlasFloat}` (or `Union{Float16, BlasFloat}` where appropriate) systematically in a separate PR.

PR #896, opened 2025-12-17
Remove disclaimer from Forwards mode docs

MasonProtter:
The docs still claim that Forwards Mode hasn't been implemented yet, so this message should be removed. Not sure if there's any other things in this document that need updating now that it actually exists


Technici4n:
Do you want to also go through the rest of the doc and do a quick update pass? Given that the rest is also somewhat outdated.

MasonProtter:
Ah okay, I don't think I understand the design well enough to know which stuff is outdated or not.

PR #801, opened 2025-10-07
Support Memory in BLAS

willtebbutt:
Resolves #787 .

PR #898, opened 2025-12-18
frule testing fix

AstitvaAggarwal:
Fixes part 1 (ie, Blas-3 ) of issue [#862](https://github.com/chalk-lab/Mooncake.jl/issues/862)

### _To reproduce issue:_ 
Run on Julia v1.12.1/v1.11.7 (I've tried it mainly on v1.12.1 till now) :  [Call](https://github.com/chalk-lab/Mooncake.jl/blob/bd1e5fe42778140dec3ce64277684ae1be10d788/src/rrules/blas.jl#L1822) via the blas-3 [testset](https://github.com/chalk-lab/Mooncake.jl/blob/main/test/rrules/blas_level_3.jl).  (note - that you would only want to run the tests for [`Float32`](https://github.com/chalk-lab/Mooncake.jl/blob/bd1e5fe42778140dec3ce64277684ae1be10d788/src/rrules/blas.jl#L1825) as they have precision upto only 1e-7) 

MWE: 
```


function Mooncake.derived_rule_test_cases(rng_ctor, ::Val{:blas_level_3})
    t_flags = ['N', 'T', 'C']
    aliased_gemm! = (tA, tB, a, b, A, C) -> BLAS.gemm!(tA, tB, a, A, A, b, C)
    realPs = [Float32]
    # tests error out only with Float32
    # , Float64] 
    Ps = [realPs..., complex.(realPs)...]
    rng = rng_ctor(123)

    test_cases = vcat(

        #
        # BLAS LEVEL 3
        #

        # aliased gemm!
        map_prod(t_flags, t_flags, Ps) do (tA, tB, P)
            As = blas_matrices(rng, P, 5, 5)
            Bs = blas_matrices(rng, P, 5, 5)
            a = randn(rng, P)
            b = randn(rng, P)
            return map_prod(As, Bs) do (A, B)
                (false, :none, nothing, aliased_gemm!, tA, tB, a, b, A, B)
            end
        end...,
    )
    memory = Any[]
    return test_cases, memory
end


rng_ctor = StableRNGs.StableRNG
mode = Mooncake.ForwardMode
random_seed = StableRNGs.LehmerRNG(state=0x000000000000000000000000000000f7)

test_cases, memory = Mooncake.TestUtils.test_hook(Mooncake.derived_rule_test_cases, rng_ctor, Val(:blas_level_3), Mooncake.ForwardMode) do
      Mooncake.derived_rule_test_cases(rng_ctor, Val(:blas_level_3))
 end

GC.@preserve memory @testset "$mode, $f, $(typeof(x))" for (
    interface_only, perf_flag, _, f, x...
) in test_cases
    Mooncake.TestUtils.test_rule(
        random_seed, f, x...; interface_only, perf_flag, is_primitive=false, mode
    )
end

```

### Julia, machine version info
```

julia> versioninfo()
Julia Version 1.12.1
Commit ba1e628ee4 (2025-10-17 13:02 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 8 × 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz   
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, tigerlake)
  GC: Built with stock GC
Threads: 4 default, 1 interactive, 4 GC (on 8 virtual cores)
Environment:
  JULIA_EDITOR = code
  JULIA_VSCODE_REPL = 1
  JULIA_NUM_THREADS = 4


```

### _More about the issue:_ 
The current `Mooncake.test_frule_correctness` was taking the JVP's projection across an unnormalized random vector for testing AD vs FD results.

 
- This leads to a larger error's propagation while testing using `isapprox` with `atol=1e-3` and `rtol=1e-3` as compared to when we would normalize the random vector before taking the JVP's projection on it.

- The tests check the following :  `isapprox(<y_ad, y_rand> + <x_ad, x_rand> ,  <y_fd, y_rand> + <x_fd, x_rand>) `.

- The problem arises with the JVP `y_fd` - this has the taylor remainder error + floating point truncation error.
The current epsilon choices are "enough" at handling this.  (I say "enough" as one would normally expect epsilon to be `cbrt(eps(T))` where `T=eltype(x)`, this is a derived heuristic for the optimal epsilon for minimizing total error)

- However, taking the dot product with the unnormalized random vector `y_rand` scales the total error randomly to more than the expected relative and absolute errors.
 
- Therefore one must normalize the random vectors, so that we are testing within the valid (within theoretical worst case) error bounds.

**The PR handles this change**

### DIFF
--- a/docs/src/utilities/debugging_and_mwes.md
+++ b/docs/src/utilities/debugging_and_mwes.md
@@ -33,6 +33,49 @@ will error.
 
 In any case, the point here is that `Mooncake.TestUtils.test_rule` provides a convenient way to produce and report an error.
 
+If you have a specific set of arguments that are causing issues, you can test them directly:
+```julia
+using Random
+rng = Xoshiro(123)
+Mooncake.TestUtils.test_rule(rng, sin, 5.0)
+```
+
+When debugging, it might be helpful to set the `interface_only=true` to skip the correctness tests and just check that the rule runs without error:
+```julia
+Mooncake.TestUtils.test_rule(rng, sin, 5.0; interface_only=true)
+```
+
+## Manually Running a Rule
+
+For more fine-grained debugging, you can manually run `rrule!!` to inspect intermediate values.
+Here's an example that differentiates a simple function:
+
+```julia
+using Mooncake: rrule!!, zero_fcodual
+
+x = 5.0
+
+# Run the forward pass - returns output CoDual and pullback
+# `zero_fcodual(x)` is equivalent to `CoDual(x, fdata(zero_tangent(x)))`.
+y, pb!! = rrule!!(zero_fcodual(sin), zero_fcodual(x))
+
+# Seed gradient for scalar output
+dy = 1.0
+
+# Run reverse pass - returns input cotangent/adjoint dx
+# Note: for scalar outputs, the adjoint is a plain scalar and `pb!!(1.0)` is sufficient.
+# More general patterns involving zero_rdata / increment!! apply to non-scalar outputs.
+_, dx = pb!!(dy)
+
+# The gradient should be cos(5.0) ≈ 0.28366
+isapprox(dx, cos(5.0))
+```
+
+This approach lets you:
+- Inspect the output of the forward pass `y` and `pb!!` before running the reverse pass
+- Set custom seed gradients for the output `dy`
+- Examine the computed gradient `dx` in detail
+
 ## Segfaults
 
 These are everyone's least favourite kind of problem, and they should be _extremely_ rare in Mooncake.jl.


AstitvaAggarwal:
I'm avoiding formatting as of now for reviews.

yebai:
To clarify, does this fix #862?

AstitvaAggarwal:
@yebai Blas-3 part of it, yes

Technici4n:
Hi, do you mean to target `Astitva/CI_issues` rather than `main`? You also have formatting issues.

Nice catch for the issue, indeed that check is quite suspicious. By adding a bit of printing, the issue becomes obvious:
```
_dot(ȳ, ẏ_fd) = 26.04294976964593
_dot(x̄, ẋ_fd) = -27.11828360380605
_dot(ȳ, ẏ_ad) = 26.04381544701755
_dot(x̄, ẋ_ad) = -27.11760022013914
```
i.e. we have cancellation between the x and y dot products such that the rtol check unfortunately fails.

I think that checking x and y separately would be a nicer fix, since normalizing by the norm of the cotangent isn't guaranteed to fix this sort of cancellation.

src/test_utils.jl
```diff
@@ -498,19 +498,26 @@ function test_frule_correctness(
...
     isapprox_results = map(fd_results) do result
         ẏ_fd, ẋ_fd = result
         return isapprox(
-            _dot(ȳ, ẏ_fd) + _dot(x̄, ẋ_fd),
-            _dot(ȳ, ẏ_ad) + _dot(x̄, ẋ_ad);
+            _dot(ȳnew, ẏ_fd) + _dot(x̄new, ẋ_fd),
+            _dot(ȳnew, ẏ_ad) + _dot(x̄new, ẋ_ad);
             rtol=rtol,
             atol=atol,
         )
```
Technici4n, line 508: Can we instead have a check like
```jl
        return (
            isapprox(_dot(x̄, ẋ_fd), _dot(x̄, ẋ_ad); rtol=rtol, atol=atol)
            && isapprox(_dot(ȳ, ẏ_fd), _dot(ȳ, ẏ_ad); rtol=rtol, atol=atol)
        )
```
In that case the normalization won't be needed, and it seems a lot more robust to be testing in this way.

AstitvaAggarwal: Thanks for the reviews. I was thinking about this as well, however the summation actually ends up comparing the entire JVP Arrays. The Jacobian contains a functional map that is describing the changes in `y` - the function's output and `x'` - mutation in input `x`  wrt to changes in input `x` . 

- So we are testing the system `f : X -> X * Y`.  Here the Jacobian will be `[ dy/dx ; dx'/dx ]` & taking the dot product with input change `[ ẋ ]` we get a JVP Array. 

- We can now compare the AD vs FD JVPs using a random probe `[ ȳ ; x̄ ]`  via the dot products (this would be a random normalized direction in the jacobian output space). 

- This dot product ends up being the scalar sum :  `<y, y_rand> + <x, x_rand>`. Effectively giving us a good similarity score in whatever the Jacobian's space is. Seperately comparing the dot products for `y` and `x` will actually end up being a not as good score as it's essentially the `L1` Norm.

The `L1` norm would have been nice as well, however it might be too tight. let me know what you think

Technici4n: I don't think there is an L1 norm here. I think that what I am suggesting can be seen as using probes `[ 0 ; x̄ ]` and `[ ȳ ; 0 ]`.

Technici4n: BTW I would be curious to just try it and see if any test seems to fail. If not then it's probably correct.

AstitvaAggarwal: It's not exactly incorrect as it is testing instead of  `[y ; x] * JVP_diff = 0` , just ` ( [y ; 0] + [0 ; x] ) * JVP_diff = 0` =>  `[y ; 0] * JVP_diff = 0` && `[0 ; x] * JVP_diff = 0`  and this does pass the tests. 

However, I still prefer the combined projection. In cases where the mutation `x'` is very small (<`1e-7`), the FD error can easily exceed the `rtol` for that specific variable (eg: stiff domains), causing the whole test to fail only due to x (inside the & conditional to fail).
By testing only the sum, we ensure the total JVP vector is correct relative to the scale of the entire system preventing the CI from failing over insignificant numerical errors in small scale variables.

Technici4n: For small x the `atol` should still take care of it.

Note that `[y ; 0] * JVP_diff = 0` && `[0 ; x] * JVP_diff = 0` implies that `[y ; x] * JVP_diff = 0` but not the other way around.

AstitvaAggarwal: That will require a new atol choice for every test with a stiff problem unfortunately. The current set up handles that sort of thing implicitly. 
I don't see the exact drawback with the current normalization though?

yebai: @Technici4n normalisation seems indeed a bit more principled here -- maybe I am missing anything here?

src/test_utils.jl
```diff
@@ -498,19 +498,31 @@ function test_frule_correctness(
     # such that AD and central differences agree on the answer.
     x̄ = map(Base.Fix1(randn_tangent, rng), x_primal)
     ȳ = randn_tangent(rng, y_primal)
```
yebai, line 501: ~~Since `randn_tangent` is intended for testing, we could move the normalization logic into `randn_tangent`. This would also automatically fix the GPU test failures in #899.~~

src/test_utils.jl
```diff
@@ -498,19 +498,31 @@ function test_frule_correctness(
...
+    # normalize the random probe vectors for the JVP AD & FD results.
+    # so that errors accumulated due to rng drift are minimised -
+    # while testing with isapprox()'s reltol and abstol = 1e-3.
+    total_norm_1, total_norm_2 = sqrt(_dot(x̄, x̄)), sqrt(_dot(ȳ, ȳ))
+
+    # scaled probe vectors handling div by zero case.
+    x̄_normdir, ȳ_normdir = _scale(iszero(total_norm_1) ? 1.0 : 1 / total_norm_1, x̄),
+    _scale(iszero(total_norm_2) ? 1.0 : 1 / total_norm_2, ȳ)
```
sunxd3: does data type matter in these tests? if yes, maybe use `one` and `inv`?

Technici4n: Tangent norms, scalings, etc should always be Float64

sunxd3: Make sense

yebai:
> I still would like to see the test split over x and y as it will make testing more tight and thus more robust. Let's try it and see if any test fails!

It’s worth noting that normalization does not loosen the test error bounds. Instead, the norm of the tangents grows proportionally to sqrt(N), since each element is sampled from N(0,1). As a result, when testing large vectors or matrices, the JVP error also grows with dimensionality. Normalization therefore allows us to use the same atol across a broader range of tests. If evidence shows that the post-normalization error bound is is loose, we can reduce atol.

yebai:
> You should introduce a new normalize(t) function as that might be useful in other places.

That’s a valid point. Since the same normalization is already used when testing the reverse rule (see [here](https://github.com/chalk-lab/Mooncake.jl/blob/5daa968fb1e01a8bea563cf68f2d75c889410a41/src/test_utils.jl#L542-L545)), it maybe useful add a utility to normalize these tangents for both forward and reverse model tests.

yebai:
I’ll go ahead and merge this, since the PR improves on main and fixes some test failures. Please feel free to open new PRs if better test metrics are identified.

Technici4n:
Thanks @AstitvaAggarwal for this, this should also unblock https://github.com/chalk-lab/Mooncake.jl/pull/801.

AstitvaAggarwal:
Thank you for the thorough reviews as well @Technici4n !

PR #899, opened 2025-12-19
Fix for CI CUDA tests

AstitvaAggarwal:
@yebai This looks like a mooncake bug, similar to the `BLAS` issue around random tangent generation.

This PR closes issue #862 and any related downstream issues.

### TD;LR:

This only occurs for `CuArray{ComplexF32}` because its numerical resolution is much poorer than `CuArray{ComplexF64}`. As a result, `isapprox` can fail due to the default large `rtol`, since `atol` is not explicitly set.

### Details: 

So the error here is due to the randomness involved in instantiating any new  `CuArray` via `undef`.
Sometimes, when _the stars align_ we might suddenly have an unlucky allocated `CuArray` `p` [instantiated](https://github.com/chalk-lab/Mooncake.jl/blob/df7e0dac9859c9213aef75e7dddb90b8065abf8b/test/ext/cuda/cuda.jl#L12) with very large values.

- Sometimes the GC hasnt cleaned up the allocated memory where our `CuArray` points to.
 Because of how Floating Point math works, a 'random' sequence of bits usually get interpreted as a very large number.
 
- In the `CUDA` extension tests, we call `Mooncake.TestUtils.test_tangent_interface` which numerically **tests** if  `p` and the final incremented value (`addition_p&r`) of `p` with `r` (a **non zero** random AD tangent of p) are unequal.
 
- Now, [calls](https://github.com/chalk-lab/Mooncake.jl/blob/df7e0dac9859c9213aef75e7dddb90b8065abf8b/ext/MooncakeCUDAExt.jl#L98) to `isapprox` silently give false positives when called as `isapprox(p , addition_p&r)` when `p` and `addition_p&r` are extremely large Base 10 numbers.
 
- The default kwarg rtol is insufficient (Floating point math). So we must explicitly set a fixed atol.

- MWE for this Issue : 
```
using StableRNGs: StableRNG
using Mooncake, CUDA

p = CuArray{ComplexF32, 2, CUDA.DeviceMemory}([3.88542f24+1.48321im 2.63392f-16-1.78286im -3.08188f-16+1.88441im
                                               8.2469f-17+1.67066im 1.211f28+1.69669im 4.3797f-24+1.83554im
                                               -1.76716f22-1.73316im -5.75034f30+1.87852im -2.76408f-12+1.55925im
                                               -2.4716f38+1.81721im -6.03118f-6-1.65717im -3.41957f27-1.76097im
                                               -2.27608f22+1.46884im -5.94271f8+1.57203im 0.000921748-1.6794im
                                               6.39649f13+1.86143im 1.11946f-37-1.17242im 2.77047f-25-1.75849im
                                               -127.944-1.80325im 3.18285f-39-1.50414im 0.123182+1.84845im
                                               0.278716+1.67517im 3041.28-1.92769im 9.33883f35+1.7806im])

# fails due to internal isapprox call giving a false positive.
Mooncake.test_tangent_interface(StableRNG(123456), p; interface_only=false)

# returns true
isapprox(p , p.+1.0f0)

# returns false
isapprox(p , p.+1.0f0; abs=1e-3)

```
For all random seeds, random tangents `r`, Julia versions - the above tests will always error out.

### DIFF
--- a/test/ext/cuda/cuda.jl
+++ b/test/ext/cuda/cuda.jl
@@ -9,7 +9,8 @@ using Mooncake.TestUtils: test_tangent_interface, test_tangent_splitting, test_r
     if CUDA.functional()
         # Check we can operate on CuArrays of various element types.
         @testset for ET in (Float32, Float64, ComplexF32, ComplexF64)
-            # Use undef to ensure robustness against "garbage" memory (NaNs, Infs, subnormals etc).
+            # Use `undef` to test against garbage memory (NaNs, Infs, subnormals).
+            # `randn` generates well-behaved values and can miss these edge cases.
             p = CuArray{ET,2,CUDA.DeviceMemory}(undef, 8, 8)
             test_tangent_interface(StableRNG(123456), p; interface_only=false)
             test_tangent_splitting(StableRNG(123456), p)


yebai:
That would indeed explain the random failures.

yebai:
@AstitvaAggarwal, the same bug is observed agaain [here](https://buildkite.com/julialang/mooncake-dot-jl/builds/2810/steps/canvas?jid=019b4c71-69e8-44a3-87cc-39bf8119d0a7#019b4c71-69e8-44a3-87cc-39bf8119d0a7/L370) so this is not yet resolved. 

You may find `StableRNGs` handy to reproduce the bug.

AstitvaAggarwal:
Oh Yeah, I had deleted my message because of this. Thanks

ext/MooncakeCUDAExt.jl
```diff
@@ -38,6 +38,9 @@ import Mooncake.TestUtils:
 const CuFloatArray = CuArray{<:IEEEFloat}
 const CuComplexArray = CuArray{<:Complex{<:IEEEFloat}}
 
+# Overload Base.eps for Complex types. (Despite the type piracy, its needed to set atol in isapprox calls)
```
Technici4n: Is this really needed with `eps(T)` below?

Technici4n: Ah now I understand. Maybe we should introduce a function such as `_eps` to avoid the type piracy?

AstitvaAggarwal: yes, ive added that now

AstitvaAggarwal:
Added a detailed writeup, MWE for the issue in my PR's description.

sunxd3:
The error left seems to relate to https://github.com/JuliaGPU/CUDA.jl/issues/2946 and https://github.com/JuliaGPU/CUDA.jl/issues/2886?

sunxd3:
also, any reason https://github.com/chalk-lab/Mooncake.jl/blob/df7e0dac9859c9213aef75e7dddb90b8065abf8b/test/ext/cuda/cuda.jl#L12 stays `undef`, maybe we should use random value?

AstitvaAggarwal:
I think its because using `CUDA.randn` guarantees that AD tests only work with "nicer" numbers and not something like an array full of NaNs, extremely large floats, or subnormal numbers (often appear in undef memory) which could cause programs to crash.

It tests Mooncake's robustness on true edge cases.

sunxd3:
That makes sense, thanks. (Maybe worth adding a comment?)

sunxd3:
Looks like the GPU CIs are all passing, this is good to go from my end

src/utils.jl
```diff
@@ -476,3 +476,16 @@ _copy(x::Tuple) = map(_copy, x)
 _copy(x::NamedTuple) = map(_copy, x)
 _copy(x::Ref{T}) where {T} = isassigned(x) ? Ref{T}(_copy(x[])) : Ref{T}()
 _copy(x::Type) = x
+
+"""
+    _eps(x)
+
+Outputs the machine epsilon `ϵ` for the Type `x` is parameterised on.
+Avoids type piracy by defining a package helper function for `Base.eps`.
+Handles types like `Complex{<:Base.IEEEFloat}` etc. which do not have an overload for `Base.eps`.
```
devmotion: Maybe instead of a custom function you could use `eps(real(T))`?

test/ext/cuda/cuda.jl
```diff
@@ -9,6 +9,7 @@ using Mooncake.TestUtils: test_tangent_interface, test_tangent_splitting, test_r
     if CUDA.functional()
         # Check we can operate on CuArrays of various element types.
         @testset for ET in (Float32, Float64, ComplexF32, ComplexF64)
+            # Use undef to ensure robustness against "garbage" memory (NaNs, Infs, subnormals etc).
```
yebai:
```suggestion
            # Use `undef` to test against garbage memory (NaNs, Infs, subnormals).
            # `randn` generates well-behaved values and can miss these edge cases.
```

yebai:
Cc @kshyatt since you worked on this code recently.

src/utils.jl
AstitvaAggarwal: thanks!

PR #904, opened 2025-12-22
rename `rrules` to `rules`: they have both forward and reverse rules

yebai:


test/runtests.jl
```diff
@@ -24,47 +24,47 @@ include("front_matter.jl")
         include("config.jl")
         include("developer_tools.jl")
         include("test_utils.jl")
-    elseif test_group == "rrules/array_legacy"
+    elseif test_group == "rules/array_legacy"
         include(joinpath("rrules", "array_legacy.jl"))
```
Technici4n: Outdated reference, no?

PR #905, opened 2025-12-22
move all tangent code to a tangents_types folder

yebai:
Mooncake consists of three main components:

1. An autograd compiler
2. Chain rules
3. Tangent and cotangent types

This PR ensures that each component lives in its own folder for improved clarity. Organizing them into separate modules may be possible, but that is left for future work.

For posterity: (2) + (3) ≈ ChainRules2.


Technici4n:
This seems a little verbose, no? How about `tangents/`? After all it's `Tangent` that is used in both forward and reverse mode

yebai:
How about `tangents_and_cotangents/`? I’d prioritize clarity and accuracy here.

Technici4n:
Well, there's nothing else in the code that refers to cotangents I believe so I find it a bit confusing? But it's no big deal either way so if you find it clearer why not

yebai:
> Well, there's nothing else in the code that refers to cotangents 

Okay, that is a good point!

PR #911, opened 2025-12-29
Add rrule for rand/rand! functions

### LINKED ISSUE #888: Incorrect average gradient using rand
MWE:

```julia
julia> using Enzyme, Mooncake, DifferentiationInterface, Random

julia> function f()
           rng = Xoshiro(42)
           f(v) = rand(rng) * v
           return sum(DifferentiationInterface.gradient(f, AutoMooncake(), 1.) for _ in 1:10000)/10000
       end
f (generic function with 2 methods)

julia> f()
0.6293451231425107

julia> function f()
           rng = Xoshiro(42)
           f(v) = rand(rng) * v
           return sum(DifferentiationInterface.gradient(f, AutoEnzyme(), 1.) for _ in 1:10000)/10000
       end
f (generic function with 2 methods)

julia> f()
0.5019776665850358
```

seems like Mooncake doesn't return the correct gradient.

ameligrana:
Fixes #888

Adds `rrule`s for `rand`/`rand!` functions so that RNG state advances correctly during AD.

Now this works:

```julia
julia> using Mooncake, DifferentiationInterface, Random

julia> function f()
           rng = Xoshiro(42)
           f(v) = rand(rng) * v
           return sum(DifferentiationInterface.gradient(f, AutoMooncake(), 1.) for _ in 1:10000)/10000
       end
f (generic function with 2 methods)

julia> f()
0.5048290869789588  # this PR (correct ~0.5)
0.6293451231425107  # main (incorrect, RNG state was being reset)
```

## Changes

- Added `@zero_derivative` rules for `rand(rng)` and `rand(rng, T)` (scalar variants) so scalar `rand` advances RNG state.
- Added `@is_primitive` + `frule!!`/`rrule!!` for `rand!(rng, A)` on `SpecialisedRNGs` (Xoshiro, TaskLocalRNG, MersenneTwister), mirroring the existing rules for `randn!` and `randexp!`. The `rrule!!` saves and restores the array contents in the reverse pass so gradients flow correctly.
- Added `@zero_derivative` rules for `Random.XoshiroSimd.xoshiro_bulk_simd` and `xoshiro_bulk_nosimd`, which use `llvmcall` intrinsics that Mooncake cannot differentiate. These are triggered when `rand(rng, dims...)` is called (e.g. `rand(Xoshiro(123), 9)`), routing through a 3-argument `rand!(rng, A, sampler)` path rather than the 2-argument primitive. This fixes the test case `x -> x .* rand(Xoshiro(123), size(x)...)`.
- Added derived rule test cases for all new `rand`/`rand!` variants.
- Added a `"Random state advancement"` test asserting Mooncake does not reset RNG state after differentiating `rand`/`randn`/`rand!`/`randn!`.

### DIFF
--- a/src/rules/random.jl
+++ b/src/rules/random.jl
@@ -1,7 +1,6 @@
-# Contains a ccall, which must be avoided.
+# The following functions contain ccall or llvmcall which cannot be differentiated.
 @zero_derivative MinimalCtx Tuple{Type{MersenneTwister},Any}
-
-# XoshiroSimd bulk operations use llvmcall which cannot be differentiated by Mooncake.
+# xoshiro_bulk_simd/nosimd(rng, dst::Ptr{UInt8}, len::Int64, eltype, [simd_width,] bits2float)
 @zero_derivative MinimalCtx Tuple{
     typeof(Random.XoshiroSimd.xoshiro_bulk_simd),
     Union{TaskLocalRNG,Xoshiro},
--- a/test/rules/random.jl
+++ b/test/rules/random.jl
@@ -1,4 +1,4 @@
-@testset "randn" begin
+@testset "random" begin
     TestUtils.run_rule_test_cases(StableRNG, Val(:random))
 end
 
@@ -6,10 +6,10 @@ end
     rng = Xoshiro(123)
     for f in (rand!, rand, randn!, randn)
         rng2 = deepcopy(rng)
-        args = f in (Random.rand!, Random.randn!) ? (rng, [0.0]) : (rng,)
+        args = f in (rand!, randn!) ? (rng, [0.0]) : (rng,)
         rule = Mooncake.build_rrule(f, args...)
         @test rng == rng2
-        if f in (Random.rand!, Random.randn!)
+        if f in (rand!, randn!)
             Mooncake.value_and_pullback!!(rule, [1.0], f, args...)
         else
             Mooncake.value_and_gradient!!(rule, f, args...)


ameligrana:
Maybe we can also add

```julia
@zero_derivative MinimalCtx Tuple{typeof(rand),Type{UInt64}}
@zero_derivative MinimalCtx Tuple{typeof(rand),KnownRNGs,Type{UInt64}}
```

so that it works fine also in e.g. this case

```julia
julia> function f()
                  rng = Xoshiro(42)
                  f(v) = rand(rng, 1:10) * v
                  return sum(DifferentiationInterface.gradient(f, AutoMooncake(), 1.) for _ in 1:100000)/100000
              end

julia> f()
5.51794
```

?

yebai:
Please feel free to merge and release.

yebai:
Perhaps add a unit test based on the above example?

Technici4n:
> Maybe we can also add
> 
> ```julia
> @zero_derivative MinimalCtx Tuple{typeof(rand),Type{UInt64}}
> @zero_derivative MinimalCtx Tuple{typeof(rand),KnownRNGs,Type{UInt64}}
> ```

Why not! The usual philosophy is to add rules for the lowest-level functions, so if these are the lowest-level we should add more rules. Otherwise if they are all implemented using a lower-level function it is preferable to write a rule for that lower-level function. I am not familiar with the internals of Random so I leave it up to you. :) I am also happy to merge this PR as is if you would already like it to go through?

ameligrana:
ok, perfect, actually adding it, I will also add some tests in a moment

ameligrana:
I added the basic tests already implemented for the other functions. Actually I'm not really sure where to add the other tests since Mooncake seems to have an involved testing infrastructure, and at the same time, digging a bit into it, the infrastructure doesn't seem to be able to test the average of multiple computations, I think the PR should be okay already like this, though if someone has any suggestion on how to add tests for this, I will add these other tests.

yebai:
A simple test that assert Mooncake doen't undo random state after calling `rand!/randn/rand` would be helpful. Please feel free to add it after this 

https://github.com/chalk-lab/Mooncake.jl/blob/main/test/rules/random.jl#L3

ameligrana:
added that

yebai:
In case it is helpful, command like

```bash
julia --project=. -e 'import Pkg; Pkg.test(; test_args=ARGS)' -- rules/random
```

allows one to run a specific group of test (ie, `rules/random`) here. The complete list of test groups can be found at

https://github.com/chalk-lab/Mooncake.jl/blob/0b55297d1418502eac42b69a6b3a252264288e5b/test/runtests.jl#L4-L71

ameligrana:
thank you for the help @yebai ! Forgot to move this forward.

yebai:
No problem -- thanks for the fix!

Issue #916, opened 2025-12-30
World age issues in nested autograd


yebai:
The following *misty closure* heuristic appears to introduce genuine problems, likely limited to Julia 1.11:

https://github.com/chalk-lab/Mooncake.jl/blob/f93752101d1984e9ccd3f83348ab1d725ca8994e/src/rules/misty_closures.jl#L20-L38

Some examples: 

- ~~DI second-order tests fail ([CI run](https://github.com/chalk-lab/Mooncake.jl/actions/runs/20608215996/job/59187979053?pr=915)). The failure pattern is inconsistent with an out-of-memory condition. Instead, Mooncake likely encounters an opaque-closure bug in Julia 1.11. The issue is resolved when transitioning to Julia 1.12.~~ 
  - The DI crash is [reproducible](https://github.com/chalk-lab/Mooncake.jl/actions/runs/20608773382/job/59189484535) after removing the misty closure heuristic (d857774), adding weight that this is an opaque-closure bug in Julia 1.11. This is potentially the same issue as discussed in https://github.com/chalk-lab/MistyClosures.jl/pull/12#issue-3278662295
- In DI Hessian computations, repeated preparation (e.g. `DontPrepareInner()`) triggers a world-age error ([issue discussion](https://github.com/chalk-lab/Mooncake.jl/issues/632#issuecomment-3700826967)). This is fixed by https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/948. 
  - One can modify MistyClosure to carry a back reference of the original Julia method so we can re-derive rules in the current world age, instead of the heuristic. Re-deriving rules would, in theory, be the right thing to do.

yebai:
> In DI Hessian computations, repeated preparation (e.g. DontPrepareInner()) triggers a world-age error (https://github.com/chalk-lab/Mooncake.jl/issues/632#issuecomment-3700826967). This is fixed by https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/948.

@sunxd3, is it clear why repeated DI preparation still causes world age issues even with the `_dual_mc` trick? It would be great to get to the bottom of it.

sunxd3:
Not to me immediately, but digging deeper is definitely needed

PR #926, opened 2026-01-06
Fix world-age mismatch in _dual_mc by using IR's valid_worlds

sunxd3:
Fixes https://github.com/chalk-lab/Mooncake.jl/issues/916

The error reproduce code is at https://github.com/chalk-lab/Mooncake.jl/issues/632#issuecomment-3700826967
Locally, the error is gone on 1.11 and 1.12

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.4.197"
+version = "0.4.198"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"


src/rules/misty_closures.jl
```diff
@@ -31,10 +31,22 @@ end
...
-    mc_world = UInt(p.oc.world)
-    interp = MooncakeInterpreter(DefaultCtx, ForwardMode; world=mc_world)
-    return build_frule(interp, p; skip_world_age_check=true)
+@static if VERSION > v"1.12-"
+    function _dual_mc(p::MistyClosure)
+        # Use the IR's valid_worlds.max_world instead of oc.world to avoid world age mismatch.
+        # The oc.world can be slightly newer than valid_worlds.max_world if methods were
+        # defined between IR generation and OpaqueClosure creation. Using max_world ensures
+        # we're within the valid range while still having access to all methods the IR needs.
+        mc_world = UInt(p.ir[].valid_worlds.max_world)
```
yebai, line 40: Can we move the version bound to this line only since the rest seems identical? This would help readbility.

sunxd3: done

yebai:
Good investigation and fix, @sunxd3! 

I suggest we find a minimal working example and add a test for it. This would be very helpful, since the world age issue is tricky to debug, and having a unit test would help guard against regressions.

sunxd3:
Thanks @yebai, very reasonable suggestion. ~~It's a tricky issue — the error only occurs on a fresh Julia session, by this I mean: if we execute~~

```julia
try
    # code from 632 comment
catch _
end

# code from 632 comment
```

~~, it won't error. This is because the first time the code block runs, methods are compiled, so at the time of the second code block runs, the world age is consistent.~~ _this is false, see comment below._

I added a test that mimic the problematic behavior, but it feels a bit tautological. Let me know if you think we should go with something else.

yebai:
The test looks sound. 

To clarify, even with the world age error, compilation in the try-catch-end block can still produce a compiled rule that help later calls?

sunxd3:
That's what I meant, but now I realized I made a mistake. Sorry for the confusion.

With this code (slightly simplified from https://github.com/chalk-lab/Mooncake.jl/issues/632#issuecomment-3700826967)

<details><summary>test code</summary>
<p>

```julia
using Mooncake: Mooncake

module MooncakeDemo
    function objective(x)
        #   (GAMSFiles.checkaxes)(x, (Base.OneTo(10),))
        objvar = (((((((((((((((((((((((((((x[1] * x[1] + x[10] * x[10]) * (x[1] * x[1] + x[10] * x[10]) - 4 * x[1]) + 3) + (x[2] * x[2] + x[10] * x[10]) * (x[2] * x[2] + x[10] * x[10])) - 4 * x[2]) + 3) + (x[3] * x[3] + x[10] * x[10]) * (x[3] * x[3] + x[10] * x[10])) - 4 * x[3]) + 3) + (x[4] * x[4] + x[10] * x[10]) * (x[4] * x[4] + x[10] * x[10])) - 4 * x[4]) + 3) + (x[5] * x[5] + x[10] * x[10]) * (x[5] * x[5] + x[10] * x[10])) - 4 * x[5]) + 3) + (x[6] * x[6] + x[10] * x[10]) * (x[6] * x[6] + x[10] * x[10])) - 4 * x[6]) + 3) + (x[7] * x[7] + x[10] * x[10]) * (x[7] * x[7] + x[10] * x[10])) - 4 * x[7]) + 3) + (x[8] * x[8] + x[10] * x[10]) * (x[8] * x[8] + x[10] * x[10])) - 4 * x[8]) + 3) + (x[9] * x[9] + x[10] * x[10]) * (x[9] * x[9] + x[10] * x[10])) - 4 * x[9]) + 3) - 0
        return objvar
  end
end
x0 = [0.0; fill(1.0, 9)]
f = MooncakeDemo.objective

backend2 = SecondOrder(AutoMooncakeForward(), AutoMooncake())
preph = prepare_hessian(f, backend2, x0)

# directly call `hessian` - works
hessian(f, preph, backend2, x0)

# closure - worldage issue
∇²f(x) = hessian(f, preph, backend2, x)
∇²f(x0)
```

</p>
</details>

I ran the closure once, saw the error, then I ran `hessian` directly, the error is gone. So I thought it's the error that fixed the worldage problem. But in fact, the try-catch-end block **doesn't** help preventing errors for the closure call. This is a pretty sloppy overlook, sorry for this.

The good news is that, running the closure version produces the worldage error consistently, I'll see if I can make a test without DI dependency.

Additionally, merging main seems to cause some allocation regression, I'll need to look into those too.

test/ext/differentiation_interface_second_order/differentiation_interface_second_order.jl
```diff
@@ -20,3 +21,34 @@ test_differentiation(
...
+function gams_objective(x)
+    #! format: off
+    objvar = (((((((((((((((((((((((((((x[1] * x[1] + x[10] * x[10]) * (x[1] * x[1] + x[10] * x[10]) - 4 * x[1]) + 3) + (x[2] * x[2] + x[10] * x[10]) * (x[2] * x[2] + x[10] * x[10])) - 4 * x[2]) + 3) + (x[3] * x[3] + x[10] * x[10]) * (x[3] * x[3] + x[10] * x[10])) - 4 * x[3]) + 3) + (x[4] * x[4] + x[10] * x[10]) * (x[4] * x[4] + x[10] * x[10])) - 4 * x[4]) + 3) + (x[5] * x[5] + x[10] * x[10]) * (x[5] * x[5] + x[10] * x[10])) - 4 * x[5]) + 3) + (x[6] * x[6] + x[10] * x[10]) * (x[6] * x[6] + x[10] * x[10])) - 4 * x[6]) + 3) + (x[7] * x[7] + x[10] * x[10]) * (x[7] * x[7] + x[10] * x[10])) - 4 * x[7]) + 3) + (x[8] * x[8] + x[10] * x[10]) * (x[8] * x[8] + x[10] * x[10])) - 4 * x[8]) + 3) + (x[9] * x[9] + x[10] * x[10]) * (x[9] * x[9] + x[10] * x[10])) - 4 * x[9]) + 3) - 0
+    #! format: on
+    return objvar
+end
+end
+
+@static if VERSION > v"1.12-"
+    @testset "world age fix with closure (#916)" begin
```
yebai, line 44:
```suggestion
    const DI = DifferentiationInterface
    DI.inner_preparation_behavior(::AutoMooncakeForward) = DI.DontPrepareInner()
    @testset "world age fix with closure (#916)" begin
```

yebai:
Please feel free to merge.

PR #927, opened 2026-01-06
More second-order tests

yebai:
Enable second-order tests on versions 1.10 and 1.12. Version 1.11 is excluded due to an opaque closure bug.


yebai:
@AstitvaAggarwal please feel free to merge and release after tests pass.

sunxd3:
there is also a CI turn over concern here, enabling all the second order tests means the DI second-order CI can take up to about an hour to finish, while the longest running other CI finish around 30 mins. could be a motivation for more optimization

Issue #807, opened 2025-10-13
`nan` issues


yebai:
See, https://github.com/chalk-lab/Mooncake.jl/commit/b8c538b0ee084b8ce1ced66d897f7d227d3a74a4#commitcomment-167818443

AstitvaAggarwal:
@yebai @Technici4n , Note that in Forward Mode AD the issue of tangent poisoning due to singularies, `NaN` does not occur due to independent tangent propagation for each input (unlike Reverse Mode) . But if the mathematically correct tangent is something `Inf/NaN` it could still cause a mathematically expected fail in downstream use despite being no direct fault/improper tangent propagation within the AD library per se.

Just noticed that `ForwardDiff.jl` in non `nan_safe` mode returns a mathematically incorrect tangent. It should have returned `NaN` correctly like in `Mooncake.jl`.
```
julia> function f(x)
           a = x
           b = sqrt(abs(x-1.0))
           return b
       end

julia> DifferentiationInterface.derivative(f, AutoForwardDiff(), 1.0)
Inf

julia> DifferentiationInterface.derivative(f, AutoMooncakeForward(), 1.0)
NaN
```

But yeah I think if possible, we could add something similar to `nan_safe` mode (depends on Preferences.jl) for Mooncake as well. This would allow in Forward/Reverse Mode, returning **both** mathematically correct results  **and** something much more safe for ML, optimization routines, depending on what the user wants to work with.  I think JAX also does something similar...

Technici4n:
I don't understand the justification for why forward and reverse mode should give different results. For a sufficiently complex function, it seems to me that a NaN in the middle of the computational graph will poison tangents, regardless of whether you are going forward or in reverse.

AstitvaAggarwal:
This example in the [commit's](https://github.com/chalk-lab/Mooncake.jl/issues/807#issue-3511658870) discussion is a bit misleading and threw me off,
```
julia> function f(x)
           a = x
           b = sqrt(abs(x-1.0))
           return b
       end

julia> DifferentiationInterface.derivative(f, AutoForwardDiff(), 1.0)
Inf

julia> DifferentiationInterface.derivative(f, AutoMooncakeForward(), 1.0)
NaN

julia> function f(x)
           a = x
           b = sqrt(abs(x-1.0))
           return a
       end
f (generic function with 1 method)

julia> DifferentiationInterface.derivative(f, AutoMooncakeForward(), 1.0)
1.0

[...]

Issue #455, opened 2025-02-03
Activity Analysis


willtebbutt:
#452 highlighted a situation in which our lack of activity analysis can cause substantial slow-downs in performance. This purpose of this issue is to sketch out how activity analysis might be implemented in Mooncake.

Todo:
- [ ] explain what activity analysis lets you do
- [ ] explain what you need to consider when implementing it in a mutation-friendly AD system
- [ ] explain how it might be implemented in Mooncake

willtebbutt:
> But if the analogy with ForwardDiff is valid, such a write attempt will just error?

Yeah, I think you're right that we can ensure that the behaviour is to error. For example, the rule for `setindex!` (or whatever the low-level primitive is called) would error if the array being written to is inactive, but the array whose value you are writing to it is active.

In terms of design, I suspect that we would be better off having a wrapper type that explicitly states that a value is inactive, otherwise we risk issues inside nested AD (I think). This is a detail we can figure out though.

Note that adding this kind of functionality would make it possible to resolve https://github.com/compintell/Mooncake.jl/issues/412 . The various SpecialFunctions.jl rules that we cannot currently support all involve read-only data (always floats I believe) which is typically inactive (this is the case that Zygote can support). It would be very nice to have this resolved.

willtebbutt:
I'm not sure that we can do this in general, but it will definitely have an effect on the what we need to shove in adjoints. This will be operation-dependent, so lets consider a couple of examples.

The rule for `*(::Float64, ::Float64)` is something like
```julia
function rrule!!(::CoDual{typeof(*)}, _x::CoDual{Float64}, _y::CoDual{Float64})
    x = primal(_x)
    y = primal(_y)
    z = x * y
    mul_adjoint(dz::Float64) = NoRData(), dz * y, dz * x
    return CoDual(z, NoFData(), mul_adjoint
end
```
Importantly, both `x` and `y` are required on the reverse-pass, meaning that `mul_adjoint` contains 16B of data.

Suppose that `y` is inactive. This means that we do not need to compute its gradient on the reverse-pass. In this case, the rule would be something like
```julia
function rrule!!(::CoDual{typeof(*)}, _x::CoDual{Float64}, _y::Inactive{Float64})
    x = primal(_x)
    y = primal(_y)
    z = x * y
    mul_adjoint(dz::Float64) = NoRData(), dz * y, GradNotRequired()
    return CoDual(z, NoFData(), mul_adjoint
end
```
[...]

Issue #412, opened 2024-12-07
Missing rules for SpecialFunctions


gdalle:
Hi there!
I saw this Discourse post benchmarking reverse-mode AD on a code with Bessel functions: https://discourse.julialang.org/t/speeding-up-zygote-autodiff-for-numerical-loop/123515/
I tried Mooncake on it but it was missing a rule for `bessely`. When I tried implementing it, I got a weird error message (can't reproduce it here since I'm not on my work computer), which I think may be linked to the partial implementation of `ChainRulesCore.rrule`:
https://github.com/JuliaMath/SpecialFunctions.jl/blob/ed6a36732712a71e99de397cca45a58432f22a0e/ext/SpecialFunctionsChainRulesCoreExt.jl#L97-L103
Wanna take a look? It would be fun to see how close we can get to Enzyme.

willtebbutt:
Yeah, the problem is that Mooncake can't (currently) handle `NotImplemented` stuff from ChainRules. In order to be able to do it without risking dropping gradient information, we need to be able to prove that it's safe to do so. At the minute we don't have any mechanism to do this.

This is to say that the only way to support the bessel functions currently is to finish the implementation off.

willtebbutt:
A good example is something like the method of `SpecialFunctions.gamma` found [here](https://github.com/JuliaMath/SpecialFunctions.jl/blob/ed6a36732712a71e99de397cca45a58432f22a0e/src/gamma_inc.jl#L1119) with associated `ChainRules.rrule` found [here](https://github.com/JuliaMath/SpecialFunctions.jl/blob/ed6a36732712a71e99de397cca45a58432f22a0e/ext/SpecialFunctionsChainRulesCoreExt.jl#L55).

If you inspect the rule, you will see that (for whatever reason), the authors of SpecialFunctions have not provided the gradient w.r.t. the first argument, and instead return whatever `ChainRulesCore.@not_implemented(INCOMPLETE_GAMMA_INFO)` produces, which IIRC is a `ChainRulesCore.NotImplemented`. This is useful in the situation that you actually don't need the gradient w.r.t. the first argument of this function because it is e.g. a hard-coded constant. Again, IIRC, `ChainRulesCore.NotImplemented` has the property that if you try to add it to anything, you'll get an error.

[...]

PR #931, opened 2026-01-07
`NaN`, `not_implemented` handling.

AstitvaAggarwal:
This PR is a prerequisite to understand #455 , truly solving #794 & #412  and completely address #807 .
Usefull threads: https://github.com/chalk-lab/Mooncake.jl/issues/412#issuecomment-2531646805 , https://github.com/chalk-lab/Mooncake.jl/issues/807#issuecomment-3405128846 and thier following messages, https://github.com/chalk-lab/Mooncake.jl/issues/794#issue-3473595950.

Refer PR code docstrings for more implementation specific info.

The below is a high level description.

### Formulating the Problem

We have 2 problem variables :

- The type of branch: Active (affecting relevant gradient calculations) / Dead (not affecting final gradient)
_**Removing Dead branches via activity analysis is only an Efficiency issue.**_

- The type of Math: Broken/Healthy 
 _**broken math is critical & leads to direct AD poisoning.**_

(Activity analysis is not required for correctness if dead branches are NaN-safe or guarded, but it is required for performance and robustness in the general case)

### Problem Details

**Broken math** can be :

1. `NaN` gradients. This can appear for functions:
	1. throughout thier domain.
	2. only at Singular domain points.

2. Gradient not implemented yet due to it being intractable.

**So Finally -> We are dealing with (Active branch + Broken Math) and (Dead branch + Broken Math) only.**

### Solutions

Now that the problem has been diagnosed, we move onto the solutions:

1.1. These must behave like healthy math unless they are actually used in an active branch.
Enable `NaN` safe for all adjacent active AD calculations.

1.2. Enable `NaN` safe **only** for gradient calculations at **singular inputs**.
These gradients have to be wrapped with the macro that conditionally informs the user that its a `NaN` and normally continue gradient calculations.

2. Implement a wrapper/macro informing user that we dont have gradient for this variable,
Loudly Error out if AD tries to use such tangent types in downstream gradient calculations.

**NOTE :** 

> 
> 1. for BOTH 1.1 and 1.2, If the user's logic leads to gradient `NaN` calculated in the active branch, we need to simply return `NaN` wrapper for mathematical correctness - this just means the active branch/function gradient is mathematically `NaN`,  AD should not “fix” that.
> 
> 2. `Inf` is valid by AD, Math so we leave this. If this converts to a `NaN` downstream -> this must be handled as a "singular point" (refer 1.2 above) exactly at the evaluation which causes it - we write a rule for this.
> 

**CURRENT STATE OF PR**

- `MooncakeNaN` must only exist in tangent world (NaN poisoning exists here) .
- Handles any dead branch in the Tangent space for numerically continuous programs. 
 
- **TODO**: More overloads for `NotImplementedTangent` and `MooncakeNaN`.
- **TODO**: Write a similar macro for primal world, allowing dynamic rulegen (pullback uses `MooncakeNaN`) for Dead branches in Primal World. (cover singular/discontinuous functions)
- **TODO**:  Handle nested AD derivatives (common in Numerical Solvers ) - needs custom tangent type.
- **TODO**: Allow `guard_nan` to handle internal context switching for functions with singular points? (although this reeks of activity analysis, I think a optimal soluiton for this is 100% possible).

- _**I have added docstrings in the PR for more exact working examples - handles all Dead branch `NaN`**_

**TLDR**
 
1. Handle general usage of **Unhandled/Intractable** analytical gradients within *special rules* with informative reasoning.

2. Dead branch + Broken Math has been solved. (Holds when **all** dead branches are finally annotated). *Edge case* : Any incorrectly annotated active branch will result in incorrect final gradients (`guard_nan` ensures this never happens)
 
3. Active branch + Broken Math isnt actually "broken", the program being differentiated will correctly give out `NaN` according to the user's logic. `Mooncake` is innocent and will correctly still return `NaN` as the calculated gradient.
 
4. The true safeguard for **TLDR** point 2 in the worst case is relying **only** on Activity Analysis and not a *perfect* Engineer/Scientist/User. 

**Important PR comments :**
1. https://github.com/chalk-lab/Mooncake.jl/pull/931#issuecomment-3723848878


src/tangents/tangents.jl
```diff
@@ -1483,6 +1483,199 @@ end
...
+
+NOTE: THIS MUST ONLY BE USED WITHIN RULES RIGHT NOW.
+Once TODO is addressed, then can also use freely in User/autodiffed code.
+"""
+struct MooncakeNaN
+    msg::String
+end
+
+# Arithmetic with itself
+import Base: +, -, *, /, ^
```
penelopeysm: instead of `import Base: foo` and then defining `foo(...) = ...`, it's easier and also clearer to just write `Base.foo(...) = ...`, since you don't have to hunt down the import to find out where the function was imported from.

src/tangents/tangents.jl
```diff
@@ -1483,6 +1483,199 @@ end
...
+    x_val = primal(x)
+    val = f(x_val)
+
+    function pb!!(::Float64)
+        a_tangent = 1.0
+        b_tangent = @MooncakeNaN_coverage(abs(x_val-1)^(-0.5) * sign(x_val-1), "NaN seen at stationary input point 1.0")
+        return NoRData(), a_tangent + b_tangent
+    end
+    return zero_fcodual(val), pb!!
+end
```
penelopeysm, line 1573: you can surround code examples in triple-backtick blocks, and it will be displayed more nicely in text editors and/or generated documentation output:

````
```julia
# code...
```
````

File: src/tangents/tangents.jl [hunk 3]
```diff
@@ -1483,6 +1483,199 @@ end
...
+
+DifferentiationInterface.derivative(ftest, AutoMooncake(), 1.0) -> returns MooncakeNaN("filler") (mathematically correct)
+# Must error out if @MooncakeNaN_coverage macro used incorrectly within rrule
+
+"""
+
+macro MooncakeNaN_coverage(expr, msg::String)
+    quote
+        possiblyNaN = $(esc(expr))
```
penelopeysm: the macro name is a bit unconventional for Julia which tends to use `snake_case` identifiers, so if you wanted to use these words, it would typically be `mooncake_nan_coverage`. I don't think `nan` needs to be capitalised as Base has things like `isnan` which don't capitalise it.

although personally I think 'coverage' is not the right word - something else like `guard_nan` might be better?

In general, though, I'm not sure that this needs to be a macro. I think an ordinary function would be fine. I would also skip the info message since it would be quite annoying to see multiple times

```julia
function guard_nan(val, nanmsg="")
    if isnan(val)
        return MooncakeNaN(nanmsg)
    else
        throw(ArgumentError(...))
    end
end

File: src/tangents/tangents.jl [hunk 4]
```diff
@@ -1483,6 +1483,199 @@ end
...
+2> The AD CFG ALWAYS TREATS the Variable WHOSE tangent is annotated with NotImplementedTangent as a DISJOINT NODE.
+3> If the program tries to use it explicitly or implicitly during tangent propagation, Mooncake must error Loudly.
+
+NOTE : This is also a tool for RULES ONLY.
+
+"""
+
+struct NotImplementedTangent
+    msg::String
+end
```
penelopeysm: right now there's only ever one string so this doesn't need to wrap a string.

for the exception throwing itself you could also consider doing the following https://docs.julialang.org/en/v1/base/io-network/#Base.showerror

```julia
struct UnimplementedTangentException <: Exception end

Base.showerror(io, ::UnimplementedTangent) = println(io, "UnimplementedTangentException: tangent not implemented for operation")
```

src/tangents/tangents.jl
```diff
@@ -1483,6 +1483,199 @@ end
...
+LinearAlgebra.dot(N::MooncakeNaN, x::MooncakeNaN) = N
+
+"""
+
+A wrapper for Masking any active/dead code branches with `MooncakeNaNs`.
+Usefull to highlight code branches/outputs in RULES ONLY with internal evaluations which may possibly return NaNs.
+
+NOTE :
+Can be used in regular functions ONLY WHEN `MooncakeNaN` has valid custom `MooncakeNaN` tangent.
+Refer above `MooncakeNaN` struct's TODO
```
penelopeysm: In technical writing it's better to avoid capitalising words in the middle of sentences, if you really need to emphasise you can put it in `**bold**` or `*italic*`.

src/tangents/tangents.jl
```diff
@@ -1483,6 +1483,199 @@ end
...
+^(N::MooncakeNaN, ::MooncakeNaN) = N
+
+# Arithmetic with numbers (not order invariant)
++(x::Number, ::MooncakeNaN) = x
++(::MooncakeNaN, x::Number) = x
+-(x::Number, ::MooncakeNaN) = x
+-(::MooncakeNaN, x::Number) = -x
+*(x::Number, N::MooncakeNaN) = N
+*(N::MooncakeNaN, x::Number) = N
+/(x::Number, ::MooncakeNaN) = Inf   # optional
```
penelopeysm: seems to me that it is better to return the MooncakeNaN, also since `Inf` is a `Float64` it doesn't follow that any `Number` divided by a NaN should become a `Float64`

```julia
julia > 2.0 / NaN
NaN
```

src/tangents/tangents.jl
AstitvaAggarwal: actually I found the way `ChainRulesCore` implements [this](https://github.com/JuliaDiff/ChainRulesCore.jl/blob/004e43ed79c09f2fea5da4b764531b2989a9ff3a/src/tangent_types/notimplemented.jl#L2) usefull, so I have included it. This will for sure help in the use-cases where a more unique "technical" explanation about the gradient is to be shown.

This struct and its usage is still very WIP, so Im still working around it much more.
But I will keep this review unresolved for future reference. Thanks!

src/tangents/tangents.jl
AstitvaAggarwal: The macro usage feels more neat to use imo but the major reason is that I actually have more functionality (have it planned out) that is yet to be included for `guard_nan` so this is just baby steps towards that.

src/tangents/tangents.jl
AstitvaAggarwal: Yes this is interesting, so basically : 

1. In Primal space : `NaN` behaves as is for correctness. 
2. In tangent space : `NaN` must behave like `0` in dead branches (to avoid NaN poisoning).
3. In tangent space : `NaN` behaves as is, in active branches (for correctness).

So because the current `MooncakeNaN` is only limited to the Tangent world and currently I only handle all dead branches cases, it must be treated as a `0`, therefore the above evalution must return `Inf`.

Thanks for pointing this out! I will add a note about this for general Information.

AstitvaAggarwal:
@penelopeysm  thank you very much for your reviews.
I have not acted upon some suggestions due to the current functionality being a step towards to the complete implementation.
The PR is mainly the "skeleton" of major parts of it (It covers all Dead Branches cases) with the main goal being - getting across the root problem at hand and the proposed solution. 

But I will keep track of them for the future changes. Thanks!

penelopeysm:
Of course, that's alright :)

Technici4n:
Hi, what exactly are you trying to achieve here? At a glance, defining a type for NaN is quite unusual, and given that one of the key design goals for Mooncake is to have a single tangent type per primal type, this feels like a weird direction to move towards.

My take on the NaN issues would be to take inspiration from established Python frameworks like PyTorch, which to my understanding allows NaNs to poison gradients, but uses activity analysis to avoid issues like https://github.com/chalk-lab/Mooncake.jl/issues/794.

AstitvaAggarwal:
@Technici4n, so `NaN` poisoning is the Issue. This can either occur in active branches or Dead branches.

In active branches this will always result in `NaN` for all the gradient results - This is mathematically correct (think a multivariate function having `NaN` gradient for any one variable, all other variable gradients must also be `NaN`). Activity analysis cannot interfere with this, if it does we will get incorrect gradients.

In dead branches this will always result in `NaN` for all the gradient results - This is mathematically incorrect. (the multivariate function does not depend on the `NaN` gradient variable, so all other variable gradients must not be tampered with).

Now, you can : 
1. write code without dead branches (no useless parts of code)
2. manually mark these dead branches using the failsafe `guard_nan` macro.
3. Use Activity analysis(AA) for removing the dead branches using IR completely.
4. Allow differentiating through dead branches and only use AA for dealing with `NaN` causing dead branches (will end up using `guard_nan`, `MooncakeNaN` anyways).

AA in general is automatic and will make overall performance better But option 4. is the optimal possibility imo.
But afaik AA can never be smart enough to know before the program execution itself i.e. which branch will be `NaN` or not (this holds especially for conditionals - or runtime dependent conditions). 
Therefore it cannot truly remove all `NaN` ever unless the branch is truly "dead" (non conditional etc.), i.e. its straight up something like a=90 and this variable is never used.


For conditionals:

```
function h(x)
    c = some_singular_function(x)  # NaN for some x
    if x > 0
        x *= c
    else
        x *= 2.0
    end
    return x
end

@code_lowered h(x) # x is singular point
CodeInfo(
1 ─      x@_4 = x@_2
│        c = Main.some_singular_function(x) # NaN
│   %3 = x@_4 > 0
└──      goto #3 if not %3
2 ─      x@_4 = x@_4 * c
└──      goto #4
3 ─      x@_4 = x@_4 * 2.0
4 ┄      return x@_4
)

# set c=NaN above and run :
julia> DifferentiationInterface.derivative(h, AutoMooncake(), 1.0)     
NaN

julia> DifferentiationInterface.derivative(h, AutoMooncake(), 2.0)     
NaN

```

@Technici4n I have updated the example, you can run this and check it out. AA afaik basically cant handle discontinuous or singular functions like `h` when it comes to `NaN` poisoning (because the conditional branch is a active<->dead branch depending on what domain point `h` acts upon. 
Functions like `h` are basically discontinuous functions in math.

Only `guard_nan`/dynamic rulegen safety macro can solve this. (Description TODO)

gdalle:
I don't have bandwidth to review this in detail but I share Bruno's _extreme_ skepticism at the idea of introducing a new `NaN` type, if only because it will make rules very type-unstable if they ever try to use it, and probably ruin performance.

Technici4n:
I don't think the goal should be to remove all NaNs. If Mooncake returns a NaN "by mistake" it means that the user needs to double-check. The goal should probably be to avoid NaNs when possible, and let the user deal with it otherwise. What should absolutely be avoided is returning a non-NaN derivative when it should be NaN.

The idea with AA is simply to avoid the most common source of NaNs, and build a way to express that rules should ignore derivatives wrt. some arguments. This could also be nicely combined with a "stop gradient" feature to force some arguments to be considered constant by Mooncake.

AstitvaAggarwal:
> I don't think the goal should be to remove all NaNs. If Mooncake returns a NaN "by mistake" it means that the user needs to double-check. The goal should probably be to avoid NaNs when possible, and let the user deal with it otherwise. What should absolutely be avoided is returning a non-NaN derivative when it should be NaN.

I am not removing correct results in `NaN` at all. please see https://github.com/chalk-lab/Mooncake.jl/pull/931#issue-3789594837 (TLDR section Point 3, additionally the whole description gives a better context also). 

> The idea with AA is simply to avoid the most common source of NaNs, and build a way to express that rules should ignore derivatives wrt. some arguments. This could also be nicely combined with a "stop gradient" feature to force some arguments to be considered constant by Mooncake.

AA as mentioned in desciption is only a efficiency feature, it can help only partially in avoiding `NaN` (see https://github.com/chalk-lab/Mooncake.jl/pull/931#issuecomment-3723848878 explanation on conditional). 
point 4 of above link is the true solution.


> I don't have bandwidth to review this in detail but I share Bruno's _extreme_ skepticism at the idea of introducing a new `NaN` type, if only because it will make rules very type-unstable if they ever try to use it, and probably ruin performance.

No worries just wanted to keep you in the loop. I agree : *but* only rules for discontinous functions that use `MooncakeNaN` will be type unstable. This is in description's TODO but I think one clear solution is to just make every `MooncakeNaN` object of the type which is corresponding function's domain type.

yebai:
Let's handle `Nan` and `NotImplemented` in seprate PRs. 

The `NaN` solution needs very careful investigation so let's not rush it.

Technici4n:
> I am not removing correct results in `NaN` at all.

What I meant is that returning `NaN` may be an acceptable outcome, even if the function is mathematically differentiable. It's annoying but not the end of the world, especially if there exists a workaround for users.

> AA as mentioned in desciption is only a efficiency feature, it can help only partially in avoiding `NaN` (see [#931 (comment)](https://github.com/chalk-lab/Mooncake.jl/pull/931#issuecomment-3723848878) explanation on conditional). point 4 of above link is the true solution.

Well, AA is also a correctness feature since it will avoid poisoning from inactive variables? Although that is of course not the main intention behind it, there I fully agree. But it would be a nice benefit. I don't really understand what you mean with point 4, potentially because I don't understand to what problem it is a solution.

One thing to also consider, which this PR might be overlooking but I am really not sure there (do users not need to put macro calls in their code? or maybe it's exclusive to rules?), is that many users will probably want to remain generic in the AD backend that they are using, so as much as possible should work out of the box.

AstitvaAggarwal:
> What I meant is that returning `NaN` may be an acceptable outcome, even if the function is mathematically differentiable. It's annoying but not the end of the world, especially if there exists a workaround for users.

Yeah the PR allows all math correct stuff to work as is.

> Well, AA is also a correctness feature since it will avoid poisoning from inactive variables? Although that is of course not the main intention behind it, there I fully agree. But it would be a nice benefit. I don't really understand what you mean with point 4, potentially because I don't understand to what problem it is a solution.

Yes it helps, but alone it wouldn't solve the full problem (misses conditional dead branches) of removing *all* singular dead branches to avoid `NaN` poisoning. It needs the PR's code to mark any suspicious conditional dead branches.

> One thing to also consider, which this PR might be overlooking but I am really not sure there (do users not need to put macro calls in their code? or maybe it's exclusive to rules?), is that many users will probably want to remain generic in the AD backend that they are using, so as much as possible should work out of the box.

Actually the PR covers it. refer line `1498` - `src/tangents/tangents.jl`.

yebai:
On closer inspection of the code, it appears that `NotImplemented` cannot be made to work with Mooncake because tangents are accumulated eagerly during the reverse pass (see this [comment](https://github.com/chalk-lab/Mooncake.jl/issues/412#issuecomment-2531837372)).

If the motivation for introducing `NotImplemented` is to enable importing the existing [Bessel function rules](https://github.com/JuliaMath/SpecialFunctions.jl/blob/31e3c1a1ff05612d8730a38b330df67d8ff77a91/ext/SpecialFunctionsChainRulesCoreExt.jl#L214-L239), a sensible and safe compromise would be to return NaN-valued tangents for arguments whose derivatives are not implemented. In cases where gradients with respect to the first argument of the Bessel functions are not required, these NaN-valued tangents are irrelevant and invisiable to the user. In cases where such gradients are required, the resulting NaNs make the failure explicit and signal that these gradients should not be used.

AstitvaAggarwal:
Just FYI, this PR does not handle activity analysis. It only allows for a solid way of handling `NaN` in all programs handled by Mooncake.

> On closer inspection of the code, it appears that `NotImplemented` cannot be made to work with Mooncake because tangents are accumulated eagerly during the reverse pass (see this [comment](https://github.com/chalk-lab/Mooncake.jl/issues/412#issuecomment-2531837372)).

Yes I am well aware of that comment and was actively still working NotimplementedTangent to handle such cases.

> If the motivation for introducing `NotImplemented` is to enable importing the existing [Bessel function rules](https://github.com/JuliaMath/SpecialFunctions.jl/blob/31e3c1a1ff05612d8730a38b330df67d8ff77a91/ext/SpecialFunctionsChainRulesCoreExt.jl#L214-L239), a sensible and safe compromise would be to return NaN-valued tangents for arguments whose derivatives are not implemented. In cases where gradients with respect to the first argument of the Bessel functions are not required, these NaN-valued tangents are irrelevant and invisiable to the user. In cases where such gradients are required, the resulting NaNs make the failure explicit and signal that these gradients should not be used.

Unfortunately No, this will cause NaN poisoning. We are injecting NaNs into all , even correct gradients.

Below is some code for SpecialFunctions.jl Rules Ive been working on. This is an example supporting my point :

```

using SpecialFunctions: gamma_inc

@is_primitive DefaultCtx Tuple{typeof(gamma_inc),<:IEEEFloat,<:IEEEFloat,<:Integer}

function tesfunc(x)
    y = gamma_inc(x, x, 1)
    return x
end

function rrule!!(
    ::CoDual{typeof(gamma_inc)}, a::CoDual{P}, x::CoDual{P}, IND::CoDual{<:Integer}
) where {P<:IEEEFloat}
    ap, xp, indp = primal(a), primal(x), primal(IND)
    y = gamma_inc(ap, xp, indp)
    function gamma_inc_pb!!(dy::Tuple{P,P})
        z = exp(-xp) * xp^(ap - 1) / gamma(ap)
        ∂x = (indp == 0 || indp == 1) ? z : -z

        return NoRData(), NaN, dy[2] * ∂x, NoRData()
    end
    return zero_fcodual(y), gamma_inc_pb!!
end


julia> function testfunc(x)
                 y = gamma_inc(x, x, 1)
                 return x
           end
testfunc (generic function with 1 method)

julia> cache=Mooncake.prepare_pullback_cache(testfunc,1.0)

julia> Mooncake.value_and_pullback!!(cache,1.0,testfunc,1.0)
(1.0, (Mooncake.NoTangent(), NaN))

julia> 
```

we can only return 0 - but this isnt safe as it leads to possible future gradients being completely wrong.

AstitvaAggarwal:
Also I had moved all code for NotImplementedTangent https://github.com/chalk-lab/Mooncake.jl/pull/931/commits/4e8b6c7c3f151adbb357e0aa7c4cd498eeb49162 to the SpecialFunctions PR code. Not sure why this PR was closed.

PR #933, opened 2026-01-07
Add an example using native Mooncake API

yebai:

### DIFF
--- a/docs/src/interface.md
+++ b/docs/src/interface.md
@@ -19,7 +19,7 @@ g(x) = sum(abs2, x)
 x_complex = [1.0 + 2.0im, 3.0 + 4.0im]
 ```
 
-With `friendly_tangents=true`, gradients use the same types as inputs:
+With `friendly_tangents = false`, gradients use the same types as tangents.
 
 ```@example interface
 cache = MC.prepare_gradient_cache(g, x_complex; friendly_tangents=true)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,14 +25,21 @@ There are several ways to interact with `Mooncake.jl`.
...
+
+# Using DifferentiationInterface API
+# Reverse-mode AD. For forward-mode AD, use `AutoMooncakeForward()`
+backend = DI.AutoMooncake()
+prep = DI.prepare_gradient(f, backend, x)
+DI.gradient(f, prep, backend, x)
+
+# Using native Mooncake API
+cache_friendly = MC.prepare_gradient_cache(f, x; friendly_tangents=true)
+val, grad = MC.value_and_gradient!!(cache_friendly, f, x)
```
penelopeysm: hmm - I think maybe we can move these three lines of code to below lines 46-48?

docs/src/interface.md
```diff
@@ -5,6 +5,50 @@ for some reason DifferentiationInterface.jl does not suffice.
...
+```
+
+With `friendly_tangents=true`, gradients use the same types as inputs:
+
+```@example interface
+cache = MC.prepare_gradient_cache(g, x_complex; friendly_tangents=true)
+val, grad = MC.value_and_gradient!!(cache, g, x_complex)
+```
+
+With `friendly_tangents=false`, gradients use the same types as tangents:
```
penelopeysm, line 29: Is there a performance difference? It might be useful to clarify (regardless of what the answer is)

docs/src/interface.md
```diff
@@ -5,6 +5,50 @@ for some reason DifferentiationInterface.jl does not suffice.
...
+`args_to_zero` is optional; when provided, it specifies a true/false value
+for each argument (e.g., `g`, `x_complex`), allowing tangent zeroing to be skipped
+per argument when the value is constant:
+
+```@example interface
+cache = MC.prepare_gradient_cache(g, x_complex; friendly_tangents = true)
+val, grad = MC.value_and_gradient!!(
+    cache,
+    g,
+    x_complex,
```
penelopeysm:
```suggestion
    x_complex;
```

README.md
penelopeysm: i would make a github suggestion but the UI won't let me :-)

docs/src/interface.md
yebai: There shouldn't -- otherwise it should be considered as a bug.

README.md
yebai: Done.

yebai:
This should be ready to go.

docs/src/interface.md
penelopeysm: Sure, I just think it's useful to mention as people may have the same question & it's good to reassure them that getting nicer output types doesn't compromise performance.

penelopeysm: Otherwise LGTM!

yebai: Added a note below.

PR #938, opened 2026-01-10
Handle NotImplemented Tangents & Rules for SpecialFunctions.jl.

AstitvaAggarwal:
The functions are quite loosely dispatching back in `SpecialFunctions.jl` .

This PR includes :

1. All missing rules for `SpecialFunctions` - closing #412.
2. Handling `ChainRulesCore.NotImplemented` Tangents, similar cases. (needed for 1)
3. proposed approach can be extended to solve `NaN` poisoning. #807 
4. rules for Holomorphic functions which requires `Complex` support. (follows 1)

```md
## Handling `ChainRulesCore.NotImplemented` Tangents in Imported Rules
Mooncake uses a *masking* trick to handle
`ChainRulesCore.NotImplemented` tangents and partial derivatives.
**NOTE:**
A missing partial derivative is irrelevant if it is multiplied by the zero
element of the corresponding tangent or cotangent space.
The result is therefore either correct or explicitly marked as unknown
(`NaN`).
### Forward mode (pushforward)
For `f(x, y)` with unimplemented `∂f/∂x`:
    ḟ = (∂f/∂x)·ẋ + (∂f/∂y)·ẏ
- If `ẋ == 0`, the contribution from `x` vanishes:
      ḟ = (∂f/∂y)·ẏ
- If `ẋ != 0`, the missing derivative is required ⇒ `ḟ = NaN`.
### Reverse mode (pullback)
Given upstream cotangent `f̄`:
    x̄ = (∂f/∂x)'·f̄
    ȳ = (∂f/∂y)'·f̄
- If `f̄ == 0`, then `x̄ = 0` even if `∂f/∂x` is not implemented.
- If `f̄ != 0`, then `x̄ = NaN`, while `ȳ` is computed normally.
### Notes
- “Zero” refers to the additive identity of the tangent/cotangent space.
- This trick relies on `NaN` and therefore applies only to
  floating-point tangent spaces and their compositions
  (e.g. arrays of floats).
  This restriction exists because `NaN`, `NaN32`, and `NaN16`
  live exclusively in `AbstractFloat` spaces.
- `NotImplemented` indicates a missing AD rule, not a non-differentiable
  function. Any resulting `NaN`s are handled using the same masking
  principle: they affect the result only when multiplied by a nonzero
  tangent or cotangent, and are otherwise safely ignored.
```

### DIFF
--- a/test/integration_testing/dispatch_doctor/dispatch_doctor.jl
+++ b/test/integration_testing/dispatch_doctor/dispatch_doctor.jl
@@ -65,10 +65,13 @@ function TestUtils.test_hook(f, ::typeof(Mooncake.compute_oc_signature), x...)
     allow_unstable(f)
 end
 
-include(joinpath(@__DIR__, "..", "..", "front_matter.jl"))
-
-include(joinpath(@__DIR__, "..", "..", "utils.jl"))
-include(joinpath(@__DIR__, "..", "..", joinpath("tangents", "tangents.jl")))
-include(joinpath(@__DIR__, "..", "..", joinpath("tangents", "codual.jl")))
-include(joinpath(@__DIR__, "..", "..", "stack.jl"))
-include(joinpath(@__DIR__, "..", "..", "interface.jl"))
+# DispatchDoctor triggers a segfault on Julia 1.12.
+@static if VERSION < v"1.12-"
+    include(joinpath(@__DIR__, "..", "..", "front_matter.jl"))
+    
+    include(joinpath(@__DIR__, "..", "..", "utils.jl"))
+    include(joinpath(@__DIR__, "..", "..", joinpath("tangents", "tangents.jl")))
+    include(joinpath(@__DIR__, "..", "..", joinpath("tangents", "codual.jl")))
+    include(joinpath(@__DIR__, "..", "..", "stack.jl"))
+    include(joinpath(@__DIR__, "..", "..", "interface.jl"))
+end


sunxd3:
Thanks Astitva, should I worry about the CI failures or are they expected?

AstitvaAggarwal:
Yes, those failures are valid. It's just that special functions dispatch is very large so it's missing out something somewhere. Will get it fixed soon

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +67,418 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
 
 @zero_derivative DefaultCtx Tuple{typeof(logfactorial),Integer}
 
+"""
```
Technici4n, line 60: I don't find this documentation very readable with all the abbreviations.

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +67,418 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
...
+
+
+**Note** - This will only work for Float point values & it's compositions (Complex etc) due to NaN, NaN32 & NaN16 living only AbstractFloat's Space.
+
+**Final Note** - NaN issues is also solved in a similar manner.
+
+"""
+
+# NaN fillers for Floating Points Numbers
+function Mooncake.increment_and_get_rdata!(
```
Technici4n: Shouldn't this go into Mooncake core? It doesn't look specific to SpecialFunctions

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +67,418 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
...
+    typeof(expint),Union{IEEEFloat,<:Complex},Union{IEEEFloat,<:Complex}
+}
+@from_rrule DefaultCtx Tuple{
+    typeof(expintx),Union{IEEEFloat,<:Complex},Union{IEEEFloat,<:Complex}
+}
+
+# 3 arg gamma_inc
+@from_rrule DefaultCtx Tuple{typeof(gamma_inc),IEEEFloat,IEEEFloat,Integer}
+
+# 3 arg gamma_inc (1st arg gradient Intractable)
```
Technici4n: what is an "intractable" gradient?

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +67,418 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
...
+
+function frule!!(
+    ::Dual{typeof(gamma_inc)}, a::Dual{T}, x::Dual{P}, IND::Dual{I}
+) where {T<:IEEEFloat,P<:IEEEFloat,I<:Integer}
+    ap, xp, indp = primal(a), primal(x), primal(IND)
+    da, dx = tangent(a), tangent(x)
+
+    # ∂p/∂a - Intractable Gradient
+    ∂a = if da != zero_tangent(da)
+        @info "Please use Finite Differences. Intractable Derivative for gamma_inc wrt to a - setting ∂f/∂a to NaN."
```
Technici4n: This could probably get quite spammy, I'm not sure it's the best way to do it. (And "intractable derivative" is not a standard term I believe?)

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +67,418 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
...
+
+        function frule!!(
+            ::Dual{typeof($func)}, a::Dual{T}, x::Dual{P}
+        ) where {L<:IEEEFloat,T<:Union{L,Complex{L}},P<:IEEEFloat}
+            ap, xp = primal(a), primal(x)
+            da, dx = tangent(a), tangent(x)
+
+            y = $func(ap, xp)
+            primal_type = typeof(y) # to ensure final Dual Tangent type is valid
+            f_sym = $(QuoteNode(func))
```
Technici4n: This will potentially resolve the exact used branch at runtime. It would be preferable to do the branching before interpolating into the generated expression.

test/ext/special_functions/special_functions.jl
```diff
@@ -75,4 +77,78 @@ using Mooncake.TestUtils: test_rule
...
+                    # # 3 arg gamma_inc (IND is 0/1, tangent a is 0 for AD but an approximation for testing FD)
+                    (:none, x -> gamma_inc(T(3), x, 0), P(2)),
+                    (:none, x -> gamma_inc(T(3), x, 1), P(2)),
+
+                    # 2 arg Standard Bessel & Hankel (1st arg gradient Intractable)
+                    (:none, x -> besselj(T(3), x), P(1.5)),
+                    (:none, x -> besseli(T(3), x), P(1.5)),
+                    (:none, x -> bessely(T(3), x), P(1.5)),
+                    (:none, x -> besselk(T(3), x), P(1.5)),
+                    # (:none, x -> hankelh1(T(3), x), P(1.5)),     #FIX
```
Technici4n: Remember to fix the missing tests

test/ext/special_functions/special_functions.jl
```diff
@@ -75,4 +77,78 @@ using Mooncake.TestUtils: test_rule
     )
         test_rule(StableRNG(123456), f, x...; perf_flag, is_primitive=false)
     end
+
+    @testset "Primitive SpecialFunctions with Intractable Gradients" begin
+        first_arg_types = [Float64, Float32, Float16]
+        second_arg_types = [Float64, Float32]
+
+        Real_type_tests = [(T, P) for T in first_arg_types, P in second_arg_types]
```
Technici4n: You could look into `map_prod` if you wanted to make this a bit cleaner.

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +67,418 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
...
+    @eval begin
+        @is_primitive DefaultCtx ForwardMode Tuple{
+            typeof($func),T,P
+        } where {L<:IEEEFloat,T<:Union{L,Complex{L}},P<:IEEEFloat}
+
+        function frule!!(
+            ::Dual{typeof($func)}, a::Dual{T}, x::Dual{P}
+        ) where {L<:IEEEFloat,T<:Union{L,Complex{L}},P<:IEEEFloat}
+            ap, xp = primal(a), primal(x)
+            da, dx = tangent(a), tangent(x)
```
Technici4n: Using `extract` or even `numberify` (to convert complex tangents back to complex numbers) would likely be cleaner, e.g. `a, da = numberify(a_da)`.

ext/MooncakeSpecialFunctionsExt.jl
AstitvaAggarwal: so sometimes the gradient formula is something like an integral itself that cannot be solved (gamma function wrt a), therefore intractable.
Computation is too expensive or unstable/No closed-form expression exists.

ext/MooncakeSpecialFunctionsExt.jl
AstitvaAggarwal: carrying over the intractable gradient term from above for Reverse Mode to forward mode - intractable derivative

AstitvaAggarwal: Another option is to just error when AD tries to diff wrt to a with such logging?

ext/MooncakeSpecialFunctionsExt.jl
Technici4n: Isn't that basically "not implemented", at the end of the day? 😉

yebai: I also prefer "not implemented" as the terminology here. That is clear regardless the true causes, eg, lack of implemntation, closured form equations, or expensive compuational costs.

ext/MooncakeSpecialFunctionsExt.jl
Technici4n: Or maybe use `maxlog=1` in the macro

AstitvaAggarwal: avoiding this change, as even if the singular inputs are different - the error message does not repeat.

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +56,461 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
 
 @zero_derivative DefaultCtx Tuple{typeof(logfactorial),Integer}
 
+"""
+    Handling ChainRuleCore.`NotImplemented` Tangents with @info logging.
```
Technici4n: I'm still confused as to what exactly this is trying to say. This is only true for Mooncake-specific rules, right? It doesn't seem to apply to imported ChainRules at least.

It seems like it could be replaced by a concise comment in `notimplemented_tangent_guard`.

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +56,461 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
...
+}
+@from_rrule DefaultCtx Tuple{
+    typeof(expintx),Union{IEEEFloat,<:Complex},Union{IEEEFloat,<:Complex}
+}
+
+# 3 arg gamma_inc
+@from_rrule DefaultCtx Tuple{typeof(gamma_inc),IEEEFloat,IEEEFloat,Integer}
+
+# handle frule return type according to primal type. 
+function real_or_complex_valued(y, primal_eltype, dy_val)
```
Technici4n: Better use dispatch, see `_rdata` in `rules/blas.jl` for an example.

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +56,461 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
...
+    end
+end
+
+# 3 arg gamma_inc (1st arg gradient Not Implemented)
+@is_primitive DefaultCtx ForwardMode Tuple{typeof(gamma_inc),IEEEFloat,IEEEFloat,Integer}
+
+function frule!!(
+    ::Dual{typeof(gamma_inc)}, a::Dual{T}, x::Dual{P}, IND::Dual{I}
+) where {T<:IEEEFloat,P<:IEEEFloat,I<:Integer}
+    a_info, x_info, IND_info = map(extract, (a, x, IND))
```
Technici4n: Better be explicit, this also makes the function more readable:
```suggestion
function frule!!(
    ::Dual{typeof(gamma_inc)}, _a::Dual{T}, _x::Dual{P}, _IND::Dual{I}
) where {T<:IEEEFloat,P<:IEEEFloat,I<:Integer}
    a, da = extract(_a)
    x, dx = extract(_x)
    IND = primal(_IND)
```
(needs further changes in the body, but this is the pattern I would recommend)

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +56,461 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
...
+    typeof(gamma),
+    Union{IEEEFloat,Complex{<:IEEEFloat}},
+    Union{IEEEFloat,Complex{<:IEEEFloat}},
+}
+function frule!!(
+    ::Dual{typeof(gamma)}, a::Dual{T}, x::Dual{P}
+) where {L<:IEEEFloat,T<:Union{L,Complex{L}},P<:Union{IEEEFloat,Complex{<:IEEEFloat}}}
+    a_info, x_info = map(extract, (a, x))
+    y = gamma(a_info[1], x_info[1]) # primal is always complex for complex inputs.
+    primal_eltype = eltype(y isa Complex ? y.re : y)
```
Technici4n, line 195: `isa Complex` is weird, there should be a way to write this without it.

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +56,461 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
...
+
+    ∂a = Mooncake.notimplemented_tangent_guard(a_info[2], :gamma)  # ∂f/∂a - NotImplemented Gradient
+    ∂x = -exp((a_info[1] - 1) * log(x_info[1]) - x_info[1])    # ∂f/∂x
+
+    # Ignore tangent(a) - NotImplemented Gradient
+    dy_val = ∂a + ∂x * (P <: Complex ? numberify(x_info[2]) : x_info[2])
```
Technici4n: `numberify` will also handle real numbers, no need for the `P <: Complex` check.

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +56,461 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
...
+    primal_eltype = eltype(y isa Complex ? y.re : y)     # to ensure final Dual Tangent type is valid
+
+    # ∂f/∂v - NotImplemented Gradient
+    ∂v = Mooncake.notimplemented_tangent_guard(v_info[2], :besselix)
+    # ∂f/∂x - Recurrence relations for derivatives wrt x.
+    ∂x_1 = (besselix(v_info[1] - 1, x_info[1]) + besselix(v_info[1] + 1, x_info[1])) / 2
+    ∂x_2 = -sign(real(x_info[1])) * y
+
+    # Non Holomorphic scaling
+    dy_val =
```
Technici4n: I'd recommend enclosing the RHS in parentheses, otherwise moving the terms around could accidentally make the expression end early (e.g. if the `+` were put on a new line).

src/tools_for_rules.jl
```diff
@@ -362,6 +367,23 @@ function mooncake_tangent(p::P, t::T) where {P<:Tuple,T<:CRC.Tangent}
     return tangent_type(P) == NoTangent ? NoTangent() : map(mooncake_tangent, p, t.backing)
 end
 
+# Mooncake style tangent from CRC Complex Tangents.
```
Technici4n: Why these comments, they're kinda obvious, no?

src/tools_for_rules.jl
```diff
@@ -362,6 +367,23 @@ function mooncake_tangent(p::P, t::T) where {P<:Tuple,T<:CRC.Tangent}
...
+function mooncake_tangent(
+    p::T, cr_tangent::CRC.NotImplemented
+) where {P<:IEEEFloat,T<:Complex{P}}
+    return mooncake_tangent(p, T(P(NaN), P(NaN)))
+end
+
+# Mooncake tangent from NotImplemented CRC Tangent.
+function mooncake_tangent(p::T, cr_tangent::CRC.NotImplemented) where {T<:IEEEFloat}
+    return T(NaN)
+end
```
Technici4n, line 383: If you really wanted to comment to draw the reader's attention, I would comment that not implemented tangents are converted to `NaN`.

src/tools_for_rules.jl
```diff
@@ -404,6 +450,35 @@ function increment_and_get_rdata!(f, r, t)
...
+    notimplemented_tangent_guard(da::Mooncake.Tangent, f_sym::Symbol)
+
+If da is nonzero and participates in AD, the method returns a `NaN` filled data structure for whatever tangent da is given.
+In case da is a zero tangent return zero-ed da but in a compatible way for immediate complex algebra within Mooncake rules.
+
+"""
+# This cannot have a method for Ints as NaN is only defined for Floats.
+function notimplemented_tangent_guard(da::L, f_sym::Symbol) where {L<:Base.IEEEFloat}
+    return if da != zero_tangent(da)
+        @info "Please use Finite Differences. Derivative Not NotImplemented for $f_sym wrt to a - setting ∂f/∂a to NaN."
```
Technici4n: It's a bit weird to write "a" in the error message given that this is meant to be a rather generic helper.

src/tools_for_rules.jl
```diff
@@ -404,6 +450,35 @@ function increment_and_get_rdata!(f, r, t)
...
 
+"""
+    notimplemented_tangent_guard(da::Mooncake.Tangent, f_sym::Symbol)
+
+If da is nonzero and participates in AD, the method returns a `NaN` filled data structure for whatever tangent da is given.
+In case da is a zero tangent return zero-ed da but in a compatible way for immediate complex algebra within Mooncake rules.
+
+"""
+# This cannot have a method for Ints as NaN is only defined for Floats.
+function notimplemented_tangent_guard(da::L, f_sym::Symbol) where {L<:Base.IEEEFloat}
```
Technici4n, line 470: This is a good idea. How about turning it into a macro? Then `maxlog=1` could be used to ensure that each rule only triggers logging once.

src/tools_for_rules.jl
```diff
@@ -404,6 +450,35 @@ function increment_and_get_rdata!(f, r, t)
...
+    else
+        L(0)
+    end
+end
+
+function notimplemented_tangent_guard(
+    da::Mooncake.Tangent{@NamedTuple{re::L,im::L}}, f_sym::Symbol
+) where {L<:Base.IEEEFloat}
+    # re, im tangents
+    return if da.fields != zero_tangent(da).fields.fields
```
Technici4n: This is probably constructing a tangent for `da`, which is quite weird. Maybe `_dot(da, da) != 0` works?

src/tools_for_rules.jl
AstitvaAggarwal: But the same rule can be called again on another singular input value ?

Technici4n: Hmm, I would maybe add "This message is only printed once" then, but still do `maxlog=1`. I think we should avoid generating potentially thousands of messages, especially since the user can't do anything about the actual messages. A single message would be enough to make the user aware that this can be a source of NaNs

AstitvaAggarwal: i get that for sure, but the new log message will be needed to actually distinguish if the NaN returned is valid or not.

AstitvaAggarwal:
Thanks for the thorough reviews @Technici4n ! I think from my side this is ready

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +57,489 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
...
+    @info "Please use Finite Differences. NotImplemented Derivative for f wrt to a - setting ∂f/∂a to NaN."
+end
+
+df = (∂f/∂x)*dx + (∂f/∂y)*dy
+
+if user/program inputs dx != 0 :
+    ∂f = NaN     # @info used + upstream derivatives are simply `NaN` (we dint know the derivative anyways)
+
+if user/program inputs dx == 0 :
+    ∂f = (∂f/∂y)*dy    # upstream derivatives are correct as they dint depend on the "NotImplemented" derivative.
```
sunxd3:
```suggestion
    ∂f = (∂f/∂y)*dy    # upstream derivatives are correct as they don't depend on the "NotImplemented" derivative.
```

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +57,489 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
...
+pullback for f simply returns `NaN`, dy
+
+**Bottomline** : All downstream branches get correct gradients to work, In case they actively use dx they are set to `NaN` + Information given (unknown gradient).
+
+
+**Note** - This will only work for Float point values & it's compositions (Complex etc) due to NaN, NaN32 & NaN16 living only AbstractFloat's Space.
+
+**Final Note** - NaN issues is also solved in a similar manner.
+
+"""
```
sunxd3, line 114: Correct me if I am wrong. This is not just for special functions, right? Maybe it's better to adapt into a proper Mooncake doc? 

`dx` `dy` etc need to be explained

src/tools_for_rules.jl
```diff
@@ -404,6 +446,35 @@ function increment_and_get_rdata!(f, r, t)
...
 end
 
+"""
+    notimplemented_tangent_guard(da::Mooncake.Tangent, f_sym::Symbol)
+
+If da is nonzero and participates in AD, the method returns a `NaN` filled data structure for whatever tangent da is given.
+In case da is a zero tangent return zero-ed da but in a compatible way for immediate complex algebra within Mooncake rules.
+
+"""
+# This cannot have a method for Ints as NaN is only defined for Floats.
```
sunxd3: there can't be empty lines or comments between a doc block and function definition, otherwise the doc won't be treated as the doc for this function

sunxd3 (commented):
Some nits on code comments

ext/MooncakeSpecialFunctionsExt.jl
AstitvaAggarwal: hmm yeah, but that should be for a followup PR as that doc page would also include details about handling `NaN` in the similar way. also this PR is too large anyways haha

yebai: I pushed a few commits to clarify these notes.

src/tools_for_rules.jl
```diff
@@ -404,6 +446,36 @@ function increment_and_get_rdata!(f, r, t)
...
 
+"""
+    notimplemented_tangent_guard(da::Mooncake.Tangent, f_sym::Symbol)
+
+If da is nonzero and participates in AD, the method returns a `NaN` filled data structure for whatever tangent da is given.
+In case da is a zero tangent return zero-ed da but in a compatible way for immediate complex algebra within Mooncake rules.
+
+NOTE: This cannot have a method for `Int`s as `NaN` is only defined for `AbstractFloat`s.
+
+"""
```
yebai, line 469:
```suggestion
"""
    notimplemented_tangent_guard(da::Mooncake.Tangent, f_sym::Symbol)

Guards the use of a tangent associated with a
`ChainRulesCore.NotImplemented` derivative.

If `da` is nonzero and participates in automatic differentiation, this
method returns a `NaN`-filled data structure matching the shape and type
of `da`, signaling an unknown derivative.

If `da` is a zero tangent, a zero-valued tangent is returned in a form
compatible with immediate algebraic composition inside Mooncake rules.

This masking ensures that missing derivatives only affect results when
they are mathematically required.

!!! note
    This function is defined only for floating-point tangent spaces.
    It cannot support `Int` tangents, since `NaN` is only defined for
    `AbstractFloat` types.
"""
```

src/tools_for_rules.jl
yebai: I'd suggest drop the warning message as it can be very noisy. Users will know since returned value is `NaN`.

src/tools_for_rules.jl
```diff
@@ -404,6 +446,36 @@ function increment_and_get_rdata!(f, r, t)
...
+function notimplemented_tangent_guard(da::L, f_sym::Symbol) where {L<:Base.IEEEFloat}
+    return if _dot(da, da) != L(0)
+        @info "Please use Finite Differences. Derivative Not NotImplemented for $f_sym wrt to argument of type $L - setting ∂f/∂arg to NaN."
```
yebai:
```suggestion
```

src/tools_for_rules.jl
```diff
@@ -404,6 +446,36 @@ function increment_and_get_rdata!(f, r, t)
...
+        L(0)
+    end
+end
+
+function notimplemented_tangent_guard(
+    da::Mooncake.Tangent{@NamedTuple{re::L,im::L}}, f_sym::Symbol
+) where {L<:Base.IEEEFloat}
+    # re, im tangents
+    return if _dot(da, da) != L(0)
+        @info "Please use Finite Differences. Derivative Not NotImplemented for $f_sym wrt to argument of type Complex{$L} - setting ∂f/∂arg to NaN."
```
yebai:
```suggestion
```

test/ext/special_functions/special_functions.jl
```diff
@@ -75,4 +77,73 @@ using Mooncake.TestUtils: test_rule
     )
         test_rule(StableRNG(123456), f, x...; perf_flag, is_primitive=false)
     end
+
+    @testset "Primitive SpecialFunctions with NotImplemented Gradients" begin
+        first_arg_types = [Float64, Float32]
+        second_arg_types = [Float64, Float32]
+        Real_type_tests = map_prod(identity, first_arg_types, second_arg_types)
+
+        # Gradient calculations for fields excluding fields with `NotImplemented` gradient type.
```
yebai:
```suggestion
        # Gradient calculations for fields excluding those with `NotImplemented` gradients
```

test/ext/special_functions/special_functions.jl
```diff
@@ -75,4 +77,73 @@ using Mooncake.TestUtils: test_rule
...
+                    (:none, x -> hankelh1x(T(2), x), P(1.5)),
+                    (:none, x -> hankelh2x(T(2), x), P(1.5)),
+
+                    # 2 arg Gamma & Exponential Integrals (1st arg gradient NotImplemented)
+                    (:none, x -> gamma(T(3), x), P(1.5)),
+                    (:none, x -> loggamma(T(3), x), P(1.5)),
+                    (:none, x -> expintx(T(3), x), P(0.5)),
+                    (:none, x -> expint(T(3), x), P(0.5)),
+
+                    # Complex Number args
```
yebai:
```suggestion
                    # Complex number args
```

test/ext/special_functions/special_functions.jl
```diff
@@ -75,4 +77,73 @@ using Mooncake.TestUtils: test_rule
...
+                    (:none, x -> hankelh1(T(3), Complex(x, x)), P(1.5)),
+                    (:none, x -> hankelh2(T(3), Complex(x, x)), P(1.5)),
+                    (:none, x -> besselix(P(0.5), Complex(x, x)), P(1.5)),
+                    (:none, x -> besseljx(P(0.5), Complex(x, x)), P(1.5)),
+                    (:none, x -> besselkx(P(0.5), Complex(x, x)), P(1.5)),
+                    (:none, x -> besselyx(P(0.5), Complex(x, x)), P(1.5)),
+                    (:none, x -> hankelh1x(T(0.5), Complex(x, x)), P(1.5)),
+                    (:none, x -> hankelh2x(T(0.5), Complex(x, x)), P(1.5)),
+
+                    # both args for below functions can be complex
```
yebai:
```suggestion
                    # Both args of the functions below can be complex
```

test/ext/special_functions/special_functions.jl
```diff
@@ -75,4 +77,73 @@ using Mooncake.TestUtils: test_rule
...
+                    (:none, x -> expint(T(3), Complex(x, x)), P(0.5)),
+                    (:none, x -> gamma(Complex(T(3), T(3)), x), P(1.5)),
+                    (:none, x -> loggamma(Complex(T(3), T(3)), x), P(1.5)),
+                    (:none, x -> expintx(Complex(T(3), T(3)), x), P(0.5)),
+                    (:none, x -> expint(Complex(T(3), T(3)), x), P(0.5)),
+                ]
+            end...,
+        )
+
+            # is_primitive = false to test closures over SpecialFunctions.
```
yebai:
```suggestion
            # Use `is_primitive = false` when testing closures over `SpecialFunctions`
```

src/tools_for_rules.jl
```diff
@@ -331,6 +331,11 @@ to_cr_tangent(t::MutableTangent) = CRC.Tangent{Any}(; map(to_cr_tangent, t.field
 to_cr_tangent(t::Tuple) = CRC.Tangent{Any}(map(to_cr_tangent, t)...)
 to_cr_tangent(nt::NamedTuple) = CRC.Tangent{Any}(; map(to_cr_tangent, nt)...)
 
+# Mooncake Complex Number Tangent to CRC style tangent.
```
yebai:
```suggestion
# Convert Mooncake complex-number tangents to `ChainRulesCore`-style tangents.
```

src/tools_for_rules.jl
```diff
@@ -362,6 +367,21 @@ function mooncake_tangent(p::P, t::T) where {P<:Tuple,T<:CRC.Tangent}
     return tangent_type(P) == NoTangent ? NoTangent() : map(mooncake_tangent, p, t.backing)
 end
 
+function mooncake_tangent(p::T, cr_tangent::T) where {P<:IEEEFloat,T<:Complex{P}}
+    return Tangent((re=real(cr_tangent), im=imag(cr_tangent)))
+end
+
+# NotImplemented CRC tangents are converted to NaN.
```
yebai:
```suggestion
# Convert `ChainRulesCore.NotImplemented` tangents to Mooncake-style `NaN`-valued tangents.
```

src/tools_for_rules.jl
```diff
@@ -391,6 +411,28 @@ function increment_and_get_rdata!(f, r, t::CRC.Thunk)
     return increment_and_get_rdata!(f, r, CRC.unthunk(t))
 end
 
+# Mooncake Complex tangents using CRC Complex Tangents.
```
yebai:
```suggestion
# Adding ChainRulesCore-style tangents to existing Mooncake-style tangents
```

src/tools_for_rules.jl
```diff
@@ -391,6 +411,28 @@ function increment_and_get_rdata!(f, r, t::CRC.Thunk)
...
+function increment_and_get_rdata!(
+    f::NoFData, r::Mooncake.RData{@NamedTuple{re::T,im::T}}, t::Complex{T}
+) where {T<:IEEEFloat}
+    return RData((re=real(t) + r.data.re, im=imag(t) + r.data.im))
+end
+
+# return NaN filled tangents for when CRC Complex Tangent is NotImplemented.
```
yebai:
```suggestion
# Return `NaN`-filled Mooncake-style tangents when a ChainRulesCore complex tangent is `NotImplemented`.
```

yebai:
Thanks @AstitvaAggarwal -- I left detailed review comments above.

ext/MooncakeSpecialFunctionsExt.jl
```diff
@@ -43,4 +57,500 @@ import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
 
 @zero_derivative DefaultCtx Tuple{typeof(logfactorial),Integer}
 
+"""
+## Handling `ChainRulesCore.NotImplemented` Tangents in Imported Rules
+
+Mooncake uses a *masking* trick to handle
+`ChainRulesCore.NotImplemented` tangents and partial derivatives.
```
yebai:
```suggestion
`ChainRulesCore.NotImplemented` partial derivatives.
```

src/tools_for_rules.jl
yebai: @AstitvaAggarwal These suggested changes have not yet been committed or accepted. Please
only click “Resolve” once the suggestion has been either accepted or
explicitly rejected (with a stated reason).

AstitvaAggarwal: okay

AstitvaAggarwal:
Thank you @yebai @Technici4n and @sunxd3 for your reviews! This was a big PR

yebai:
@Technici4n @sunxd3, apologies if any of your requested changes remain unresolved. I had to merge this PR because it was impossible to push my local changes to a fork.


@AstitvaAggarwal, please use branches within this repository for future PRs. Please also double check unresolved comments from Xianda and Bruno.

ext/MooncakeSpecialFunctionsExt.jl
AstitvaAggarwal: This is the best I could think of, do you have a something in mind?

ext/MooncakeSpecialFunctionsExt.jl
AstitvaAggarwal: Done 👍

src/tools_for_rules.jl
AstitvaAggarwal: Done

src/tools_for_rules.jl
AstitvaAggarwal: Thanks! Done

AstitvaAggarwal: Done

ext/MooncakeSpecialFunctionsExt.jl
AstitvaAggarwal: I think this is addressed now, the docstring have been edited.

PR #950, opened 2026-01-18
fix: SIGILL issue from combining foldable with stability analysis

MilesCranmer:
@yebai I think this should fix https://github.com/MilesCranmer/DispatchDoctor.jl/issues/103.

It could be an issue in Julia itself but this seems to silence the SIGILL.

### DIFF
--- a/ext/MooncakeSpecialFunctionsExt.jl
+++ b/ext/MooncakeSpecialFunctionsExt.jl
@@ -2,12 +2,26 @@ module MooncakeSpecialFunctionsExt
 
 using SpecialFunctions, Mooncake
 using Base: IEEEFloat
-
-import Mooncake: DefaultCtx, @from_chainrules, @zero_derivative
+import ChainRulesCore as CRC
+import Mooncake:
+    DefaultCtx,
+    @from_chainrules,
+    @from_rrule,
+    @zero_derivative,
+    @is_primitive,
+    Dual,
+    frule!!,
+    Tangent,
+    primal,
+    notimplemented_tangent_guard,
+    ForwardMode,
+    numberify,
+    extract
 
 @from_chainrules DefaultCtx Tuple{typeof(airyai),IEEEFloat}
 @from_chainrules DefaultCtx Tuple{typeof(airyaix),IEEEFloat}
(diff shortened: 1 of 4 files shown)


yebai:
Thanks, @MilesCranmer.

It is surprising that this seemingly innocuous line results in segmentation faults.

MilesCranmer:
Ok I read a bit more into this. My guess is that it's because these functions are not actually `:foldable` when DispatchDoctor is active. This is for a few reasons:

1. `:foldable` implies `:nortcall` which says that `Core.Compiler.return_type` is never used. However, DispatchDoctor relies on `Core.Compiler.return_type`. So this is an invalid effect for the compiler to assume.
2. Then, I think `:consistent` (also implied by `foldable`) is also incorrect. Because, when two things happen:

    1. The inferred type is not concrete, and
    2. DispatchDoctor `@stable` is active

    A function that is type unstable, when analyzed by DD, will throw. However, whether it throws or not can be turned off by the `allow_unstable(::Function)` function context. So this actually violates the `:consistent` effect which is that it always throws the same way. 

3. Finally, it's also incompatible with `:effect_free` since the warn mode might print to stderr. Though that isn't applicable here.



Actually, perhaps a better way to solve this would be to call:

```julia
DispatchDoctor.register_macro!(Symbol("@foldable"), DispatchDoctor.IncompatibleMacro)
```

which would basically turn every `@foldable` into an `@unstable @foldable`.


Or alternatively, perhaps `@foldable` could be turned off when DispatchDoctor is active, and then unstable types could still be caught for stuff like `tangent_type` (which might be happening here)?

yebai:
Both could work in principle. 

Given that `tangent_type` is part of the compiler, the benefit of making sure it is type-stable / performant only improve time-to-first-gradient. Not sure how useful that could be. 

Given all these, I lean towards 

```julia
DispatchDoctor.register_macro!(Symbol("@foldable"), DispatchDoctor.IncompatibleMacro)
DispatchDoctor.register_macro!(Symbol("@mooncake_overlay"), DispatchDoctor.IncompatibleMacro)
```

MilesCranmer:
ok done!

yebai:
Thanks @MilesCranmer!

PR #942, opened 2026-01-14
support matrix-vector or vector-matrix in gemm

yebai:
Fix https://github.com/chalk-lab/Mooncake.jl/issues/622 and other related issues.

TODO: 

- [x] support vector-vector, vector-matrix, matrix-vector cases
- [x] unit tests for all cases
- [x] fix all tests

Issue #486, opened 2025-02-14
Changes for 0.5


willtebbutt:
There are a couple of changes that I want to make that are breaking. In particular,
1. use `Config` directly in `build_rrule`, `prepare_gradient_cache`, and `prepare_pullback_cache`, etc, rather than passing in kwargs, and
2. refactor `prepare_pullback_cache` so that it's possible to use inside DI without requiring a call to `zero_tangent`.

This will bring immediately advantages to DI's interface with Mooncake. For example, if we add fields to `Config` (in a non-breaking way), DI won't need to change to support them. It should also reduce the amount of `kwargs...` expressions floating around in the Mooncake codebase, which can only be a good thing.

Technically non-breaking changes I plan include:
1. removing "convenience" methods of `build_rrule`. I'm anticipating that all users should generally be using `prepare_{gradient,pullback}_cache`, so the convenience methods are redundant. If people need to use this (internal) function, they will obviously still be able to.

I expect to add to this list before actually making the release.

If anyone happens to know of any unsatisfactory bits of Mooncake which, whether technically breaking or just "probably disruptive to someone", ought to go into 0.5, please feel free to mention them here!

aj-fleming:
Far beyond the scope of this issue, but what would happen for Mooncake to need to support the differentiation of vector-valued functions?

willtebbutt:
If by support differentiating vector-valued functions you mean compute Jacobian-like things, then I would suggest taking a look at DifferentiationInterface -- It's default Jacobian computation functionality already works for Mooncake :)

willtebbutt:
As discussed in https://discourse.julialang.org/t/large-memory-consumption-when-using-mooncake-via-differentiationinterface-for-gaussian-process-optimisation/127289/8?u=willtebbutt , there's not currently a good way to clear the cached `OpaqueClosure`s that Mooncake keeps around. We should probably add a function to the interface which promises to completely clear out any cached data that Mooncake currently holds on to.

Technici4n:
Other idea: Set `friendly_tangents=true` by default.

PR #953, opened 2026-01-19
Make `Complex{<:IEEEFloat}` its own tangent type

### LINKED ISSUE #944: Complex float tangent types
## Make `tangent_type` for `Complex{<:BLASFloat}` return the primal complex type
*(avoid `Tangent` wrapper / `<:Number` issues)*

### Context / motivation

In several parts of the [QuantumKitHub](https://github.com/QuantumKitHub) ecosystem we depend on AD for computing gradients of quite complicated loss functions, typically in the context of quantum physics.
There, wave functions often involve complex numbers, so supporting complex numbers is essential for our purposes.
Additionally, our methods involve quite a lot of linear algebra, for which we rely on highly specialized, performance-critical implementations for BLAS-compatible element types (e.g. `Float32/64`, `ComplexF32/64`).
As we are making the move from Zygote to Mooncake (thanks a lot for your wonderful work!), we would like to port over some of the "reverse rules" we have already written, which are often crafted with performance in mind.

Currently, for complex BLAS types, the tangent type appears to be generated as a `Tangent{...}`-style wrapper that is therefore not a `<:Number`.
This has a large surface-area impact: many generic numeric and array operations (and many of our performance-specialized methods) assume element types behave like numbers and/or match existing method signatures like `<:DenseVector{ComplexF64}`.

In practice we want to keep using `Vector{ComplexF64}` / `StridedArray{ComplexF64}` (and friends) so we can:

* dispatch to specialized kernels without proliferating methods for tangent wrapper types,
* use BLAS and other strided-array code paths without extra bridging

Additionally, purely from a mathematical and practical point of view, it does feel somewhat strange to have these tangent types show up, having different behavior for tangent vectors of real and complex numbers.

For a concrete example of the repercussions of the current choice, see e.g. https://github.com/Jutho/StridedViews.jl/pull/19, which requires some non-trivial type gymnastics with reinterpreted arrays, even though we really just have a strided vector of complex numbers.
I'm not really looking for concrete solutions to that particular problem though, this is more of a comment on a recurring issue that looks a little bit like strange design choice from the outside.

### Proposed change

For real BLAS floats, it seems natural that the tangent space is representable by the same scalar type. I’d like to ask whether the same can hold for the complex BLAS variants.

Concretely, would it make sense to define something along the lines of:

```julia
tangent_type(::Type{Complex{T}}) where {T<:BLASFloat} = Complex{T}
# (or whichever BLASFloat alias Mooncake uses for Float32/64)
```

(or more generally: `tangent_type(::Type{Complex{T}}) = Complex{tangent_type(T)}` and have that reduce to `Complex{T}` for BLASFloat `T`).

This would preserve:

* correct additive behavior (componentwise),
* a natural representation for tangents as complex numbers (encoding a 2D real tangent vector),
* easy interop with BLAS/strided-array code and existing dispatch.

So, maybe some questions I have for you are:

1. Are there fundamental reasons that make `tangent_type(Complex{<:BLASFloat}) = Complex{...}` undesirable or incorrect in Mooncake’s design?
2. If the current `Tangent` wrapper is intentional to represent tangents in a “struct-of-components” form, is there a way to make that wrapper behave more like a scalar (e.g. `<:Number`, or provide a lightweight “scalar tangent” path) specifically for BLAS complex types?

lkdvos:
This is the start of my attempt to fix #944.
The core change here is to have `tangent_type(::Type{Complex{T}}) where {T<:IEEEFloat} = Complex{T}`. 

In principle this can later be extended towards other values of `T`, but this requires `tangent_type(T) <: Real`, so for the purpose of this PR I will err on the side of caution and start with a smaller footprint.

Considering versions, I feel like this could be considered a rather substantial change, that might warrant a minor version bump, but on the other hand this shouldn't be making any big changes to the externally visible interface, so I don't really know what should go here.

### DIFF
--- a/ext/MooncakeCUDAExt.jl
+++ b/ext/MooncakeCUDAExt.jl
@@ -82,17 +82,6 @@ function randn_tangent_internal(rng::AbstractRNG, x::CuMaybeComplexArray, dict::
     dict[x] = t
     return t
 end
-# function randn_tangent_internal(
-#     rng::AbstractRNG, x::CuArray{T}, dict::MaybeCache
-# ) where {T<:Complex}
-#     haskey(dict, x) && return dict[x]::tangent_type(typeof(x))
-#     t = tangent_type(typeof(x))(undef, size(x))
-#     t_ = reinterpret(T, t)
-#     th = randn(rng, T, size(x)...)
-#     t_ .= CuArray(th)
-#     dict[x] = t
-#     return t
-# end
 function TestUtils.has_equal_data_internal(
     x::P, y::P, equal_undefs::Bool, d::Dict{Tuple{UInt,UInt},Bool}
 ) where {P<:CuMaybeComplexArray}
--- a/src/rules/blas.jl
+++ b/src/rules/blas.jl
@@ -839,11 +839,12 @@ end
     } where {T<:BlasFloat},
 )
 
+# Helper function to avoid NaN poisoning caused due to adding undef or non initialized C matrices.
 function ifelse_nan(cond, left::P, right::P) where {P<:BlasFloat}
     return isnan(cond) * left + !isnan(cond) * right
 end
 
-function frule!!(
+@inline function frule!!(
     ::Dual{typeof(BLAS.gemm!)},
     transA::Dual{Char},
     transB::Dual{Char},
@@ -1637,9 +1638,6 @@ function hand_written_rule_test_cases(rng_ctor, ::Val{:blas_level_3})
     realPs = [Float64, Float32]
     Ps = [realPs..., complex.(realPs)...]
 
-    _make_codual(x, dx) = CoDual(x, dx)
-    _make_codual(x::Complex{<:IEEEFloat}, dx) = CoDual(x, dx)
-
     test_cases = Any[]
 
     #
@@ -1669,11 +1667,9 @@ function hand_written_rule_test_cases(rng_ctor, ::Val{:blas_level_3})
                 Cs = blas_matrices(rng, P, 3, 5)
 
                 return map(As, Bs, Cs) do A, B, C
-                    a_da = _make_codual(P(α), P(dα))
-                    b_db = _make_codual(P(β), P(dβ))
-                    (
-                        false, :stability, nothing, BLAS.gemm!, tA, tB, a_da, A, B, b_db, C
-                    )
+                    a_da = CoDual(P(α), P(dα))
+                    b_db = CoDual(P(β), P(dβ))
+                    (false, perf_flag, nothing, BLAS.gemm!, tA, tB, a_da, A, B, b_db, C)
                 end
             end
         end...,
@@ -1694,8 +1690,8 @@ function hand_written_rule_test_cases(rng_ctor, ::Val{:blas_level_3})
                 Cs = blas_matrices(rng, P, 3, 1)
 
                 return map(As, Bs, Cs) do A, B, C
-                    a_da = _make_codual(P(α), P(dα))
-                    b_db = _make_codual(P(β), P(dβ))
+                    a_da = CoDual(P(α), P(dα))
+                    b_db = CoDual(P(β), P(dβ))
                     (
                         false, perf_flag, nothing, BLAS.gemm!, tA, 'N', a_da, A, B, b_db, C
                     )
@@ -1721,8 +1717,8 @@ function hand_written_rule_test_cases(rng_ctor, ::Val{:blas_level_3})
                 Cs = blas_matrices(rng, P, 1, 5)
 
                 return map(As, Bs, Cs) do A, B, C
-                    a_da = _make_codual(P(α), P(dα))
-                    b_db = _make_codual(P(β), P(dβ))
+                    a_da = CoDual(P(α), P(dα))
+                    b_db = CoDual(P(β), P(dβ))
                     (false, perf_flag, nothing, BLAS.gemm!, tA, tB, a_da, A, B, b_db, C)
                 end
             end
@@ -1744,8 +1740,8 @@ function hand_written_rule_test_cases(rng_ctor, ::Val{:blas_level_3})
                 Cs = blas_matrices(rng, P, 1, 1)
 
                 return map(As, Bs, Cs) do A, B, C
-                    a_da = _make_codual(P(α), P(dα))
-                    b_db = _make_codual(P(β), P(dβ))
+                    a_da = CoDual(P(α), P(dα))
+                    b_db = CoDual(P(β), P(dβ))
                     (
                         false, perf_flag, nothing, BLAS.gemm!, tA, 'N', a_da, A, B, b_db, C
                     )
@@ -1769,7 +1765,7 @@ function hand_written_rule_test_cases(rng_ctor, ::Val{:blas_level_3})
                 As = blas_matrices(rng, P, R, R)
                 Bs = blas_matrices(rng, P, M, N)
                 return map(As, Bs) do A, B
-                    α_dα = _make_codual(randn(rng, P), P(dα))
+                    α_dα = CoDual(randn(rng, P), P(dα))
                     # 1.10 fails to infer part of a matmat product in the pullback
                     perf_flag = VERSION < v"1.11-" ? :none : :stability
                     (
--- a/src/rules/complex.jl
+++ b/src/rules/complex.jl
@@ -20,16 +20,15 @@ randn_tangent_internal(rng::AbstractRNG, p::CF, ::MaybeCache) = randn(rng, typeo
 __verify_fdata_value(::IdDict{Any,Nothing}, ::P, ::P) where {P<:CF} = nothing
 _verify_rdata_value(::P, ::P) where {P<:CF} = nothing
 
-increment!!(t::T, s::T) where {T<:CF} = t + s
 increment_internal!!(::IncCache, t::T, s::T) where {T<:CF} = t + s
 
 function increment_field!!(x::CF{P}, re_or_im::P, ::Val{FieldName}) where {P,FieldName}
-    return if FieldName === :re
+    return if (FieldName === :re) || (FieldName == 1)
         complex(real(x) + re_or_im, imag(x))
-    elseif FieldName === :im
+    elseif (FieldName === :im) || (FieldName == 2)
         complex(real(x), re_or_im + imag(x))
     else
-        throw(ArgumentError(lazy"Unkown field `$Fieldname` for type `$(CF{P})`)"))
+        throw(ArgumentError(lazy"Unkown field `$FieldName` for type `$(CF{P})`)"))
     end
 end
 
--- a/test/rules/complex.jl
+++ b/test/rules/complex.jl
@@ -1,11 +1,7 @@
 @testset "complex" begin
     rng = sr(123)
     p = Complex{Float64}(5.0, 4.0)
-    TestUtils.test_tangent_interface(rng, p)
-    TestUtils.test_tangent_splitting(rng, p)
-    # TestUtils.run_rule_test_cases(StableRNG, Val(:complex))
+    TestUtils.test_data(rng, p)
     p = Complex{Float32}(5.0, 4.0)
-    TestUtils.test_tangent_interface(rng, p)
-    TestUtils.test_tangent_splitting(rng, p)
-    # TestUtils.run_rule_test_cases(StableRNG, Val(:complex))
+    TestUtils.test_data(rng, p)
 end


Technici4n:
Yes, this will potentially need a minor bump. To be discussed once it's about to be merged I suppose.

Regarding the allocs, could it be that `tangent_type(::Type{CF{P}})` is not specialized? https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing

src/rules/complex.jl
```diff
@@ -0,0 +1,75 @@
...
+
+    function lgetfield_Complex_pullback(Δy_rdata)
+        if FieldName === :re
+            Δx = complex(Δy_rdata, zero(Δy_rdata))
+        elseif FieldName === :im
+            Δx = complex(zero(Δy_rdata), Δy_rdata)
+        end
+        return NoRData(), Δx, NoRData()
+    end
+    return y_cd, lgetfield_A_pullback
```
Confusio: `lgetfield_A_pullback` should be `lgetfield_Complex_pullback` here?

lkdvos: Thank you, indeed!

lkdvos:
So it turns out that even the main branch is giving me these allocations, but only on my mac. I've tested on linux, and no such allocations are happening.
Since this is therefore probably unrelated to this PR, I will ignore this for now, but happy to report an issue if this could be helpful?

> Regarding the allocs, could it be that tangent_type(::Type{CF{P}}) is not specialized?

This was also what I was initially thinking, but I think that `f(::Type{CF{P}}) where {P}` actually does specialize due to the `where {P}` clause, so this should be fine (can confirm on my linux machine).

Anyways, I'll investigate the remainder of the rules etc, I think at least the tangent type tests are now passing 🎉

.github/workflows/CI.yml
```diff
@@ -45,6 +45,7 @@ jobs:
           'rules/random',
           'rules/tasks',
           'rules/twice_precision',
+          'rules/complex',
```
Technici4n: Nit: keep it sorted :D

src/rules/complex.jl
```diff
@@ -0,0 +1,101 @@
+const ComplexFloat{P<:IEEEFloat} = Complex{P}
+const CF{P} = ComplexFloat{P}
```
Technici4n, line 2: Could we support an arbitrary `P` for `Complex{P}`? While we won't be able to test it, I think it would be extremely unexpected to suddenly encounter the `Tangent{@NamedTuple{...}}` for some types and a `Complex` for floats. Is this fine @yebai?

src/rules/blas.jl
```diff
@@ -29,54 +29,45 @@ primal. This operation is not guaranteed to be possible for all array types, but
 possible for all array types of interest so far.
 """
 function arrayify(
-    x::Union{Dual{A},CoDual{A}}
-) where {T<:Union{IEEEFloat,BlasFloat},A<:Union{AbstractArray{T},Ptr{<:T}}}
+        x::Union{Dual{A}, CoDual{A}}
+    ) where {T <: Union{IEEEFloat, BlasFloat}, A <: Union{AbstractArray{T}, Ptr{<:T}}}
     return arrayify(primal(x), tangent(x))  # NOTE: for complex numbers, tangents are reinterpreted to Complex
```
Technici4n: This comment can now be removed!

src/rules/blas.jl
```diff
@@ -99,61 +90,57 @@ function viewify(
...
     xinds = 1:incx:(incx * n)
     return view(x, xinds), view(dx, xinds)
 end
 
-numberify(x::BlasRealFloat) = x
-function numberify(x::Tangent{@NamedTuple{re::P,im::P}}) where {P<:BlasRealFloat}
-    complex(x.fields.re, x.fields.im)
-end
+numberify(x::BlasFloat) = x
 numberify(x::Dual) = primal(x), numberify(tangent(x))
```
Technici4n: `numberify` can be removed, and usages for (Co)Duals can be replaced by `extract`.

src/rules/blas.jl
```diff
@@ -99,61 +90,57 @@ function viewify(
...
-_rdata(x::BlasRealFloat) = x
-_rdata(x::BlasComplexFloat) = RData((; re=real(x), im=imag(x)))
+_rdata(x::BlasFloat) = x
```
Technici4n: `_rdata` can be removed, this was a workaround to convert `Complex` numbers back to the corresponding RData in reverse mode.

src/rules/blas.jl
```diff
@@ -1704,6 +1693,7 @@ function hand_written_rule_test_cases(rng_ctor, ::Val{:blas_level_3})
     Ps = [realPs..., complex.(realPs)...]
 
     _make_codual(x, dx) = CoDual(x, dx)
+    _make_codual(x::Complex{<:IEEEFloat}, dx) = CoDual(x, dx)
     _make_codual(x::Complex, dx) = CoDual(x, Tangent((; re=real(dx), im=imag(dx))))
```
Technici4n: This one can be deleted. More generally I would suggest a search for `re`/`im` in the codebase to see if there are other such overloads that can be removed.

src/tools_for_rules.jl
```diff
@@ -323,6 +323,7 @@ end
 Convert a Mooncake tangent into a type that ChainRules.jl `rrule`s expect to see.
 """
 to_cr_tangent(t::IEEEFloat) = t
+to_cr_tangent(t::Complex{<:IEEEFloat}) = t
```
Technici4n, line 326: The `c::Tangent{@NamedTuple{re::T,im::T}}` one below can be removed.

src/Mooncake.jl
```diff
@@ -151,6 +151,7 @@ include(joinpath("rules", "new.jl"))
 include(joinpath("rules", "random.jl"))
 include(joinpath("rules", "tasks.jl"))
 include(joinpath("rules", "twice_precision.jl"))
+include(joinpath("rules", "complex.jl"))
```
Technici4n: nit: keep them ordered

src/rules/complex.jl
yebai: It is better to relax this gradually, justified by necessity.

Technici4n: It seems to increase the amount of code a little for no benefits. I could see Mooncake treating all `<:Real` types as numbers automatically in the future (so with the same tangent and rdata type, and no fdata). That would also address `BFloat16` support fairly easily.

yebai: The tangent types are acceptable here, but writing sufficiently flexible rules may be challenging. Experience with prior AD systems suggests that if these cases cannot be exercised in CI, they are likely to cause problems (correctness issues or hard-to-debug errors). 

This is not a hard rule, but the point is that someone needs to think about them very carefully before putting them into implementation. 

EDIT: see https://github.com/chalk-lab/Mooncake.jl/issues/961

lkdvos: Am I understanding this correct that we can, for the purpose of this PR, keep the scope limited to the `IEEEFloat` types?
While I am not necessarily opposed to further broadening the scope, I don't feel super comfortable at the moment to do so, mostly since we lack tests for that. In other words, I think for that we should first add tests that are currently working, then alter the tangent type, and then verify that didn't break anything.

The main issue I can currently see with actually making `tangent_type(Complex{T}) = Complex{tangent_type(T))` is that we have no guarantee that `tangent_type(::Type{<:Real})::Type{<:Real}`, and this is required to be able to put it as a type parameter in the `Complex` again.
For example:
```julia
using Mooncake
struct MyReal <: Real
    x::Float64
end
Complex{tangent_type(MyReal)}
ERROR: TypeError: in Complex, in T, expected T<:Real, got Type{Mooncake.Tangent{@NamedTuple{x::Float64}}}
```

So, unless we can guarantee that all `T <: Real` have tangents `<: Real`, we can't actually achieve this just yet.

src/tools_for_rules.jl
lkdvos: I might be mistaken, but I think we still need that for the cases where `!(T <: IEEEFloat)`?

lkdvos:
Apologies for the repeated force-push etc, I got myself into some rebase/merge issues and decided to just squash the entire thing and go from there.

This being said, I think this might now be in a position to get a true review, I'd be very interested in hearing your thoughts.

Some comments I had myself:

- There is a somewhat odd random `import` statement in the line quoted below, which makes it such that keeping the `include` statements sorted actually does make a difference. I would suggest to either always qualify, or put this `import` in the Mooncake.jl file, as this has subtle namespace consequences that might be unintended or unexpected.
https://github.com/chalk-lab/Mooncake.jl/blob/ae44b48d85dad52fb951a917af67191dbd481024/src/rules/misty_closures.jl#L99

- Repeating something I put in a reply to a comment so it doesn't get lost in the black hole of github comments: for the purpose of this PR I suggest to only consider `Complex{<:IEEEFloat}`. The main reason against fully generalizing this to arbitrary `<:Real` is the following example of something that would break if I tried to implement this in full generality. I'm happy to revisit this later, but I don't think this should block this PR.
```julia
using Mooncake
struct MyReal <: Real
    x::Float64
end
Complex{tangent_type(MyReal)}
ERROR: TypeError: in Complex, in T, expected T<:Real, got Type{Mooncake.Tangent{@NamedTuple{x::Float64}}}
```

ext/MooncakeCUDAExt.jl
```diff
@@ -68,164 +64,79 @@ tangent_type(::Type{CuContext}) = NoTangent
...
     haskey(dict, x) && return dict[x]::tangent_type(typeof(x))
-    t = tangent_type(typeof(x))(undef, size(x))
-    t_ = reinterpret(T, t)
-    th = randn(rng, T, size(x)...)
-    t_ .= CuArray(th)
+    t = CuArray(randn(rng, eltype(x), size(x)...))
     dict[x] = t
     return t
 end
+# function randn_tangent_internal(
```
Technici4n: leftover?

src/rules/blas.jl
```diff
@@ -1729,7 +1671,9 @@ function hand_written_rule_test_cases(rng_ctor, ::Val{:blas_level_3})
                 return map(As, Bs, Cs) do A, B, C
                     a_da = _make_codual(P(α), P(dα))
                     b_db = _make_codual(P(β), P(dβ))
-                    (false, perf_flag, nothing, BLAS.gemm!, tA, tB, a_da, A, B, b_db, C)
+                    (
+                        false, :stability, nothing, BLAS.gemm!, tA, tB, a_da, A, B, b_db, C
```
Technici4n: It's interesting that the perf_flag could be dropped, I guess why not.

src/rules/complex.jl
```diff
@@ -0,0 +1,101 @@
...
+zero_rdata_from_type(P::Type{<:CF}) = zero(P)
+
+set_to_zero_internal!!(::SetToZeroCache, p::CF) = zero(p)
+
+randn_tangent_internal(rng::AbstractRNG, p::CF, ::MaybeCache) = randn(rng, typeof(p))
+
+__verify_fdata_value(::IdDict{Any,Nothing}, ::P, ::P) where {P<:CF} = nothing
+_verify_rdata_value(::P, ::P) where {P<:CF} = nothing
+
+increment!!(t::T, s::T) where {T<:CF} = t + s
```
Technici4n: this one shouldn't be needed

test/interface.jl
```diff
@@ -129,11 +129,9 @@ using Mooncake:
             z̄ = 0.3 + 0.7im
 
             cache = Mooncake.prepare_pullback_cache(testf, z)
```
Technici4n, line 131: I guess this now makes it a bad test for friendly tangents, but if you don't want to update it it's also fine. (What would one even use).

test/rules/complex.jl
```diff
@@ -0,0 +1,11 @@
+@testset "complex" begin
+    rng = sr(123)
+    p = Complex{Float64}(5.0, 4.0)
+    TestUtils.test_tangent_interface(rng, p)
+    TestUtils.test_tangent_splitting(rng, p)
```
Technici4n: Shouldn't `TestUtils.test_data` be enough to handle both?

Technici4n (commented):
Looks great! Restricting to IEEEFloat for now is fine of course.

yebai:
Looks good overall—can you double-check whether any changes in #942 were reverted due to a weird git issue?

src/rules/blas.jl
lkdvos: This is definitely an unintended change, I'll revert since 1. I don't know what this means and 2. if this should be changed it is probably better left for a dedicated PR

test/interface.jl
lkdvos: I also have to admit that I didn't try to understand what was going on here and just made the changes so everything runs. I'll leave this as-is since I indeed don't really know what we would even use instead, but let me know if you do have inspiration, happy to try something else in that case

lkdvos:
Thank you for the review comments!
I double-checked for the merge and I am now pretty confident everything is in order (I was missing a small comment and an `@inline` annotation indeed, good catch!).
I've addressed the remainder of @Technici4n's comments as well, most were indeed just a matter of small oversights on my part or me not fully grasping everything in the internals of the library.
The main thing that still remains is deciding what kind of version change should be associated to this PR, otherwise I think this should be ready (assuming everything passes now, I will look into it if not)

Technici4n:
LGTM now, thank you. I'll let @yebai figure out the versioning. 😄 If we go for 0.5 (which we probably should) this might also be a good opportunity to address some of the items in https://github.com/chalk-lab/Mooncake.jl/issues/486

yebai:
Please feel free to merge and release. 

Bruno: Please feel free to make a breaking release based on this PR and set `friendly_tangent=true` by default.

Technici4n:
Okay thanks I'll be working on that tomorrow then!

Technici4n:
I will already merge this PR and should have #972 ready in a few hours -- this includes finding a new example for friendly/unfriendly tangents, so we can have a 0.5 release with both PRs by this evening. :)

yebai:
For posterity, the type constraint in this PR does not work as
expected due to a counterintuitive Julia behaviour: type aliasing does
not propagate parameter constraints.

For example:

```julia
const ComplexFloat{P<:Base.IEEEFloat} = Complex{P}
const CF{P} = ComplexFloat{P}
```

Methods such as

- `tangent_type(::Type{CF{P}}) where {P}`
- `tangent_type(::Type{NoFData}, ::Type{CF{P}}) where {P}`

(and even direct uses of `CF{P})` will accept `P = Int`, despite the
original `P<:IEEEFloat` constraint on `ComplexFloat{P}`. Aliases do not
enforce the bounds of the types they refer to.

As a result, enforcing the constraint directly on the concrete type,
i.e. `tangent_type(::Type{CF{P}}) where {P<:IEEEFloat}`, is
necessary. This change (ad0c3c2) actually resolves the issue reported in
https://github.com/chalk-lab/Mooncake.jl/issues/974.

Issue #914, opened 2025-12-30
Prevent inlining in optimize_ir! during forward over reverse


yebai:
Ref
- https://github.com/chalk-lab/Mooncake.jl/pull/878#issuecomment-3666924752
- https://github.com/chalk-lab/Mooncake.jl/pull/878#issuecomment-3676865618

PR #954, opened 2026-01-19
Prevent inlining of rules call during f-o-r

sunxd3:
Fix https://github.com/chalk-lab/Mooncake.jl/issues/914

To summarize the problem, higher-order differentiation (e.g., forward-over-reverse for Hessians) fails because primitive rule calls get inlined during IR optimization, exposing their internals which have no rules defined (because we defined rules for the function that got inlined).

The core idea of this PR is marking primitive `rrule!!` calls with `NoInlineCallInfo` after inference but before inlining in `optimise_ir!` (and teach `BugPatchInterpreter` to respect the oninlining marker). This prevents the inliner from substituting the rule body. To enable this without impacting first-order AD, a new `noinline_rules` parameter (default `false`) is added and threaded through the AD infrastructure(`build_rrule`/`build_frule` → `generate_ir` → `ADInfo`/`DualInfo` → `optimise_ir!`) and stored in `LazyDerivedRule`/`DynamicDerivedRule` for nested rule building.

The `noinline_rules` mechanism prevents inlining, but forward mode still sees `rrule!!(f, ...)` as a regular call and will differentiate through its body. So this PR also made changes to `@zero_derivative`: the macro now registers `rrule!!(f, ...)` as a forward-mode primitive with an `frule!!` returning zero, preventing forward mode from differentiating through the `rrule!!` body.

The main files to review are `ir_utils.jl` and `tools_for_rules.jl`.

_The DITest on 1.11 is expected to fail at this point (because I commented out the patch). The new argument (`noinline_rules`) will need to passed to reverse mode rule generation, some options are: 1. via `Config` 2. Mooncake expose `Hessian` interface directly 3. better ideas?._

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -95,7 +95,6 @@ jobs:
           {test_type: 'ext', label: 'luxlib'},
           {test_type: 'ext', label: 'nnlib'},
           {test_type: 'ext', label: 'special_functions'},
-          {test_type: 'integration_testing', label: 'format'},
           {test_type: 'integration_testing', label: 'array'},
           {test_type: 'integration_testing', label: 'bijectors'},
           {test_type: 'integration_testing', label: 'diff_tests'},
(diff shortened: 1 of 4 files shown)


sunxd3:
closed in favor of #962

PR #956, opened 2026-01-20
Chunked Forward Mode proof of concept

Technici4n:
Very much a POC ;)

This is what I had in mind: adding an `N` parameter to `Dual` and storing the tangents in an `NTuple`.

### DIFF
--- a/src/tangents/dual.jl
+++ b/src/tangents/dual.jl
@@ -7,14 +7,14 @@ is computed, and `tangent` the direction in which it is computed.
 
 Must satisfy `tangent_type(P) == T`.
 """
-struct Dual{P,T}
+struct Dual{P,T,N}
     primal::P
-    tangent::T
+    tangents::NTuple{N,T}
 end
 
 primal(x::Dual) = x.primal
-tangent(x::Dual) = x.tangent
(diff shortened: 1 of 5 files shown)


devmotion:
> adding an N parameter to Dual and storing the tangents in an NTuple.

This is was ForwardDiff is doing as well (the only difference would be that it wraps the NTuple of tangents in a custom Partials type, presumably to be able to eg define `zero` for the tuple of tangents without committing ype piracy).

Technici4n:
Yes, I was inspired by ForwardDiff, although I can't say for sure if it was conscious or not. 😄

I'm not sure there would be a large benefit to wrapping the tangents tuple in another struct. I am also trying to keep `Dual` as close to `CoDual` as possible.

src/tangents/dual.jl
```diff
@@ -7,14 +7,14 @@ is computed, and `tangent` the direction in which it is computed.
 
 Must satisfy `tangent_type(P) == T`.
 """
-struct Dual{P,T}
+struct Dual{P,T,N}
     primal::P
-    tangent::T
+    tangents::NTuple{N,T}
```
AstitvaAggarwal: This interleaves data in memory especially for struct tangent types-which would prevents the compiler from using SIMD.

PR #962, opened 2026-01-21
Use `build_primitive_rrule` wrapper to block inlining

sunxd3:
Fix https://github.com/chalk-lab/Mooncake.jl/issues/914

Higher-order differentiation (e.g., forward-over-reverse for Hessians) fails because primitive rule calls get inlined during IR optimization, exposing their internals which have no rules defined.

This PR uses `build_primitive_rrule` to return a wrapper struct `PrimitiveRRule{Sig, Tmaybeinline}` instead of directly returning `rrule!!`. When `Tmaybeinline=false`, the wrapper forces a `@noinline` call boundary around `rrule!!`, preventing the compiler from inlining the rule body.

A `maybeinline_primitive` parameter (default `true`) is added to `ADInfo` and threaded through both reverse and forward mode infrastructure. When set to `false` (for higher-order AD), primitives are wrapped with the noinline variant.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.5.1"
+version = "0.5.2"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/test/ext/dynamic_expressions/Project.toml
+++ b/test/ext/dynamic_expressions/Project.toml
@@ -1,6 +1,5 @@
 [deps]
 AllocCheck = "9b6a8646-10ed-4001-bbdc-1d2f46dfbb1a"
-DifferentiationInterface = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 DynamicExpressions = "a40a106e-89c9-4ca8-8020-a735e8728b6b"
 JET = "c3a54625-cd67-489e-a8e7-0a5a0ff4e31b"
 Mooncake = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
--- a/test/ext/dynamic_expressions/dynamic_expressions.jl
+++ b/test/ext/dynamic_expressions/dynamic_expressions.jl
@@ -3,14 +3,13 @@ Pkg.activate(@__DIR__)
 Pkg.develop(; path=joinpath(@__DIR__, "..", "..", ".."))
 
 using Mooncake
-using Mooncake: Mooncake
+using Mooncake: Mooncake, prepare_gradient_cache, value_and_gradient!!
 using Mooncake.TestUtils
 using Mooncake.TestUtils: test_rule, test_data
 using Optim: Optim
 using DynamicExpressions
 using DynamicExpressions: Nullable
 using StableRNGs: StableRNG
-using DifferentiationInterface: AutoMooncake, gradient, prepare_gradient
 
 # Needed for certain parts of TestUtils
 using JET: JET
@@ -46,16 +45,17 @@ end
         eval_sum = let f = f
             X -> sum(f(X))
         end
-        backend = AutoMooncake(; config=nothing)
-        prep = prepare_gradient(eval_sum, backend, X)
-        dX = gradient(eval_sum, prep, backend, X)
+
+        cache = prepare_gradient_cache(eval_sum, X)
+        y, dX = value_and_gradient!!(cache, eval_sum, X)
+        d_f, d_X = dX
 
         # analytic derivative: df/dx1 = 1, df/dx2 = -sin(x2 - 0.2), df/dx3 = 0
         dX_ref = zeros(size(X))
         dX_ref[1, :] .= 1
         dX_ref[2, :] .= -sin.(X[2, :] .- 0.2)
         # third row already zero
-        @test isapprox(dX, dX_ref; rtol=1e-10, atol=0)
+        @test isapprox(d_X, dX_ref; rtol=1e-10, atol=0)
     end
 end
 
@@ -74,11 +74,13 @@ end
                 f -> sum(f(X))
             end
 
-            backend = AutoMooncake(; config=nothing)
-            prep = prepare_gradient(eval_sum, backend, expr)
-            dexpr = gradient(eval_sum, prep, backend, expr)
+            cache = prepare_gradient_cache(
+                eval_sum, expr; config=Mooncake.Config(; friendly_tangents=true)
+            )
+            y, dfexpr = value_and_gradient!!(cache, eval_sum, expr)
+            d_f, d_expr = dfexpr
 
-            const_tangent = dexpr.fields.tree.children[2].x.val
+            const_tangent = d_expr.tree.children[2].x.val
             @test const_tangent ≈ N
         end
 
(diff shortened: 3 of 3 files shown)


src/interpreter/reverse_mode.jl
```diff
@@ -704,12 +709,16 @@ function make_ad_stmts!(stmt::Expr, line::ID, info::ADInfo)
         sig = Tuple{arg_types...}
         interp = info.interp
         raw_rule = if is_primitive(context_type(interp), ReverseMode, sig, interp.world)
-            rrule!! # intrinsic / builtin / thing we provably have rule for
+            if info.noinline_primitives
+                build_primitive_rrule(sig)
+            else
+                rrule!!
+            end # intrinsic / builtin / thing we provably have rule for
```
yebai: Can we also refactor to pass only `info::ADInfo` around since it contains all the information? 

```suggestion
            build_primitive_rrule(sig; maybeinline_primitive = info.maybeinline_primitive)
```

src/Mooncake.jl
```diff
@@ -104,7 +104,13 @@ callable `struct` with type parameters which are the result of this computation.
 context, the motivation for using this function is the same as that of using staged
 programming (e.g. via `@generated` functions) more generally.
 """
-build_primitive_rrule(::Type{<:Tuple}) = rrule!!
+struct PrimitiveRRule{Sig} end
```
yebai:
```suggestion
struct PrimitiveRRule{Sig, Tmaybeinline} end
```

src/Mooncake.jl
```diff
@@ -104,7 +104,13 @@ callable `struct` with type parameters which are the result of this computation.
...
+
+@noinline function (rule::PrimitiveRRule{Sig})(args...) where {Sig}
+    return tuple_splat(rrule!!, args)
+end
```
yebai: @inline `build_primitive_rrule` is also for other purposes, so it's better to let `build_primitive_rrule` always return a wrapper then let the wrapper decide whether to inline primitives or not. For example 


```suggestion
function (rule::PrimitiveRRule{Sig, Tmaybeinline})(args...) where {Sig, Tmaybeinline < Bool}
    return if Tmaybeinline
              tuple_splat(rrule!!, args)
         else
              @noinline tuple_splat(rrule!!, args)
         end 
end
```

src/Mooncake.jl
```diff
@@ -104,7 +104,13 @@ callable `struct` with type parameters which are the result of this computation.
...
+
+build_primitive_rrule(sig::Type{<:Tuple}) = PrimitiveRRule{sig}()
```
yebai:
```suggestion
build_primitive_rrule(sig::Type{<:Tuple}, maybeinline_primitive::Bool) = PrimitiveRRule{sig, maybeinline_primitive}()
```

src/interpreter/reverse_mode.jl
```diff
@@ -131,6 +131,7 @@ struct ADInfo
     arg_rdata_ref_ids::Dict{Argument,ID}
     ssa_rdata_ref_ids::Dict{ID,ID}
     debug_mode::Bool
+    noinline_primitives::Bool
```
yebai: If `maybeinline_primitive=true`, maybe inline; otherwise never inlines. 

```suggestion
    maybeinline_primitive::Bool
```

yebai (commented):
Thanks @sunxd3 -- it looks sound. 

For consistency, I suggest we add the same `maybeinline_primitive` flag to ForwardMode. Alternatively, we can default this to true in `build_primative_rule` and remove the need to pass around the `maybeinline_primitive` flag.

src/Mooncake.jl
```diff
@@ -82,15 +82,57 @@ pb!!(1.0)
...
+    return Tuple{Vararg{Any,nparams}}
+end
+
+@inline function (rule::PrimitiveRRule{Sig,true})(args...) where {Sig}
+    return tuple_splat(rrule!!, args)
+end
+
+@noinline function (rule::PrimitiveRRule{Sig,false})(args...) where {Sig}
+    return rrule!!(args...)
+end
```
yebai:
```suggestion
@inline function (rule::PrimitiveRRule{Sig,false})(args...) where {Sig}
    return tuple_splat_noinline(rrule!!, args)
end

# illustrative code below
# we need to put these in the right place

for N in 1:256
    @eval function tuple_splat_noinline(f, x::Tuple{Vararg{Any,$N}})
        @noinline f($(map(n -> :(getfield(x, $n)), 1:N)...))
    end
end

@generated function tuple_splat_noinline(f, x::Tuple)
    Expr(
        :block,
        Expr(
            :meta, :noinline,
            Expr(
                :call,
                :f,
                map(n -> :(getfield(x, $n)), 1:length(x.parameters))...
            )
        )
    )
end
```

yebai:
@sunxd3, can we pass `maybeinline_primitive=true` via `Mooncake.Config`, which is now part of DI's interface (ie, `AutoMooncake(Mooncake.Config(maybeinline_primitive=true))` is passed through to `prepare_gradient_cache`)?

src/config.jl
```diff
@@ -17,9 +17,14 @@ Configuration struct for use with `ADTypes.AutoMooncake`.
     The tangent is converted from/to the friendly representation at the interface level,
     so all Mooncake internal computations and rule implementations always use the
     [`tangent_type`](@ref) representation.
+- `maybeinline_primitive::Bool=true`: if `false`, primitive rules are wrapped with a
```
Technici4n, line 20: This doesn't seem useful to external users, why is it exposed here?

src/utils.jl
```diff
@@ -63,6 +63,60 @@ end
...
+The two-argument form calls `f(x[1], x[2], ..., x[N])`.
+The three-argument form calls `f(v, x[1], x[2], ..., x[N])`.
+
+This function is marked `@inline`, allowing the call to be inlined into the caller.
+
+See also: [`tuple_splat_noinline`](@ref)
+"""
+tuple_splat
+
+for N in 1:256
```
Technici4n, line 85: Why, is the `@generated` one below not sufficient?

sunxd3:
Yeah, that sounds like a good idea. I'd love to see how that works out

yebai:
> Could we not instead skip optimise_ir! when we know that the rule will be differentiated over? Potentially moving the optimization to the frule for _build_rule? 

IIRC, there may be performance implications to watch if you skip `optimise_ir! ` until forward-mode. But it is hard to anticipate how much that would be.

MasonProtter:
This should probably just be done via a pass that sets whether or not the `rrule!!` itself should be inlined. 

There's a few ways to do this, but one rather coarse (and easy) one would be to add a little logic to this check: https://github.com/chalk-lab/Mooncake.jl/blob/main/src/interpreter/abstract_interpretation.jl#L164-L182 and then if we're in an higher-order AD context, then you also emit the `NoInlineCallInfo` for `rrule!!`. 

This would at least be significantly simpler

sunxd3:
https://github.com/chalk-lab/Mooncake.jl/pull/954 is an alternative that's more more inline to what Mason is suggesting

src/utils.jl
yebai: There are patterns we already use in Mooncake, likely for performance and specialisation reasons. See, https://github.com/chalk-lab/Mooncake.jl/blob/7ec22cb8fc580988d92a6f7e0f979ce5ed669b55/src/utils.jl#L52-L56

src/config.jl
yebai: DifferentiationInterface uses it.

yebai:
I support merging this PR. The core mechanism is sound, and the use of the `maybeinline_primitive ` flag is motivated by the need to minimise performance overhead. I am content to wait a few additional days before merging; however, I will proceed unless an alternative PR is submitted to address this.

Technici4n:
@yebai I will see what I can do. While this PR fixes a real problem with f-o-r, leaking workarounds into normal reverse mode and into the Mooncake API will make future evolution more difficult.

Technici4n:
Alright here it is: https://github.com/chalk-lab/Mooncake.jl/pull/990

sunxd3:
close in favor of #990

sunxd3:
https://github.com/chalk-lab/Mooncake.jl/blob/d040047575db51d5675d8ff3d482c4962fd7c809/src/Mooncake.jl#L86-L180 still might be a good reference for future

PR #967, opened 2026-01-22
Support adjoint and transpose for CuArrays

kshyatt:
Added the missing `frule!!`s for `_new_`. A little unsure of what I'm doing so feedback very welcome!


yebai:
@kshyatt All type constructors are lowered to `_new_` (see the
[documentation](https://chalk-lab.github.io/Mooncake.jl/stable/developer_documentation/internal_docstrings/#Mooncake._new_-Union{Tuple{N},%20Tuple{T},%20Tuple{Type{T},%20Vararg{Any,%20N}}}%20where%20{T,%20N})).
Consequently, we only need to define rules for `_new_` on tangent types (see,
for example, [this implementation](https://github.com/chalk-lab/Mooncake.jl/blob/cc0e38c1a2ffde0e094ba04304ec22c728d1fb84/src/rules/twice_precision.jl#L65-L78).
Once such rules are in place, the code below can be removed.

https://github.com/chalk-lab/Mooncake.jl/blob/cc0e38c1a2ffde0e094ba04304ec22c728d1fb84/ext/MooncakeCUDAExt.jl#L262-L288

In addition, I do not think we need a rule for `_new_` unless the constructed type
is a tangent type. In this PR, the `Adjoint` type may not be a tangent type; if
that is the case, then no rules are required for it.

Please feel free to reach out to @sunxd3 if you have any further questions. 

EDIT: see,  https://github.com/chalk-lab/Mooncake.jl/pull/970

PR #970, opened 2026-01-22
clarifying rules for _new_, lgetfield, and lsetfield! in customised tangents #968

### LINKED ISSUE #968: clarifying rules for `_new_`, `lgetfield`, and `lsetfield!` in customised tangents
On the customised tangent [page](https://chalk-lab.github.io/Mooncake.jl/stable/developer_documentation/custom_tangent_type/#6.-From-%22It-Works!%22-to-Passing-[TestUtils.test_data](@ref)), it is stated that rules for the above functions
are required. However, the rationale is not entirely clear. In particular, it
would be helpful to explain that Mooncake implicitly lowers the constructors of
all types to `_new_`. As a result, there is no need to write rules for individual
constructors; instead, rules should be written for `_new_`.

As for `lgetfield` and `lsetfield!`, the motivation is type stability

```
    lgetfield(x, f::Val)

An implementation of `getfield` in which the field `f` is specified statically via a
`Val`. This enables the implementation to be type-stable even when it is not
possible to constant-propagate `f`. Moreover, it enables the pullback to also be type-stable.

It will always be the case that

getfield(x, :f) === lgetfield(x, Val(:f))
getfield(x, 2) === lgetfield(x, Val(2))

This approach is identical to the one taken by `Zygote.jl` to circumvent the same problem.
`Zygote.jl` calls the function `literal_getfield`, while we call it `lgetfield`.
```

yebai:
Fix #968

### DIFF
--- a/docs/src/developer_documentation/forwards_mode_design.md
+++ b/docs/src/developer_documentation/forwards_mode_design.md
@@ -243,7 +243,7 @@ There are three cases to consider, in order of preference:
 Primitives:
 
 If `is_primitive` returns `true` when applied to the signature constructed from the static types of `f` and `x`, then we simply replace the expression with `Expr(:call, frule!!, f, x...)`, regardless whether we have an `:invoke` or `:call` expression.
-(Due to the [Standardisation](@ref) steps, it regularly happens that we see `:call` expressions in which we actually do know enough type information to do this, e.g. for `Mooncake._new_` `:call` expressions).
+(Due to the [Standardisation](@ref standardisation) steps, it regularly happens that we see `:call` expressions in which we actually do know enough type information to do this, e.g. for `Mooncake._new_` `:call` expressions).
 
 Static Dispatch:
 


yebai:
@kshyatt, can you also take a look at whether this is helpful?

docs/src/developer_documentation/custom_tangent_type.md
```diff
@@ -218,6 +218,37 @@ You must provide adjoints for every `getfield`/`lgetfield` variant that appears
 | [`_new_`](@ref)      | `A(x)`, `A(x, a::A)`, `A(x, nothing)`—three separate `rrule!!` methods     |
 | [`lsetfield!`](@ref) | `(A, Val{:field}, new_value)` including both Symbol & Int field IDs        |
 
+##### Why These Primitives?
+
+**`_new_`**: Mooncake implicitly lowers the constructors of all types to `_new_`. As a result, there is no need to write rules for individual constructors; instead, rules should be written for `_new_`. This provides a uniform interface for differentiating object construction.
```
sunxd3:
```suggestion
**`_new_`**: Mooncake's IR normalization rewrites `Expr(:new, ...)` (the lowered form of composite type construction) into calls to `_new_`. This provides a uniform interface for differentiating object construction. In practice, you typically only need rules for `_new_` to cover the allocation step, rather than writing a rule per constructor method, but any additional logic inside constructors still needs rules as usual. Splatted constructions are normalized to `_splat_new_` in the same way. You can inspect `@code_lowered A(1.0)` to see the raw `:new` form. See the Standardisation section in `docs/src/developer_documentation/forwards_mode_design.md` and the implementation in `src/interpreter/ir_normalisation.jl` for details.
```

docs/src/developer_documentation/custom_tangent_type.md
```diff
@@ -218,6 +218,37 @@ You must provide adjoints for every `getfield`/`lgetfield` variant that appears
...
+
+For example, the constructor call `A(1.0)` is lowered to:
+```julia
+_new_(A{Float64}, 1.0)
```
sunxd3:
```suggestion
_new_(A{Float64}, 1.0, nothing)
```

docs/src/developer_documentation/custom_tangent_type.md
```diff
@@ -218,6 +218,37 @@ You must provide adjoints for every `getfield`/`lgetfield` variant that appears
...
+```
+This means you only need to write rules for `_new_` to handle all constructor variants, rather than writing separate rules for `A(::T)`.
```
sunxd3:
```suggestion
This means `_new_` covers the `:new` portion across constructor variants, rather than writing separate rules just for `A(::T)`.
```
slightly more accurate (in case of constructor might use additional logic that does require a rule)

docs/src/developer_documentation/custom_tangent_type.md
```diff
@@ -218,6 +218,37 @@ You must provide adjoints for every `getfield`/`lgetfield` variant that appears
...
+
+**`lgetfield` and `lsetfield!`**: These functions are designed for type stability. The standard `getfield(x, :f)` with a symbol argument is not type-stable when the field cannot be constant-propagated. `lgetfield` addresses this by using `Val` to specify the field statically:
+
+```julia
+lgetfield(x, f::Val)
+```
+
```
sunxd3, line 239:
```suggestion

The analogous mutating form is `lsetfield!(x, Val(:f), v)`, which corresponds to `setfield!(x, :f, v)` when the field name is a compile-time constant. This only applies to mutable structs, since `setfield!` is invalid for immutable ones. Mooncake's IR normalization also rewrites literal-field `setfield!` calls to `lsetfield!` for the same reason.

```

docs/src/developer_documentation/custom_tangent_type.md
yebai: How about below? It clarifies a bit more why `_new_` doesn't always need rules. Is it still accurate? 

```suggestion
**`_new_`**: Mooncake’s IR normalisation rewrites `Expr(:new, ...)`—the lowered representation of composite type construction—into calls to [`_new_`](@ref). This establishes a uniform interface for differentiating object construction. In practice, Mooncake can differentiate through `_new_` calls directly, so rules typically only need to be defined for logic inside `_new_` that Mooncake cannot differentiate through, rather than for each individual constructor method.

Splatted constructions are normalised analogously to [`_splat_new_`](@ref). The raw `:new` form can be inspected using `@code_lowered A(1.0)`. For details, see the [*standardisation*](@ref standardisation) section of forward differentiation and the implementation in `src/interpreter/ir_normalisation.jl`.
```

docs/src/developer_documentation/custom_tangent_type.md
yebai: @sunxd3, is there a reason we don't need rules for `setfield!`? IIRC, the page doesn't mention `setfield!`.

docs/src/developer_documentation/custom_tangent_type.md
sunxd3: I like the change (but maybe make the paragraph break more consistent)

docs/src/developer_documentation/custom_tangent_type.md
sunxd3: I think `setfield!` [always delegates](https://github.com/chalk-lab/Mooncake.jl/blob/b224566835c829772dd7c008189c9957073f1ba8/src/rules/builtins.jl#L1019-L1026) to `lsetfield!`, while `getfield` [doesn't always](https://github.com/chalk-lab/Mooncake.jl/blob/b224566835c829772dd7c008189c9957073f1ba8/src/rules/builtins.jl#L923-L987).

Issue #961, opened 2026-01-21
Canonicalising tangent types to simplify `frule!!`/`rrule!!` rules


yebai:
For several GEMM-like BLAS rules, Mooncake uses an explicit normalisation step inside `rrule!!` to collapse different arrays and tangent types into a smaller set of canonical representations. This allows a single rule to handle many argument-type combinations without duplicating logic.
   
**Example: `rrule!!` with `CoDual` types**

Consider an rrule!! invoked on CoDual inputs, each carrying a primal value and an associated tangent:

```julia
y, pullback!! = rrule!!(
    CoDual(f, Δf),
    CoDual(x₁, Δx₁),
    CoDual(x₂, Δx₂),
)
```

In Mooncake, `rrule!!` is called with CoDual-wrapped arguments, including the function itself, to carry both primal values and tangents. `Δf` is shown here for completeness, but is not essential to this example. Assume the incoming tangents use different concrete array representations (schematically shown as SparseArray and DenseArray).

```julia
Δx₁ :: SparseArray  
Δx₂ :: DenseArray
```

Inside the rule, these can be normalised explicitly at the boundary:

```julia
# Tf is the primal type of f, and TΔf is its associated tangent type.
function rrule!!(::CoDual(Tf, TΔf), x₁::CoDual, x₂::CoDual)
    # arrayify defines the canonical representation expected by the rule.
    Δx₁′ = arrayify(tangent(x₁))
    Δx₂′ = arrayify(tangent(x₂))

    y = f(primal(x₁), primal(x₂))

    function pullback!!(Δy)
        return NoTangent(),
               Δx₁′ * Δy,
               Δx₂′ * Δy
    end

    return y, pullback!!
end
```

After normalisation, the remainder of the rule can assume a single, well-defined tangent representation, avoiding case splits on mixed tangent types.

**Connection to Base.promote_rule**

Internal tangent utilities, such as `matrixify`, `arrayify` and `numberify` play the same conceptual role as

```julia
Base.promote_rule(::Type{Type1}, ::Type{Type2}) = Type1
```

[...]

PR #971, opened 2026-01-22
Canonicalising tangent types

yebai:
Fix https://github.com/chalk-lab/Mooncake.jl/issues/961

### DIFF
--- a/docs/src/developer_documentation/custom_tangent_type.md
+++ b/docs/src/developer_documentation/custom_tangent_type.md
@@ -220,7 +220,7 @@ You must provide adjoints for every `getfield`/`lgetfield` variant that appears 
 
 ##### Why These Primitives?
 
-**`_new_`**: Mooncake’s IR normalisation rewrites `Expr(:new, ...)`—the lowered representation of composite type construction—into calls to [`_new_`](@ref). Because many `_new_` calls can be differentiated using generic rules (see `src/rules/new.jl`), explicit differentiation rules are only required for object-construction logic that cannot be handled automatically. Such rules can therefore be written against `_new_` itself, rather than against individual constructor methods.
+**`_new_`**: Mooncake’s IR normalisation rewrites `Expr(:new, ...)`—the lowered representation of composite type construction—into calls to [`_new_`](@ref). Because many `_new_` calls can be differentiated using generic rules (see [`src/rules/new.jl`](https://github.com/chalk-lab/Mooncake.jl/blob/main/src/rules/new.jl)), explicit differentiation rules are only required for object-construction logic that cannot be handled automatically. Such rules can therefore be written against `_new_` itself, rather than against individual constructor methods.
 
 For example, the constructor call `A(1.0)` is lowered to:
 ```julia


Technici4n:
This reads a bit too AI generated and I'm not sure how relevant it is outside of the BLAS rules 😅

yebai:
Thanks, @Technici4n. This normalisation pattern is broadly useful beyond `blas` rules—for example, in `kron` rules. Additionally, documenting it reduces reliance on any single critical contributor.

EDIT: I'll leave this open for a few days so more people can chip in.

yebai:
This should be ready!

docs/src/developer_documentation/custom_tangent_type.md
```diff
@@ -220,7 +220,7 @@ You must provide adjoints for every `getfield`/`lgetfield` variant that appears
 
 ##### Why These Primitives?
 
-**`_new_`**: Mooncake’s IR normalisation rewrites `Expr(:new, ...)`—the lowered representation of composite type construction—into calls to [`_new_`](@ref). Because many `_new_` calls can be differentiated directly, differentiation rules are only required for object-construction logic that cannot be handled automatically, and can be written against `_new_` rather than individual constructor methods.
+**`_new_`**: Mooncake’s IR normalisation rewrites `Expr(:new, ...)`—the lowered representation of composite type construction—into calls to [`_new_`](@ref). Because many `_new_` calls can be differentiated using generic rules (see `src/rules/new.jl`), explicit differentiation rules are only required for object-construction logic that cannot be handled automatically. Such rules can therefore be written against `_new_` itself, rather than against individual constructor methods.
```
kshyatt: Could we add a proper link to `src/rules/new.jl`?

docs/src/utilities/defining_rules.md
```diff
@@ -42,4 +42,43 @@ Then implement a method of one of the following:
 ```@docs; canonical=false
 Mooncake.rrule!!
 Mooncake.build_primitive_rrule
-```
\ No newline at end of file
+```
+
+## Canonicalising Tangent Types
+
+For some differentiation rules, Mooncake performs an explicit canonicalisation step inside `rrule!!` that collapses heterogeneous array and tangent types into a small set of canonical representations. By canonicalising at the rule boundary, a single implementation can support many combinations of argument and tangent types without duplicating logic or relying on complex dispatch. This allows the remainder of the rule to assume a single, well-defined tangent representation.
```
kshyatt, line 49: Do we need to only use `matrixify` for `rrule!!`s or is it necessary also in forward mode? Can this be made more explicit?

Technici4n:
It's worth noting that these functions are considered internal. But I suppose that at this stage the entire infrastructure to define rules is considered internal anyway. 😉

docs/src/utilities/defining_rules.md
yebai: It's not specific to reverse mode, but forward mode rules are often much more straightforward. 

EDIT: Added ref to `frule!!` in https://github.com/chalk-lab/Mooncake.jl/pull/978

Issue #275, opened 2024-09-28
Renaming CoDual and separate used


willtebbutt:
Currently `CoDual` is used for two purposes:
1. to pair a primal and its fdata, and
2. to pair a primal and its tangent.

Having two uses for the same type, only distinguishable by the context, is not good design. It was laziness on my part when I introduced fdata / rdata.

A better design would be to have two distinct types: `PrimalFDataPair` and `PrimalTangentPair`. These will be abbreviated to `PFPair` and `PTPair` respectively (otheriwse the code is going to become a _lot_ more verbose).

Introducing this distinction will also substantially reduce the chances of bugs in interface code (value_and_gradient!!, DifferentiationInterface.jl's wrappers, etc) -- currently it's very straightforward to accidentally make a `CoDual` containing a primal + tangent, when you ought to have produce a `CoDual` containing a primal + fdata.

There are a couple of functions which will need to change -- for example `zero_fcodual` and `zero_codual` will become `zero_primal_fdata_pair` and `zero_primal_tangent_pair`. More verbose names are acceptable for these, as they do not appear in _literally_ every `rrule!!` signature. There are probably some more that will appear during refactoring.

Technici4n:
I think this is a good idea, probably using `CoDual` for primal-tangent pairs and `FCoDual` for primal-fdata pairs; I would advise against using `Dual` for the former to make sure that we don't mixup forward and reverse mode quantities.

yebai:
`CoDual` and `FCoDual` are good names. 

It's also good to assume `Dual` is only for forward-mode and never appears inside `CoDual` / `FCoDual`. Similarly, tangents in `Dual` should always be explicit (`Tangent` or its chunked variant); `FData` should not be permitted as a tangent in `Dual`.

PR #972, opened 2026-01-23
Make friendly tangents the default

Technici4n:
Preparing for 0.5. 😉

This includes the most important changes listed in https://github.com/chalk-lab/Mooncake.jl/issues/486, except for #275 which we can probably leave to 0.6 (e.g. at the same time as chunked forward mode).

### DIFF
--- a/docs/src/interface.md
+++ b/docs/src/interface.md
@@ -30,14 +30,24 @@ With `friendly_tangents=true` (the default), gradients use the same types as the
 cache = MC.prepare_gradient_cache(g, x_eval)
 val, grad = MC.value_and_gradient!!(cache, g, x_eval)
 ```
+This produces a tuple containing the value of the function (here `0.5`) and the gradient
+(here `(g, SimplePair(2.0, 4.0))`).
+The first part of the gradient is the gradient wrt. `g` itself;
+it contains no data and thus gets returned as-is.
+The second part of the gradient is the gradient wrt. `x`: `SimplePair(2.0, 4.0)`.
 
 In case of issues with friendly tangents, gradients can be returned using the Mooncake-internal
-representation (by setting `friendly_tangents=false` in the config):
+representation by setting `friendly_tangents=false` in the config:
 
 ```@example interface
 cache = MC.prepare_gradient_cache(g, x_eval; config=MC.Config(friendly_tangents=false))
 val, grad = MC.value_and_gradient!!(cache, g, x_eval)
 ```
+The corresponding gradient is `(Mooncake.NoTangent(), Mooncake.Tangent{@NamedTuple{x1::Float64, x2::Float64}}((x1 = 2.0, x2 = 4.0)))`.
+Indeed, `g` contains no differentiable data its gradient will be `NoTangent()`.
+And since `SimpePair` contains differentiable data, its gradient is represented using a `@NamedTuple{x1::Float64, x2::Float64}`
+wrapped in a `Tangent` object.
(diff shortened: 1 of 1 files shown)


docs/src/interface.md
```diff
@@ -19,17 +19,18 @@ g(x) = sum(abs2, x)
 x_complex = [1.0 + 2.0im, 3.0 + 4.0im]
 ```
```
Technici4n, line 26: We need to find a better example than complex numbers after #953. Maybe a simple custom struct with two fields.

src/interface.jl
```diff
@@ -155,7 +155,7 @@ function value_and_pullback!!(
 end
 
 """
-    value_and_gradient!!(rule, f, x...; friendly_tangents=false)
+    value_and_gradient!!(rule, f, x...; friendly_tangents=true)
```
Technici4n, line 158: Making friendly tangents true by default means that the function will be returned, is that fine?

E.g. for `value_and_gradient!!(rule, f, 5.0)` in one of the tests, previously we would return
```jl
(15.0, (NoTangent(), 3.0))
```
But now we will return
```jl
(15.0, (Main.ToolsForRulesResources.overlay_tester, 3.0))
```

yebai: It's probably okay to override `tangent_to_primal!!` for some types to improve clarity. 

~~Mooncake.tangent_to_primal!!(primal, ::Mooncake.NoTangent) = Mooncake.NoTangent()~~

```julia
Mooncake.tangent_to_primal!!(::Type{typeof(T)}, ::Mooncake.NoTangent) where T<:Function = 
    Mooncake.NoTangent()
```

Note that `friendly_tangent` means we return something that is "user-friendly", so we may not need to constrain ourselves to the primal space.

Technici4n: I don't think it's a good idea since that is not the correct primal type. More practically this will also break nesting, e.g. non-differentiable fields inside tangent types (e.g. the U or L symbol in a `Symmetric` for example). So we'll probably have to live with this but I suppose I could add a quick sentence in the docs.

yebai: I can live with that, but note that the example above is restricted to primal types whose tangent is `NoTangent`, so it does not affect cases such as `Symmetric`.

`Mooncake.tangent_to_primal!!(primal, tangent)` may be worth revisiting in the future—for example, replacing it with an interface such as `Mooncake.friendly_tangent!!(friendly_tangent, tangent)`. In hindsight, using primal types to represent tangents is not entirely natural. 

*EDIT*: I tweaked the above example to restrict the primal type to subtypes of functions.

Technici4n:
Ready now I believe! The docs and DI tests will fail to build. Once this is merged I will look into updating DI.

yebai:
It may be a good idea to delay defaulting to the user-friendly tangent until 0.6, after we sort out the remaining design issues.

Technici4n:
Sure, we can defer this PR.

yebai:
Sorry for the confusion: It's okay to review and merge this PR for release 0.5, but it may be better to keep `friendly_tangents=false` as the default until 0.6.

Technici4n:
Ah right, that makes sense. I'll get it on it tomorrow then :)

yebai:
Closed in favour of https://github.com/chalk-lab/Mooncake.jl/pull/1103, which is safer, as it returns raw tangents by default and restricts friendly tangents to selected types. Friendly tangents can be enabled as the default in the next breaking release.

PR #976, opened 2026-01-25
NaN handling for builtins, low level math primitives.

AstitvaAggarwal:
~~Follows from discussions on `NaN` poisoning, completes https://github.com/chalk-lab/Mooncake.jl/pull/797.
Allows for manual handling for issue https://github.com/chalk-lab/Mooncake.jl/issues/807, (Activity analysis is the next step for fixing `NaN` poisoning in `non primitive` singular function rules)~~

~~Through this PR, Mooncake will by default always handle `NaN` poisoning as long as the singular function in question has a primitive rule written.~~

TL;DR: 

This PR improves NaN handling for low-level math primitives and builtins. In addition, it separates the handling of NaN values from explicitly non-differentiable points in Mooncake. Currently, both cases are represented as NaN, which conflates distinct semantics and obscures intent.

We propose introducing distinct placeholders—for example, `nan_tangent_guard` to represent NaN propagation and `nondifferentiable_tangent_guard` to represent true non-differentiability (these are similar to [`notimplemented_tangent_guard`](https://github.com/chalk-lab/Mooncake.jl/blob/7ec22cb8fc580988d92a6f7e0f979ce5ed669b55/src/tools_for_rules.jl#L449-L465)). While both methods may return `NaN` and may share the same zero-masking strategy to prevent gradient poisoning, they remain semantically distinct.

Both cases can be overridden by overloading the relevant functions (ie, `*_guard`), allowing users to customise how NaN values and non-differentiable points are handled.

### DIFF
--- a/src/rules/builtins.jl
+++ b/src/rules/builtins.jl
@@ -654,11 +654,10 @@ end
 function frule!!(::Dual{typeof(sqrt_llvm)}, x)
     _x, dx = extract(x)
     y = sqrt_llvm(_x)
-    dy = dx / (2 * y)
+    dy = nan_tangent_guard(dx, dx / (2 * y))
     return Dual(y, dy)
 end
 function rrule!!(::CoDual{typeof(sqrt_llvm)}, x::CoDual{P}) where {P}
-    _x = primal(x)
     _y = sqrt_llvm(primal(x))
     function llvm_sqrt_pullback!!(dy)
         dx = nan_tangent_guard(dy, dy / (2 * _y))
@@ -669,8 +668,9 @@ end
 
 @intrinsic sqrt_llvm_fast
 function frule!!(::Dual{typeof(sqrt_llvm_fast)}, x)
-    y = sqrt_llvm_fast(primal(x))
-    dy = tangent(x) / (2 * y)
+    _x, dx = extract(x)
+    y = sqrt_llvm_fast(_x)
+    dy = nan_tangent_guard(dx, dx / (2 * y))
     return Dual(y, dy)
 end
 function rrule!!(::CoDual{typeof(sqrt_llvm_fast)}, x::CoDual{P}) where {P}
--- a/src/rules/low_level_maths.jl
+++ b/src/rules/low_level_maths.jl
@@ -95,7 +95,9 @@ function frule!!(::Dual{typeof(log)}, b::Dual{P}, x::Dual{P}) where {P<:IEEEFloa
     _x, dx = extract(x)
     y = log(_b, _x)
     log_b = log(_b)
-    return Dual(y, -db * y / (log_b * _b) + dx * (inv(_x) / log_b))
+    dy_b = nan_tangent_guard(db, -db * y / (log_b * _b))
+    dy_x = nan_tangent_guard(dx, dx * (inv(_x) / log_b))
+    return Dual(y, dy_b + dy_x)
 end
 function rrule!!(::CoDual{typeof(log)}, b::CoDual{P}, x::CoDual{P}) where {P<:IEEEFloat}
     y = log(primal(b), primal(x))
@@ -112,7 +114,8 @@ end
 function frule!!(::Dual{typeof(log)}, x::Dual{P}) where {P<:IEEEFloat}
     _x, dx = extract(x)
     y = log(_x)
-    return Dual(y, dx / _x)
+    dy = nan_tangent_guard(dx, dx / _x)
+    return Dual(y, dy)
 end
 function rrule!!(::CoDual{typeof(log)}, x::CoDual{P}) where {P<:IEEEFloat}
     y = log(primal(x))
@@ -127,7 +130,8 @@ end
 function frule!!(::Dual{typeof(sqrt)}, x::Dual{P}) where {P<:IEEEFloat}
     _x, dx = extract(x)
     y = sqrt(_x)
-    return Dual(y, dx / (2 * y))
+    dy = nan_tangent_guard(dx, dx / (2 * y))
+    return Dual(y, dy)
 end
 function rrule!!(::CoDual{typeof(sqrt)}, x::CoDual{P}) where {P<:IEEEFloat}
     y = sqrt(primal(x))
@@ -142,7 +146,8 @@ end
 function frule!!(::Dual{typeof(cbrt)}, x::Dual{P}) where {P<:IEEEFloat}
     _x, dx = extract(x)
     y = cbrt(_x)
-    return Dual(y, dx / (3 * y^2))
+    dy = nan_tangent_guard(dx, dx / (3 * y^2))
+    return Dual(y, dy)
 end
 function rrule!!(::CoDual{typeof(cbrt)}, x::CoDual{P}) where {P<:IEEEFloat}
     y = cbrt(primal(x))
@@ -156,7 +161,9 @@ end
 @is_primitive MinimalCtx Tuple{typeof(log10),P} where {P<:IEEEFloat}
 function frule!!(::Dual{typeof(log10)}, x::Dual{P}) where {P<:IEEEFloat}
     _x, dx = extract(x)
-    return Dual(log10(_x), dx / (_x * log(P(10))))
+    y = log10(_x)
+    dy = nan_tangent_guard(dx, dx / (_x * log(P(10))))
+    return Dual(y, dy)
 end
 function rrule!!(::CoDual{typeof(log10)}, x::CoDual{P}) where {P<:IEEEFloat}
     y = log10(primal(x))
@@ -170,7 +177,9 @@ end
 @is_primitive MinimalCtx Tuple{typeof(log2),P} where {P<:IEEEFloat}
 function frule!!(::Dual{typeof(log2)}, x::Dual{P}) where {P<:IEEEFloat}
     _x, dx = extract(x)
-    return Dual(log2(_x), dx / (_x * log(P(2))))
+    y = log2(_x)
+    dy = nan_tangent_guard(dx, dx / (_x * log(P(2))))
+    return Dual(y, dy)
 end
 function rrule!!(::CoDual{typeof(log2)}, x::CoDual{P}) where {P<:IEEEFloat}
     y = log2(primal(x))
@@ -184,7 +193,9 @@ end
 @is_primitive MinimalCtx Tuple{typeof(log1p),P} where {P<:IEEEFloat}
 function frule!!(::Dual{typeof(log1p)}, x::Dual{P}) where {P<:IEEEFloat}
     _x, dx = extract(x)
-    return Dual(log1p(_x), dx / (1 + _x))
+    y = log1p(_x)
+    dy = nan_tangent_guard(dx, dx / (1 + _x))
+    return Dual(y, dy)
 end
 function rrule!!(::CoDual{typeof(log1p)}, x::CoDual{P}) where {P<:IEEEFloat}
     y = log1p(primal(x))
@@ -211,7 +222,7 @@ function frule!!(
     ::Dual{typeof(hypot)}, x::Dual{P}, xs::Vararg{Dual{P}}
 ) where {P<:IEEEFloat}
     h = hypot(primal(x), map(primal, xs)...)
-    dh = sum(primal(a) * tangent(a) for a in (x, xs...)) / h
+    dh = sum(nan_tangent_guard(tangent(a), primal(a) * tangent(a)) for a in (x, xs...)) / h
     return Dual(h, dh)
 end
 function rrule!!(


yebai:
@nsiccha, would you be able to help review this?

yebai:
@nsiccha, any concrete suggestions on how to improve that?

AstitvaAggarwal:
yes @nsiccha, we must correctly return NaNs if the user actually wants to involve singular functions in gradients. This is mathematically correct behaviour

nsiccha:
@yebai just add the `iszero(dy)` condition to the conditions that have been there before. My question is whether this is behaviour that we want to change (this PR) or conserve.

@AstitvaAggarwal Do we have to do this? Going of https://juliadiff.org/ChainRulesCore.jl/dev/maths/nondiff_points.html, we don't have to and also shouldn't.

AstitvaAggarwal:
the current state of code where we end up returning `0` is fine for making things "just work". but Imo its better to return to the user what they actually work with and non a misleading answer like `0` (unless they explicitly dont involve that nonsensical calculation). we should follow Math and not what "might" the user prefer.

In optimization for example if they end up continuing with `0` for a gradient value, and simply continue - sure they will get some answer (but that trajectory wont be valid). 

TLDR: `NaN poisoning` is the issue. A step further - in case users want to deal with `NaN`s etc. in whatever way they want - They are free to `@mooncake_overlay` their function. But Mooncake must only return correct math.

Further reference: https://juliadiff.org/ForwardDiff.jl/stable/user/advanced/#Fixing-NaN/Inf-Issues

src/rules/low_level_maths.jl
```diff
@@ -341,3 +341,21 @@ function hand_written_rule_test_cases(rng_ctor, ::Val{:low_level_maths})
...
+#     a = x
+#     b = sqrt(abs(x - 1.0))
+#     return a
+# end
+
+# for backend in (AutoForwardDiff(), AutoEnzyme(), AutoZygote(), AutoReverseDiff())
+#     @info "$backend" pullback(sqrt, backend, 0.0, (1.0,)) pullback(
+#         sqrt, backend, 0.0, (0.0,)
+#     ) pullback(f, backend, 1.0, (1.0,)) pullback(f, backend, 1.0, (0.0,))
+# end
```
yebai:
```suggestion
```

test/rules/builtins.jl
```diff
@@ -94,3 +94,23 @@ foo_throws(e) = throw(e)
...
+    function builtins_nantester(f, args)
+        a = f(args)
+        b = args
+        return b
+    end
+
+    for (f, args) in test_cases
+        _, grad = value_and_gradient!!(
+            prepare_gradient_cache(builtins_nantester, f, args), builtins_nantester, f, args
+        )
```
yebai: For code clarity, let's do this in two steps: 

```suggestion
        cache = prepare_gradient_cache(builtins_nantester, f, args)
        _, grad = value_and_gradient!!(cache, builtins_nantester, f, args)
```

test/rules/builtins.jl
```diff
@@ -94,3 +94,23 @@ foo_throws(e) = throw(e)
...
     end
 end
+
+@testset "NaN handling in builtins rrules" begin
+    test_cases = vcat(map([Float16, Float32, Float64]) do T
+        cases = [(Base.sqrt_llvm, T(0)), (Base.sqrt_llvm_fast, T(0))]
+        return cases
+    end...)
+
+    function builtins_nantester(f, args)
```
yebai, line 105:
```suggestion
    # Test cases for avoiding `NaN` poisoning. 
    #  See https://github.com/chalk-lab/Mooncake.jl/issues/807 
    function builtins_nantester(f, args)
```

test/rules/low_level_maths.jl
```diff
@@ -14,17 +14,25 @@
...
                     (hypot, (T(0), T(0), T(0))),
-                    # builtins
-                    (Base.sqrt_llvm, T(0)),
-                    (Base.sqrt_llvm_fast, T(0)),
                 ]
                 return cases
             end...,
         )
+
+        function low_level_maths_nantester(f, args)
```
yebai, line 24:
```suggestion
        # Test cases for avoiding `NaN` poisoning. 
        #  See https://github.com/chalk-lab/Mooncake.jl/issues/807
        function low_level_maths_nantester(f, args)
```

test/rules/low_level_maths.jl
```diff
@@ -14,17 +14,25 @@
...
         for (f, args) in test_cases
-            cache = prepare_gradient_cache(f, args...)
-            _, grad = value_and_gradient!!(cache, f, args...)
-            @test all(iszero, grad[2:end])
+            _, grad = value_and_gradient!!(
+                prepare_gradient_cache(low_level_maths_nantester, f, args),
+                low_level_maths_nantester,
+                f,
+                args,
+            )
```
yebai: For code clarity, let's do this in two steps: 
```suggestion
            cache = prepare_gradient_cache(low_level_maths_nantester, f, args)
            _, grad = value_and_gradient!!(
                cache,
                low_level_maths_nantester,
                f,
                args,
            )
```

yebai:
> TLDR: NaN poisoning is the issue. A step further - in case users want to deal with NaNs etc. in whatever way they want - They are free to @mooncake_overlay their function. But Mooncake must only return correct math.

I am okay with that. It is occasionally inconvenient, but it is mathematically consistent. It's also true that users can use `@mooncake_overlay` to override the behaviour. 

It's worth noting that this PR returns `NaN`s for non-differentiable points,  and differs from the ChainRules norms pointed out by @nsiccha.

yebai:
Cc @devmotion and @Technici4n in case they have any comments.

nsiccha:
I guess as the "return 0 instead of Inf" logic was actually added by @AstitvaAggarwal himself (https://github.com/chalk-lab/Mooncake.jl/commit/b8c538b0ee084b8ce1ced66d897f7d227d3a74a4), I guess we can revert the changes from that commit and go with how it's implemented now - I'll review everything else tomorrow.

test/rules/builtins.jl
```diff
@@ -94,3 +94,24 @@ foo_throws(e) = throw(e)
         )
     end
 end
+
+@testset "NaN handling in builtins rrules" begin
+    test_cases = vcat(map([Float16, Float32, Float64]) do T
```
nsiccha: That could be mapreduce(vcat, ....) no?

Technici4n:
If my understanding is correct, this basically implements "strong zeros" (also called NaN safe mode in ForwardDiff), correct? If so that's indeed a good solution for now IMO!

AstitvaAggarwal:
yes, we basically default to that without any performance loss.

test/rules/builtins.jl
AstitvaAggarwal: point 2 seems to be an issue with how, mooncake works with transformations and IR code. This happens even before a ruleset is derived.

AstitvaAggarwal:
Ohh that's a good idea

AstitvaAggarwal:
> One high-level improvement is to introduce a named constant:
> 
> ```julia
> const NaG = 0  # NaG = not a gradient
> ```
> 
> This constant can then be used to replace
> 
> ```julia
> iszero(dy) ? dy : dy / (2 * _y)
> ```
> 
> with
> 
> ```julia
> iszero(dy) ? NaG .*(1+.dy) : dy / (2 * _y)
> ```
> 
> The advantage is that the sentinel value becomes explicit and easily configurable; for example, users may override it via `Mooncake.NaG = Inf` if desired.

This would unfortunately interfere in avoiding `NaN poisoning` where gradients are not even used. Therefore this is unsafe. for non differentiability i think overlaying is fine.

AstitvaAggarwal:
@yebai, returning `iszero(dy) ? NaG .*(1+.dy) : dy / (2 * _y)` does not do anything. When `iszero(dy)` the product does not even participate in the final gradient for the main function.


> One high-level improvement is to introduce a named constant:
> 
> ```julia
> const NaG = 0  # NaG = not a gradient
> ```
> 
> This constant can then be used to replace
> 
> ```julia
> iszero(dy) ? dy : dy / (2 * _y)
> ```
> 
> with
> 
> ```julia
> iszero(dy) ? NaG .*(1+.dy) : dy / (2 * _y)
> ```
> 
> The advantage is that the sentinel value becomes explicit and easily configurable; for example, users may override it via `Mooncake.NaG = Inf` if desired.

AstitvaAggarwal:
I have now added the feature to not only handle `NaN poisoning` but also cases when correctly returned `NaN` is not easy to work with and a user chosen value is preferred. 

This is independant from `poisoning` handling and allows us to keep the current functionality added, explained here : https://github.com/chalk-lab/Mooncake.jl/pull/797#issuecomment-3377850155

src/interface.jl
```diff
@@ -1,3 +1,57 @@
...
+set_nan_filler!(Inf)
+
+# Reset to default
+set_nan_filler!(0.0)
+```
+
+See also: [`get_nan_filler`](@ref)
+"""
+set_nan_filler!(val) = (_nan_filler[] = val)
+
```
yebai: This can be defined as a function to avoid the overhead associated with reference and dereference operations. Custom behaviour can be introduced by overloading this function.

NB: GitHub corrupts the long code suggestion below; please view the Markdown source instead.

```suggestion
"""
    nan_gradient_sentinel(::Type{T}) where T

Return the sentinel value used to replace NaN gradients in reverse-mode AD rules
for gradients of type `T`.

When differentiating non-differentiable functions (e.g., `sqrt` at 0 or `log` at 0),
the mathematical gradient is undefined (`NaN`). This function provides a fallback
value so that the reverse pass can continue.

By default, the sentinel is `T(NaN)` for floating-point types. Users may overload
this function for custom types or alternative NaN-handling policies.

# Example
```julia
# Default behaviour
nan_gradient_sentinel(Float64)  # returns NaN

# Custom behaviour for Float64
nan_gradient_sentinel(::Type{Float64}) = Inf
```
"""
@inline nan_gradient_sentinel(::Type{T}) where {T} = T(NaN)

"""
    nan_gradient_sentinel(x)

Convenience method that returns the sentinel value based on the type of x.

# Example
```julia
x = 1.0
nan_gradient_sentinel(x)  # returns NaN
```
“””
@inline nan_gradient_sentinel(x) = nan_gradient_sentinel(typeof(x))
```

File: src/interface.jl [hunk 8]
```diff
@@ -1,3 +1,57 @@
...
+# After changing the sentinel
+set_nan_filler!(Inf)
+get_nan_filler()  # returns Inf
+```
+"""
+get_nan_filler() = _nan_filler[]
+
+"""
+    set_nan_filler!(value)
```
Technici4n: What is the motivation for being able to fill with anything else than NaN?

src/interface.jl
```diff
@@ -1,3 +1,57 @@
...
+```julia
+# Default behavior
+get_nan_filler()  # returns NaN
+
+# After changing the sentinel
+set_nan_filler!(Inf)
+get_nan_filler()  # returns Inf
+```
+"""
+get_nan_filler() = _nan_filler[]
```
Technici4n: Echoing @yebai's comment, this is likely to be very slow at runtime. It would be preferable to have a constant such that it could be inlined. Or just write `NaN` directly in the rules unless there is a good reason for making the filler configurable.

src/interface.jl
AstitvaAggarwal: https://github.com/chalk-lab/Mooncake.jl/pull/797#issuecomment-3377850155

src/interface.jl
AstitvaAggarwal: this is addressed now.

src/interface.jl
Technici4n: That comment reads quite AI generated, is quite long, and doesn't really explain the motivation. The idea is to be able to set the nan filler to 0? Returning a "NaN" value of `0` when `dy` is not 0 doesn't sound like a good idea to me, and it also seems to go beyond the scope of this PR...

yebai: This is not just about representing `NaN` values, but also about being explicit about where a function is non-differentiable; see  
https://juliadiff.org/ChainRulesCore.jl/dev/maths/nondiff_points.html

The goal is to introduce separate placeholders for `NaN` values and for non-differentiable points. At present, both cases are conflated into `NaN`, which is the source of the confusion. Ideally, we would provide distinct placeholders, such as `nan_tangent_guard` and `nondifferentiable_tangent_guard`.

src/interface.jl
```diff
@@ -1,3 +1,68 @@
...
+```julia
+nan_guard(0.0, NaN)  # returns 0.0
+nan_guard(1.0, NaN)  # returns nan_gradient_sentinel(Float64)
+nan_guard(2.0, 3.0)  # returns 3.0
+```
+"""
+@inline function nan_guard(dy::L, grad::T) where {L,T}
+    iszero(dy) && return dy
+    return isnan(grad) ? nan_gradient_sentinel(T) : grad
+end
```
yebai: @AstitvaAggarwal It would be fine to unify `nan_gradient_sentinel` into `nan_guard` (or `nan_tangent_guard`, as suggested above) in order to reduce an extra layer of indirection. Please use a separate function—e.g. `nondifferentiable_tangent_guard`—to clearly distinguish non-differentiable cases from `NaN` handling. Both cases could utilise the same zero masking to avoid gradient poisoning. 

```suggestion
@inline function nan_tangent_guard(dy::L, grad::T) where {L,T}
    iszero(dy) && return dy
    return isnan(grad) ? T(NaN) : grad
end

@inline function nondifferentiable_tangent_guard(dy::L, grad::T) where {L,T}
    return iszero(dy) ? dy : T(NaN)
end
```

yebai (changes requested):
Good work, @AstitvaAggarwal —one final refactoring suggestion below. Let's have different placeholders for `nan` and `nondifferentiable`.

nsiccha:
I agree with @Technici4n that the `_guard` functions really don't belong into this PR.

yebai:
It looks like several closely related ideas, i.e., handling `NaN` poisoning, handling non-differentiable points and an explicit mechanism for `_guard`, were raised by different reviews. These are good points, and it's also good we now have a clear [picture](https://github.com/chalk-lab/Mooncake.jl/pull/976#issue-3854046002) of the solution.  

I'm happy to let @AstitvaAggarwal complete the additional comments and then review again.

src/interface.jl
AstitvaAggarwal: addressing for clarity - have addressed this. Yes there is a reason to keep it configurable, as mentioned in the previous reviews here https://github.com/chalk-lab/Mooncake.jl/pull/976#issuecomment-3811934414

AstitvaAggarwal:
> I agree with @Technici4n that the `_guard` functions really don't belong into this PR.

Im not sure where he mentions this?

src/rules/low_level_maths.jl
```diff
@@ -101,9 +101,9 @@ function rrule!!(::CoDual{typeof(log)}, b::CoDual{P}, x::CoDual{P}) where {P<:IE
     y = log(primal(b), primal(x))
     function log_adjoint(dy::P)
         log_b = log(primal(b))
-        return NoRData(),
-        ifelse(iszero(primal(b)) || iszero(log_b), P(0), -dy * y / (log_b * primal(b))),
-        ifelse(iszero(primal(x)) || iszero(log_b), P(0), dy / (primal(x) * log_b))
+        grad_args = (-dy * y / (log_b * primal(b))), (dy / (primal(x) * log_b))
```
Technici4n: Nit: better write two calls to nan_guard than this, it's more readable

src/interface.jl
```diff
@@ -1,3 +1,68 @@
+"""
+    nan_gradient_sentinel(::Type{T}) where T
```
Technici4n: This likely doesn't belong in the public interface.

src/interface.jl
```diff
@@ -1,3 +1,68 @@
...
+
+Return the sentinel value used to replace NaN gradients in reverse-mode AD rules
+for gradients of type `T`.
+
+When differentiating non-differentiable functions (e.g., `sqrt` at 0 or `log` at 0),
+the mathematical gradient is undefined (NaN). This function provides a fallback
+value so that the reverse pass can continue.
+
+By default, the sentinel is `T(NaN)` for floating-point types. Users may overload
+this function for custom types or alternative NaN-handling policies.
```
Technici4n: Probably not since it would be type piracy.

src/interface.jl
```diff
@@ -1,3 +1,68 @@
...
+# Example
+```julia
+nan_guard(0.0, NaN)  # returns 0.0
+nan_guard(1.0, NaN)  # returns nan_gradient_sentinel(Float64)
+nan_guard(2.0, 3.0)  # returns 3.0
+```
+"""
+@inline function nan_guard(dy::L, grad::T) where {L,T}
+    iszero(dy) && return dy
+    return isnan(grad) ? nan_gradient_sentinel(T) : grad
```
Technici4n: Why not return grad directly and get rid of this sentinel stuff?

AstitvaAggarwal:
To all reviewers, lets revisit the core problem for clarity :

I specifically limited this PR to `NaN` handling to address the longstanding [issue](https://github.com/chalk-lab/Mooncake.jl/issues/807).

In @nsiccha's [review](https://github.com/chalk-lab/Mooncake.jl/pull/976#issuecomment-3811934414), the suggestion is given to resolve all of `Non Differentiability` within this PR. I [address](https://github.com/chalk-lab/Mooncake.jl/pull/976#issuecomment-3812094969) it, keeping the current PR limited to only `NaN` handling.

Addressing `Non Differentiability` as this has complicated all other discussions. A function is `Non Differentiable` only : 

1. If the gradient at the point is `NaN`  -  I am addressing this.
2. If the left and right limits of the gradient at the point are unequal - mostly out of PR scope (rule level/overlay?).

---

About `NaN` handling - Only functions that are `Non Differentiable` will cause the appearance of `NaN` : 

1. Causing `NaN poisoning` if `grad` not involved in gradient calculation (`dy=0`) but `NaN` is still numerically returned.
2. Showing clear `Non Differentiability` if `grad` is involved in gradient calculation (`dy!=0`).

Therefore, **at the rule level** `nondifferentiable_tangent_guard` & `nan_tangent_guard` do not solve the problem as the **same** problem (occurance of `NaN`) changes it's nature (nondiff vs `NaN` poison) based on what `dy` is (user controls this).

To solve this, a single interface method `nan_guard` covers the calculated gradient's value (based on above 2 cases) : 

1. Set `grad=0` if `isnan(grad) & iszero(dy)`  -  stops primitive rule level `NaN` poisoning.
2. Set `grad=user_sentinel_value` if `isnan(grad) & !iszero(dy)`  -  allows for _correct math_ (`NaN`) **or** optimizer, solver [friendly](https://juliadiff.org/ChainRulesCore.jl/v1.14/maths/nondiff_points.html#Local-Minima-/-Maxima) `0` etc.

---

Finally, I kindly request all reviewers to please give specific points/reviews to address like [this](https://github.com/chalk-lab/Mooncake.jl/pull/976#discussion_r2764918191), it keeps thing clear.
This helps in avoiding making interpretation vague and overcomplicating things.

src/interface.jl
AstitvaAggarwal: because grad has to be mathematically correct + sometimes optimiser friendly.

src/interface.jl
AstitvaAggarwal: How though, NaN32, NaN16 exist?  And this function is limited to Mooncake

Technici4n: Overloading the function for custom types would be fine, but overloading it for alternative NaN-handling policies would be type piracy. However, I am now questioning why this function would be used for custom types? The usual Mooncake policy means that it should only be used for the types that Mooncake explicitly supports in its rules. So it might be simpler to inline the function and replace all callsites by `T(NaN)`.

src/interface.jl
Technici4n: What I am saying is essentially that `isnan(grad) ? nan : grad` can equivalently be written as `grad`, especially if/once `nan_gradient_sentinel(T)` is replaced by `T(NaN)`.

AstitvaAggarwal: `nan_gradient_sentinel(T)` is needed, though. For above stated reason

src/interface.jl
AstitvaAggarwal: Oh! I understand the type piracy logic, yes I'll just change that allowing `T(some global variable)` and add a few more changes

But responding to the rest : `nan_gradient_sentinel(T)` may return non float values also, like some object /error message etc. we don't lose anything by keeping flexibility for overloads to future specific types

nsiccha:
I actually still don't really see the utility of the `_guard` functions.

A: Mooncake actually already seems to follow [ChainRules.jl's recommendations for non-differentiable points](https://juliadiff.org/ChainRulesCore.jl/dev/maths/nondiff_points.html) - except for the recent changes made by @AstitvaAggarwal in https://github.com/chalk-lab/Mooncake.jl/commit/b8c538b0ee084b8ce1ced66d897f7d227d3a74a4. See for example [the current handling of `abs_float`](https://github.com/chalk-lab/Mooncake.jl/blob/0146ccd2c7c25f52f273ee697e684025553d5ed8/src/rules/builtins.jl#L173-L177). I don't think we should suddenly change it for some functions and make the handling of non-differentiable points inconsistent across mooncake.
B: The [current implementation of `nan_tangent_guard`](https://github.com/chalk-lab/Mooncake.jl/blob/80956b103236cedf0b2717d20a8cbfcbc2740211/src/interface.jl#L61-L64) doesn't actually do what it's supposed to do as far as I understood its purpose. In the setting where it's called (e.g. [here for `sqrt_llvm`](https://github.com/chalk-lab/Mooncake.jl/blob/80956b103236cedf0b2717d20a8cbfcbc2740211/src/rules/builtins.jl#L664)), it will return i) 0 if dy is 0 - that's good, ii) return Inf if dy != 0 and y = 0 - this will may still lead to NaN/Inf poisoning down the line, iii) return something potentially non-NaN if dy != 0 and dy = NaN or y = NaN or dy = Inf and y = Inf - I don't think we have fully  thought this through yet, and probably should do that before merging such a change.

src/interface.jl
AstitvaAggarwal: done

src/rules/low_level_maths.jl
AstitvaAggarwal: done

AstitvaAggarwal:
fyi, Im removing the current code refactor and switching over to the changes as mentioned here: https://github.com/chalk-lab/Mooncake.jl/pull/976#discussion_r2764918191.

src/interface.jl
```diff
@@ -1,3 +1,24 @@
+"""
+    nan_tangent_guard(dy::L, grad::T) where {L,T}
+
+Guard against NaN gradients in automatic differentiation. Returns `dy` if `dy=0` signifying that the gradient is not involved in total
+gradient calculation, otherwise returns `grad` unless it's NaN (in which case returns correct type wrapped `T(NaN)`).
```
```suggestion
    nan_tangent_guard(dy::L, grad::T) where {L,T}

Guard against NaN propagation in automatic differentiation.

When `dy = 0`, the corresponding gradient does not contribute to the total
gradient, so a zero tangent is returned to prevent NaN poisoning.

Otherwise, return `grad`.

Note that this does not fully eliminate gradient poisoning; it relies on
zero masking (i.e., a strong zero with `dy = 0`) to reduce NaN propagation.
```

src/interface.jl
```diff
@@ -1,3 +1,24 @@
...
+@inline function nan_tangent_guard(dy::L, grad::T) where {L,T}
+    iszero(dy) && return dy
+    return isnan(grad) ? T(NaN) : grad
+end
+
+"""
+    nondifferentiable_tangent_guard(dy::L, grad::T) where {L,T}
+
+Handle functions at non-differentiable domain points. Returns `dy` if `dy=0` signifying that the gradient is not involved in total
+gradient calculation, otherwise returns `T(NaN)` to signal the function is non differentiable.
```
```suggestion
Handle functions at non-differentiable domain points.

If `dy = 0`, the gradient does not contribute to the total gradient
calculation, so a zero tangent is returned.

Otherwise, return `T(NaN)` to signal that the function is
non-differentiable at this point.

This behaviour may be overloaded to return alternative values
(e.g., a zero tangent) when required.

See:
https://juliadiff.org/ChainRulesCore.jl/dev/maths/nondiff_points.html
```

src/interface.jl
```diff
@@ -1,3 +1,24 @@
+"""
+    nan_tangent_guard(dy::L, grad::T) where {L,T}
+
+Guard against NaN gradients in automatic differentiation. Returns `dy` if `dy=0` signifying that the gradient is not involved in total
+gradient calculation, otherwise returns `grad` unless it's NaN (in which case returns correct type wrapped `T(NaN)`).
+"""
+@inline function nan_tangent_guard(dy::L, grad::T) where {L,T}
+    iszero(dy) && return dy
+    return isnan(grad) ? T(NaN) : grad
```
yebai: `T(NaN)` only works for IEEEFloats etc, and will throw an error for composite types. `dy .* NaN` is more robust. 

```suggestion
    # `dy .* NaN` returns a NaN tangent with the same type as `dy`.
    return isnan(grad) ? dy .* NaN : grad
```

src/interface.jl
```diff
@@ -1,3 +1,24 @@
...
+end
+
+"""
+    nondifferentiable_tangent_guard(dy::L, grad::T) where {L,T}
+
+Handle functions at non-differentiable domain points. Returns `dy` if `dy=0` signifying that the gradient is not involved in total
+gradient calculation, otherwise returns `T(NaN)` to signal the function is non differentiable. 
+"""
+@inline function nondifferentiable_tangent_guard(dy::L, grad::T) where {L,T}
+    return iszero(dy) ? dy : T(NaN)
```
```suggestion
    `dy .* NaN` returns a NaN tangent with the same type as `dy`.
    return iszero(dy) ? dy : dy .* NaN
```

src/interface.jl
```diff
@@ -1,3 +1,24 @@
+"""
+    nan_tangent_guard(dy::L, grad::T) where {L,T}
+
+Guard against NaN gradients in automatic differentiation. Returns `dy` if `dy=0` signifying that the gradient is not involved in total
+gradient calculation, otherwise returns `grad` unless it's NaN (in which case returns correct type wrapped `T(NaN)`).
+"""
+@inline function nan_tangent_guard(dy::L, grad::T) where {L,T}
```
Technici4n: These functions should be moved to a different file, they are not part of the public interface.

src/interface.jl
Technici4n:
```suggestion
    iszero(dy) && return zero(T)
    return grad
```

src/interface.jl
Technici4n:
```suggestion
    return iszero(dy) ? zero(T) : T(NaN)
```
in case T != L

src/interface.jl
yebai: Good point. Maybe keep them together with `nonimplemented_tangent_guard`?

src/interface.jl
yebai: Good point.

src/interface.jl
yebai: Nice catch.

src/tools_for_rules.jl
```diff
@@ -441,6 +441,42 @@ function increment_and_get_rdata!(f, r, t)
...
+When `dy = 0`, the corresponding gradient does not contribute to the total  
+gradient, so a zero tangent is returned to prevent NaN poisoning.  
+
+Otherwise, return `grad`.  
+
+Note that this does not fully eliminate gradient poisoning; it relies on  
+zero masking (i.e., a strong zero with `dy = 0`) to reduce NaN propagation.
+"""
+@inline function nan_tangent_guard(dy::L, grad::T) where {L,T}
+    iszero(dy) && return dy
```
yebai: Perhaps this is more robust as it ensures the returned value has the same tangent type of `grad`? 

```suggestion
    _T = eltype(T)
    grad .= iszero(dy) ? _T(0) : _T(NaN)
```

src/tools_for_rules.jl
```diff
@@ -441,6 +441,42 @@ function increment_and_get_rdata!(f, r, t)
...
+Otherwise, return `T(NaN)` to signal that the function is
+non-differentiable at this point.
+This behaviour may be overloaded to return alternative values
+(e.g., a zero tangent) when required.
+See:
+https://juliadiff.org/ChainRulesCore.jl/dev/maths/nondiff_points.html.
+"""
+@inline function nondifferentiable_tangent_guard(dy::L, grad::T) where {L,T}
+    #  `dy .* NaN` returns a NaN tangent with the same type as `dy`.
+    return iszero(dy) ? dy : dy .* eltype(T)(NaN)
```
yebai:
```suggestion
    _T = eltype(T)
    grad .= iszero(dy) ? _T(0) : _T(NaN)
    return grad
```

src/tools_for_rules.jl
```diff
@@ -441,6 +441,42 @@ function increment_and_get_rdata!(f, r, t)
...
+If `dy = 0`, the gradient does not contribute to the total gradient
+calculation, so a zero tangent is returned.
+Otherwise, return `T(NaN)` to signal that the function is
+non-differentiable at this point.
+This behaviour may be overloaded to return alternative values
+(e.g., a zero tangent) when required.
+See:
+https://juliadiff.org/ChainRulesCore.jl/dev/maths/nondiff_points.html.
+"""
+@inline function nondifferentiable_tangent_guard(dy::L, grad::T) where {L,T}
```
yebai:
```suggestion
@inline function nondifferentiable_tangent_guard(
    dy::L,
    grad::T,
) where {Tf<:Union{Base.IEEEFloat,Array{<:Base.IEEEFloat}},L<:Tf,T<:Tf}
```

src/tools_for_rules.jl
AstitvaAggarwal: I think lets just stick to zero() for now? this should handle all relevant rdata for Complex, Int, IEEEfloat.

src/rules/low_level_maths.jl
```diff
@@ -101,9 +101,9 @@ function rrule!!(::CoDual{typeof(log)}, b::CoDual{P}, x::CoDual{P}) where {P<:IE
     y = log(primal(b), primal(x))
     function log_adjoint(dy::P)
         log_b = log(primal(b))
-        return NoRData(),
-        ifelse(iszero(primal(b)) || iszero(log_b), P(0), -dy * y / (log_b * primal(b))),
-        ifelse(iszero(primal(x)) || iszero(log_b), P(0), dy / (primal(x) * log_b))
+        grad_b = nan_tangent_guard(dy, -dy * y / (log_b * primal(b)))
+        grad_x = nan_tangent_guard(dy, dy / (primal(x) * log_b))
+        return NoRData(), grad_b, grad_x
```
yebai: Let's keep the notation consistent: 

```suggestion
        db = nan_tangent_guard(dy, -dy * y / (log_b * primal(b)))
        dx = nan_tangent_guard(dy, dy / (primal(x) * log_b))
        return NoRData(), db, dx
```

src/rules/low_level_maths.jl
```diff
@@ -117,7 +117,8 @@ end
 function rrule!!(::CoDual{typeof(log)}, x::CoDual{P}) where {P<:IEEEFloat}
     y = log(primal(x))
     function log_adjoint(dy::P)
-        return NoRData(), ifelse(iszero(primal(x)), P(0), dy / primal(x))
+        grad_x = nan_tangent_guard(dy, dy / primal(x))
+        return NoRData(), grad_x
```
yebai:
```suggestion
        dx = nan_tangent_guard(dy, dy / primal(x))
        return NoRData(), dx
```

src/rules/low_level_maths.jl
```diff
@@ -131,7 +132,8 @@ end
 function rrule!!(::CoDual{typeof(sqrt)}, x::CoDual{P}) where {P<:IEEEFloat}
     y = sqrt(primal(x))
     function sqrt_adjoint(dy::P)
-        return NoRData(), ifelse(iszero(primal(x)), P(0), dy / (2 * y))
+        grad_x = nan_tangent_guard(dy, dy / (2 * y))
+        return NoRData(), grad_x
```
yebai:
```suggestion
        dx = nan_tangent_guard(dy, dy / (2 * y))
        return NoRData(), dx
```

src/rules/low_level_maths.jl
```diff
@@ -145,7 +147,8 @@ end
 function rrule!!(::CoDual{typeof(cbrt)}, x::CoDual{P}) where {P<:IEEEFloat}
     y = cbrt(primal(x))
     function cbrt_adjoint(dy::P)
-        return NoRData(), ifelse(iszero(y), P(0), dy / (3 * y^2))
+        grad_x = nan_tangent_guard(dy, dy / (3 * y^2))
+        return NoRData(), grad_x
```
yebai:
```suggestion
        dx = nan_tangent_guard(dy, dy / (3 * y^2))
        return NoRData(), dx
```

src/rules/low_level_maths.jl
```diff
@@ -158,7 +161,8 @@ end
 function rrule!!(::CoDual{typeof(log10)}, x::CoDual{P}) where {P<:IEEEFloat}
     y = log10(primal(x))
     function log10_adjoint(dy::P)
-        return NoRData(), ifelse(iszero(primal(x)), P(0), dy / (primal(x) * log(P(10))))
+        grad_x = nan_tangent_guard(dy, dy / (primal(x) * log(P(10))))
+        return NoRData(), grad_x
```
yebai:
```suggestion
        dx = nan_tangent_guard(dy, dy / (primal(x) * log(P(10))))
        return NoRData(), dx
```

src/rules/low_level_maths.jl
```diff
@@ -171,7 +175,8 @@ end
 function rrule!!(::CoDual{typeof(log2)}, x::CoDual{P}) where {P<:IEEEFloat}
     y = log2(primal(x))
     function log2_adjoint(dy::P)
-        return NoRData(), ifelse(iszero(primal(x)), P(0), dy / (primal(x) * log(P(2))))
+        grad_x = nan_tangent_guard(dy, dy / (primal(x) * log(P(2))))
+        return NoRData(), grad_x
```
yebai:
```suggestion
        dx = nan_tangent_guard(dy, dy / (primal(x) * log(P(2))))
        return NoRData(), dx
```

src/rules/low_level_maths.jl
```diff
@@ -184,7 +189,8 @@ end
 function rrule!!(::CoDual{typeof(log1p)}, x::CoDual{P}) where {P<:IEEEFloat}
     y = log1p(primal(x))
     function log1p_adjoint(dy::P)
-        return NoRData(), ifelse(iszero(1 + primal(x)), P(0), dy / (1 + primal(x)))
+        grad_x = nan_tangent_guard(dy, dy / (1 + primal(x)))
+        return NoRData(), grad_x
```
yebai:
```suggestion
        dx = nan_tangent_guard(dy, dy / (1 + primal(x)))
        return NoRData(), dx
```

src/rules/low_level_maths.jl
```diff
@@ -213,7 +219,7 @@ function rrule!!(
 ) where {P<:IEEEFloat}
     h = hypot(primal(x), map(primal, xs)...)
     function hypot_pb!!(dh::P)
-        grads = map(a -> ifelse(iszero(h), P(0), dh * (primal(a) / h)), (x, xs...))
+        grads = map(a -> nan_tangent_guard(dh, dh * (primal(a) / h)), (x, xs...))
         return NoRData(), grads...
```
yebai:
```suggestion
        tangents = map(a -> nan_tangent_guard(dh, dh * (primal(a) / h)), (x, xs...))
        return NoRData(), tangents...
```

yebai:
@nsiccha @Technici4n this should be ready if you want take a final look. We plan to merge this in the next 2-3 days.

src/tools_for_rules.jl
```diff
@@ -442,41 +442,86 @@ function increment_and_get_rdata!(f, r, t)
...
+"""
+    nondifferentiable_tangent_guard(dy::L, tangent::T) where {L,T}
+
+Handle functions at non-differentiable domain points. See:
+https://juliadiff.org/ChainRulesCore.jl/dev/maths/nondiff_points.html
+
+If `dy = 0`, the gradient does not contribute to the total gradient
+calculation, so a zero tangent is returned.
+
+Otherwise, return `T(NaN)` to signal that the function is
```
nsiccha: Your reference above quite literally says at the very top in a callout block (emphasis mine) "If the function is not-differentiable choose to return something useful rather than erroring. [...]. **Throwing an error, or returning NaN is generally the least useful option.**"

src/rules/builtins.jl
```diff
@@ -650,14 +652,18 @@ end
...
+    dy = dx / (2 * y)
     return Dual(y, dy)
 end
 function rrule!!(::CoDual{typeof(sqrt_llvm)}, x::CoDual{P}) where {P}
     _x = primal(x)
     _y = sqrt_llvm(primal(x))
-    llvm_sqrt_pullback!!(dy) = NoRData(), ifelse(iszero(_y), P(0), dy / (2 * _y))
+    function llvm_sqrt_pullback!!(dy)
+        grad_x = nan_tangent_guard(dy, dy / (2 * _y))
+        return NoRData(), grad_x
```
nsiccha: To make this agree with Hong's changes below:

```suggestion
        dx = nan_tangent_guard(dy, dy / (2 * _y))
        return NoRData(), dx
```

src/rules/builtins.jl
```diff
@@ -669,7 +675,10 @@ function frule!!(::Dual{typeof(sqrt_llvm_fast)}, x)
 end
 function rrule!!(::CoDual{typeof(sqrt_llvm_fast)}, x::CoDual{P}) where {P}
     _y = sqrt_llvm_fast(primal(x))
-    llvm_sqrt_fast_pullback!!(dy) = NoRData(), ifelse(iszero(_y), P(0), dy / (2 * _y))
+    function llvm_sqrt_fast_pullback!!(dy)
+        grad_x = nan_tangent_guard(dy, dy / (2 * _y))
+        return NoRData(), grad_x
```
nsiccha: Same as above.

```suggestion
        dx = nan_tangent_guard(dy, dy / (2 * _y))
        return NoRData(), dx
```

nsiccha (commented):
I have voiced my conerns before.

src/tools_for_rules.jl
yebai: Yes, we can set the returned value to `0`. 

For clarity, I think part of the reason for ChainRules suggesting against `NaN` is to avoid gradient poisoning.

yebai:
```suggestion
Otherwise, return `T(0)` to signal that the function is
```

src/tools_for_rules.jl
```diff
@@ -442,41 +442,86 @@ function increment_and_get_rdata!(f, r, t)
...
+non-differentiable at this point.
+
+This behaviour may be overloaded to return alternative values
+(e.g., a zero tangent) when required.
```
yebai:
```suggestion
(e.g., a nan-tangent) when required. 
```

src/tools_for_rules.jl
```diff
@@ -442,41 +442,86 @@ function increment_and_get_rdata!(f, r, t)
...
+"""
+@inline function nondifferentiable_tangent_guard(
+    dy::L,
+    tangent::T,
+) where {
+    L<:Union{Base.IEEEFloat,Complex{<:Base.IEEEFloat}},
+    T<:Union{Base.IEEEFloat,Complex{<:Base.IEEEFloat}},
+}
+    _tangent = iszero(dy) ? T(0) : T(NaN)
+    return _tangent
```
yebai:
```suggestion
    return if iszero(dy)
        T(0)
    else
        T(0)
    end
```

src/tools_for_rules.jl
```diff
@@ -442,41 +442,86 @@ function increment_and_get_rdata!(f, r, t)
...
+"""
+@inline function nan_tangent_guard(
+    dy::L,
+    tangent::T,
+) where {
+    L<:Union{Base.IEEEFloat,Complex{<:Base.IEEEFloat}},
+    T<:Union{Base.IEEEFloat,Complex{<:Base.IEEEFloat}},
+}
+    _tangent = iszero(dy) ? T(0) : tangent
+    return _tangent
```
yebai:
```suggestion
    return if iszero(dy)
        T(0)
    else
        tangent
    end
```

yebai:
For posterity: in `nondifferentiable_guard(dy, tangent)`, this PR returns `0` when `dy == 0`; otherwise, it returns the user-provided `tangent`.

src/rules/builtins.jl
```diff
@@ -650,14 +652,18 @@ end
 
 @intrinsic sqrt_llvm
 function frule!!(::Dual{typeof(sqrt_llvm)}, x)
-    y = sqrt_llvm(primal(x))
-    dy = tangent(x) / (2 * y)
+    _x, dx = extract(x)
+    y = sqrt_llvm(_x)
+    dy = dx / (2 * y)
```
yebai:
```suggestion
    dy = nan_tangent_guard(dx, dx / (2 * y))
```

src/rules/builtins.jl
yebai:
```suggestion
    _x, dx = extract(x)
    y = sqrt_llvm_fast(_x)
    dy = nan_tangent_guard(dx, dx / (2 * y))
```

src/rules/low_level_maths.jl
yebai:
```suggestion
    dy = nan_tangent_guard(dx, dx / (2 * y))
    return Dual(y, dy)
```

PR #980, opened 2026-01-28
Add caching and friendly tangents in forward mode

### LINKED ISSUE #917: Expand the forward-mode interface to support caching and friendly tangents
To be consistent with reverse-mode, and allow DI to only use the public API. :)

(As suggested in https://github.com/chalk-lab/Mooncake.jl/pull/860#issuecomment-3694824041)

gdalle:
Fixes #917

Inspired and triggered by @Technici4n's DI contribution in https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/957, might as well make it available in Mooncake instead of using (even more) internals.

Remaining questions:

- Is the tuple interface the one we want?
- Are there enough tests?
- Are there enough docs?
- Is this breaking? I don't think so even though we rename `rule` to `cache` in the existing `value_and_derivative!!` API

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.5.0"
+version = "0.5.1"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
@@ -92,4 +92,15 @@ StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["AllocCheck", "Aqua", "BenchmarkTools", "DiffTests", "JET", "Logging", "Pkg", "Revise", "StableRNGs", "Test"]
+test = [
+    "AllocCheck",
+    "Aqua",
+    "BenchmarkTools",
+    "DiffTests",
+    "JET",
+    "Logging",
+    "Pkg",
+    "Revise",
+    "StableRNGs",
+    "Test",
+]


src/interface.jl
```diff
@@ -658,20 +658,87 @@ function value_and_gradient!!(
...
 """
     prepare_derivative_cache(fx...; config=Mooncake.Config())
 
 Returns a cache used with [`value_and_derivative!!`](@ref). See that function for more info.
 """
-@unstable function prepare_derivative_cache(fx...; config=Config())
-    build_frule(fx...; config.debug_mode, config.silence_debug_messages)
+@unstable function prepare_derivative_cache(f, x::Vararg{Any,N}; config=Config()) where {N}
+    fx = (f, x...)
+    y = f(x...)
```
Technici4n: Could this be moved to the friendly tangents-exclusive branch?

src/interface.jl
```diff
@@ -658,20 +658,87 @@ function value_and_gradient!!(
...
+
+    # Check that the output of `fx` is supported.
+    __exclude_func_with_unsupported_output(fx)
```
Technici4n: Is this really needed for forward-mode? To be fair I don't know the specifics of reverse with aliasing.

src/interface.jl
```diff
@@ -658,20 +658,87 @@ function value_and_gradient!!(
...
+
+    # Construct rule and tangents.
+    rule = build_frule(fx...; config.debug_mode, config.silence_debug_messages)
+
+    # Construct cache for output. Check that `_copy_to_output!!`ing appears to work.
+    input_tangents = map(zero_tangent, fx)
+    output_primal = _copy_output(y)
```
Technici4n: Specific to friendly tangents

src/interface.jl
```diff
@@ -658,20 +658,87 @@ function value_and_gradient!!(
...
 """
-    value_and_derivative!!(rule::R, f::Dual, x::Vararg{Dual,N})
+    value_and_derivative!!(cache::ForwardCache, f::Dual, x::Vararg{Dual,N})
 
 Returns a `Dual` containing the result of applying forward-mode AD to compute the (Frechet)
 derivative of `primal(f)` at the primal values in `x` in the direction of the tangent values
 in `f` and `x`.
 """
-value_and_derivative!!(rule::R, fx::Vararg{Dual,N}) where {R,N} = rule(fx...)
+value_and_derivative!!(cache::ForwardCache, fx::Vararg{Dual,N}) where {N} =
```
Technici4n, line 693: Shouldn't this function be deprecated?

src/interface.jl
```diff
@@ -658,20 +658,87 @@ function value_and_gradient!!(
...
+
+Tuples are used as inputs and outputs instead of `Dual` numbers to accommodate the case where internal Mooncake tangent types do not coincide with tangents provided by the user (in which case we translate between "friendly tangents" and internal tangents using cache storage).
+
+!!! info
+    `cache` must be the output of [`prepare_derivative_cache`](@ref), and (fields of) `f` and `x` must be of the same size and shape as those used to construct the `cache`. This is to ensure that the gradient can be written to the memory allocated when the `cache` was built.
+
+!!! warning
+    `cache` owns any mutable state returned by this function, meaning that mutable components of values returned by it will be mutated if you run this function again with different arguments. Therefore, if you need to keep the values returned by this function around over multiple calls to this function with the same `cache`, you should take a copy (using `copy` or `deepcopy`) of them before calling again.
+"""
+function value_and_derivative!!(
```
Technici4n, line 709: Should this function be called `value_and_pushforward!!`? Not sure.

src/interface.jl
```diff
@@ -658,20 +658,87 @@ function value_and_gradient!!(
...
 Returns a `Dual` containing the result of applying forward-mode AD to compute the (Frechet)
 derivative of `primal(f)` at the primal values in `x` in the direction of the tangent values
 in `f` and `x`.
 """
-value_and_derivative!!(rule::R, fx::Vararg{Dual,N}) where {R,N} = rule(fx...)
+value_and_derivative!!(cache::ForwardCache, fx::Vararg{Dual,N}) where {N} =
+    cache.rule(fx...)  # TODO: handle friendly tangents for the output here?
+
+"""
+    value_and_derivative!!(cache::ForwardCache, (f, df), (x, dx), ...)
```
Technici4n, line 697: I was initially thinking that `(f, x, ...), (df, dx, ...)` might be preferable, but actually this makes sense to me too!

Technici4n, line 697: Remember to update the docs as well (`interface.md` I believe?)

src/interface.jl
gdalle: I don't really know what it's for, I just lifted it from the pullback preparation

src/interface.jl
gdalle: We already had this debate and Mooncake settled on "derivative" (against my vote but that ship has sailed)

src/interface.jl
gdalle: I see two issues with your alternative:

First, users may accidentally switch the order of arguments or forget one when they list the tangents afterwards, so keeping primal and tangent together is safer.

Second, we cannot have two `Vararg` in the same signature (I tried), so it would have to look like this
```julia
function value_and_derivative!!(cache, args::Vararg{Any,N}) where {N}
    primals = args[begin:N÷2]
    tangents = args[(N÷2+1):end]
    # etc
end
```
and I'm not sure whether operations like `N ÷ 2 + 1` can be performed at compile-time, which might cause type instabilities. Note that

src/interface.jl
gdalle: I wasn't aware that the pullback equivalent was deprecated, so yes, probably

src/interface.jl
gdalle: I will remove it for now

src/interface.jl
gdalle: Wait, I don't see any deprecation warnings anywhere in the package, so probably not?
Removing or renaming this method would be breaking anyway

Technici4n: I guess we can keep it around

src/interface.jl
Technici4n: That is quite unfortunate, since by analogy with `value_and_gradient!!` I would expect the following names:
- `value_and_pushforward!!` for a general pushforward, but it requires both `x` and `dx`;
- `value_and_derivative!!`: for functions that accept a single IEEEFloat argument, will internally set `dx=1` and compute the pushforward.

But whatever I guess 🤷

src/interface.jl
Technici4n: Yes, both of your points are valid. What you are proposing in this PR looks better than my suggestion.

Technici4n (approved):
This seems good to go, can you just update the signatures at the end of `docs/src/interface.md`?

src/interface.jl
gdalle: I agree, and that's the terminology I picked for DI, but Will or Hong (not sure) felt strongly otherwise and Mooncake followed

src/interface.jl
gdalle: The reason I didn't is because if we want to avoid breaking changes until Mooncake 0.6, the existing API with `Dual` inputs has to keep working, and keep being part of the public docs. That's why I didn't specialize the signature (but we could add a deprecation warning after all)

Technici4n:
Oops I didn't wait for the tests to pass it seems. But it should be fine and will unblock DI.

gdalle:
@Technici4n let's still wait to see if the ambiguity issue was resolved and if code coverage was 100% before releasing Mooncake v0.5.1?

Technici4n:
Indeed, formatting failed. I see a segfault for basic-lts as well, not sure if it's related or not. Let me know if I should try to fix some of these!

gdalle:
Feel free to open a PR to fix the formatting. As for the segfault, do you have the permissions for relaunching the CI job? Sometimes they randomly fail

gdalle:
Took care of the formatting in #983

PR #981, opened 2026-01-29
Work on #685

### LINKED ISSUE #685: add example of simple, ℝⁿ→ℝ, scalar function to the manual
The rule framework of Mooncake.jl is very powerful but also somewhat complex, especially for people new to AD.

I think that the manual would benefit from a more gradual introduction, first defining an `ℝ→ℝ` function (eg continuing with the example of `sin`), then an `ℝ²→ℝ` bivariate function like `/`.

Now all of these are of course in [src/rrules/low_level_maths.jl](https://github.com/chalk-lab/Mooncake.jl/blob/05bbf164f0cbb86dfc28cd5650251edce3d6b646/src/rrules/low_level_maths.jl) at the very beginning, the point would be a *pedagogical* explanation, before concepts like fdata/rdata is introduced.

nsiccha:
Would fix #685

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.5.0"
+version = "0.5.2"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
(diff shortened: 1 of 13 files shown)


nsiccha:
Well well well.... Rendering of docs fails with
```
ERROR: LoadError: Unsatisfiable requirements detected for package DifferentiationInterface [a0c0ee7d]:
```

nsiccha:
In any case - @tpapp is this similar to what you envisioned? 

(Also, Mooncake's way of defining rrules is quite verbose!)

tpapp:
@nsiccha, partly yes, but also followed by 

1. an `ℝ²→ℝ` function, eg `/`, and
2. an interim example of either rdata or fdata (just one)

nsiccha:
@tpapp: Maybe you could draft a version of what you want, otherwise I'll merge this PR in the coming days. Thanks for engaging!

tpapp:
@nsiccha: thanks for doing this, and feel free to merge, currently I am busy but plan to address that point later.

PR #983, opened 2026-01-30
Fix segfault when passing friendly tangents to "non-friendly" forward cache in 1.10 and formatting

gdalle:
#980 was merged before formatting passed

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -693,7 +693,6 @@ in `f` and `x`.
 function value_and_derivative!!(cache::ForwardCache, fx::Vararg{Dual,N}) where {N}
     # TODO: check Dual coherence here like we do below?
     return cache.rule(fx...)
-    # TODO: handle friendly tangents for the output?
 end
 
 """


gdalle:
@Technici4n the segfault is happening again, no clue why

gdalle:
Here's the MWE for the segfault. It happens when we try to use `value_and_derivative!!` with friendly tangent types even though the cache was not prepared for that:

```julia
using Mooncake: Config, zero_tangent, prepare_derivative_cache, value_and_derivative!!

struct SimplePair
    x1::Float64
    x2::Float64
end

f = (x, y) -> x * y + cos(x)
g = (sp::SimplePair) -> SimplePair(f(sp.x1, sp.x2), 2.0)

x, y = 5.0, 4.0
dx, dy = 3.0, 2.0

fx_sp = (g, SimplePair(x, y))
dfx_sp = (zero_tangent(g), SimplePair(dx, dy))

cache_sp_friendly = prepare_derivative_cache(
    fx_sp...; config=Config(; friendly_tangents=true)
)
value_and_derivative!!(cache_sp_friendly, zip(fx_sp, dfx_sp)...)  # works

cache_sp_unfriendly = prepare_derivative_cache(
    fx_sp...; config=Config(; friendly_tangents=false)
)
value_and_derivative!!(cache_sp_unfriendly, zip(fx_sp, dfx_sp)...)  # segfaults on 1.10 only
```

src/interface.jl
```diff
@@ -707,7 +708,7 @@ Tuples are used as inputs and outputs instead of `Dual` numbers to accommodate t
     `cache` owns any mutable state returned by this function, meaning that mutable components of values returned by it will be mutated if you run this function again with different arguments. Therefore, if you need to keep the values returned by this function around over multiple calls to this function with the same `cache`, you should take a copy (using `copy` or `deepcopy`) of them before calling again.
 """
 function value_and_derivative!!(
-    cache::ForwardCache, f::NTuple{2,Any}, x::Vararg{<:NTuple{2,Any},N}
+    cache::ForwardCache, f::NTuple{2,Any}, x::Vararg{NTuple{2,Any},N}
```
Technici4n, line 712: is this the fix for the segfault?

Technici4n: Ah no, missed your other comment.

Technici4n:
Is the segfault maybe coming from calling an opaque closure with argument types that don't match the signature? Then it should hopefully go away with `debug_mode=true`.

src/interface.jl
gdalle: No this was just for the warnings that popped up at the beginning of the tests

gdalle:
> Is the segfault maybe coming from calling an opaque closure with argument types that don't match the signature? Then it should hopefully go away with debug_mode=true.

Maybe that was the root cause, I got it to go away by manually checking that the tangent types match the primal types inside provided `Dual` numbers. Not sure why this isn't done inside the `Dual` constructor, probably because of partially initialized structs?

src/interface.jl
```diff
@@ -725,6 +728,11 @@ function value_and_derivative!!(
     end
 
     input_duals = map(Dual, input_primals, input_tangents)
+
+    if !friendly_tangents  # in friendly mode, conversion should assure tangent coherence
```
gdalle:
```suggestion
    if !friendly_tangents  # in friendly mode, conversion should ensure tangent coherence
```

src/interface.jl
```diff
@@ -725,6 +728,11 @@ function value_and_derivative!!(
     end
 
     input_duals = map(Dual, input_primals, input_tangents)
+
+    if !friendly_tangents  # in friendly mode, conversion should ensure tangent coherence
+        error_if_incorrect_dual_types(input_duals...)
```
Technici4n, line 732: This should already be checked by debug mode. However if we don't do this and users forget to pass `friendly_tangents=true` they will just get a segfault in 1.10. So I suppose better be safe than sorry. This should hopefully not have any influence on the runtime performance if the compiler does a good job.

src/interface.jl
```diff
@@ -690,8 +690,11 @@ Returns a `Dual` containing the result of applying forward-mode AD to compute th
 derivative of `primal(f)` at the primal values in `x` in the direction of the tangent values
 in `f` and `x`.
 """
-value_and_derivative!!(cache::ForwardCache, fx::Vararg{Dual,N}) where {N} =
-    cache.rule(fx...)  # TODO: handle friendly tangents for the output here?
+function value_and_derivative!!(cache::ForwardCache, fx::Vararg{Dual,N}) where {N}
+    # TODO: check Dual coherence here like we do below?
+    return cache.rule(fx...)
+    # TODO: handle friendly tangents for the output?
```
Technici4n: I don't think we'd do that here. If the user is able to pass `Dual`s as input they should also be able to process a `Dual` coming out, IMO.

src/interface.jl
gdalle: Indeed I just checked and debug mode allows a similar diagnosis, but the default being `debug_mode=false` and `friendly_tangents=false`, I find the `ArgumentError` more explicit than a segfault.

src/interface.jl
gdalle: Alright then. Do we add a deprecation warning to this method?

gdalle: I'd wait a little longer, just to check that DI integration works well. Deprecation warnings can always be added at a later release

Technici4n: I'd maybe just remove `# TODO: handle friendly tangents for the output?` for now? The deprecation can come later and keeping this method around doesn't hurt atm. We'll probably get to revisit the forward-mode interface once chunking is a thing.

gdalle: done

gdalle:
Are we good vis-a-vis failing tests?

Technici4n:
Yes, good. Same failures as the Mooncake 0.5 PR + a spurious `rules/blas_level_3-1.12-x64` failure.

Technici4n:
Thank you for the quick fixup, let's get this released!

Issue #58, opened 2024-01-22
Performance Bugs


willtebbutt:
Examples which allocate that should not:

- [ ] `(false, :allocs, nothing, TestResources.kron!, randn(400, 400), Diagonal(randn(20)), randn(20, 20))` (to do with `increment!!` allocating for non isbits types contained in `PossiblyUninitTangent`)
- [ ] `(false, :stability, nothing, lsetfield!, TestResources.FullyInitMutableStruct(5.0, [1.0, 2.0]), Val(:y), [1.0, 3.0, 4.0])` (to do with increment implementation for `PossiblyUninitTangent`.

willtebbutt:
Note: at the time of writing, there are roughly 80 elements in shared data for the `_kron!` function, and I believe that there should only really be 3.

PR #990, opened 2026-02-05
Rearrange forward-over-reverse

Technici4n:
This PR rearranges forward-over-reverse so that it operates with an frule for `build_derived_rrule` which I think is cleaner. This should also be a good approach to make forward-over-forward-over-reverse or other fun things work. 😉

Summary of changes:
- Split up `build_rrule` such that we can use `@zero_derivative` on `build_rrule_checks` and define an frule!! for `build_derived_rrule`.
- The frule!! for `build_derived_rrule` now handles forward-over-reverse. The "primal" IR (for reverse) is fully optimized, whereas the "dual" IR (for forward-over-reverse) is optimized using a forward-mode interpreter, such that forward-mode primitives don't get inlined! This was verified to work in the `broadcast sum of squares` test, and fixes https://github.com/chalk-lab/Mooncake.jl/issues/914.
- Fix vararg signature flattening which was missing in forward-mode and was causing debug mode to fail for vararg functions.
- Fix dynamic frule using `_typeof` instead of `Base._stable_typeof`.
- We are now able to cleanly differentiate through `prepare_pullback_cache`, although this is quite inefficient. In the future it should be possible to transparently cache the rules at the lower levels of AD.
- Fix `test_rule` using `false` instead of `debug_mode` when checking the rule cache.
- Bail out of debug mode validation when a vector is in the middle of resizing (fixes failure of the frule for `push!` in Julia v1.11, for vectors of mutable types)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.5.9"
+version = "0.5.10"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"


Technici4n:
Now this works as well: 😄
```jl
function _compute_val_grad(x::Float64)
    f(x) = x^4
    cache = prepare_gradient_cache(f, x)
    val, grad = value_and_gradient!!(cache, f, x)
    return val, grad[2]
end

function _second_der(x::Float64)
    frule = build_frule(_compute_val_grad, x)
    return frule(
        zero_dual(_compute_val_grad),
        Dual(x, 1.0),
    )
end
```
and returns:
```jl
Dual{Tuple{Float64, Float64}, Tuple{Float64, Float64}}((16.0, 32.0), (32.0, 48.0))
```
😉

MasonProtter:
One reason to differentiate *optimized* IR rather than unoptimized IR is that optimized IR tends to be significantly more compact, and differentiating the unoptimized IR tends to produce bigger expressions that are then later on harder to optimize. 

This is one of the main reasons that Enzyme.jl does so much better than e.g. Zygote for nested AD. 

Since fewer optimizations happen on the julia side than the LLVM side, there's fewer opportunities here when we work on optimized IR, but I suspect it's still worth it.

Technici4n:
I see yeah, that makes sense. What we can do with the refactoring from this PR is performing a standard optimization on the "primal" rrule IR, while optimizing a different copy of this IR using a forward-mode interpreter that will thus prevent the inlining of primitives, before differentiating through it.

yebai:
Looking at the benchmarking numbers, it doesn't seem to have a noticeable impact. 

> This is one of the main reasons that Enzyme.jl does so much better than e.g. Zygote for nested AD.

Enzyme (C++ backend) reflects substantial engineering effort to optimise what was initially an inefficient prototype. Consequently, claims that attributing Enzyme’s performance advantages primarily to its operation on LLVM IR—rather than to these subsequent engineering improvements—may be overstated or misleading without a clear analysis or supporting evidence from the literature.

To the best of my knowledge, SIMD is the only optimisation that materially affects performance. A discussion of this point is provided [here](https://chalk-lab.github.io/Mooncake.jl/stable/known_limitations/#Differentiating-SIMD-Code). Additionally, [this issue](https://github.com/chalk-lab/Mooncake.jl/issues/58) documents the set of optimisations implemented by Will in the Mooncake prototype.

sunxd3:
the logic makes sense to me, I think it's a great idea.

currently the code calls with `do_inline = false` which still produce optimized IR (minus the inline and following compat pass), but I saw you have some TODOs and code (`do_optimize`) regarding skip optimization. which of these two (skip inlining and skip optimize) is what you are thinking?

MasonProtter:
> Enzyme (C++ backend) reflects substantial engineering effort to optimise what was initially an inefficient prototype. Consequently, claims that attributing Enzyme’s performance advantages primarily to its operation on LLVM IR—rather than to these subsequent engineering improvements—may be overstated or misleading without a clear analysis or supporting evidence from the literature.
>
> To the best of my knowledge, SIMD is the only optimisation that materially affects performance. A discussion of this point is provided [here](https://chalk-lab.github.io/Mooncake.jl/stable/known_limitations/#Differentiating-SIMD-Code). Additionally, https://github.com/chalk-lab/Mooncake.jl/issues/58 documents the set of optimisations implemented by Will in the Mooncake prototype.

This is more or less unrelated to what I was saying. My point is that optimized IR can be significantly more compact than unoptimized IR. In a lot of simple cases, there won't be much of a difference, but in complex cases, the optimized IR can do things like eliminate control flow, eliminate branches, remove redundant function calls, and all sorts of other things that result in significantly more compact code.

Differentiating code tends to make it bigger, just due to how the chain-rule works, and so if you're always differentiating the un-optimized IR, you have to do significantly more work at compile time because there's more code to differentiate, and it makes it harder (or in some cases impossible) to remove chunks of code after differentiating that should have been removed *before* differentiation.

wsmoses:
> Looking at the benchmarking numbers, it doesn't seem to have a noticeable impact.
> 
> > This is one of the main reasons that Enzyme.jl does so much better than e.g. Zygote for nested AD.
> 
> Enzyme (C++ backend) reflects substantial engineering effort to optimise what was initially an inefficient prototype. Consequently, claims that attributing Enzyme’s performance advantages primarily to its operation on LLVM IR—rather than to these subsequent engineering improvements—may be overstated or misleading without a clear analysis or supporting evidence from the literature.
> 
> To the best of my knowledge, SIMD is the only optimisation that materially affects performance. A discussion of this point is provided [here](https://chalk-lab.github.io/Mooncake.jl/stable/known_limitations/#Differentiating-SIMD-Code). Additionally, [this issue](https://github.com/chalk-lab/Mooncake.jl/issues/58) documents the set of optimisations implemented by Will in the Mooncake prototype.

This is an inaccurate and intentionally misleading statement. Enzyme has been efficient from first release, and there have been numerous scientific studies showing the impact of optimization on AD, like https://proceedings.neurips.cc/paper_files/paper/2020/file/9332c513ef44b682e9347822c2e457ac-Paper.pdf

You are of course welcome to develop your code as you see fit, but let’s try to avoid making unsubstantiated claims

yebai:
Thanks, Billy, for the clarification. To be clear, my comment was not intended to discount your or Enzyme’s contributions.

My point is narrower: I’m not aware of work that cleanly isolates LLVM-level optimisation as the primary driver of Enzyme’s performance advantages in practice, as distinct from the substantial engineering and algorithmic improvements. If there are studies that directly analyse this distinction, I’d be very interested to read them.

The NeurIPS paper you link demonstrates the importance of optimisation for AD systems in general, which I agree with. However, it does not, as far as I can tell, provide evidence that operating at LLVM IR—rather than higher-level IRs used by mainstream deep learning frameworks—is the dominant factor behind Enzyme’s observed performance, particularly given that those systems also achieve state-of-the-art results. The example in Figure 1 illustrates a well-known interaction between loop-invariant code motion and reverse-mode AD when an O(N) computation is naively duplicated inside a loop. While this is a valid demonstration of how optimisation ordering can asymptotically change complexity, it relies on a deliberately structured case where the primal code is already asymptotically suboptimal prior to LICM, and where the adjoint introduces a dependence that blocks an otherwise trivial transformation. As such, it demonstrates the *possibility* of asymptotic degradation under a particular optimisation pipeline, but does not by itself establish systematic or dominant performance advantages from operating at LLVM IR.

I agree that unsubstantiated claims should be avoided; my intent here is simply to caution against attributing Enzyme’s performance primarily to its LLVM-level formulation without a clear empirical analysis separating that factor from later engineering work.

This discussion is drifting into broader questions that are somewhat beyond the scope of this PR, so I’d be very happy to continue it in a more appropriate forum.

*For transparency:* (1) my team contributed to multiple rounds of bug discovery and minimisation for Enzyme, and I am on a joint grant that provides funding to the project; and (2) my comments above are motivated by curiosity as a researcher with a background in computational statistics and machine learning, trying to understand which factors most materially affect the quality and performance of an autodiff implementation.

MasonProtter:
> my intent here is simply to caution against attributing Enzyme’s performance primarily to its LLVM-level formulation without a clear empirical analysis separating that factor from later engineering work.

I never claimed that Enzyme gets better performance due to operating at the LLVM level. All I said is that when Enzyme works on optimized LLVM, the code shrinks more than when we operate on optimized julia IR, meaning that there are less potential benefits for Mooncake operating on optimized IR than for Enzyme operating on optimized IR (but I still think it's worth doing)

Technici4n:
Alright, this PR is now starting to look pretty good. The higher order derivative tests pass on my machine on Julia 1.11, not sure what's up with CI.

The key line is `optimized_fwd_ir = optimise_ir!(dri.fwd_ir; interp=interp_forward)` i.e. the IR is optimized before we build the frules, using a forward-mode interpreter, such that forward-mode primitives don't get inlined! This was verified to work in the `broadcast sum of squares` test.

I am on holidays for a week, but will be keeping an eye on this PR. If anyone is interested in helping, here are some things that I would like to fix before I would be 100% happy with this PR:
- Fix CI on GitHub.
- Building the frules with `debug_mode=true` fails, at least for the `x^4` example. (But enabling debug mode on the inner rrule is fine. 😉) Something vararg-related apparently... This should be fixed.
- Check which of the rules in `high_order_derivative_patches` we still need. I left a few TODOs.
- Write more tests. :)

Technici4n:
@sunxd3 @yebai this PR is finally ready, would one of you mind reviewing it? We can't quite get rid of the workaround rules since inner preparation is too slow, for the DI tests but we'll be able to cache that in a follow-up PR :)

yebai:
Excellent work -- one clarification question on the blow PR comment: perhaps expand a bit on what that means —e.g., why it is not efficient and how it can be improved next? 

> We are now able to cleanly differentiate through prepare_pullback_cache, although this is quite inefficient. In the future it should be possible to transparently cache the rules at the lower levels of AD.


EDIT: 

> Fix dynamic frule using _typeof instead of Base._stable_typeof.

Isn't it the other way around? The code replaces `_typeof` with ` Base._stable_typeof`. 

> Fix test_rule using false instead of debug_mode when checking the rule cache.

Similar to the above, the code replaces `false` with `debug_mode`.

Technici4n:
> Excellent work -- one clarification question on the blow PR comment: perhaps expand a bit on what that means —e.g., why it is not efficient and how it can be improved next?
> 
> > We are now able to cleanly differentiate through prepare_pullback_cache, although this is quite inefficient. In the future it should be possible to transparently cache the rules at the lower levels of AD.

Indeed, this is the main question/thing-to-do going forward if this PR is fine. In this PR, an frule for `build_derived_rrule` is introduced such that forward-mode primitives don't get inlined. We thus need to make sure that all calls to `build_derived_rrule` get seen by the forward pass. This means that the inner function must contain `prepare_pullback_cache`, and currently every Hessian-vector product thus builds the derived rule, which is very slow. My idea to fix this is to introduce `build_primitive_frule`, such that we will be able to instantiate a callable struct once during the outer preparation, which will be able to efficiently cache the inner preparation. This should make taking derivatives of functions which call `build_primitive_rule` reasonably efficient, and should also generalized to higher orders. In the future we could even provide a dedicated Hessian API for extra speed!

> 
> EDIT:
> 
> > Fix dynamic frule using _typeof instead of Base._stable_typeof.
> 
> Isn't it the other way around? The code replaces `_typeof` with ` Base._stable_typeof`.
> 
> > Fix test_rule using false instead of debug_mode when checking the rule cache.
> 
> Similar to the above, the code replaces `false` with `debug_mode`.

It depends on how you read it! What I mean is "fix \<bug\>" where \<bug\> = "dynamic frule using _typeof instead of Base._stable_typeof", but you read it as "fix \<bug\> \<solution\>" where \<bug\> = "dynamic frule" and \<solution\> = "using _typeof instead of Base._stable_typeof", so indeed it is the other way around.

Technici4n:
Thanks a lot! 😄

PR #991, opened 2026-02-10
Fix #955: prevent const-folding primitive calls

### LINKED ISSUE #955: `rrule!!` ignored for inlined function
In this example, my custom `rrule!!` never gets called

```julia-repl
julia> using Mooncake: Mooncake, @is_primitive, DefaultCtx, ReverseMode, CoDual, NoRData

julia> fake_grad(x, y) = x
fake_grad (generic function with 1 method)

julia> @is_primitive DefaultCtx ReverseMode Tuple{typeof(fake_grad), Any, Any}

julia> function Mooncake.rrule!!(::Mooncake.CoDual{typeof(fake_grad)}, x::CoDual, y::CoDual)
           function fake_grad_adjoint(dy)
               return NoRData(), NoRData(), dy
           end
           return CoDual(x.x, y.dx), fake_grad_adjoint
       end

julia> import DifferentiationInterface as DI

julia> DI.value_and_gradient(DI.AutoMooncake(), [1., 2., 3.]) do x
           fake_grad(1.0, 2x[1] + x[3])
       end
(1.0, [0.0, 0.0, 0.0])
```

If I force `fake_grad` to not be inlined, I get the expected answer:

```julia-repl
julia> @noinline fake_grad(x, y) = x
fake_grad (generic function with 1 method)

julia> DI.value_and_gradient(DI.AutoMooncake(), [1., 2., 3.]) do x
           fake_grad(1.0, 2x[1] + x[3])
       end
(1.0, [2.0, 0.0, 1.0])
```

sunxd3:
Fixes #955.

some description is in this comment https://github.com/chalk-lab/Mooncake.jl/pull/991#issuecomment-3880651215

### DIFF
--- a/src/interpreter/abstract_interpretation.jl
+++ b/src/interpreter/abstract_interpretation.jl
@@ -238,9 +238,9 @@ function maybe_widen_primitive_call_return_type(call::CC.CallMeta, argtypes::Vec
     rt = should_widen ? CC.widenconst(call.rt) : call.rt
 
     @static if VERSION ≥ v"1.11-"
-        return CC.CallMeta(rt, call.exct, call.effects, info)
+        return CC.CallMeta(rt, call.exct, call.effects, call.info)
     else
-        return CC.CallMeta(rt, call.effects, info)
+        return CC.CallMeta(rt, call.effects, call.info)
     end
 end
 


src/interpreter/abstract_interpretation.jl
```diff
@@ -197,11 +204,21 @@ function any_matches_primitive(applicable, C, M, world)
...
+function should_widen_primitive_call_return_type(rt, argtypes::Vector{Any})
+    rt isa CC.Const || return false
+    # `argtypes` usually includes the callee in position 1; we only care about runtime arguments.
+    for n in 2:length(argtypes)
+        argtypes[n] isa CC.Const || return true
+    end
+    return false
+end
+
+function rewrap_callmeta(call::CC.CallMeta, info::CC.CallInfo; widen_rt::Bool=false)
```
yebai: Positional argument should do the trick since `rewrap_callmeta` is only called twice and both are passed with `widen_rt` flag. 

```suggestion
function rewrap_callmeta(call::CC.CallMeta, info::CC.CallInfo, widen_rt::Bool)
```

src/interpreter/abstract_interpretation.jl
```diff
@@ -169,13 +169,20 @@ function Core.Compiler.abstract_call_gf_by_type(
...
+                return rewrap_callmeta(call, info; widen_rt=widen_rt)
             else
                 return CC.Future{CC.CallMeta}(
                     ret::CC.Future, interp, sv
                 ) do call, interp, sv
                     info = NoInlineCallInfo(call.info, atype)
-                    return rewrap_callmeta(call, info)
+                    # See comment in the non-Future branch above.
+                    widen_rt = should_widen_primitive_call_return_type(call.rt, argtypes)
+                    return rewrap_callmeta(call, info; widen_rt=widen_rt)
```
yebai:
```suggestion
                    return rewrap_callmeta(call, info, widen_rt)
```

Technici4n:
What exactly is the problem in #955? It doesn't look related to constants, does it? Or is x considered constant in the scope of the function?

sunxd3:
yeah, that's pretty much the reason

```julia
using Mooncake
using Mooncake: DefaultCtx, ReverseMode, CoDual, NoRData, get_interpreter, generate_ir
using InteractiveUtils

fake_grad(x, y) = x

f(a, b) = fake_grad(1.0, 2a + b)

a = 2.0
b = 3.0
```

```julia
julia> @code_warntype optimize=false f(a, b)
MethodInstance for f(::Float64, ::Float64)
  from f(a, b) @ Main REPL[15]:1
Arguments
  #self#::Core.Const(Main.f)
  a::Float64
  b::Float64
Body::Float64
1 ─ %1 = Main.fake_grad::Core.Const(Main.fake_grad)
│   %2 = Main.:+::Core.Const(+)
│   %3 = Main.:*::Core.Const(*)
│   %4 = (%3)(2, a)::Float64
│   %5 = (%2)(%4, b)::Float64
│   %6 = (%1)(1.0, %5)::Core.Const(1.0)
└──      return %6
```

<details><summary>rule def</summary>
<p>

```julia
Mooncake.@is_primitive DefaultCtx ReverseMode Tuple{typeof(fake_grad),Any,Any}

function Mooncake.rrule!!(::CoDual{typeof(fake_grad)}, x::CoDual, y::CoDual)
    pb(dy) = (NoRData(), NoRData(), dy)
    return CoDual(x.x, y.dx), pb
end
```

</p>
</details> 

```julia
julia> interp = get_interpreter(ReverseMode); dri = generate_ir(interp, Tuple{typeof(f),Float64,Float64}; do_inline=false);

julia> display(dri.rvs_ir)
 1 ─      nothing::Nothing                                                                                                         │
 2 ─      goto #3                                                                                                                  │
 3 ─      nothing::Nothing                                                                                                         │
 4 ─      goto #5                                                                                                                  │
 5 ─ %5 =   dynamic (Mooncake.increment!!)($(QuoteNode(0.0)), 0.0)::Float64                                                        │
 │   %6 =   dynamic (Mooncake.increment!!)($(QuoteNode(0.0)), 0.0)::Float64                                                        │
 │   %7 =   builtin (tuple)($(QuoteNode(NoRData())), %5, %6)::Tuple{NoRData, Float64, Float64}                                     │
 └──      return %7  
```

docs/src/developer_documentation/ir_representation.md
```diff
@@ -416,18 +416,15 @@ Since the name associated to instructions does not change when you insert anothe
 ```jldoctest my_factorial
 julia> using Mooncake.BasicBlockCode: ID, new_inst
 
-julia> new_id = ID() # this produces a new unique `ID`.
-ID(108)
+julia> new_id = ID(); # this produces a new unique `ID`.
```
sunxd3, line 419: the hardcoded ID is a bit problematic, so I use a semicolon. of course this is loss of info and test, but maybe it's okay

src/interpreter/abstract_interpretation.jl
```diff
@@ -197,11 +204,21 @@ function any_matches_primitive(applicable, C, M, world)
     false
 end
 
-function rewrap_callmeta(call::CC.CallMeta, info::CC.CallInfo)
+function should_widen_primitive_call_return_type(rt, argtypes::Vector{Any})
```
yebai: Some additional docstrings and examples for this utility (eg, cases it returns true and false) would be very helpful for people unfamiliar with Julia's internals.

src/rules/avoiding_non_differentiable_code.jl
```diff
@@ -17,6 +17,10 @@ end
 @zero_derivative MinimalCtx Tuple{Type{Float16},Any,RoundingMode}
 @zero_derivative MinimalCtx Tuple{typeof(==),Type,Type}
 
+# `count` returns an `Int` and is non-differentiable; differentiating through its
+# implementation can also lead to avoidable allocations on LTS (see integration tests).
```
yebai: Better not to mention "integration tests" or be more specific about what it refers to. 

```suggestion
# Optional rule to avoid unnecessary allocations on LTS
```

test/interpreter/abstract_interpretation.jl
```diff
@@ -8,6 +8,19 @@ contains_non_primitive(x) = @inline non_primitive(x)
 contains_primitive_behind_call(x) = @inline contains_primitive(x)
 union_split_primitive_call(x::Ref{Union{Float64,Float32}}) = @inline a_primitive(x[])
 
+# Issue #955: if a primitive call's return value is inferred as `Const`, the compiler can
+# fold the call away entirely. That makes the primitive invisible to Mooncake, so its
+# custom `rrule!!` never runs.
```
yebai:
```suggestion
# Issue #955: if a primitive call's return value is inferred as `Const`,
# the compiler may fold the call away entirely. This makes the primitive 
# invisible to Mooncake,  so its custom `rrule!!` never runs.
```

src/interpreter/abstract_interpretation.jl
```diff
@@ -169,13 +169,20 @@ function Core.Compiler.abstract_call_gf_by_type(
             @static if VERSION < v"1.12-"
                 call = ret::CC.CallMeta
                 info = NoInlineCallInfo(call.info, atype)
-                return rewrap_callmeta(call, info)
+                # Primitives must remain visible in the caller IR so we can apply their
+                # custom rules. Prevent inlining *and* disable const-prop of the return
+                # value (otherwise `compact!` can fold the call away entirely, e.g. a
+                # primitive whose inferred return is `Const`).
+                widen_rt = should_widen_primitive_call_return_type(call.rt, argtypes)
```
yebai:
```suggestion
                widen_rt = should_widen_primitive_call_return_type(call, argtypes)
```

src/interpreter/abstract_interpretation.jl
```diff
@@ -169,13 +169,20 @@ function Core.Compiler.abstract_call_gf_by_type(
...
+                return rewrap_callmeta(call, info, widen_rt)
             else
                 return CC.Future{CC.CallMeta}(
                     ret::CC.Future, interp, sv
                 ) do call, interp, sv
                     info = NoInlineCallInfo(call.info, atype)
-                    return rewrap_callmeta(call, info)
+                    # See comment in the non-Future branch above.
+                    widen_rt = should_widen_primitive_call_return_type(call.rt, argtypes)
```
yebai:
```suggestion
                    widen_rt = should_widen_primitive_call_return_type(call, argtypes)
```

src/interpreter/abstract_interpretation.jl
```diff
@@ -197,11 +204,21 @@ function any_matches_primitive(applicable, C, M, world)
     false
 end
 
-function rewrap_callmeta(call::CC.CallMeta, info::CC.CallInfo)
+function should_widen_primitive_call_return_type(rt, argtypes::Vector{Any})
+    rt isa CC.Const || return false
```
yebai:
```suggestion
function should_widen_primitive_call_return_type(call::CC.CallMeta, argtypes::Vector{Any})
    call.rt isa CC.Const || return false
```

yebai (approved):
Looks good -- I have a few comments below, mostly about clarity.

src/interpreter/abstract_interpretation.jl
```diff
@@ -197,11 +204,21 @@ function any_matches_primitive(applicable, C, M, world)
...
+    # `argtypes` usually includes the callee in position 1; we only care about runtime arguments.
+    for n in 2:length(argtypes)
+        argtypes[n] isa CC.Const || return true
+    end
+    return false
+end
+
+function rewrap_callmeta(call::CC.CallMeta, info::CC.CallInfo, widen_rt::Bool)
+    rt = widen_rt ? CC.widenconst(call.rt) : call.rt
```
yebai:
```suggestion
    # If `widen_rt` is true, widen `call.rt` with `CC.widenconst`,  
    # discarding any `Const` information to prevent constant folding;  
    # otherwise, use `call.rt` unchanged.
    #
    # This drops value-level information tracked by inference  
    # (e.g. `Const(3)` → `Int`), ensuring the call is not folded away.  
    # For example, if `call.rt` is inferred as `Const(42)`, widening  
    # yields `Int`, preserving the call so downstream rules still apply.
    rt = widen_rt ? CC.widenconst(call.rt) : call.rt
```

src/interpreter/abstract_interpretation.jl
sunxd3: added a block of comments https://github.com/chalk-lab/Mooncake.jl/blob/13ec0da97ebcd14fc7adb80f08cc7e1445c979ae/src/interpreter/abstract_interpretation.jl#L207-L244

Technici4n:
Looking back at #955, is there even an issue? It's unexpected that the compiler decided to constant fold the function entirely, but does it cause problems in practice? (After all the values are constant)

yebai:
cc @simeonschaub

yebai:
> Looking back at https://github.com/chalk-lab/Mooncake.jl/issues/955, is there even an issue? It's unexpected that the compiler decided to constant fold the function entirely, but does it cause problems in practice? (After all the values are constant)

In some contexts, this likely reflects a genuine issue, so addressing it is preferable.

More broadly, this appears to be one of three scenarios in which a primitive may be (constant-)folded or inlined, and thus become invisible to Mooncake:

1. **Inlining**  
   The call is replaced by its body. The primitive no longer appears as a call; only its internal operations remain.

2. **Constant propagation / constant folding**  (eg, #955)
   If the result can be computed at compile time, the entire call is replaced by a literal constant.

3. **Compile-time evaluation of effect-free calls with known arguments**  
   When a function is effect-free and all arguments are known at compile time, the compiler evaluates the call during compilation and replaces it with its result.

After this PR, cases (2) and (3) should be addressed. Case (1) has been addressed early on. ~~Case (3), however, remains to be identified and addressed.~~

yebai:
I refactored Xianda’s code to introduce minimal changes relative to main. We want to encourage this approach to (1) reduce the risk of accidental bugs and (2) make code changes easier to review and track in PRs.

simeonschaub:
> Looking back at https://github.com/chalk-lab/Mooncake.jl/issues/955, is there even an issue? It's unexpected that the compiler decided to constant fold the function entirely, but does it cause problems in practice? (After all the values are constant)

My example is certainly a bit of an odd one, since I explicitly return a "wrong" gradient in the reverse pass, so I don't know how much of this is considered officially supported. It seems like this issue mainly crops up for very simple functions, so I am not actually sure how much of a problem this would cause in the wild, but it certainly stumped me at first that my rule wasn't applied

yebai:
@sunxd3 I'm happy with this PR. Perhaps wait for 1-2 days before merging and releasing so more can chip in with alternative views,

Issue #969, opened 2026-01-22
Blas level 3 CI issues on Julia v1.12


AstitvaAggarwal:
see for example: https://github.com/chalk-lab/Mooncake.jl/actions/runs/21254113214/job/61163632081?pr=967.
This failure seems to keep randomnly coming up after https://github.com/chalk-lab/Mooncake.jl/pull/942.

yebai:
The rules/blas_level_3 test group is likely experiencing out-of-memory kills on GitHub Actions runners. The job runs for ~35 minutes before receiving a SIGTERM, indicating that the runner's 7GB memory limit is exceeded by the matrix-matrix operations being tested. Recommended fix: split the BLAS level 3 tests into smaller groups to reduce memory pressure.

yebai:
@sunxd3 can you take a look please?

sunxd3:
I investigated this a bit.

This CI does pass locally and on [GHA](https://github.com/chalk-lab/Mooncake.jl/actions/runs/21864407760/job/63101699848?pr=991) (at least it's not failing deterministically).

~It is possible that some upstream fixes have already bailed us out (e.g. from JET or Julia Stdlibs). I'll keep an eye on this. We can keep it open for slightly a longer while.~ (this is a misunderstanding)~

It is failing sporadically and it looks like resource-related kill, likely OOM.

sunxd3:
For a quick fix, splitting the tests in to two groups should work.

But the real question is why this is 1.12 specific (1.10, 1.11 runs pretty well with half the time).

sunxd3:
More failed runs: https://github.com/chalk-lab/Mooncake.jl/actions/runs/22118810295/job/63933686978

sunxd3:
it doesn't look like the test is OOM kill, but [here](https://gist.github.com/sunxd3/550684db502ad938f32d1953dc960bef) is some investigations for why 1.12 might be more memory-intensive than the previous two versions (this is written with LLM, so might be better read with assist from an agent)

PR #997, opened 2026-02-17
Tuple handling for `Mooncake.increment_and_get_rdata!`

AstitvaAggarwal:
Is a basic dispatch, to be used in for example: https://github.com/SciML/Integrals.jl/actions/runs/22083336973/job/63812815712?pr=319.

**Description** : `Mooncake.increment_and_get_rdata!` has no method for dispatching on tuple tangents, i.e. when the cotangent returned from a `ChainRulesCore.rrule` is a Tuple. 

**MWE** :
```
using Mooncake, ChainRulesCore

# Case 1: plain float tuple - fdata is a NoFData, rdata is Tuple
r = (1.0, 2.0)
f_no = Mooncake.fdata(r)
t = ChainRulesCore.Tangent{typeof(r)}(0.1, 0.2)
Mooncake.increment_and_get_rdata!(f_no, r, t)

# Case 2: tuple of arrays - fdata is a Tuple, rdata is NoRData
f = ([1.0, 1.0], [2.0, 2.0])
r_no = Mooncake.rdata(f)
t = ChainRulesCore.Tangent{typeof(f)}([0.1, 0.1], [0.2, 0.2])
Mooncake.increment_and_get_rdata!(f, r_no, t)

```

### DIFF
--- a/src/tools_for_rules.jl
+++ b/src/tools_for_rules.jl
@@ -413,7 +413,10 @@ function increment_and_get_rdata!(f, r, t::CRC.Thunk)
     return increment_and_get_rdata!(f, r, CRC.unthunk(t))
 end
 
-# The need for these dispatches comes up for example in: https://github.com/SciML/Integrals.jl/pull/319 while handling Domain's gradients.
+# Tuple tangents from ChainRulesCore require special handling because tuple elements
+# may be a mix of types: some with only rdata (e.g., scalars), some with only fdata
+# (e.g., arrays), and some with both. These four dispatches for increment_and_get_rdata!
+# handle all the possible cases for when the ChainRulesCore.Tangent has Tuple type data.
 function increment_and_get_rdata!(f, r, t::CRC.Tangent{P,<:Tuple}) where {P}
     return increment_and_get_rdata!(f, r, t.backing)
 end
@@ -422,7 +425,7 @@ function increment_and_get_rdata!(f::NoFData, r::Tuple, t::Tuple)
 end
 function increment_and_get_rdata!(f::Tuple, r::NoRData, t::Tuple)
     increment!!(f, t)
-    return r
+    return NoRData()
 end
 function increment_and_get_rdata!(f::Tuple, r::Tuple, t::Tuple)
     return map((fi, ri, ti) -> increment_and_get_rdata!(fi, ri, ti), f, r, t)


sunxd3:
can you give a MWE and description of the problem?

sunxd3:
Naive question: are there use cases for supporting 'Tuple{T1, T2}'? (Asking because you decide to support the more general case)

AstitvaAggarwal:
Yes a tuple can be heterogeneous, for example:

```
f = ((1.0, 1.0), 2.0)  # nested tuple and Float64
```

Here the first element is a tuple and the second is a float ( both have all tangent data only in rdata), so the general type `Tuple` or `Tuple{T1, T2}` is the correct common case. Restricting to Tuple{T, T} would miss this.

sunxd3:
How might this show up in realistic scenarios?

AstitvaAggarwal:
> How might this show up in realistic scenarios?

I believe chainrules tangent data being tuples is common.

One thing I can think of is for example: the tangent of the domains of variables (in cases like change in variables during integration etc). This can be the single / multivariate case (tuple of vectors or just tuples).

Apart from this, I'm pretty sure the current missing dispatch of `increment_and_get_rdata!` will cause errors anywhere the chainrules tangent is a tuple.

sunxd3:
Sounds good, add some tests for the general case?

AstitvaAggarwal:
Also added mixed tangent (fdata, rdata) data cases handling + tests. Tuples should be totally covered now.

AstitvaAggarwal:
the possible blas failures are unrelated i [think](https://github.com/chalk-lab/Mooncake.jl/issues/969)

src/tools_for_rules.jl
```diff
@@ -413,6 +413,21 @@ function increment_and_get_rdata!(f, r, t::CRC.Thunk)
     return increment_and_get_rdata!(f, r, CRC.unthunk(t))
 end
 
+# The need for these dispatches comes up for example in: https://github.com/SciML/Integrals.jl/pull/319 while handling Domain's gradients.
```
sunxd3: maybe make this a bit more clear what the problem is, also better to break off the long line

src/tools_for_rules.jl
```diff
@@ -413,6 +413,21 @@ function increment_and_get_rdata!(f, r, t::CRC.Thunk)
...
+function increment_and_get_rdata!(f, r, t::CRC.Tangent{P,<:Tuple}) where {P}
+    return increment_and_get_rdata!(f, r, t.backing)
+end
+function increment_and_get_rdata!(f::NoFData, r::Tuple, t::Tuple)
+    return map((ri, ti) -> increment_and_get_rdata!(f, ri, ti), r, t)
+end
+function increment_and_get_rdata!(f::Tuple, r::NoRData, t::Tuple)
+    increment!!(f, t)
+    return r
```
sunxd3:
```suggestion
    return NoRData()
```

PR #1006, opened 2026-02-25
Add rrule for BridgeStan's log_density function

nsiccha:
Adds a MooncakeBridgeStanExt package extension providing rrule!! for BridgeStan.log_density.

BridgeStan's log_density calls into compiled Stan C++ code, so Mooncake cannot differentiate through it generically. This extension uses the existing log_density_gradient! API — which Stan already computes analytically — for an exact, efficient reverse-mode rule.

- tangent_type(StanModel) = NoTangent: StanModel holds C pointers Mooncake must not recurse into
- Two rrule!! methods: positional log_density(sm, q) and kwarg log_density(sm, q; propto, jacobian)
- Tests cover both call forms across all propto/jacobian combinations

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -88,6 +88,7 @@ jobs:
       fail-fast: false
       matrix:
         test_group: [
+          {test_type: 'ext', label: 'bridgestan'},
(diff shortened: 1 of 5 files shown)


yebai:
This could be added to the BridgeStan repo.

PR #993, opened 2026-02-11
Mooncake Dispatch for `CRC.rrule_via_ad`

AstitvaAggarwal:
To be merged after https://github.com/chalk-lab/Mooncake.jl/pull/965
Addresses #848  for Reverse Mode AD.

PR #965, opened 2026-01-21
`Lux` Kernel rules + Mooncake `CRC.RuleConfig`

AstitvaAggarwal:
Closes [LuxDL/Lux.jl#1472](https://github.com/LuxDL/Lux.jl/issues/1472), addresses the second part under lux here [link](https://github.com/chalk-lab/Mooncake.jl/issues/278). Will be followed by another PR in Lux.jl adding Mooncake.jl in its Benchmark Tests.

1. Add more Lux Kernel rules - Mooncake can now diff through almost all of Lux.
2.  Adds ChainRules.RuleConfig [Superpower](https://juliadiff.org/ChainRulesCore.jl/stable/rule_author/superpowers/ruleconfig.html) capabilities for Mooncake - We can now write and import ChainRulesCore rules which specifically work with Mooncake (default) or some other AD backend. This is chosen by setting the cfg kwarg in @from_chainrules etc (default is Mooncake.MooncakeRuleConfig())

**TODO** : Add MooncakeRuleConfig dispatch for CRC.rrule_via_ad. Is ready here: [chalk-lab/Mooncake.jl#993](https://github.com/chalk-lab/Mooncake.jl/pull/993).

**Note:** 
Currently we dont directly test the following in mooncake (due to below mentioned limitations).
But `Mooncake`  still correctly works with the `ChainRulesCore` rules, the application of these rules is tested in Lux's CI :

- `activation!!`, `bias_activation!!` - circular referencing within the function fails the test framework's address mapping assertion.

- `alpha_dropout` & `dropout_dot_mul` - Test framework (FD) perturbs all inputs, but rules return NoTangent() for masks/noise, causing FD vs AD mismatch.

Further,

- the added `cfg` arg/kwarg in the relevant Mooncake methods/macros is required to accommodate the usage of chainrule rules that depend on `CRC.RuleConfig` arg info.

- Following this PR, we require https://github.com/chalk-lab/Mooncake.jl/pull/993

PR #1007, opened 2026-02-25
Native Rule for LuxLib batchnorm

AstitvaAggarwal:
Preferred alternative to https://github.com/chalk-lab/Mooncake.jl/pull/965, https://github.com/chalk-lab/Mooncake.jl/pull/993 for `LuxLib.Impl.batchnorm_affine_normalize_internal`.

- It's `ChainRulesCore.rule` uses `ruleconfig, rrule_via_ad` as seen [here](https://github.com/LuxDL/Lux.jl/blob/5744a4afccd4959b438ac671bfdeefc18d4077a9/lib/LuxLib/src/impl/batchnorm.jl#L405).

- Starting relatively small for clarity by updating the current [workaround](https://github.com/chalk-lab/Mooncake.jl/blob/358dd72be44465c2bfd2f2347f6fdbdbea1f6086/ext/MooncakeLuxLibExt.jl#L57).

- Added missing Tests covering full `LuxLib.Optional` dispatch for `β`, `γ` arguments.

- This is the first PR in a series of PRs that adds rules for core `LuxLib`  optimized methods.

### DIFF
--- a/ext/MooncakeLuxLibExt.jl
+++ b/ext/MooncakeLuxLibExt.jl
@@ -94,7 +94,6 @@ function _batchnorm_affine_normalize_identity(
 end
 
 # Native Mooncake rule for batchnorm_affine_normalize_internal.
-# `MinimalCtx` to avoid upstream CRC.rrule which uses `rrule_via_ad` which is avoided by Mooncake for performance.
 @is_primitive MinimalCtx Tuple{
     typeof(_batchnorm_affine_normalize_identity),
     AbstractInternalArrayOpMode,


ext/MooncakeLuxLibExt.jl
```diff
@@ -66,90 +80,79 @@ import LuxLib.Impl:
...
-    σ²::AbstractVector,
-    γ::Union{Nothing,AbstractVector},
-    β::Union{Nothing,AbstractVector},
-    ϵ::Real,
-) where {F,xT}
-    y = batchnorm_affine_normalize_internal(opmode, identity, x, μ, σ², γ, β, ϵ)
-    LuxLib.Impl.activation!(y, opmode, act, y)
-    return y
-end
+    act_cache = prepare_pullback_cache(_act, y_pre)
```
sunxd3: **Performance: `prepare_pullback_cache` called inside `rrule!!`**

`prepare_pullback_cache` is designed as a one-time setup function. Calling it here means every batchnorm forward pass pays for:
- `deepcopy` of all arguments (via `__exclude_func_with_unsupported_output`)
- A redundant primal evaluation of the activation on the full array
- Fresh tangent allocation (`map(zero_tangent, fx)`)
- A full warmup forward + backward pass through the compiled rule
- Output cache allocation

The `build_rrule` compilation itself IS cached (via `interp.oc_cache`), but everything else is not. The `Cache` object is meant to be created once and reused across many `value_and_pullback!!` calls.

Also note that `value_and_pullback!!` (line 135) re-runs the activation forward pass internally, so `_act.(y_pre)` is effectively computed twice per call.

**Suggestion:** Either hoist the cache outside the `rrule!!` body (e.g. keyed by `(F, eltype, size)`), use `build_rrule` directly and manage forward/reverse manually, or reconsider whether the old `@mooncake_overlay` approach (which let Mooncake trace through the activation natively with zero per-call setup) was actually cheaper.

ext/MooncakeLuxLibExt.jl
```diff
@@ -54,9 +70,7 @@ Mooncake.@zero_adjoint DefaultCtx Tuple{
     typeof(LuxLib.Impl.generate_dropout_mask),AbstractRNG,Any,Any,Any,Any
 }
 
-# This is a really horrible hack that we need to do until Mooncake is able to support the
-# call-back-into-ad interface that ChainRules exposes.
-
+# Native Mooncake Rules for ChainRules that use RuleConfig, rrule_via_ad functionality.
```
sunxd3: **Misleading comment**

This says the new code uses `RuleConfig`/`rrule_via_ad`, but it actually **replaces** that mechanism with a native Mooncake `rrule!!`.

Suggested:
```julia
# Native Mooncake rule for batchnorm_affine_normalize_internal.
# Upstream CRC.rrule uses rrule_via_ad (callback into AD) which Mooncake
# cannot handle, so we provide a direct rrule!! with manual activation pullback.
```

ext/MooncakeLuxLibExt.jl
```diff
@@ -6,7 +6,23 @@ using Base: IEEEFloat
...
+    NoTangent,
+    zero_tangent,
+    primal,
+    @is_primitive,
+    NoRData,
+    extract,
+    zero_fcodual,
+    tangent,
+    prepare_pullback_cache,
+    value_and_pullback!!
```
sunxd3: **Unused imports**

`zero_fcodual` and `tangent` are not used anywhere in the file. `NoTangent` is also unused — the code uses `CRC.NoTangent` instead.

```suggestion
using Mooncake:
    @from_rrule,
    DefaultCtx,
    MinimalCtx,
    @mooncake_overlay,
    CoDual,
    zero_tangent,
    primal,
    @is_primitive,
    NoRData,
    extract,
    prepare_pullback_cache,
    value_and_pullback!!
```

ext/MooncakeLuxLibExt.jl
```diff
@@ -66,90 +80,79 @@ import LuxLib.Impl:
...
-        promote_type(
-            safe_eltype(x), safe_eltype(μ), safe_eltype(σ²), safe_eltype(γ), safe_eltype(β)
-        ),
-    )
-    γ′ = similar(
-        x, promote_type(safe_eltype(γ), safe_eltype(σ²), safe_eltype(ϵ)), size(x, N - 1)
-    )
-
-    batchnorm_affine_normalize_internal!(y, opmode, identity, x, μ, σ², γ, β, ϵ, γ′)
+@is_primitive MinimalCtx Tuple{
```
sunxd3, line 97: **MinimalCtx is the right choice — but worth a comment**

The upstream `CRC.rrule` uses `rrule_via_ad` which Mooncake cannot trace, so AD would *fail* without this rule — justifying `MinimalCtx` over `DefaultCtx`. A one-liner above this line would help future readers:

```julia
# MinimalCtx: upstream CRC.rrule uses rrule_via_ad which Mooncake cannot trace
```

ext/MooncakeLuxLibExt.jl
```diff
@@ -66,90 +80,79 @@ import LuxLib.Impl:
...
+        ∂x, ∂μ, ∂σ², ∂γ, ∂β = ∇batchnorm_affine_normalize(
+            _opmode, ∂y_pre, _x, _μ, _σ², _γ, _β, _ϵ, γ′
+        )
+
+        # accumulate gradients
+        x̄ .+= ∂x
+        μ̄ .+= ∂μ
+        σ²̄ .+= ∂σ²
+        # γ, β may have NoTangent gradients for primal=nothing
+        ∂γ isa CRC.NoTangent || (γ̄ .+= ∂γ)
```
sunxd3: **Minor: cross-framework type guard**

This mixes ChainRulesCore types into a Mooncake rule. Consider checking the primal directly instead:

```julia
primal(γ) === nothing || (γ̄ .+= ∂γ)
primal(β) === nothing || (β̄ .+= ∂β)
```

This avoids depending on what type `∇batchnorm_affine_normalize` happens to return.

ext/MooncakeLuxLibExt.jl
```diff
@@ -66,90 +80,79 @@ import LuxLib.Impl:
...
+        ∂β isa CRC.NoTangent || (β̄ .+= ∂β)
+
+        return NoRData(),
```
sunxd3, line 145: **Minor: `ϵ` rdata type**

The 9th `NoRData()` here corresponds to `ϵ::CoDual{<:Real}`. For `Float64`, the formal rdata type should be `Float64`, so this should arguably be `zero_rdata(primal(ϵ))` instead of `NoRData()`. Works in practice since `ϵ` is always a constant, but technically a type mismatch.

test/ext/luxlib/luxlib.jl
```diff
@@ -62,6 +62,23 @@ using Mooncake.TestUtils: test_rule
...
+                    )
+                end,
+                LuxLib.LoopedArrayOp(),
+                Lux.relu,
+                randn(5, 4, 3),
+                randn(4),
+                rand(4) .+ 1.0,
+                randn(4),
+                randn(4),
+            ),
```
sunxd3: **Test coverage is thin relative to code change scope**

The new `rrule!!` handles arbitrary activation functions `F` and all `AbstractInternalArrayOpMode`s, but only `relu` + `LoopedArrayOp` are tested. The γ/β Optional coverage is correct (LuxLib only supports both-nothing or both-vector), but consider a Cartesian product matching the `fused_dense` tests below (line 83-102):

```julia
vec(map(Iterators.product(
    [LuxLib.LoopedArrayOp(), LuxLib.GenericBroadcastOp{Lux.CPUDevice()}()],
    [(nothing, nothing), (randn(4), randn(4))],
    [Lux.relu, tanh, identity],
)) do (opmode, (gamma, beta), activation)
    (false, :none, false,
     (opmode, act, x, m, s, g, b) -> LuxLib.Impl.batchnorm_affine_normalize_internal(
         opmode, act, x, m, s, g, b, 1e-3),
     opmode, activation, randn(5,4,3), randn(4), rand(4).+1.0, gamma, beta)
end)
```

The `identity` case is especially important — it's the boundary case where the activation pullback should be a no-op.

yebai:
@avik-pal, we decided to rewrite some CRC Lux rules as native Mooncake rules to reduce indirection and complexity. It would be helpful if you could review this PR and #1017 so these rules are written in good Lux practise and performance.

test/ext/luxlib/luxlib.jl
AstitvaAggarwal: I have not written tests for `GenericBroadcastOp` as `batchnorm_affine_normalize_internal!` does not have a dispatch for `LuxLib.GenericBroadcastOp`.

ext/MooncakeLuxLibExt.jl
AstitvaAggarwal: The old `@mooncake_overlay`approach needs a native rule as discussed.

ext/MooncakeLuxLibExt.jl
```diff
@@ -66,90 +77,98 @@ import LuxLib.Impl:
...
+    LuxLib.Optional{<:AbstractVector},
+    Real,
+}
 
-    ∇batchnorm_affine_normalize_internal = LuxLib.Impl.@closure Δ -> begin
-        ∂x, ∂μ, ∂σ², ∂γ, ∂β = ∇batchnorm_affine_normalize(opmode, Δ, x, μ, σ², γ, β, ϵ, γ′)
-        ∂∅ = CRC.NoTangent()
-        return ∂∅, ∂∅, ∂∅, 𝒫x(∂x), 𝒫μ(∂μ), 𝒫σ²(∂σ²), 𝒫γ(∂γ), 𝒫β(∂β), ∂∅
+function Mooncake.rrule!!(
+    ::CoDual{typeof(_batchnorm_affine_normalize_identity)},
```
AstitvaAggarwal, line 109: @sunxd3 Now the Mooncake Rule totally lives within Mooncake. Ive added a AD helper mapping function to totally isolate the affine transform (without type piracy or performance loss via cache creation or pullback/double forward passes) while the activation is diffed through the chain rule directly in Mooncake.

ext/MooncakeLuxLibExt.jl
```diff
@@ -66,90 +77,98 @@ import LuxLib.Impl:
...
-        AbstractVector,
-        LuxLib.Optional{<:AbstractVector},
-        LuxLib.Optional{<:AbstractVector},
-        Real,
-    },
-)
-
+# Overlay for `batchnorm_affine_normalize_internal` to use Mooncake AD's mapping function -> `_batchnorm_affine_normalize_identity`
+# the chosen `act` function's gradient is derived by Mooncake normally.
 @mooncake_overlay function batchnorm_affine_normalize_internal(
```
AstitvaAggarwal, line 157: Maybe this approach can be done for all the `rrule_via_ad` usage cases, this will make almost everything much better.

ext/MooncakeLuxLibExt.jl
```diff
@@ -6,7 +6,21 @@ using Base: IEEEFloat
...
+    DefaultCtx,
+    MinimalCtx,
+    @mooncake_overlay,
+    CoDual,
+    zero_tangent,
+    primal,
+    @is_primitive,
+    NoRData,
+    extract,
+    prepare_pullback_cache,
```
sunxd3:
```suggestion
```

ext/MooncakeLuxLibExt.jl
```diff
@@ -6,7 +6,21 @@ using Base: IEEEFloat
...
+    value_and_pullback!!,
```
sunxd3:
```suggestion
```

ext/MooncakeLuxLibExt.jl
```diff
@@ -66,90 +77,98 @@ import LuxLib.Impl:
 
 import ChainRulesCore as CRC
```
sunxd3:
```suggestion
```

ext/MooncakeLuxLibExt.jl
```diff
@@ -54,102 +66,105 @@ Mooncake.@zero_adjoint DefaultCtx Tuple{
...
     batchnorm_affine_normalize_internal,
     batchnorm_affine_normalize_internal!,
     ∇batchnorm_affine_normalize,
     AbstractInternalArrayOpMode
 
-import ChainRulesCore as CRC
-
-function CRC.rrule(
-    ::typeof(batchnorm_affine_normalize_internal),
+# AD helper mapping function for the Lux affine transform.
```
yebai: This comment is confusing — avoiding mentioning AD is more self-contained. 

```suggestion
# Helper function for the Lux affine transform.
```

ext/MooncakeLuxLibExt.jl
```diff
@@ -54,102 +66,105 @@ Mooncake.@zero_adjoint DefaultCtx Tuple{
...
+    y = similar(x, PT)
     batchnorm_affine_normalize_internal!(y, opmode, identity, x, μ, σ², γ, β, ϵ, γ′)
+    return y
+end
 
-    𝒫x, 𝒫μ, 𝒫σ² = CRC.ProjectTo(x), CRC.ProjectTo(μ), CRC.ProjectTo(σ²)
-    𝒫γ = γ === nothing ? identity : CRC.ProjectTo(γ)
-    𝒫β = β === nothing ? identity : CRC.ProjectTo(β)
+# Native Mooncake rule for batchnorm_affine_normalize_internal.
+# `MinimalCtx` to avoid upstream CRC.rrule which uses `rrule_via_ad` which is avoided by Mooncake for performance.
```
yebai: The sentence on `MinimalCtx` adds unnecessary details that are not helpful for clarity. The general rule is to either explain it clearly or avoid diving into the specifics. 

```suggestion
```

ext/MooncakeLuxLibExt.jl
```diff
@@ -54,102 +66,105 @@ Mooncake.@zero_adjoint DefaultCtx Tuple{
...
-        AbstractVector,
-        AbstractVector,
-        LuxLib.Optional{<:AbstractVector},
-        LuxLib.Optional{<:AbstractVector},
-        Real,
-    },
-)
-
+# Overlay for `batchnorm_affine_normalize_internal` to use Mooncake AD's mapping function -> `_batchnorm_affine_normalize_identity`
+# the chosen `act` function's gradient is derived by Mooncake normally.
```
yebai:
```suggestion
# Overlay for `batchnorm_affine_normalize_internal`
#  - Use Mooncake’s helper function `_batchnorm_affine_normalize_identity`  
#     and its manually written rule.
#  - Let Mooncake differentiate through the broadcasted `act` function.
```

yebai (commented):
Good work -- see below for a few clarity comments.

PR #1026, opened 2026-03-05
Mark interface functions with `@inline`

### LINKED ISSUE #1020: Severe perf degradations with functions as arguments
Passing a function as an argument to a differentiated function is really slow. Here's an example. I'm using Mooncake 0.5.13, Julia 1.12.5, macOS; although I can confirm that the same is observed on Julia 1.11.

```julia
using Mooncake, Chairmarks

foo1(xs::AbstractVector{<:Real}, f::Any) = f(xs)
xs = [1.0, 2.0, 3.0]

foo1(xs, sum)

gc1 = Mooncake.prepare_gradient_cache(foo1, xs, sum)
@be Mooncake.value_and_gradient!!($gc1, $foo1, $xs, $sum)
#=
Benchmark: 2816 samples with 75 evaluations
 min    376.107 ns (8 allocs: 240 bytes)
 median 393.333 ns (8 allocs: 240 bytes)
 mean   448.517 ns (8 allocs: 240 bytes, 0.04% gc time)
 max    148.838 μs (8 allocs: 240 bytes, 99.29% gc time)
=#
```

Note that changing the type parameter in `foo1` from `Any` to `Function` or even `typeof(sum)` doesn't help at all.

There are a couple of ways to work around this. One is to replace `sum` with a callable struct:

```julia
struct Sum end
(s::Sum)(xs::AbstractVector{<:Real}) = sum(xs)
s = Sum()
gc2 = Mooncake.prepare_gradient_cache(foo1, xs, s)

@be Mooncake.value_and_gradient!!($gc2, $foo1, $xs, $s)
#=
Benchmark: 3552 samples with 269 evaluations
 min    85.346 ns
 median 95.416 ns
 mean   97.905 ns
 max    209.881 ns
=#
```

The other is to wrap the function itself inside a callable struct, which turns out to be even faster:

```julia
struct FooStruct{F}
    f::F
end
(f::FooStruct)(xs::AbstractVector{<:Real}) = f.f(xs)
foo2 = FooStruct(sum)
foo2(xs)
gc3 = Mooncake.prepare_gradient_cache(foo2, xs)
@be Mooncake.value_and_gradient!!($gc3, $foo2, $xs)
#=
Benchmark: 3302 samples with 471 evaluations
 min    55.202 ns
 median 59.803 ns
 mean   60.729 ns
 max    114.119 ns
=#
```

I'm pretty sure that this is responsible for some level of overheads on DynamicPPL models, because the function we differentiate (`logdensity_at`, see https://github.com/TuringLang/DynamicPPL.jl/blob/efe25a85f053aaf37dee9f5cbba52cd7f3ac070f/src/logdensityfunction.jl#L349-L362) pretty much always takes a function like `DynamicPPL.getlogjoint_internal` as an argument.

For example, on this trivial model there are quite substantial overheads (DynamicPPL version is 0.40.6):

```julia
using DynamicPPL, Distributions, Mooncake, Chairmarks, ADTypes, LogDensityProblems

@model f() = x ~ Normal()
ldf = LogDensityFunction(f(), getlogjoint_internal, LinkAll(); adtype=AutoMooncake())
x = [1.0]
@be LogDensityProblems.logdensity_and_gradient($ldf, $x)
#=
julia> @be LogDensityProblems.logdensity_and_gradient($ldf, $x)
Benchmark: 2798 samples with 38 evaluations
 min    741.237 ns (16 allocs: 736 bytes)
 median 758.789 ns (16 allocs: 736 bytes)
 mean   880.780 ns (16 allocs: 736 bytes, 0.04% gc time)
 max    309.650 μs (16 allocs: 736 bytes, 99.32% gc time)
=#
```

Now, Claude suggested sticking an `@inline` on `value_and_gradient!!` on this method:

https://github.com/chalk-lab/Mooncake.jl/blob/38b6d4b489739abdc827e951e4be1c192f197349/src/interface.jl#L640

````diff
diff --git a/src/interface.jl b/src/interface.jl
index e83c35e92..7e3b413a1 100644
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -637,7 +637,7 @@ value_and_gradient!!(cache, f, x, y)
 (4.0, (NoTangent(), [1.0, 1.0], [2.0, 2.0]))
 ```
 """
-function value_and_gradient!!(
+@inline function value_and_gradient!!(
     cache::Cache,
     f::F,
     x::Vararg{Any,N};
````

That in fact completely solves the problem (even though I think Claude arrived at this via wrong reasoning). All three of the `sum` examples above benchmark to around 45 ns and 0 allocs, and the DynamicPPL model is now

```julia
julia> @be LogDensityProblems.logdensity_and_gradient($ldf, $x)
Benchmark: 3066 samples with 288 evaluations
 min    98.524 ns (2 allocs: 64 bytes)
 median 105.469 ns (2 allocs: 64 bytes)
 mean   106.563 ns (2 allocs: 64 bytes)
 max    279.514 ns (2 allocs: 64 bytes)
```

I don't know why that makes a difference, and I'm not convinced that Claude's argument was correct, but my guess is that there are issues with specialisation of Vararg on function-valued arguments (https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing)...?.

penelopeysm:
Another way of fixing #1020. This also fixes all the issues.

```julia
using Mooncake, Chairmarks

foo1(xs::AbstractVector{<:Real}, f::Any) = f(xs)
xs = [1.0, 2.0, 3.0]

foo1(xs, sum)

gc1 = Mooncake.prepare_gradient_cache(foo1, xs, sum)
@be Mooncake.value_and_gradient!!($gc1, $foo1, $xs, $sum)

#=
Benchmark: 3416 samples with 1661 evaluations
 min    16.105 ns
 median 16.657 ns
 mean   16.709 ns
 max    27.669 ns
=#

using DynamicPPL, Distributions, Mooncake, Chairmarks, ADTypes, LogDensityProblems

@model f() = x ~ Normal()
ldf = LogDensityFunction(f(), getlogjoint_internal, LinkAll(); adtype=AutoMooncake())
x = [1.0]
@be LogDensityProblems.logdensity_and_gradient($ldf, $x)

#=
Benchmark: 2960 samples with 279 evaluations
 min    103.645 ns (2 allocs: 64 bytes)
 median 105.287 ns (2 allocs: 64 bytes)
 mean   113.797 ns (2 allocs: 64 bytes, 0.03% gc time)
 max    22.737 μs (2 allocs: 64 bytes, 98.91% gc time)
=#
```

Closes #1025
Closes #1020

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -179,8 +179,6 @@ value_and_gradient!!(rule, f, x, y)
 (4.0, (NoTangent(), [1.0, 1.0], [2.0, 2.0]))
 ```
 """
-# @inline forces specialisation on Vararg with function-valued arguments, avoiding severe
-# perf regressions. See https://github.com/chalk-lab/Mooncake.jl/issues/1020.
 @inline function value_and_gradient!!(
     rule::R, fx::Vararg{Any,N}; friendly_tangents=false
 ) where {R,N}
@@ -554,8 +552,6 @@ Mooncake.value_and_pullback!!(cache, 1.0, f, x, y)
 (4.0, (NoTangent(), [1.0, 1.0], [2.0, 2.0]))
 ```
 """
-# @inline forces specialisation on Vararg with function-valued arguments, avoiding severe
-# perf regressions. See https://github.com/chalk-lab/Mooncake.jl/issues/1020.
 @inline function value_and_pullback!!(
     cache::Cache,
     ȳ,
@@ -643,8 +639,6 @@ value_and_gradient!!(cache, f, x, y)
 (4.0, (NoTangent(), [1.0, 1.0], [2.0, 2.0]))
 ```
 """
-# @inline forces specialisation on Vararg with function-valued arguments, avoiding severe
-# perf regressions. See https://github.com/chalk-lab/Mooncake.jl/issues/1020.
 @inline function value_and_gradient!!(
     cache::Cache,
     f::F,
@@ -677,8 +671,6 @@ end
 
 Returns a cache used with [`value_and_derivative!!`](@ref). See that function for more info.
 """
-# @inline forces specialisation on Vararg with function-valued arguments, avoiding severe
-# perf regressions. See https://github.com/chalk-lab/Mooncake.jl/issues/1020.
 @unstable @inline function prepare_derivative_cache(
     f, x::Vararg{Any,N}; config=Config()
 ) where {N}
--- a/test/interface.jl
+++ b/test/interface.jl
@@ -42,6 +42,10 @@ end
             ((x...) -> x[1] + x[2], randn(Float64), randn(Float64)),
             (sum, randn(10)),
             (x -> (x .*= 2; sum(x)), randn(10)),
+            # Regression test for https://github.com/chalk-lab/Mooncake.jl/issues/1020:
+            # passing a function-valued arg previously caused perf regressions due to
+            # missing specialisation; @inline on the interface functions fixes this.
+            ((xs, f) -> f(xs), randn(10), sum),
         ]
             kwargs = (debug_mode=false, silence_debug_messages=true)
             rule = build_rrule(fargs...; kwargs...)


yebai:
@penelopeysm, is there a way to have a unit/regression test for this function? It looks quite vulnerable to accidental breakage.

penelopeysm:
Honestly, I'm not sure how to test this. You could do some heuristic by checking allocations on value_and_gradient, I suppose -- but it feels more fragile than one might wish it to be.

yebai (approved):
Looks good — I added a comment to the changes as a reminder!

PR #1031, opened 2026-03-05
Reduce peak memory in blas_level_3 tests

### LINKED ISSUE #969: Blas level 3 CI issues on Julia v1.12

yebai:
Fixes #969

### 1. Split `blas_level_3` tests by element type (`src/rules/blas.jl`, `test/rules/blas_level_3.jl`)

The `rules/blas_level_3` CI job on Julia 1.12 x64 was being SIGTERM'd after ~38 min. All test-case tuples for a `Val` set are constructed before any test runs, so every tuple's primal objects are live simultaneously. Julia 1.12 added a 5th memory-backed variant in `blas_vectors`/`blas_matrices`, making the monolithic set ~25% larger.

**Fix:** split by `BlasFloat` element type. Each `Val(:blas_level_3_P)` set runs all level-3 BLAS variants for one type, so GC can reclaim one type's objects before the next type's are allocated. Peak live object count drops ~2.4×.

| Val | HW cases | Derived cases |
|-----|----------|---------------|
| `:blas_level_3_Float64` | 2 560 | 225 |
| `:blas_level_3_Float32` | 2 560 | 225 |
| `:blas_level_3_ComplexF64` | 10 020 | 225 |
| `:blas_level_3_ComplexF32` | 10 020 | 225 |

Test-case bodies are identical to `main`; only the dispatch organisation changes.

### 2. Clarify `GC.@preserve` comment (`src/test_utils.jl`)

The inline comment on `GC.@preserve memory` is moved above its line and expanded to explain why preservation is needed for tests involving pointer-backed types.

### DIFF
--- a/src/test_utils.jl
+++ b/src/test_utils.jl
@@ -1040,12 +1040,10 @@ function run_hand_written_rule_test_cases(rng_ctor, v::Val, mode::Type{<:Mode})
     test_cases, memory = test_hook(Mooncake.hand_written_rule_test_cases, rng_ctor, v) do
         Mooncake.hand_written_rule_test_cases(rng_ctor, v)
     end
-    GC.@preserve memory @testset "$f, $(_typeof(x))" for (  # keep pointer-backed objects alive
-        interface_only,
-        perf_flag,
-        _,
-        f,
-        x...,
+    # GC.@preserve keeps backing objects alive for tests involving pointer-backed
+    # types: without it, the GC may collect them mid-test.
+    GC.@preserve memory @testset "$f, $(_typeof(x))" for (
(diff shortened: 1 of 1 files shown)


yebai:
Hopefully this will now reduce the peak mem enough—can split further if needed.

sunxd3 (approved):
Looks good to me, this makes a lot of sense

Issue #1016, opened 2026-02-28
friendly_tangents=true fails with _copy_output TypeError when closure captures FunctionWrapper


ChrisRackauckas-Claude:
## Description

`prepare_pullback_cache` with `Config(friendly_tangents=true)` throws a `TypeError` when the differentiated closure captures a `FunctionWrapper`. The `_copy_output` function recurses into the `FunctionWrapper` internal fields and tries to `new` a `Core.TypeName`, which fails.

Without `friendly_tangents`, everything works fine.

## MWE

```julia
import Mooncake
import OrdinaryDiffEq as ODE
import SciMLBase

function lorenz!(du, u, p, t)
    du[1] = p[1] * (u[2] - u[1])
    du[2] = u[1] * (p[2] - u[3]) - u[2]
    du[3] = u[1] * u[2] - p[3] * u[3]
end

u = [1.0, 0.0, 0.0]
p = [10.0, 28.0, 8/3]

# ODE solve wraps f! in FunctionWrappersWrapper (via AutoSpecialize)
prob = ODE.ODEProblem(lorenz!, u, (0.0, 1.0), p)
sol = ODE.solve(prob, ODE.Tsit5())

# After solve, sol.prob.f.f is a FunctionWrappersWrapper containing FunctionWrapper
f_wrapped = sol.prob.f

# Closure capturing the FunctionWrappersWrapper-based ODEFunction
pf = let f = f_wrapped
    function (out, u, _p, t)
        f(out, u, _p, t)
        return out
    end
end

dy_mem = zero(u)
λ_mem = zero(u)

# Works without friendly_tangents:
cache = Mooncake.prepare_pullback_cache(pf, dy_mem, u, p, 0.0)
λ_mem .= [1.0, 1.0, 1.0]
dy, _ = Mooncake.value_and_pullback!!(cache, λ_mem, pf, dy_mem, u, p, 0.0)
@show cache.tangents[4]  # [-1.0, 1.0, 0.0]  ✓

# Fails with friendly_tangents=true:
config = Mooncake.Config(; friendly_tangents=true)
cache2 = Mooncake.prepare_pullback_cache(pf, dy_mem, u, p, 0.0; config=config)
# TypeError: in new, expected DataType, got Type{Symbol}
```

## Error

```
TypeError: in new, expected DataType, got Type{Symbol}

Stacktrace:
  [1] _copy_output_mutable_cartesian
    @ Mooncake/src/interface.jl:396
[...]

PR #1033, opened 2026-03-06
Fix `_copy_output` for opaque mutable types

yebai:
Fix https://github.com/chalk-lab/Mooncake.jl/issues/1016#issuecomment-4006300973 reported by @Parvm1102

```julia
using Mooncake
Mooncake._copy_output(Dict(:x => 1))
# TypeError: in new, expected DataType, got Type{Symbol}

struct DataStore
    _n::Int
    _data::Dict{Symbol, Any}
end
f(x, ds) = sum(x) * Float32(ds._n)
x = randn(Float32, 3, 3)
ds = DataStore(3, Dict{Symbol,Any}(:x => x))
Mooncake.prepare_gradient_cache(f, x, ds; config=Mooncake.Config(friendly_tangents=true))  # FAILS
```

## Changes

- `_copy_output`: fix `nfields` call; add `nf == 0` early-return; add methods for `Type`, `Core.TypeName`, `Module`.
- `_copy_to_output!!`: same fixes, plus `nf == 0 && return src` (not `dst`) so stale cached values are not returned.
 - `_copy_to_output!!`: fall back to `_copy_output(src[i])` when `dst[i]` is uninitialized. 
 - Tests: every new testset covers both `_copy_output` and `_copy_to_output!!`, including a struct-containing-Dict 
round-trip test.

## Extensibility

For types that need custom copy behaviour, overload `_copy_output` and `_copy_to_output!!` directly. `Core.SimpleVector` is an example already in the codebase:

```julia
_copy_output(x::SimpleVector) = Core.svec([map(_copy_output, x_sub) for x_sub in x]...)

function _copy_to_output!!(dst::SimpleVector, src::SimpleVector)
    return Core.svec(map(_copy_to_output!!, dst, src)...)
end
```

### DIFF
--- a/test/interface.jl
+++ b/test/interface.jl
@@ -194,9 +194,7 @@ end
                 SimplePair(2x.x1 * x̄.x1 + x.x2 * x̄.x2, cos(x.x2) * x̄.x1 + x.x1 * x̄.x2),
             )
 
-            # Regression test: _copy_output must use nfields(src), not nfields(P);
-            # nfields(P) returns the wrong field count, causing jl_new_structv to
-            # raise "invalid struct allocation".
+            # Regression test for "invalid struct allocation" and `TypeError` error. See #1024.
             struct ImmutableWithNothingFields
                 a::Float64
                 b::Float64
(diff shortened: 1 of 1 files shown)


src/interface.jl
```diff
@@ -315,11 +316,14 @@ end
...
-    # nfields(P) counts fields of the DataType object itself; use nfields(src) or
-    # fieldcount(P) to get the number of fields declared by struct P.
+    # nfields(src) not nfields(P): the latter counts fields of the
+    # DataType object itself.
     nf = nfields(src)
 
     if ismutable(src)
+        # No Julia-visible fields (e.g. Symbol, String): nothing to update.
+        # Overload _copy_to_output!! to customise.
+        nf == 0 && return src
```
Technici4n: Any reason to limit this to mutable types?

yebai: Not really - Copilot thought this was enough to fix the bug.

yebai:
@AstitvaAggarwal, please feel free to merge and release after you review this.

AstitvaAggarwal:
GTM, the PR looked fine. If anything comes up I'll take a go at it. (Sorry for the late reviews, I had this on my todo)

PR #1040, opened 2026-03-07
Implement `@reverse_from_forward` to define a reverse rule when the forward rule exists

### LINKED ISSUE #548: `ForwardOptimise` for low dimentional scalar and high-order functions
ReverseDiff `ForwardOptimize` explains the idea very well: 

```julia
"""
    ReverseDiff.@forward(f)(args::Real...)
    ReverseDiff.@forward f(args::Real...) = ...
    ReverseDiff.@forward f = (args::Real...) -> ...

Declare that the given function should be differentiated using forward mode automatic
differentiation. Note that the macro can be used at either the definition site or at the
call site of `f`. Currently, only `length(args) <= 2` is supported. **Note that, if `f` is
defined within another function `g`, `f` should not close over any differentiable input of
`g`.** By using this macro, you are providing a guarantee that this property holds true.

This macro can be very beneficial for performance when intermediate functions in your
computation are low dimensional scalar functions, because it minimizes the number of
instructions that must be recorded to the tape. For example, take the function
`sigmoid(n) = 1. / (1. + exp(-n))`. Normally, using ReverseDiff to differentiate this
function would require recording 4 instructions (`-`, `exp`, `+`, and `/`). However, if we
apply the `@forward` macro, only one instruction will be recorded (`sigmoid`). The `sigmoid`
function will then be differentiated using ForwardDiff's `Dual` number type.

This is also beneficial for higher-order elementwise function application. ReverseDiff
overloads `map`/`broadcast` to dispatch on `@forward`-applied functions. For example,
`map(@forward(f), x)` will usually be more performant than `map(f, x)`.

ReverseDiff overloads many Base scalar functions to behave as `@forward` functions by
default. A full list is given by `DiffRules.diffrules()`.
"""
macro forward(ex)
    return esc(annotate_func_expr(:ForwardOptimize, __module__, ex))
end

```

Notes: 

- This feature builds on https://github.com/compintell/Mooncake.jl/issues/337
- It might also help reduce compilation time for reverse-mode autograd

gdalle:
Fixes #548:

- [x] Add a new macro `@reverse_from_forward` which defines a primitive `rrule!!` from an existing `frule!!`
- [x] Implement it for functions `y = f(x...)` where inputs are `Union{IEEEFloat, Array{<:IEEEFloat}}` and outputs are amenable to `LinearAlgebra.dot`
- [x] Generalize it to more types, for instance by using `Mooncake._dot` (can be done in a later PR)
- [x] Give friendly error messages when the macro is misused
- [x] Add thorough testing using `Mooncake.TestUtils`
  - [x] For single-argument functions
  - [x] For multi-argument functions
- [ ] Test type-stability of the generated code => no need, done by DispatchDoctor

### DIFF
--- a/src/reverse_from_forward.jl
+++ b/src/reverse_from_forward.jl
@@ -64,22 +64,23 @@ function (forward_mode_rrule!!::ForwardModeRRule!!)(
             else
                 @assert args[i] isa Array{<:IEEEFloat} # TODO: relax
                 # iterate over input dimensions
-                tangent(args_codual[i]) .+= map(eachindex(args[i])) do j
-                    # create perturbation of scalar dimension j of argument i
-                    args_dual_one_ij = (
-                        args_dual_zero[begin:(i - 1)]...,
-                        Dual(args[i], basis(args[i], j)),
-                        args_dual_zero[(i + 1):end]...,
-                    )
-                    # compute partial derivative with respect to dimension j of argument i
-                    y_dual_one_ij = _frule!!(f_dual, args_dual_one_ij...)
-                    partial_derivative_ij = tangent(y_dual_one_ij)
-                    # deduce one component of the pullback using a dot product
-                    rdata_ij =
-                        dot(fdata(partial_derivative_ij), tangent(y_codual)) +
-                        dot(rdata(partial_derivative_ij), dy_rdata)
-                    return rdata_ij
-                end
+                tangent(args_codual[i]) .+=
+                    map(eachindex(args[i])) do j
+                        # create perturbation of scalar dimension j of argument i
+                        args_dual_one_ij = (
+                            args_dual_zero[begin:(i - 1)]...,
+                            Dual(args[i], basis(args[i], j)),
+                            args_dual_zero[(i + 1):end]...,
+                        )
+                        # compute partial derivative with respect to dimension j of argument i
+                        y_dual_one_ij = _frule!!(f_dual, args_dual_one_ij...)
+                        partial_derivative_ij = tangent(y_dual_one_ij)
+                        # deduce one component of the pullback using a dot product
+                        rdata_ij =
+                            dot(fdata(partial_derivative_ij), tangent(y_codual)) +
+                            dot(rdata(partial_derivative_ij), dy_rdata)
+                        return rdata_ij
+                    end
                 return nothing
             end
         end


AstitvaAggarwal:
Since this was on my Todo as well, we can maybe get this PR merged (after reviews) and in separate PRs I expand incrementally.

gdalle:
I'd be happy to expand too, just wanted to get a first round of feedback on the general direction

AstitvaAggarwal:
Yeah all good, ive got some pending work on the Lux side of things as well haha. Thanks!

gdalle:
I wonder what should happen if the `rrule!!` method already exists for the function at hand

yebai:
> I wonder what should happen if the rrule!! method already exists for the function at hand

These rules are standard Julia funcitons so overloading them will override existing rules, which is fine here.

gdalle:
Would love reviews from @AstitvaAggarwal and @Technici4n (can't request them via GitHub because I'm not a repo member)

gdalle:
Strange that the new tests pass on `x64` but fail on `lts-x86`

src/wrong_mode.jl
```diff
@@ -0,0 +1,160 @@
...
+                "`Mooncake.@reverse_from_forward` does not support output type `$(typeof(y))` with input type `$(typeof(x))`.",
+            ),
+        )
+    end
+end
+
+function basis(a::Array{<:IEEEFloat}, i)
+    # TODO: fix for immutable arrays
+    # TODO: fix for GPU arrays
+    b = similar(a)
```
AstitvaAggarwal: To avoid garbage values (this causes CI on lts x86 to fail)
```suggestion
    b = zeros(eltype(a), size(a))
```

src/wrong_mode.jl
```diff
@@ -0,0 +1,160 @@
...
+                "`Mooncake.@reverse_from_forward` does not support output type `$(typeof(y))` with input type `$(typeof(x))`.",
+            ),
+        )
+    end
+end
+
+function basis(a::Array{<:IEEEFloat}, i)
+    # TODO: fix for immutable arrays
+    # TODO: fix for GPU arrays
+    b = zeros(eltype(a), size(a))
```
gdalle:
```suggestion
    b = zero(a)
```

yebai:
Thanks, Guillaume. A few high-level comments:

1. `build_primitive_rrule` could be overloaded to redirect to any rule method (see eg, https://github.com/chalk-lab/Mooncake.jl/issues/548#issuecomment-4005642117). This would avoid the need to override existing `rrule!!` implementations and offer a straightforward improvement toward a more robust design.  
2. Supporting multiple arguments is necessary. This can be addressed in a separate PR if needed.  
3. We should review ReverseDiff’s list of functions that use `ForwardOptimise` and include them here as well.

gdalle:
1. Done.
2. Done.
3. I think this is best done at a later stage, because:
    - It would require an additional dependency on [DiffRules.jl](https://github.com/JuliaDiff/DiffRules.jl)
    - I'm not sure whether the performance benefits for Mooncake will be as tangible as those observed with ReverseDiff. Thorough benchmarking seems necessary before deciding on the concrete list of functions, especially because many of them probably already have a reverse rule.

src/tangents/fwds_rvs_data.jl
```diff
@@ -10,6 +10,8 @@ Base.copy(::NoFData) = NoFData()
 
 increment_internal!!(::IncCache, ::NoFData, ::NoFData) = NoFData()
 
+LinearAlgebra.dot(::NoFData, ::NoFData) = false
```
Technici4n, line 13: we could consider using `Mooncake._dot` if it works here

src/Mooncake.jl
```diff
@@ -165,6 +166,7 @@ include(joinpath("rules", "high_order_derivative_patches.jl"))
 
 include("config.jl")
 include("developer_tools.jl")
+include("wrong_mode.jl")
```
Technici4n: I find this name a bit confusing 😄

src/reverse_from_forward.jl
```diff
@@ -0,0 +1,136 @@
...
+
+function (forward_mode_rrule!!::ForwardModeRRule!!)(
+    f_codual::CoDual{F},
+    args_codual::Vararg{<:CoDual{<:Union{IEEEFloat,<:Array{<:IEEEFloat}}},N},  # TODO: relax
+) where {F,N}
+    (; _frule!!) = forward_mode_rrule!!
+    f = primal(f_codual)
+    args = map(primal, args_codual)
+
+    if tangent_type(F) != NoTangent
```
Technici4n, line 18: In principle this could be checked at build time, but leaving it here should also be fine.

src/tangents/fwds_rvs_data.jl
gdalle: The reason I didn't is because it seems to systematically allocate a dictionary at the time, which is a bit silly to multiply two scalars (the most common use case)

src/Mooncake.jl
gdalle: I put it here in case there is a symmetric macro to simulate forward mode through reverse mode. This is essentially using the wrong autodiff mode to differentiate something (but I have no issue renaming it)

src/reverse_from_forward.jl
gdalle: I didn't figure out how to do it outside of a generated function, but it shouldn't be too bad since it is clearly inferrable

src/tangents/fwds_rvs_data.jl
Technici4n: We could easily add `isbitstype(P) ? NoCache() : IdDict()` to it

src/Mooncake.jl
gdalle: Renamed

src/tangents/fwds_rvs_data.jl
gdalle: I think we can leave it for subsequent improvements

gdalle:
This should be ready

yebai:
Three optimisation commits on top of the original implementation:
a hand-written `frule!!` for `f(x::Float64)` and `f(x::Vector{Float64})`, differentiated
via `value_and_gradient!!` after a warm-up compile.

```julia
# scalar
f_scalar(x::Float64) = sin(x) + x^2 * cos(x)
@reverse_from_forward Tuple{typeof(f_scalar), Float64}

# vector
f_vec(x::Vector{Float64}) = sum(xi -> sin(xi) + xi^2*cos(xi), x)
@reverse_from_forward Tuple{typeof(f_vec), Vector{Float64}}
```

**Scalar input** (1 forward pass in pullback)

| | `843b3e2` basis-buf | `b91b541` ntuple | `a354109` Ref{T} | `d5e3bd5` latest |
|---|---|---|---|---|
| derived reverse | 28.6 ns / 0 alloc | 28.1 ns / 0 alloc | 27.8 ns / 0 alloc | 27.9 ns / 0 alloc |
| `@from_forward` | 16.4 ns / 0 alloc | 17.4 ns / 0 alloc | 15.7 ns / 0 alloc | 16.3 ns / 0 alloc |

**Vector input** (n forward passes in pullback — expected O(n) scaling)

| n | derived reverse | `843b3e2` basis-buf | `b91b541` ntuple | `a354109` Ref{T} | `d5e3bd5` latest |
|---|---|---|---|---|---|
| 1 | 43 ns / 0 alloc | 209 ns / 12 alloc | 94 ns / 6 alloc | 41 ns / 0 alloc | 50 ns / 0 alloc |
| 4 | 90 ns / 0 alloc | 889 ns / 30 alloc | 459 ns / 6 alloc | 359 ns / 0 alloc | 354 ns / 0 alloc |
| 16 | 305 ns / 0 alloc | 6.6 μs / 102 alloc | 4.8 μs / 6 alloc | 4.5 μs / 0 alloc | 4.4 μs / 0 alloc |
| 64 | 985 ns / 0 alloc | 83 μs / 390 alloc | 70 μs / 6 alloc | 68 μs / 0 alloc | 70 μs / 0 alloc |

Each commit eliminated one source of allocations:
- **`843b3e2`**: replaced per-element `basis(args[i], j)` (fresh array each call) with a
  single pre-allocated buffer — cuts O(n) → O(1) allocations in the basis loop, but 6
  fixed allocations remain (one `zero(args[i])` per call).
- **`b91b541`**: replaced tuple slicing `args_dual_zero[begin:(i-1)]...` with
  `ntuple(Val(N))` — removes 2 of the 6 fixed allocations; 6 remain from buffer setup.
- **`a354109`**: moved buffer allocation into `build_primitive_rrule` via a `@generated`
  function that produces a typed `NTuple` of `Ref{T}` slots; buffers are lazily sized on
  the first call and reused thereafter — **0 allocations per call**.
- **`d5e3bd5`**: correctness fixes, robustness improvements, docs, and Julia 1.10 compat —
  allocation count unchanged (0), runtime unchanged.

The O(n) runtime scaling for vector inputs is fundamental (n forward passes per pullback),
not an artefact of allocation.

gdalle:
I'll review these LLM optimizations tomorrow

yebai:
I'm happy to move them to a separate PR if needed -- GitHub doesn't support creating PRs against this branch since it is from a fork. In addition, I'd suggest keeping these tools together with `tools_for_rules.jl`. 

The crash is likely unrelated to this PR but a Julia opaque closure bug.

gdalle:
Sorry it was my day off, I set aside time tomorrow morning for a final pass on Claude's changes

src/tangents/fwds_rvs_data.jl
```diff
@@ -415,7 +413,10 @@ Base.copy(::NoRData) = NoRData()
 
 @inline increment_field!!(::NoRData, y, ::Val) = NoRData()
 
-LinearAlgebra.dot(::NoRData, ::NoRData) = false
+# Required so that _dot dispatches correctly: No{F,R}Data carry no information so their
+# contribution to any inner product is zero.
+_dot_internal(::MaybeCache, ::NoFData, ::NoFData) = 0.0
+_dot_internal(::MaybeCache, ::NoRData, ::NoRData) = 0.0
```
gdalle: I used `false` on purpose because it is the "smallest" zero available in Julia, and will automatically be promoted to whatever floating-point type makes sense. There is no reason to force `_dot_internal` to return a `Float64`.

src/tangents/tangents.jl
```diff
@@ -1007,7 +1007,9 @@ Should be defined for all standard tangent types.
 Inner product between tangents `t` and `s`. Must return a `Float64`.
```
gdalle, line 1007: It would be wise to remove the constraint of returning a `Float64` (see above).

src/tangents/tangents.jl
```diff
@@ -1007,7 +1007,9 @@ Should be defined for all standard tangent types.
...
 Always available because all tangent types correspond to finite-dimensional vector spaces.
 """
-_dot(t::T, s::T) where {T} = _dot_internal(IdDict{Any,Any}(), t, s)::Float64
+function _dot(t::T, s::T) where {T}
+    _dot_internal(isbitstype(T) ? NoCache() : IdDict{Any,Any}(), t, s)::Float64
```
gdalle: Can we also avoid caching for `Array`s and similar types? I seem to remember that's already what we do with the gradient operator.

src/tools_for_rules.jl
```diff
@@ -887,3 +887,198 @@ Equivalent to `@from_chainrules ctx sig has_kwargs ReverseMode`. See
...
     return _from_chainrules_impl(ctx, sig, has_kwargs, ReverseMode)
 end
+
+#
+# Functionality supporting @from_forward
+#
+
+# Returns a typed tuple with one entry per argument (excluding the function).
+# `nothing` for IEEEFloat args; `Ref{Array{T,D}}` holding an empty array for array args.
+# The empty array is sized on first use and resized when the input shape changes.
```
gdalle: Does this kind of adaptive pre-allocation happen for any other rule in Mooncake?

src/tools_for_rules.jl
```diff
@@ -887,3 +887,198 @@ Equivalent to `@from_chainrules ctx sig has_kwargs ReverseMode`. See
...
+                ),
+            )
+        end
+    end
+    return :(($(exprs...),))
+end
+
+# `_zero_bufs[i]`: zero-tangent buffers for the primal forward pass (one per arg).
+# `_basis_bufs[i]`: unit-perturbation buffers for the pullback (one per arg).
+# Both are `nothing` for scalar args and `Ref{Array}` for array args.
```
gdalle: We could also use `zero` and `one` for scalar args

src/tools_for_rules.jl
```diff
@@ -887,3 +887,198 @@ Equivalent to `@from_chainrules ctx sig has_kwargs ReverseMode`. See
...
+struct ForwardModeRRule!!{FR,ZB<:Tuple,BB<:Tuple}
```
gdalle: Is there a realistic case where `ZB != BB`?

src/tools_for_rules.jl
```diff
@@ -887,3 +887,198 @@ Equivalent to `@from_chainrules ctx sig has_kwargs ReverseMode`. See
...
+    _frule!!::FR
+    _zero_bufs::ZB
+    _basis_bufs::BB
+end
+
+function __verify_sig(rule::ForwardModeRRule!!, fx)
+    # TODO: modify `fx`
+    return __verify_sig(rule._frule!!, fx)
```
gdalle: If we leave it like this, `__verify_sig` will try to verify the `_frule!!` on an input made of `CoDual`, which will error whenever the verification is non-trivial.

src/tools_for_rules.jl
```diff
@@ -887,3 +887,198 @@ Equivalent to `@from_chainrules ctx sig has_kwargs ReverseMode`. See
...
+                args_dual_one_i = ntuple(Val(N)) do k
+                    k == i ? Dual(args[i], one(args[i])) : args_dual_zero[k]
+                end
+                y_dual_one_i = _frule!!(f_dual, args_dual_one_i...)
+                return _dot(tangent(y_dual_one_i), dy_full)
+            end
+        end
+
+        # Array args: fdata via one forward pass per element (column-by-column VJP).
+        for i in 1:N
```
gdalle: This is not a good idea, I used `ntuple` on purpose so that the loop would be unrolled and each iteration specialized according to the type of `args[i]`. If we use `for`, we bring type instability into the loop. This "fix" makes me skeptical about Claude's other "fixes" for type stability, and I suggest we add some tests with `@inferred` (at the very least), or ideally `JET.@test_opt`.

src/tools_for_rules.jl
```diff
@@ -887,3 +887,198 @@ Equivalent to `@from_chainrules ctx sig has_kwargs ReverseMode`. See
...
+            tangent(args_codual[i]) isa NoFData && continue
+            @assert args[i] isa Array{<:IEEEFloat} # TODO: relax
+            b_ref = fmr._basis_bufs[i]
+            if size(b_ref[]) != size(args[i])
+                b_ref[] = zero(args[i])
+            end
+            t = tangent(args_codual[i])
+            if length(t) != length(args[i])
+                resize!(t, length(args[i]))
```
gdalle: This won't work for general `Array`s, only for `Vector`s:
```julia
julia> methods(resize!)
# 5 methods for generic function "resize!" from Base:
 [1] resize!(compact::Compiler.IncrementalCompact, nnewnodes::Int64)
     @ ../usr/share/julia/Compiler/src/ssair/ir.jl:1739
 [2] resize!(stmts::Compiler.InstructionStream, len)
     @ ../usr/share/julia/Compiler/src/ssair/ir.jl:291
 [3] resize!(D::Compiler.DFSTree, n::Integer)
     @ ../usr/share/julia/Compiler/src/ssair/domtree.jl:113
 [4] resize!(B::BitVector, n::Integer)
     @ bitarray.jl:809
 [5] resize!(a::Vector, nl_::Integer)
     @ array.jl:1478
```
But in general, I think this line won't be hit? Why would the provided tangent have a different length than the primal?

src/tools_for_rules.jl
```diff
@@ -887,3 +887,198 @@ Equivalent to `@from_chainrules ctx sig has_kwargs ReverseMode`. See
...
+# 1. Mark the function as a forward-mode primitive.
+@is_primitive DefaultCtx ForwardMode Tuple{typeof(mysin),Float64}
+
+# 2. Implement the forward rule.
+function Mooncake.frule!!(::Dual{typeof(mysin)}, x::Dual{Float64})
+    v = primal(x)
+    return Dual(sin(v), cos(v) * tangent(x))
+end
+
+# 3. Derive the reverse rule automatically from the forward rule.
```
gdalle: We should warn that step 1 and 2 are optional (readers of this docstring may not know what a "derived" `frule!!` is in Mooncake terminology)

src/tools_for_rules.jl
```diff
@@ -887,3 +887,198 @@ Equivalent to `@from_chainrules ctx sig has_kwargs ReverseMode`. See
...
+    functions that use exception handling for domain checks (e.g. throwing on invalid
+    input), where direct reverse-mode AD would fail.
+
+!!! warning
+    This macro is still experimental and has strict limitations:
+        - The function must have all its arguments and its output of type `<:Base.IEEEFloat` or `Array{<:Base.IEEEFloat}` when called (the signature passed to `@from_forward` may use abstract supertypes, but concrete call sites must satisfy this constraint)
+        - The function must not close over any data (its own tangent type must be `NoTangent`)
+        - The function must not mutate any of its arguments, because the pullback runs several forward passes and relies on the arguments remaining unchanged between them
+        - The `frule!!` must not mutate its tangent inputs: the pullback reuses zero-tangent buffers across forward passes and relies on them remaining zero
+        - The rule is **not thread-safe**: it holds shared mutable buffers that are reused across calls. Do not call the same cached rule concurrently from multiple threads.
```
gdalle: That would be rather easy to fix if we want to. We could just use `zero(b_ref[])` every time we need it.
Also, even if we extend allowed types, the rule will fail on GPU arrays and other arrays where we cannot access individual elements.

test/tools_for_rules.jl
```diff
@@ -160,6 +161,132 @@ end
...
+using Mooncake:
+    Dual,
+    DefaultCtx,
+    ForwardMode,
+    ReverseMode,
+    @is_primitive,
+    @from_forward,
+    primal,
+    tangent
+
```
gdalle: I think the comment you deleted here was useful to understand what goes on:
```
# we define rules for fi that pretend fi_true was run instead, to check that Mooncake doesn't recurse inside the function
# we still use fi_true in the Mooncake.TestUtils.test_rule
```

test/tools_for_rules.jl
```diff
@@ -389,4 +516,125 @@ end
...
+            @testset "Array input, scalar output: zero allocations" begin
+                test_rule(
+                    StableRNG(63),
+                    R.fperf_v,
+                    [3.0, 5.0];
+                    is_primitive=false,
+                    perf_flag=:allocs,
+                )
+            end
+            @testset "Buffer resize: correctness with variable-size arrays" begin
```
gdalle: This should be tested with matrices too

src/tools_for_rules.jl
```diff
@@ -929,3 +929,198 @@ Equivalent to `@from_chainrules ctx sig has_kwargs ReverseMode`. See
...
+        dargs_rdata = ntuple(Val(N)) do i
+            if rdata_type(typeof(args[i])) == NoRData
+                return NoRData()
+            else
+                @assert args[i] isa IEEEFloat # TODO: relax
+                args_dual_one_i = ntuple(Val(N)) do k
+                    k == i ? Dual(args[i], one(args[i])) : args_dual_zero[k]
+                end
+                y_dual_one_i = _frule!!(f_dual, args_dual_one_i...)
+                return _dot(tangent(y_dual_one_i), dy_full)
```
sunxd3: `_dot`'s type annotation enforces returning `Float64`, a type cast is safer.

```suggestion
                return typeof(args[i])(_dot(tangent(y_dual_one_i), dy_full))
```

gdalle: This forced conversion should be removed, see my earlier comment

yebai:
Addressed all review comments:

- Merged `ZB`/`BB` type params into single `B` in `ForwardModeRRule!!`
- `__verify_sig` now returns `nothing` (frule sig validated at build time)
- Replaced `for i in 1:N` with `ntuple(Val(N))` for type-stable array pullback
- Fixed `size(t)` check instead of `length(t)` for ND array tangent resize
- Added `typeof(args[i])(...)` cast for correct Float16/Float32 scalar rdata
- Simplified docstring example (prerequisite frule can be derived automatically)
- Added matrix-to-matrix test (`f4m`) with correct forward rule and numerical validation

src/tools_for_rules.jl
```diff
@@ -960,15 +960,16 @@ end
...
-    _basis_bufs::BB
+    _zero_bufs::B
+    _basis_bufs::B
 end
 
-function __verify_sig(rule::ForwardModeRRule!!, fx)
-    # TODO: modify `fx`
-    return __verify_sig(rule._frule!!, fx)
+function __verify_sig(::ForwardModeRRule!!, fx)
+    # The _frule!! signature is already validated during build_frule. Passing CoDual-derived
```
gdalle: Then why does this verification mechanism exist at all for reverse mode? And what happens if the user writes the forward rule as a method of `frule!!` (i.e. makes the function a forward-mode primitive instead of deriving the rule)?

src/tools_for_rules.jl
```diff
@@ -1025,23 +1026,20 @@ function (fmr::ForwardModeRRule!!)(
                     k == i ? Dual(args[i], one(args[i])) : args_dual_zero[k]
                 end
                 y_dual_one_i = _frule!!(f_dual, args_dual_one_i...)
-                return _dot(tangent(y_dual_one_i), dy_full)
+                return typeof(args[i])(_dot(tangent(y_dual_one_i), dy_full))
```
gdalle: As stated above, it makes no sense to force `_dot` to return a `Float64`. This fixes the symptom but not the root cause

src/tools_for_rules.jl
```diff
@@ -1075,16 +1074,6 @@ calling this macro.
...
-# 1. Mark the function as a forward-mode primitive.
-@is_primitive DefaultCtx ForwardMode Tuple{typeof(mysin),Float64}
-
-# 2. Implement the forward rule.
-function Mooncake.frule!!(::Dual{typeof(mysin)}, x::Dual{Float64})
-    v = primal(x)
-    return Dual(sin(v), cos(v) * tangent(x))
-end
-
-# 3. Derive the reverse rule automatically from the forward rule.
```
gdalle: I don't think these are useless lines of examples, some users will also need to define `frule!!` themselves

test/tools_for_rules.jl
```diff
@@ -283,6 +295,8 @@ end
 @from_forward Tuple{typeof(f4),Vector{Float64}}
 @from_forward Tuple{typeof(f4_true),Vector{Float64}}
 
+@from_forward Tuple{typeof(f4m),Matrix{Float64}}
```
gdalle: This is the function with which resizing should be tested

yebai:
It's probably best for @gdalle to complete this PR on his own, then the Mooncake team can review it. 

The extra changes I created can be found in https://github.com/chalk-lab/Mooncake.jl/tree/fix/from_forward if that is useful.

gdalle:
Hi, can you explain why this PR was closed after you explicitly and repeatedly asked me to work on it?

PR #1043, opened 2026-03-08
NNlib activation function rules

AstitvaAggarwal:
Current NNlib rrules have incomplete handling for `logsumexp`, `softmax` and `logsoftmax`.

The following PR adds :

- Updated Native Mooncake rrules (current `@from_rrule` fails for Adjoint/Transpose because `increment_and_get_rdata!` overload is not present, is unsafe to implement.) 
- tests for rrules and helper methods
- Unblocks Lux.jl Integration. see https://github.com/LuxDL/Lux.jl/actions/runs/22794547300/job/66139118086?pr=1664, https://github.com/LuxDL/Lux.jl/actions/runs/22794547300/job/66139118021?pr=1664.
- future proofs current rules for common activation functions

### DIFF
--- a/src/tangents/codual.jl
+++ b/src/tangents/codual.jl
@@ -43,6 +43,10 @@ function _codual_internal(::Type{P}, f::F, extractor::E) where {P,F,E}
     P == Union{} && return Union{}
     P == DataType && return CoDual
     P isa Union && return Union{f(P.a),f(P.b)}
+    # Use `isa` not `<:`: generators like `NTuple{N,Int} where N` are instances of
+    # UnionAll but not subtypes of it (`NTuple{N,Int} where N <: UnionAll` is false).
+    # `P == UnionAll` handles the UnionAll metatype itself (`UnionAll isa UnionAll` is false).
+    (P isa UnionAll || P == UnionAll) && return CoDual # P is abstract, tangent type unknown.
 
     if P <: Tuple && !all(isconcretetype, (P.parameters...,))
         field_types = (P.parameters...,)
@@ -54,7 +58,6 @@ function _codual_internal(::Type{P}, f::F, extractor::E) where {P,F,E}
         end
     end
 
-    P <: UnionAll && return CoDual # P is abstract, so we don't know its tangent type.
     return isconcretetype(P) ? CoDual{P,extractor(P)} : CoDual
 end
 
--- a/src/tangents/dual.jl
+++ b/src/tangents/dual.jl
@@ -29,9 +29,27 @@ zero_dual(x) = Dual(x, zero_tangent(x))
 randn_dual(rng::AbstractRNG, x) = Dual(x, randn_tangent(rng, x))
 
 function dual_type(::Type{P}) where {P}
+    P == Union{} && return Union{}
     P == DataType && return Dual
     P isa Union && return Union{dual_type(P.a),dual_type(P.b)}
-    P <: UnionAll && return Dual # P is abstract, so we don't know its tangent type.
+    # Use `isa` not `<:`: generators like `NTuple{N,Int} where N` are instances of
+    # UnionAll but not subtypes of it (`NTuple{N,Int} where N <: UnionAll` is false).
+    # `P == UnionAll` handles the UnionAll metatype itself (`UnionAll isa UnionAll` is false).
+    (P isa UnionAll || P == UnionAll) && return Dual # P is abstract, tangent type unknown.
+
+    # Union Splitting
+    if P <: Tuple && !all(isconcretetype, (P.parameters...,))
+        field_types = (P.parameters...,)
+        union_fields = _findall(Base.Fix2(isa, Union), field_types)
+
+        # If there is exactly one Union field, split it to help the compiler
+        if length(union_fields) == 1 &&
+            all(p -> p isa Union || isconcretetype(p), field_types)
+            P_split = split_union_tuple_type(field_types)
+            return Union{dual_type(P_split.a),dual_type(P_split.b)}
+        end
+    end
+
     return isconcretetype(P) ? Dual{P,tangent_type(P)} : Dual
 end
 
--- a/test/front_matter.jl
+++ b/test/front_matter.jl
@@ -46,7 +46,8 @@ using Mooncake:
     Mode,
     ForwardMode,
     ReverseMode,
-    MistyClosureTangent
+    MistyClosureTangent,
+    dual_type
 
 using Mooncake:
     CC,
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -7,6 +7,7 @@ include("front_matter.jl")
         include(joinpath("tangents", "tangents.jl"))
         include(joinpath("tangents", "fwds_rvs_data.jl"))
         include(joinpath("tangents", "codual.jl"))
+        include(joinpath("tangents", "dual.jl"))
         include("debug_mode.jl")
         include("stack.jl")
         @testset "interpreter" begin
--- a/test/tangents/codual.jl
+++ b/test/tangents/codual.jl
@@ -20,6 +20,8 @@
             Union{CoDual{Float64,NoFData},CoDual{Int,NoFData}},
         ),
         (UnionAll, CoDual, CoDual),
+        (DataType, CoDual, CoDual),
+        (Union{}, Union{}, Union{}),
 
         # Tuples:
         # Concrete tuples:
@@ -60,6 +62,11 @@
 
         # General Abstract Tuples
         (Tuple{Any}, CoDual, CoDual),
+
+        # Abstract Vararg / NTuple UnionAll tuples (bounded and unbounded)
+        (NTuple{N,Int} where {N}, CoDual, CoDual),
+        (Tuple{Vararg{Float64,N}} where {N}, CoDual, CoDual),
+        (Tuple{Vararg{Float64}}, CoDual, CoDual),
     ]
         @test TestUtils.check_allocs(codual_type, P) == D
         @test TestUtils.check_allocs(Mooncake.fcodual_type, P) == F
--- a/test/tangents/dual.jl
+++ b/test/tangents/dual.jl
@@ -1,0 +1,49 @@
+@testset "Dual" begin
+    @test Dual(5.0, 4.0) isa Dual{Float64,Float64}
+    @test Dual(Float64, NoTangent()) isa Dual{Type{Float64},NoTangent}
+    @test zero_dual(5.0) == Dual(5.0, 0.0)
+
+    @testset "$P" for (P, D) in Any[
+        (Float64, Dual{Float64,Float64}),
+        (Int, Dual{Int,NoTangent}),
+        (Real, Dual),
+        (Any, Dual),
+        (Type{UnitRange{Int}}, Dual{Type{UnitRange{Int}},NoTangent}),
+        (Type{Tuple{T}} where {T}, Dual),
+        (Union{Float64,Int}, Union{Dual{Float64,Float64},Dual{Int,NoTangent}}),
+        (UnionAll, Dual),
+        (DataType, Dual),
+        (Union{}, Union{}),
+
+        # Tuples:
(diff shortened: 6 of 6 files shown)


AstitvaAggarwal:
Fyi it's best to include this PR in the latest release for mooncake.

AstitvaAggarwal:
@yebai  let's get this PR merged before https://github.com/chalk-lab/Mooncake.jl/pull/1044, it's easier to handle the merge conflicts for me in that PR

yebai:
A few comments from Copilot. 

1. High – `logsumexp` rule can produce incorrect primal/gradients for `Inf` inputs
    - In `ext/MooncakeNNlibExt.jl:139` and `ext/MooncakeNNlibExt.jl:160`, the new rule computes `tmp = exp.(xp .- max_)` and `y = max_ + log(s)` directly.
    - For cases like `xp = [Inf, Inf]`, `xp .- max_` becomes `NaN`, so `tmp`, `s`, and `y` become `NaN`; mathematically `logsumexp([Inf, Inf])` should be `Inf`.
  
2. High – `kw logsumexp` pullback shape/seed contract looks wrong for scalar-output kw cases
    - In `ext/MooncakeNNlibExt.jl:151`, `kw logsumexp` uses a single pullback signature `logsumexp_kw_pb!!(::NoRData)` at `ext/MooncakeNNlibExt.jl:164`.
    - If kw usage yields scalar output (for example explicit full reduction), reverse mode typically needs scalar seed `dy`, not `NoRData`.
    - Existing `LogExpFunctions` extension splits colon/non-colon kw handling specifically for this reason; this `NNlib` rule does not.

3. Medium – tests do not cover the two highest-risk edge cases above
    - `NNlib` extension tests added in `test/ext/nnlib/nnlib.jl` cover `Adjoint/Transpose` and several `dims`, but not `Inf/NaN` stability or scalar-output kw reduction for `logsumexp`.
    - That leaves regressions from items 1–2 undetected.

4. High – the new `NNlib` primitives claim all `AbstractArrays`, but the reverse accumulator only supports plain arrays plus `Adjoint/Transpose`
    - The primitive declarations are broad at `ext/MooncakeNNlibExt.jl:58`, `ext/MooncakeNNlibExt.jl:94`, and `ext/
      MooncakeNNlibExt.jl:130`.
    - But `_accum_fdata!` only has methods for `AbstractArray` and `FData` with `Adjoint/Transpose` at `ext/
      MooncakeNNlibExt.jl:44`.
    - That means wrappers like `view`, `reshape`, `reinterpret`, `Diagonal`, or triangular wrappers are likely to hit `MethodError` in reverse mode once these new primitives are selected, even though `Mooncake` already has wrapper-handling machinery elsewhere such as `src/rules/blas.jl:41`.
    - The current tests only add `Adjoint/Transpose`; they do not cover those other wrappers.
    - Perhaps use `arrayify` on input arguments inside the rule?

ext/MooncakeNNlibExt.jl
```diff
@@ -19,9 +39,135 @@ const SupportedArray{P,N} = Union{Array{P,N},AbstractGPUArray{P,N}}
...
+_accum_fdata!(xf::AbstractArray, ::AbstractArray{T}, g::T) where {T<:IEEEFloat} = xf .+= g
+function _accum_fdata!(xf::FData, ::Adjoint{T}, g::AbstractArray{T}) where {T<:IEEEFloat}
+    return xf.data.parent .+= g'
+end
+function _accum_fdata!(xf::FData, ::Transpose{T}, g::AbstractArray{T}) where {T<:IEEEFloat}
+    return xf.data.parent .+= transpose(g)
+end
+
+# logsoftmax rrules
+@is_primitive MinimalCtx Tuple{typeof(logsoftmax),AbstractArray{<:IEEEFloat}}
```
sunxd3: the old code is defined on `SupportedArray`, is it safe to use `AbstractArray`?

ext/MooncakeNNlibExt.jl
```diff
@@ -19,9 +39,135 @@ const SupportedArray{P,N} = Union{Array{P,N},AbstractGPUArray{P,N}}
...
+    kw::CoDual{<:NamedTuple{(:dims,)}},
+    ::CoDual{typeof(logsumexp)},
+    x::CoDual{<:AbstractArray{T}},
+) where {T<:IEEEFloat}
+    dims = primal(kw).dims
+    xp = primal(x)
+    max_ = maximum(xp; dims, init=typemin(T))
+    @fastmath tmp = exp.(xp .- max_)
+    s = sum(tmp; dims)
+    @fastmath y = max_ .+ log.(s)
```
sunxd3, line 175: could `s` be a scalar here?

test/ext/nnlib/nnlib.jl
```diff
@@ -118,6 +164,29 @@ dropout_tester_3(Trng, x, p) = dropout(Trng(1), x, p; dims=(1, 2))
             _rand(rng, 3, 3, 2),
         ),
 
+        # logsumexp with Adjoint, Transpose
+        (false, :stability, true, logsumexp, _rand(rng, 2, 3)'),
+        (false, :none, true, Core.kwcall, (dims=1,), logsumexp, _rand(rng, 3, 3)'),
```
sunxd3: why `none`?

AstitvaAggarwal: The kwarg rule is type unstable for `Adjoints`, `Transpose` types when `maximum(...; dims)` is called on them. Solved in the current PR state, by `collect`ing the `Adjoint`/`Transpose` so that `maximum` does not work with a permutated view of the `AbstractArray` on `CPU` (this causes type instability)

ext/MooncakeNNlibExt.jl
AstitvaAggarwal: The point of writing Mooncake rules is to support Adjoint, Transpose and other such unique abstractarray objects. SupportedArray is not enough for many applications, ill make the current code more robust. 

Solved in latest PR state, (@yebai this also addresses your review's point 4)

ext/MooncakeNNlibExt.jl
AstitvaAggarwal: never, because `sum` is called with `dims` in the kwarg rule only.  It will be scalar in the non kwarg rule only (as no dims passed). see the docstrings for `logsumexp`, it behaves like `sum` - the primal shape depends on kwarg `dim`.   @yebai this is also the response to your review's point 2

sunxd3 (approved):

ext/MooncakeNNlibExt.jl
```diff
@@ -19,9 +39,156 @@ const SupportedArray{P,N} = Union{Array{P,N},AbstractGPUArray{P,N}}
     Tuple{typeof(dropout),AbstractRNG,SupportedArray{P},P} where {P<:IEEEFloat},
     true,
 )
-@from_rrule(MinimalCtx, Tuple{typeof(softmax),SupportedArray{<:IEEEFloat}}, true)
-@from_rrule(MinimalCtx, Tuple{typeof(logsoftmax),SupportedArray{<:IEEEFloat}}, true)
-@from_rrule(MinimalCtx, Tuple{typeof(logsumexp),SupportedArray{<:IEEEFloat}}, true)
+
+# Helper accumulator functions
```
yebai:
```suggestion
# Accumulator helper functions
```

ext/MooncakeNNlibExt.jl
```diff
@@ -19,9 +39,156 @@ const SupportedArray{P,N} = Union{Array{P,N},AbstractGPUArray{P,N}}
...
+    return error(
+        """
+    `_accum_fdata!` is not implemented for array type $(typeof(x)).
+    If you are trying to differentiate through `softmax`, `logsoftmax`, or `logsumexp`
+    with a custom array wrapper, please open an issue or add a method:
+
+        _accum_fdata!(xf::YourTangentType, ::YourArrayType{T}, g::YourGradType) where {T<:IEEEFloat}
+    """,
+    )
+end
```
yebai: Can you use `arrayify` in relevant rules to avoid these helpers? See https://chalk-lab.github.io/Mooncake.jl/stable/utilities/defining_rules/#Canonicalising-Tangent-Types

```suggestion
```

ext/MooncakeNNlibExt.jl
```diff
@@ -19,9 +39,156 @@ const SupportedArray{P,N} = Union{Array{P,N},AbstractGPUArray{P,N}}
...
+
+# logsoftmax rrules
+@is_primitive MinimalCtx Tuple{typeof(logsoftmax),AbstractArray{T}} where {T<:IEEEFloat}
```
yebai: It's better to limit these rules to `SupportedArray`. The same suggestions apply to all uses of `AbstractArray`. 

```suggestion
@is_primitive MinimalCtx Tuple{typeof(logsoftmax),SupportedArray{T}} where {T<:IEEEFloat}
```

ext/MooncakeNNlibExt.jl
```diff
@@ -19,9 +39,156 @@ const SupportedArray{P,N} = Union{Array{P,N},AbstractGPUArray{P,N}}
...
+    @fastmath y = max_ + log(s)
+    res = zero_fcodual(y)
+    function logsumexp_pb!!(dy::T)
+        _accum_fdata!(tangent(x), xp, dy .* tmp ./ s)
+        return NoRData(), NoRData()
+    end
+    return res, logsumexp_pb!!
+end
+
+function Mooncake.rrule!!(
```
yebai, line 163: Let's split this rule into two rules, one for CPU and one for GPU, if the logic differs between CPU and GPU. Clarity is more important than saving a few lines of code. 

The same suggestion applies to other rules with similar patterns.

ext/MooncakeNNlibExt.jl
AstitvaAggarwal: Unfortunately `arrayify` has no method for `Adjoint/Transpose` or other exotic `AbstractArrays`, and it falls through to the error fallback in `src/rules/blas.jl:70` which throws an error.
The `_accum_fdata!` helpers are necessary here because for example : the tangent accumulation for `Adjoint/Transpose` requires writing back into `xf.data.parent` with the appropriate adjoint/transpose operation, which `arrayify` doesn't handle.

In fact the fallback for `_accum_fdata!` is also future proof when encountering unhandled `AbstractArray` objects.

ext/MooncakeNNlibExt.jl
AstitvaAggarwal: without `AbstractArray` we will not be able to take advantage of the future proof  `_accum_fdata!` fallback.
`SupportedArray` is inadequate to handle anything beyond `Arrays, GPUArrays` etc

AstitvaAggarwal:
@yebai I have now handled and added tests for `Inf` Instability as mentioned in points 1, 3 [here](https://github.com/chalk-lab/Mooncake.jl/pull/1043#issuecomment-4025326659)

ext/MooncakeNNlibExt.jl
yebai: Let's add `Adjoint/Transpose` support for `arrayify`, then add `Adjoint/Transpose`  to `SupportedArrays` types.

ext/MooncakeNNlibExt.jl
AstitvaAggarwal: The GPU instability only comes in `logsumexp` due to `maximum` usage - no other rule is affected.

Im not sure i understand the CPU, GPU splitting for this rule, the current methods/operators dispatch correctly according to GPU, CPU usage? The `if` condition was only needed to avoid the `PermutedDimsArray` to be operated on in a unstable manner (`collect` fixes this to an `Array`)

AstitvaAggarwal:
@yebai @sunxd3 should we get this (and the other PR) merged and add the adjoint, transpose arrayify functionality + integration seperately? 

I say this because the main review points seem to be addressed already, the additional changes are not trivial and extend beyond this PRs scope (3 activation functions)

yebai:
Let's merge #1048 and #1049 into this branch first -- both are ready to go.

PR #1045, opened 2026-03-09
Minor fixes for CuArray tangent rules

yebai:
## Issues fixed

**`lgetfield` guards were wrong** (`name===2||name===:dims` → `name===1||name===:data`).
The original guard was written for an old 2-field `CuArray` where field 2 was `dims`. In CUDA.jl v5, field 2 is `maxsize` (non-differentiable) and field 1 is `data` (the differentiable `DataRef`). The inverted condition meant the `data` field was returning `NoTangent()` and every other field was being treated as differentiable — silently zero-ing all gradients through `CuArray`.

**`lgetfield` 2-arg `rrule!!` returned the wrong fdata** (`x.dx` → `x.dx.data`).
The fdata for a `DataRef` field should be the `DataRef` from the tangent CuArray, not the whole tangent CuArray itself.

**`lgetfield` 2-arg `rrule!!` had wrong `NoPullback` arity** (4 → 3 args).
The 2-arg `lgetfield` rrule takes 3 arguments; `NoPullback` was constructed for 4.

**Vararg constructor rule allocated garbage GPU memory as tangent.**
`frule!!`/`rrule!!` for `Tuple{Type{<:CuArray}, UndefInitializer, Vararg{Int,N}}` returned `P(undef, _dims)` as the tangent — a fresh uninitialized CuArray. This is semantically wrong (tangents must be zero-initialised). The Vararg form is a thin wrapper over the `Dims{N}` form, which already has a correct `@zero_derivative` primitive, so the Vararg rule is unnecessary.

**Redundant `_new_` frule!!s for `Adjoint`/`Transpose`.**
Two `frule!!` overloads for `_new_` on `Adjoint{T,CuArray}` and `Transpose{T,CuArray}` were identical to what the generic `_new_` rule provides. They were removed.

## New rules

**`_new_` rules for the DataRef-based `CuArray` inner constructor.**
The 4-argument inner constructor `CuArray{T,N,M}(data, maxsize, offset, dims)` is used by views and similar operations. Without a custom rule the generic `_new_` would construct a `Tangent{CuArray,...}` — incorrect since `tangent_type(CuArray) = CuArray`. The new rules create the tangent CuArray from `data.dx` (the `DataRef` stored in the input tangent).

**`reshape` primitive.**
CUDA.jl's `reshape` body calls `copy(DataRef{...})` for reference counting, which internally uses `llvmcall`. Mooncake cannot translate `llvmcall`, so `reshape` is made a primitive. The tangent is `reshape(x.dx, dims)` — a view into the same memory — so `NoPullback` is correct.

## Tests

- `is_primitive=false` for the 1D `CuArray(undef, n)` constructor (now traced as derived, via the `Dims{N}` primitive).
- Two new `reshape` test cases (real and complex).

### DIFF
--- a/ext/MooncakeCUDAExt.jl
+++ b/ext/MooncakeCUDAExt.jl
@@ -60,9 +60,9 @@ const CuMaybeComplexArray = Union{CuFloatArray,CuComplexArray}
 # `y = _new_(typeof(y), getfield(x, :data), getfield(x, :maxsize), getfield(x, :offset), (32, 64))`.
 # If this path grows, nearby internal storage types to watch for custom tangent handling are
 # `GPUArrays.RefCounted`, `CUDA.Managed`, `CUDA.DeviceMemory`, `CuStream`, and `CuContext`.
-# Mooncake's default tangents can easily become the wrong representation for these
-# handle-like CuArray internal types: their fields describe ownership/state/pointers, not
-# differentiable array data.
+# Mooncake's default tangents may not be buildable for these handle-like CuArray internal
+# types, because their fields can include ownership/state/pointer internals implemented as
+# new primitive types that need explicit tangent support.
 const CuDataRef = fieldtype(CuArray{Float32,1}, :data)
 
 # DataRef values are mutable reference-counted handles; preserve them as-is for fdata.


yebai:
@kshyatt, this improves the CuArray tangent by resolving all TODO items and addressing a few minor bugs. Can you see whether it sitll pass your use cases?

ext/MooncakeCUDAExt.jl
sunxd3, line 42:
```suggestion
    to_cr_tangent,
    mooncake_tangent,
```

yebai:
@AstitvaAggarwal, please feel free to merge and release.

AstitvaAggarwal:
@yebai GPU tests fail

PR #1047, opened 2026-03-09
Dual Tangent testing + Abstract `Vararg` / `NTuple` `UnionAll` handling.

AstitvaAggarwal:
This PR adds: 

- Tangent tests for `Mooncake.Dual` tangent types similar to current existing `Mooncake.CoDual` tangent tests.
- Fix handling for UnionAll generators (like `NTuple{N,Int} where {N}` , `Tuple{Vararg{Float64}}` etc.) for both `CoDual` and `Dual`.
- Adds Union type Splitting for primals called with `dual_type` (similar to  `_codual_internal`) .

This PR also unblocks https://github.com/LuxDL/Lux.jl/actions/runs/22834613426/job/66230608297?pr=1664 (This fails as `_codual_internal` cannot currently handle `UnionAll` generators.

**MWE** : 
```
using Mooncake, NNlib

x = rand(Float32, 10, 1, 1)
pd = NNlib.PoolDims(x, 2)
f = x -> sum(NNlib.lpnormpool(x, pd; p=2))

cache = Mooncake.prepare_gradient_cache(f, x)
val, grad = Mooncake.value_and_gradient!!(cache, f, x)
```

### DIFF
--- a/src/tangents/codual.jl
+++ b/src/tangents/codual.jl
@@ -38,12 +38,15 @@ zero_codual(x) = CoDual(x, zero_tangent(x))
 Equivalent to `CoDual(x, uninit_tangent(x))`.
 """
 uninit_codual(x) = CoDual(x, uninit_tangent(x))
+
 function _codual_internal(::Type{P}, f::F, extractor::E) where {P,F,E}
     P == Union{} && return Union{}
     P == DataType && return CoDual
     P isa Union && return Union{f(P.a),f(P.b)}
-    # P is abstract (a UnionAll generator or the UnionAll type itself), so we dont know its tangent type.
-    (P isa UnionAll || P == UnionAll) && return CoDual
+    # Use `isa` not `<:`: generators like `NTuple{N,Int} where N` are instances of
+    # UnionAll but not subtypes of it (`NTuple{N,Int} where N <: UnionAll` is false).
+    # `P == UnionAll` handles the UnionAll metatype itself (`UnionAll isa UnionAll` is false).
+    (P isa UnionAll || P == UnionAll) && return CoDual # P is abstract, tangent type unknown.
 
     if P <: Tuple && !all(isconcretetype, (P.parameters...,))
         field_types = (P.parameters...,)
--- a/src/tangents/dual.jl
+++ b/src/tangents/dual.jl
@@ -32,8 +32,10 @@ function dual_type(::Type{P}) where {P}
     P == Union{} && return Union{}
     P == DataType && return Dual
     P isa Union && return Union{dual_type(P.a),dual_type(P.b)}
-    # P is abstract (a UnionAll generator or the UnionAll type itself), so we dont know its tangent type.
-    (P isa UnionAll || P == UnionAll) && return Dual
+    # Use `isa` not `<:`: generators like `NTuple{N,Int} where N` are instances of
+    # UnionAll but not subtypes of it (`NTuple{N,Int} where N <: UnionAll` is false).
+    # `P == UnionAll` handles the UnionAll metatype itself (`UnionAll isa UnionAll` is false).
+    (P isa UnionAll || P == UnionAll) && return Dual # P is abstract, tangent type unknown.
 
     # Union Splitting
     if P <: Tuple && !all(isconcretetype, (P.parameters...,))
--- a/test/front_matter.jl
+++ b/test/front_matter.jl
@@ -47,8 +47,7 @@ using Mooncake:
     ForwardMode,
     ReverseMode,
     MistyClosureTangent,
-    dual_type,
-    verify_dual_type
+    dual_type
 
 using Mooncake:
     CC,


test/tangents/dual.jl
```diff
@@ -0,0 +1,48 @@
...
+                Dual{Tuple{Nothing,Float64},Tuple{NoTangent,Float64}},
+            },
+        ),
+
+        # General Abstract Tuples
+        (Tuple{Any}, Dual),
+
+        # Abstract Vararg / NTuple UnionAll tuples
+        (NTuple{N,Int} where {N}, Dual),
+        (Tuple{Vararg{Float64}} where {N}, Dual),
```
sunxd3: why not `Tuple{Vararg{Float64, N}} where {N}`

yebai:
Cc @Technici4n and @gdalle

test/tangents/dual.jl
AstitvaAggarwal: fixed, also added an additional testcase

AstitvaAggarwal:
feel free to merge @yebai

src/tangents/codual.jl
```diff
@@ -38,11 +38,12 @@ zero_codual(x) = CoDual(x, zero_tangent(x))
 Equivalent to `CoDual(x, uninit_tangent(x))`.
 """
 uninit_codual(x) = CoDual(x, uninit_tangent(x))
-
 function _codual_internal(::Type{P}, f::F, extractor::E) where {P,F,E}
     P == Union{} && return Union{}
     P == DataType && return CoDual
     P isa Union && return Union{f(P.a),f(P.b)}
+    # P is abstract (a UnionAll generator or the UnionAll type itself), so we dont know its tangent type.
+    (P isa UnionAll || P == UnionAll) && return CoDual
```
yebai: `P` is a type, so `isa` is incorrect here. 

```suggestion
    # P is abstract (a `UnionAll` generator or the `UnionAll` type itself), 
    # so we don't know its tangent type.
    P <: UnionAll && return CoDual
```

src/tangents/dual.jl
```diff
@@ -29,9 +29,25 @@ zero_dual(x) = Dual(x, zero_tangent(x))
 randn_dual(rng::AbstractRNG, x) = Dual(x, randn_tangent(rng, x))
 
 function dual_type(::Type{P}) where {P}
+    P == Union{} && return Union{}
     P == DataType && return Dual
     P isa Union && return Union{dual_type(P.a),dual_type(P.b)}
-    P <: UnionAll && return Dual # P is abstract, so we don't know its tangent type.
+    # P is abstract (a UnionAll generator or the UnionAll type itself), so we dont know its tangent type.
+    (P isa UnionAll || P == UnionAll) && return Dual
```
yebai:
```suggestion
    # P is abstract (a `UnionAll` generator or the `UnionAll` type itself), 
    # so we don't know its tangent type.
    P <: UnionAll && return CoDual
```

src/tangents/codual.jl
AstitvaAggarwal: This and the below review (also the current Mooncake code) will completely miss out on types : 

```
julia> NTuple{N,Int} where {N} isa UnionAll
true

julia> NTuple{N,Int} where {N} <: UnionAll
false
```

Types in Julia are first class objects of `UnionAll`, so we need isa check with `UnionAll` and also an explicit check for `UnionAll` itself.

src/tangents/dual.jl
AstitvaAggarwal: addressed above.

src/tangents/codual.jl
yebai: Interesting. Julia's UnionAll type is tricky. See below for more examples. 

@Technici4n may want to take a look at this PR too. 

```
julia> Type{NTuple{N,Int} where N} <: UnionAll
true

julia> NTuple{N,Int} where {N} isa UnionAll
true

julia> versioninfo()
Julia Version 1.12.5
Commit 5fe89b8ddc1 (2026-02-09 16:05 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 256 × AMD EPYC 9555 64-Core Processor
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, znver5)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 256 virtual cores)
Environment:
  JULIA_DEPOT_PATH = /homes/hg344/projects/local/x64/julia
```

AstitvaAggarwal: This further confirms my change, `Type{NTuple{...}} <: UnionAll` returns `true` because it is checking the type of the generator, but in the original function `P ` is the generator itself. Since `P` is an instance of `UnionAll` (not a subtype of it), using `P isa UnionAll` is the correct usage.

AstitvaAggarwal:
lets get this merged?

PR #1048, opened 2026-03-10
refactor: use arrayify for Adjoint/Transpose accumulation in NNlib rrules

yebai:
## Summary

Closes / improves on #1043.

### `src/rules/blas.jl`
Added two new `arrayify` methods for `Adjoint` and `Transpose`:

```julia
function arrayify(x::Adjoint{T,<:AbstractArray{T}}, dx::TangentOrFData) where {T<:Union{IEEEFloat,BlasFloat}}
    _, _dx = arrayify(x.parent, _fields(dx).parent)
    return x, adjoint(_dx)
end
function arrayify(x::Transpose{T,<:AbstractArray{T}}, dx::TangentOrFData) where {T<:Union{IEEEFloat,BlasFloat}}
    _, _dx = arrayify(x.parent, _fields(dx).parent)
    return x, transpose(_dx)
end
```

### `ext/MooncakeNNlibExt.jl`
- Each pullback now uses `arrayify` directly on the captured `CoDual`:
  ```julia
  _, dx = arrayify(x)
  dx .+= <gradient expression>
  ```
  The `CoDual` overload of `arrayify` already unpacks `primal(x)` and `tangent(x)`, so no extra arguments are needed.
- Removed now-unused `FData` import.

## Test results

All 1273 tests pass (1254 rule tests + 6 arrayify tests + 13 logsumexp stability tests).


AstitvaAggarwal:
@yebai thanks for opening this, but this will need much more thought. simply because we need to account for tangent accumulation in Adjoints, transpose matrices, handling of type instability for CPU and some other issues.

src/rules/blas.jl
```diff
@@ -62,6 +62,18 @@ function arrayify(
     _, _dx = arrayify(x.data, _fields(dx).data)
     return x, Tx(_dx)
 end
+function arrayify(
```
AstitvaAggarwal, line 65: These Arrayify methods are for all subtypes of `AbstractArray`? 
This will fail when the primal used with `arrayify` is not just `Adjoint`/`Transpose`. 
Better to dispatch for `SupportedArrays` only (like `_accum_fdata!` [did](https://github.com/chalk-lab/Mooncake.jl/pull/1048/changes#diff-934143dbdbf63bdda0914dac2bd05737124f51d058f7428b9e4410382439d3e7L57)) and provide a fall back like here: https://github.com/chalk-lab/Mooncake.jl/pull/1048/changes#diff-934143dbdbf63bdda0914dac2bd05737124f51d058f7428b9e4410382439d3e7L62

yebai: That is a misunderstanding. The `arrayify` is already type-constrained to `Adjoint/Transpose`.

AstitvaAggarwal: I was refering to the inner type within the `Transpose/Adjoint`. But I just saw https://github.com/chalk-lab/Mooncake.jl/blob/f8b97790d5fbe4a3663e74c6a3567a8f86ad85b9/src/rules/blas.jl#L70 exists, so all is good 👍

PR #1049, opened 2026-03-10
Widen SupportedArray to include Adjoint/Transpose

yebai:
## Summary

> ⚠️ Please merge after PR #1048.

Widens `SupportedArray` in `MooncakeNNlibExt` to cover `Adjoint` and `Transpose` wrappers of CPU/GPU arrays, so the custom `logsoftmax`/`softmax`/`logsumexp` rules apply to those types too.

## Changes

- Redefine `SupportedArray{P}` as a 1-parameter union: `Array{P}`, `AbstractGPUArray{P}`, and their `Adjoint`/`Transpose` wrappers
- Replace `AbstractArray{T}` with `SupportedArray{T}` in all `@is_primitive` and `rrule!!` declarations
- Inline `Union{Array{P,3},AbstractGPUArray{P,3}}` for `batched_mul` (needs `N=3`, no Adjoint/Transpose)
- Remove the `collect` workaround in `logsumexp_kw` (was only needed for CPU Adjoint/Transpose to avoid `PermutedDimsArray` instability in `maximum`)
- Drop `LinearAlgebra: Adjoint, Transpose` import from the ext
- Restore Adjoint/Transpose test cases for all three functions

### DIFF
--- a/ext/MooncakeNNlibExt.jl
+++ b/ext/MooncakeNNlibExt.jl
@@ -34,6 +34,19 @@ const SupportedArray{P} = Union{
     Transpose{P,<:Union{Array{P},AbstractGPUArray{P}}},
 }
 
+# On Julia ≤ 1.11, `maximum(x::Adjoint/Transpose; dims, init)` routes through
+# `LinearAlgebra.mapreducedim! → switch_dim12 → PermutedDimsArray`, leaving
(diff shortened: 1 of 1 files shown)


AstitvaAggarwal:
@yebai please let me handle the review suggestions within my original PR. I say this because the two PRs created need much more thought around some problems and also complicate the original PR at hand.

PR #1036, opened 2026-03-06
Add `build_primitive_frule` and use it for forward-over-reverse caching

Technici4n:
Still WIP, this seems to be causing issues with debug mode somehow. Maybe `_copy` needs to preserve aliasing.

PR #1051, opened 2026-03-10
Fix DebugFRule → DebugRRule in FoRCache and document aliasing

yebai:
Fixes two bugs in `FoRCache` that caused `Rosenbrock (debug_mode=true)` to fail (continuing #1036).

### Bug 1: `DebugFRule` → `DebugRRule`

`build_derived_rrule` returns a `DerivedRule` — a reverse-mode rule. In debug mode, `FoRCache` was wrapping it in `DebugFRule` instead of `DebugRRule`, causing the forward-mode interpreter to crash when called with `CoDual` arguments.

### Bug 2: stale tangent captures on cache hit

On a cache hit, `_copy(rule)` resets primal `Stack`s to empty (`_copy(Stack{T}) = Stack{T}()`). The old approach `deepcopy`'d the post-pass tangent, retaining grown `Vector` lengths — producing a size mismatch with the fresh primal. This is incorrect regardless of debug mode, but only caught by `DebugFRule`'s size-consistency check.

Fix: cache `(rule, fwd_dual_callable, rvs_dual_callable)` instead of `(rule, rule_tangent)`. On a cache hit, `captures_tangent` is always regenerated via `zero_tangent` on the fresh `_copy`'d primal captures, called jointly on `(fwds_oc.oc.captures, pb_oc_ref[].oc.captures)` so aliasing between fwds/rvs comms objects is preserved via `zero_tangent`'s `IdDict`.

### DIFF
--- a/src/rules/high_order_derivative_patches.jl
+++ b/src/rules/high_order_derivative_patches.jl
@@ -8,9 +8,9 @@
     typeof(build_derived_rrule),MooncakeInterpreter{C},Any,Any,Bool
 } where {C}
 
-function build_primitive_frule(sig::Type{<:Tuple{
-    typeof(build_derived_rrule),MooncakeInterpreter,Any,Any,Bool
-}})
+function build_primitive_frule(
+    sig::Type{<:Tuple{typeof(build_derived_rrule),MooncakeInterpreter,Any,Any,Bool}}
+)
     return FoRCache(Dict{ClosureCacheKey,Any}())
 end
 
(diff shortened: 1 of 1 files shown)


yebai:
@Technici4n, this hopefully identified and fixed the remaining issues #1036.

Technici4n:
Thank you Hong, that's very helpful. I will have a look shortly

Technici4n:
Thanks again!

PR #1056, opened 2026-03-11
feat(CUDA) GPU broadcast, matmul/matvecmul rules, CUBLAS/CUSPARSE primitives

yebai:
Fix https://github.com/chalk-lab/Mooncake.jl/issues/278

## Summary

Adds forward- and reverse-mode AD support for GPU (CUDA) operations in Mooncake.

## Examples

```julia
using CUDA, Mooncake, LinearAlgebra

# 1. GPU matrix multiplication
f = (A, B) -> sum(A * B)
A, B = CUDA.randn(Float32, 4, 4), CUDA.randn(Float32, 4, 4)
cache = prepare_gradient_cache(f, A, B)
_, (_, ∂A, ∂B) = value_and_gradient!!(cache, f, A, B)

# 2. Differentiating through a GPU broadcast
f = x -> sum(sin.(x) .* cos.(x))
x = CUDA.randn(Float32, 8)
cache = prepare_gradient_cache(f, x)
_, (_, ∂x) = value_and_gradient!!(cache, f, x)  # ∂x::CuArray{Float32}

# 3. GPU → CPU transfer inside the function
f = x -> sum(Array(x).^2)
x = CUDA.randn(Float32, 4)
cache = prepare_gradient_cache(f, x)
_, (_, ∂x) = value_and_gradient!!(cache, f, x)

# 4. sum with a function over a GPU array
f = x -> sum(abs2, x)
x = CUDA.randn(Float32, 8)
cache = prepare_gradient_cache(f, x)
_, (_, ∂x) = value_and_gradient!!(cache, f, x)

# 5. Complex GPU matrix multiplication
f = A -> real(sum(A * adjoint(A)))
A = CUDA.randn(ComplexF32, 4, 4)
cache = prepare_gradient_cache(f, A)
_, (_, ∂A) = value_and_gradient!!(cache, f, A)

# 6. norm — common in regularization (∂x ≈ 2x)
f = x -> norm(x)^2
x = CUDA.randn(Float32, 8)
cache = prepare_gradient_cache(f, x)
_, (_, ∂x) = value_and_gradient!!(cache, f, x)

# 7. dot product (∂x ≈ y, ∂y ≈ x)
f = (x, y) -> dot(x, y)
x, y = CUDA.randn(Float32, 4), CUDA.randn(Float32, 4)
cache = prepare_gradient_cache(f, x, y)
_, (_, ∂x, ∂y) = value_and_gradient!!(cache, f, x, y)

[...]

PR #1057, opened 2026-03-12
PR 1: register non-differentiable CUDA types and fix Task field handling

yebai:
Part 1 of #1056 

## Summary

Prerequisite for more CUDA rules: register opaque CUDA types as
non-differentiable and fix `Task` field access so Mooncake does not attempt to
trace through CUDA runtime internals.

## Changes

**`ext/MooncakeCUDAExt.jl`**
- Expand `CuDataRef` to cover all three CUDA memory kinds: `DeviceMemory`,
  `UnifiedMemory`, and `HostMemory` (previously only `DeviceMemory` was handled)
- Add `tangent_type(T) = NoTangent` for opaque CUDA types:
  `cublasStatus_t`, `cublasGemmAlgo_t`, `cublasHandle_t`, `cusparseHandle_t`,
  `CUdevice_attribute_enum`, `cudaError_enum`, `libraryPropertyType`
- Add `zero_tangent_internal` overrides for opaque `Ptr` types used by CUBLAS/CUSPARSE
- Add `@zero_derivative` for `CUBLAS.version`, `CUDA.cuDeviceGetAttribute`, and
  `CUDA.attribute`

**`src/rules/tasks.jl`**
- Mark all non-`@task`-state `Task` fields (`:storage`, `:code`, `:donenotify`,
  `:result`, `:logstate`, `:next`, `:queue`, `:sticky`, etc.) as non-differentiable
  runtime infrastructure, preventing Mooncake from attempting to trace them

**`src/rules/misc.jl`**
- Add `_get_tangent_field` helper for `NoTangent` structs (field access returns
  `NoTangent()` regardless of field name)

## Why this is needed

Without explicit `NoTangent` registrations, Mooncake tries to construct
tangent types for them, which fails with confusing errors deep inside the AD
machinery. These registrations make the failure surface early and cleanly for any
rule that touches these types.

### DIFF
--- a/ext/MooncakeCUDAExt.jl
+++ b/ext/MooncakeCUDAExt.jl
@@ -147,9 +147,9 @@ for (_cuda_opaque_t, _is_ptr) in [
     # --- other non-differentiable CUDA types ---
     (libraryPropertyType, false),
     (CUdevice_attribute_enum, false),
-    (cublasPointerMode_t, false),
-    (cublasFillMode_t, false),
-    (cublasDiagType_t, false),
+    (CUBLAS.cublasPointerMode_t, false),
+    (CUBLAS.cublasFillMode_t, false),
+    (CUBLAS.cublasDiagType_t, false),
     (cudaError_enum, false),
     # CuStream contains Ptr/Bool/CuContext fields; without NoTangent, Mooncake generates a
     # MutableTangent that propagates into task-local CUDA state → SIGILL at runtime.


ext/MooncakeCUDAExt.jl
```diff
@@ -53,26 +72,22 @@ import Mooncake.TestUtils:
...
-# `GPUArrays.RefCounted`, `CUDA.Managed`, `CUDA.DeviceMemory`, `CuStream`, and `CuContext`.
-# Mooncake's default tangents may not be buildable for these handle-like CuArray internal
-# types, because their fields can include ownership/state/pointer internals implemented as
-# new primitive types that need explicit tangent support.
-const CuDataRef = fieldtype(CuArray{Float32,1}, :data)
+
+# CuArray{T,N,M}.data is a DataRef — a reference-counted handle to the GPU memory buffer.
+# Operations like reshape and view reconstruct a CuArray from its components.
+# CuDataRef expanded to cover DeviceMemory, UnifiedMemory, and HostMemory variants.
+const CuDataRef = Union{
```
kshyatt, line 89: Why do these all have `Float32` eltype?

ext/MooncakeCUDAExt.jl
```diff
@@ -81,10 +96,79 @@ Mooncake.__verify_fdata_value(::IdDict{Any,Nothing}, ::CuDataRef, ::CuDataRef) =
...
+    (CUSPARSE.cusparseDnMatDescr, true),
+    (CUSPARSE.cusparseSpSVDescr, true),
+    (CUSPARSE.cusparseSpSMDescr, true),
+    (CUSPARSE.cusparseSpGEMMDescr, true),
+    (CUSPARSE.cusparseSpMMOpPlan, true),
+    # --- CUBLAS enum/status types ---
+    (CUBLAS.cublasOperation_t, false),
+    (CUBLAS.cublasComputeType_t, false),
+    (CUBLAS.cublasStatus_t, false),
+    (CUBLAS.cublasGemmAlgo_t, false),
```
kshyatt, line 146: There are some more that are missed here eg the diagonal chars or upper/lower for triangular methods

yebai: Can you list them explicitly? I am not familiar with CUDA.jl

ext/MooncakeCUDAExt.jl
```diff
@@ -53,26 +72,22 @@ import Mooncake.TestUtils:
...
-# If this path grows, nearby internal storage types to watch for custom tangent handling are
-# `GPUArrays.RefCounted`, `CUDA.Managed`, `CUDA.DeviceMemory`, `CuStream`, and `CuContext`.
-# Mooncake's default tangents may not be buildable for these handle-like CuArray internal
-# types, because their fields can include ownership/state/pointer internals implemented as
-# new primitive types that need explicit tangent support.
-const CuDataRef = fieldtype(CuArray{Float32,1}, :data)
+
+# CuArray{T,N,M}.data is a DataRef — a reference-counted handle to the GPU memory buffer.
+# Operations like reshape and view reconstruct a CuArray from its components.
+# CuDataRef expanded to cover DeviceMemory, UnifiedMemory, and HostMemory variants.
```
yebai:
```suggestion
# CuArray{T,N,M}.data is a DataRef — a reference-counted handle to the GPU memory buffer.
# Operations like reshape and view reconstruct a CuArray from its components:
#   `y = _new_(typeof(y), getfield(x, :data), getfield(x, :maxsize), getfield(x, :offset), dims)`.
# The tangent of data flows through these _new_ calls, so Mooncake needs lgetfield and
# _new_ rules for DataRef.
#
# CuArray{T,N,M} uses a different DataRef concrete type for each memory kind M:
#   DeviceMemory  → DataRef{Managed{DeviceMemory}}
#   UnifiedMemory → DataRef{Managed{UnifiedMemory}}
#   HostMemory    → DataRef{Managed{HostMemory}}
# DataRef does NOT depend on T or N — only on M — so three entries cover every
# CuArray{T,N,M} combination.  Missing a variant causes Mooncake to fall through to the
# generic struct handler, which tries to build tangents for DataRef's internal Ptr fields.
```

ext/MooncakeCUDAExt.jl
yebai: Updated the comments.

ext/MooncakeCUDAExt.jl
kshyatt: There's `cublasPointerMode_t`, `cublasFillMode_t`, `cublasDiagType_t`

yebai: @kshyatt, I added these to the list. Feel free to add more to this list by pushing commits to this PR or new PRs after merging this.

kshyatt: Sure, thanks for getting this rolling!

sunxd3 (approved):
I can't comment on completeness, but the current code look good to me

PR #1066, opened 2026-03-15
SLEEFPirates, LuxLib silent errors

AstitvaAggarwal:
Fix a faulty rule addition. Errors extremely silently on Lux's CI only. Its passes locally and on Mooncake's CI itself due to the nature of the function and it's special usage in Lux, LuxLib. (hijacking activation functions for faster performance)


AstitvaAggarwal:
@yebai i will release a new Mooncake version through this PRs commit.

PR #1072, opened 2026-03-16
fix(cuda): free NDual arrays in sum(f,x) rrules; factor pullback helper; fix allocs

yebai:
## Summary

Follow-up fixes to #1061 identified during code review.

- **`sum(f,x)` rrules kept full NDual `out` array alive in pullback closure.** The `materialize` and `materialize!` rrules correctly pre-extract partial slots into plain `CuArray{T}` and set `out = nothing` to let the NDual array be GC'd. The `sum(f, x)` rules did not — holding a 2× (real) or 3× (complex) sized array alive for the whole backward pass. Fixed to match the materialize pattern.

- **`materialize_pb!!` and `materialize!_pb!!` duplicated ~35 lines of pullback accumulation logic.** Factored into `_gpu_accum_pullback!` helper called by both.

- **`_check_mixed_gpu_eltype` built a heap `Vector` unconditionally** via comprehension + `unique`. Replaced with an early-exit loop that allocates nothing in the common single-type case, and only errors on the first mismatch.

- **`_unbroadcast` used `filter(d -> dims[d], 1:ndims(dx))`** which creates a heap `Vector{Int}` on every pullback call. Replaced with `ntuple` + `filter` on a `Tuple` (no `Vector` allocation for small `ndims`).


ext/MooncakeCUDAExt/MooncakeCUDAExt.jl
```diff
@@ -1222,9 +1222,19 @@ function rrule!!(
     flat_dx = _fields(tangent(x)).parent
     out = _gpu_broadcast_dual(primal(f), flat_px)
     y = sum(_gpu_dual_val, out)
+    # Pre-extract the single partial slot and free the NDual array immediately.
+    # The NDual array is (1+1)× the size of a float array; keeping it alive in the
+    # closure for the pullback doubles GPU memory use unnecessarily.
+    partial_slot = if _is_gpu_differentiable(eltype(out))
+        p = broadcast(o -> _gpu_dual_part_cx(o, 1), out)
+        p
```
AstitvaAggarwal:
```suggestion
        broadcast(o -> _gpu_dual_part_cx(o, 1), out)
```

ext/MooncakeCUDAExt/MooncakeCUDAExt.jl
```diff
@@ -1266,10 +1276,25 @@ function rrule!!(::CoDual{typeof(sum)}, f::CoDual, x::CoDual{<:CuComplexArray})
...
+    out = nothing
     function sum_f_cx_pb!!(dy)
-        if _is_gpu_differentiable(eltype(out))
+        if partial_slots !== nothing
+            p1, p2 = partial_slots
+            # p1, p2 are plain CuArray{T} of extracted partials (∂f/∂Re and ∂f/∂Im).
+            # real(conj(dy) * p) is the Wirtinger chain-rule contribution.
             dx .+= broadcast(
-                o -> complex(_gpu_adj_part(o, dy, 1), _gpu_adj_part(o, dy, 2)), out
+                (s1, s2) -> complex(real(conj(dy) * s1), real(conj(dy) * s2)), p1, p2
```
AstitvaAggarwal: avoid calling `conj(dy)` twice?

ext/MooncakeCUDAExt/MooncakeCUDAExt.jl
```diff
@@ -2213,20 +2239,82 @@ end
...
-    )
+    # Walk flat_pargs with an early-exit loop rather than building a temporary array.
+    # In the common case (all same element type), this allocates nothing.
+    first_et = nothing
+    for pa in flat_pargs
+        (
+            pa isa CuMaybeComplexArray ||
+            pa isa Adjoint{<:CuFloatOrComplex,<:CuMaybeComplexArray} ||
+            pa isa Transpose{<:CuFloatOrComplex,<:CuMaybeComplexArray}
+        ) || continue
```
AstitvaAggarwal, line 2305: Minor suggestion, but we can get this conditional in a helper check like `_is_gpu_array_arg(pa) =
    pa isa CuMaybeComplexArray ||
    pa isa Adjoint{<:CuFloatOrComplex,<:CuMaybeComplexArray} ||
    pa isa Transpose{<:CuFloatOrComplex,<:CuMaybeComplexArray}`  . This can be reused for more GPU broadcast helpers?

AstitvaAggarwal:
Apart from minor comments above, the PR looks good

yebai:
Thanks for the comments — both addressed in the latest commits:

1. **`conj(dy)` hoist**: hoisted to `cdy = conj(dy)` before the broadcast in `sum_f_cx_pb!!`.
2. **`_is_gpu_array_arg` helper**: the predicate is currently only used in one place so left inline for now; happy to extract it if it appears elsewhere.

PR #1078, opened 2026-03-16
fix(@zero_derivative): support parametric Vararg{T} and Vararg{T,N}

### LINKED ISSUE #923: more informative error message for `Vararg` in `@zero_derivative`.
`@zero_derivative` doesn't support `Vararg` and may fail sliently to produce a rule, eg, in #922. It would be helpful to improve `@zero_derivative` to warn the user if the usage contains `Vararg`, eg, 

```julia
@zero_derivative MinimalCtx Tuple{Type{<:CuArray},UndefInitializer,Vararg{N,Int}} where {N}
```

yebai:
## Summary

Fixes #923.

`@zero_derivative` only detected bare `Vararg` as a vararg signature. Signatures using `Vararg{T}` or `Vararg{T,N}` fell through to the non-vararg branch, which generated rules with `CoDual{<:Vararg{T,N}}` as a parameter type — nonsensical and never matching any dispatch, so the rule was silently a no-op.

- Adds `_is_vararg_expr`: detects bare `Vararg`, `Vararg{T}`, and `Vararg{T,N}`
- Adds `_vararg_wrapped_type`: generates `Vararg{Dual{<:T},N}` / `Vararg{CoDual{<:T},N}` correctly, including arbitrarily nested element types (e.g. `Vararg{Vector{Float64},N}`)
- Throws `ArgumentError` with an informative message if `Vararg` appears in a non-last position
- Adds `test_rule` coverage for `Vararg{T}`, `Vararg{T,N}`, nested element types, and the non-last-position error

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -186,7 +186,6 @@ function lookup_method(sig)::Union{Method,Nothing}
     world = Base.get_world_counter()
     min = Base.RefValue{UInt}(typemin(UInt))
     max = Base.RefValue{UInt}(typemax(UInt))
-    # Base._methods_by_ftype is a Julia-internal API.
     ms = Base._methods_by_ftype(sig, nothing, 1, world, true, min, max, Ptr{Int32}(C_NULL))
     return (ms isa Vector && !isempty(ms)) ? first(ms).method : nothing
 end
--- a/test/tools_for_rules.jl
+++ b/test/tools_for_rules.jl
@@ -41,6 +41,12 @@ nested_vararg_zero_tester(x::Vector{Float64}...) = 0
     typeof(nested_vararg_zero_tester),Vararg{Vector{Float64},N}
 } where {N}
 
+mixed_vararg_zero_tester(n::Int, x::Float64...) = 0
+@zero_derivative MinimalCtx Tuple{typeof(mixed_vararg_zero_tester),Int,Vararg{Float64}}
+@zero_derivative MinimalCtx Tuple{
+    typeof(mixed_vararg_zero_tester),Int,Vararg{Float64,N}
+} where {N}
+
 zero_tester_forward_only(x) = 0
 @zero_derivative MinimalCtx Tuple{typeof(zero_tester_forward_only),Float64} ForwardMode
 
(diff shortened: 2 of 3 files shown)


src/utils.jl
```diff
@@ -186,6 +186,7 @@ function lookup_method(sig)::Union{Method,Nothing}
     world = Base.get_world_counter()
     min = Base.RefValue{UInt}(typemin(UInt))
     max = Base.RefValue{UInt}(typemax(UInt))
+    # Base._methods_by_ftype is a Julia-internal API; the try/catch in showerror guards callers.
```
yebai:
```suggestion
    # Base._methods_by_ftype is a Julia-internal API.
```

test/tools_for_rules.jl
```diff
@@ -28,6 +28,19 @@ zero_tester(x) = 0
...
+
+counted_vararg_zero_tester(x::Float64...) = 0
+@zero_derivative MinimalCtx Tuple{
+    typeof(counted_vararg_zero_tester),Vararg{Float64,N}
+} where {N}
+
+nested_vararg_zero_tester(x::Vector{Float64}...) = 0
+@zero_derivative MinimalCtx Tuple{
+    typeof(nested_vararg_zero_tester),Vararg{Vector{Float64},N}
+} where {N}
```
sunxd3, line 42: These cover all three parametric Vararg forms well, but none of the new functions has a regular parameter *before* the trailing Vararg — e.g. `Tuple{typeof(f), Int, Vararg{Float64}}`. That leaves the `vcat(map(..., [1:end-1]), _vararg_wrapped_type(...))` path untested with a non-trivial prefix.

Worth adding something like:
```julia
mixed_vararg_zero_tester(n::Int, x::Float64...) = 0
@zero_derivative MinimalCtx Tuple{typeof(mixed_vararg_zero_tester), Int, Vararg{Float64}}
```
with a corresponding `test_rule` call passing `(3, 5.0, 4.0)`.

test/tools_for_rules.jl
```diff
@@ -182,6 +195,35 @@ end
...
+            perf_flag=:stability_and_allocs,
+        )
+        test_rule(
+            sr(123),
+            ToolsForRulesResources.nested_vararg_zero_tester,
+            [5.0],
+            [4.0];
+            is_primitive=true,
+            perf_flag=:stability_and_allocs,
+        )
```
sunxd3, line 233: Every new `test_rule` call passes exactly 2 vararg arguments. Could you add a call with 0 vararg args (just the function, no trailing args) for `counted_vararg_zero_tester`? That would exercise the empty-splat edge case in the generated code (`Expr(Symbol("..."), arg_names[end])` with N=0).

yebai:
Thanks @sunxd3 -- good catches. All should be resolved now!

PR #1079, opened 2026-03-16
feat: add Mooncake.stop_gradient to block gradient flow

### LINKED ISSUE #181: Stop gradient feature
Hi,

Stopping/dropping gradients is quite a common feature for a lot of applications. It would be great to have that in Tapir as well. Currently, `ChainRulesCore` provides `ignore_gradient` and `non_differentiable` as a unified interface, but pretty much only `Zygote` supports it. However, `ReverseDiff` provides its own macro: `ReverseDiff.@skip`.

yebai:
## Summary

Implements `Mooncake.stop_gradient(x)` (closes #181), analogous to `tf.stop_gradient` / `jax.lax.stop_gradient`.

- Returns `x` unchanged in the primal
- Blocks gradient flow in **reverse mode**: `rrule!!` copies the fdata (prevents aliasing of tangent storage), and the pullback returns zero rdata
- Blocks gradient flow in **forward mode**: `frule!!` returns `zero_dual(primal(x))` (zero tangent)
- Declared as `@is_primitive MinimalCtx` for both modes
- Removed the `Core.kwcall` interception — kwargs already produce a native `MethodError` with a clear message
- Added `rrule!!`/`frule!!` for `Core.throw_methoderror` (Julia 1.12+, guarded by `@static if`) and `Core.throw_inexacterror` so these builtins propagate errors correctly through AD instead of hitting `MissingRuleForBuiltinException`

### Usage

```julia
f(x) = x[1] * Mooncake.stop_gradient(x)[2]
_, (_, g) = Mooncake.value_and_gradient!!(Mooncake.build_rrule(f), f, [3.0, 4.0])
# g == [4.0, 0.0]  — only x[1] contributes; x[2] is stopped

# Multiple values via tuple
g(x, y) = sum(Mooncake.stop_gradient((x, y))[1]) + sum(Mooncake.stop_gradient((x, y))[2])
```

## Test plan

- [x] Primal pass-through, full-zero gradient, partial stop, multi-value tuple
- [x] Kwarg rejection: primal throws `MethodError`; AD-context test guarded to Julia 1.12+ (where `Core.throw_methoderror` exists)
- [x] `TestUtils.test_rule` with `interface_only=true` for scalar, array, and tuple inputs

### DIFF
--- a/src/test_utils.jl
+++ b/src/test_utils.jl
@@ -1431,6 +1431,8 @@ function test_get_tangent_field_performance(t::Union{MutableTangent,Tangent})
     end
 end
 
+# Counts the number of GC allocations made by a function call.
+#
 # A plain Vararg method does not fully specialise on Type{T} arguments — they widen to
 # DataType — which causes spurious allocations to be reported. For example:
 #
@@ -1447,27 +1449,25 @@ end
 #                                                                    ^^^^^^^^^^^ preserved
 #
 # @generated functions specialise fully on all argument types (including Type{T}), so
-# lumping everything into one Vararg gives full specialisation on both the function and
-# its arguments. @generated function bodies cannot contain closures, so the @noinline
-# measurement barrier from the old eval-loop approach is replaced by Tcount_allocs: an
-# immutable struct whose callable method is @noinline, replicating the same effect.
-# test_hook is called so that external tools (e.g. the dispatch_doctor integration test)
-# can intercept and suppress the measurement where needed.
-struct Tcount_allocs{F,Args<:Tuple}
-    f::F
-    args::Args
-end
-@noinline function (m::Tcount_allocs)()
-    stats = Base.gc_num()
-    m.f(m.args...)
-    return Base.gc_alloc_count(Base.GC_Diff(Base.gc_num(), stats))
-end
-@generated function count_allocs(f_and_x::Vararg{Any,N}) where {N}
+# packing f and its arguments into one Vararg gives per-element-type specialisation
+# without explicit numbered overloads. @generated function bodies cannot contain closures,
+# so the measurement window (gc_num / gc_alloc_count) lives in __count_allocs (inlinable,
+# no barrier) and test_hook lives in the count_allocs wrapper so that external tools
+# (e.g. dispatch_doctor) can intercept and suppress the measurement where needed.
+@generated function __count_allocs(f_and_x::Vararg{Any,N}) where {N}
+    N >= 1 ||
+        return :(error("__count_allocs requires at least one argument (the function)"))
     args = [:(f_and_x[$i]) for i in 2:N]
     quote
-        test_hook(
-            Tcount_allocs(f_and_x[1], ($(args...),)), count_allocs, f_and_x[1], $(args...)
-        )
+        stats = Base.gc_num()
+        f_and_x[1]($(args...))
+        Base.gc_alloc_count(Base.GC_Diff(Base.gc_num(), stats))
+    end
+end
+function count_allocs(f_and_x::Vararg{Any,N}; test_hook::Bool=true) where {N}
+    test_hook || return __count_allocs(f_and_x...)
+    return TestUtils.test_hook(count_allocs, f_and_x...) do
+        __count_allocs(f_and_x...)
     end
 end
 
--- a/test/integration_testing/dispatch_doctor/dispatch_doctor.jl
+++ b/test/integration_testing/dispatch_doctor/dispatch_doctor.jl
@@ -5,6 +5,9 @@ Pkg.develop(; path=joinpath(@__DIR__, "..", "..", ".."))
 using Mooncake: Mooncake, TestUtils, Tangent
 using DispatchDoctor: allow_unstable, type_instability
 
+# Each test_hook below intercepts a specific TestUtils function to suppress
+# DispatchDoctor's type-instability checks where needed. All hooks dispatch on ::typeof(f)
+# so that adding new methods to those functions automatically narrows the hook.
 TestUtils.test_hook(::Any, ::typeof(TestUtils.test_opt), ::Any...) = nothing
 TestUtils.test_hook(::Any, ::typeof(TestUtils.report_opt), tt) = nothing
 function TestUtils.test_hook(f, ::typeof(Mooncake.hand_written_rule_test_cases), ::Any...)
--- a/test/test_utils.jl
+++ b/test/test_utils.jl
@@ -132,14 +132,18 @@
         @test_throws AssertionError populate_address_map(z, z̄)
     end
     @testset "count_allocs" begin
-        # Zero-alloc functions return 0.
+        # Function form: zero-alloc functions return 0.
         @test TestUtils.count_allocs(identity, 1.0) == 0
         @test TestUtils.count_allocs(+, 1.0, 2.0) == 0
-        # Type{T} arguments are specialised correctly (not widened to DataType).
+        # Regression: non-isbits arg (heap object). An @noinline barrier would box the
+        # array when passing it through the Vararg tuple, causing a spurious allocation
+        # on Julia 1.10. The @generated design avoids the barrier entirely.
+        @test TestUtils.count_allocs(length, [1, 2, 3]) == 0
+        # Regression: Type{T} arg. A plain Vararg method widens Type{Float64} to DataType,
+        # causing a spurious boxing allocation on Julia 1.12. The @generated design
+        # specialises on element types and preserves the concrete Type{T} parameter.
+        @test TestUtils.count_allocs(isbitstype, Float64) == 0
         @test TestUtils.count_allocs(Mooncake.fdata_type, Tuple{Float64}) == 0
         @test TestUtils.count_allocs(Mooncake.fdata_type, Tuple{Vector{Float64}}) == 0
-        # Type constructor as first arg: count_allocs(Ref, 1.0) calls Ref(1.0), which
-        # heap-allocates. Verifies count_allocs works when f is a type constructor.
-        @test TestUtils.count_allocs(Ref, 1.0) >= 1
     end
 end


yebai:
@Red-Portal, can you give this a try and see whether this helps?

yebai:
@penelopeysm can you help take a look? This was motivated by a need in AdvancedVI, I think.

src/rules/misc.jl
```diff
@@ -39,6 +39,87 @@
...
+"""
+    stop_gradient(x)
+    stop_gradient(x, y, ...)
+
+Returns its argument(s) with zero gradient. Gradients will not propagate through any
+argument in the reverse pass. In the forward pass, arguments are returned unchanged.
+
+With a single argument, returns the argument directly. With multiple arguments, returns
+a tuple. With no arguments, returns an empty tuple. Keyword arguments are not supported
+and will throw an `ArgumentError`.
```
Technici4n: Would it not be cleaner to always accept exactly one argument, and let the user wrap/unwrap it in a tuple? That will also fix aliasing.

src/rules/misc.jl
```diff
@@ -39,6 +39,87 @@
...
+
+function rrule!!(::CoDual{typeof(stop_gradient)}, args::CoDual...)
+    # Copy fdata so that in-place gradient accumulation into the output does not
+    # affect the inputs' fdata (i.e., avoids aliasing of tangent storage).
+    primals = map(primal, args)
+    fdata_copies = map(x -> _copy(tangent(x)), args)
+    y_primal = length(primals) == 1 ? only(primals) : primals
+    y_fdata = length(fdata_copies) == 1 ? only(fdata_copies) : fdata_copies
+    y = CoDual(y_primal, y_fdata)
+    lzrs = map(x -> lazy_zero_rdata(primal(x)), args)
```
Technici4n: I would be a little careful here as we are losing aliasing of the primals.

yebai: See, #1081

src/rules/misc.jl
yebai: That was the first version; it was later generalised to support multi-argument functions. Let's keep as-is unless there is a stronger case against it.

penelopeysm: I agree with @Technici4n, this API is really weird, it should either always be one argument or always a tuple, but now as implemented there's this special case where I have to remember that if the tuple is exactly length 1 then it unwraps...

you can see that this is problematic and hinders the writing of generic code with e.g.

```julia
# definition copied from below
stop_gradient(args...) = length(args) == 1 ? args[1] : args

function g(args::Tuple)
    println(args)
end

function f(args...)
    g(stop_gradient(args...))
end

julia> f(1, 2)
(1, 2)

julia> f(1)
ERROR: MethodError: no method matching g(::Int64)

julia> f()
()
```

src/rules/misc.jl
```diff
@@ -39,6 +39,105 @@
...
+    the two fdata buffers diverge. For example:
+    ```julia
+    function f(x)
+        y = stop_gradient(x)  # primal(y) === x, but fdata(y) ≠ fdata(x)
+        y[1] = 2.0            # mutates x[1], but tangent goes into fdata(y)
+        return x[1] + x[2]   # reads fdata(x), which is now out of sync with fdata(y)
+    end
+    ```
+    In practice this is unlikely: `stop_gradient` is intended to detach a value, not to
+    alias it back alongside the original. See issue #1081.
```
penelopeysm: Maybe don't claim that this is unlikely in the docstring? Let people judge for themselves whether it applies to their usecase.
```suggestion
    See https://github.com/chalk-lab/Mooncake.jl/issues/1081 for more details.
```

src/rules/misc.jl
```diff
@@ -39,6 +39,105 @@
...
+ 0.0
+```
+"""
+stop_gradient(args...) = length(args) == 1 ? args[1] : args
+
+# Intercept kwarg calls without generating a positional fallback (which would overwrite
+# the method above). Defining directly on Core.kwcall avoids that conflict.
+function Core.kwcall(::NamedTuple, ::typeof(stop_gradient), args...)
+    throw(ArgumentError("stop_gradient does not support keyword arguments"))
+end
```
penelopeysm: Why is this necessary? If you call it with keyword arguments it already raises a MethodError which is the correct error to throw and has a perfect error message already, there's no need to catch that and return an ArgumentError?

```julia
julia> foo(args...) = 42
foo (generic function with 1 method)

julia> foo(; x = 3)
ERROR: MethodError: no method matching foo(; x::Int64)
This method may not support any kwargs.

Closest candidates are:
  foo(Any...) got unsupported keyword argument "x"
   @ Main REPL[30]:1

Stacktrace:
 [1] top-level scope
   @ REPL[31]:1
```

src/rules/misc.jl
```diff
@@ -39,6 +39,105 @@
...
+
+function frule!!(
+    ::Dual{typeof(Core.kwcall)},
+    ::Dual{<:NamedTuple},
+    ::Dual{typeof(stop_gradient)},
+    args::Dual...,
+)
+    throw(ArgumentError("stop_gradient does not support keyword arguments"))
+end
```
penelopeysm: Again this is unnecessary

src/rules/misc.jl
```diff
@@ -39,6 +39,105 @@
...
+
+function rrule!!(
+    ::CoDual{typeof(Core.kwcall)},
+    ::CoDual{<:NamedTuple},
+    ::CoDual{typeof(stop_gradient)},
+    args::CoDual...,
+)
+    throw(ArgumentError("stop_gradient does not support keyword arguments"))
+end
```
penelopeysm: Unnecessary

test/rules/misc.jl
```diff
@@ -1,4 +1,36 @@
...
+            sum(Mooncake.stop_gradient(x, y)[1]) + Mooncake.stop_gradient(x, y)[2]
+        c_multi = Mooncake.prepare_gradient_cache(f_multi, x, 4.0)
+        _, (_, gx, gy) = Mooncake.value_and_gradient!!(c_multi, f_multi, x, 4.0)
+        @test iszero(gx)
+        @test iszero(gy)
+
+        # Kwargs throw in both primal and AD contexts.
+        @test_throws ArgumentError Mooncake.stop_gradient(1.0; kw=1)
+        f_kw(x) = sum(Mooncake.stop_gradient(x; kw=1))
+        @test_throws ArgumentError Mooncake.prepare_gradient_cache(f_kw, [1.0, 2.0])
```
penelopeysm: Unnecessary

penelopeysm:
Claude can write code, but I'm afraid it's not writing _good_ code here!

src/rules/misc.jl
```diff
@@ -39,6 +39,105 @@
...
+# Intercept kwarg calls without generating a positional fallback (which would overwrite
+# the method above). Defining directly on Core.kwcall avoids that conflict.
+function Core.kwcall(::NamedTuple, ::typeof(stop_gradient), args...)
+    throw(ArgumentError("stop_gradient does not support keyword arguments"))
+end
+
+@is_primitive MinimalCtx Tuple{typeof(stop_gradient),Vararg{Any}}
+@is_primitive MinimalCtx Tuple{
+    typeof(Core.kwcall),NamedTuple,typeof(stop_gradient),Vararg{Any}
+}
```
penelopeysm:
```suggestion
```

this too is not needed

yebai:
> Claude can write code, but I'm afraid it's not writing good code here!

Thanks, Penny!  Claude is doing the *right* thing -- I am the one to blame for the unnecessary kwcall rules above. 

EDIT: the `rrule!!` for `kwcall` is needed due to the way reverse mode works -- otherwise, the error throwing doesn't work as intended.

yebai:
Removed the `Core.kwcall` method, its `@is_primitive`, and its `rrule!!` — the native `MethodError` is cleaner. Also added `rrule!!`/`frule!!` for `Core.throw_methoderror` and `Core.throw_inexacterror` so the error propagates correctly through AD.

Red-Portal:
@yebai Hi Hong, sorry for the delay. I am a bit out of bandwidth until next week. Sorry!

src/rules/builtins.jl
```diff
@@ -1078,6 +1078,22 @@ end
 frule!!(::Dual{typeof(throw)}, args::Dual...) = throw(map(primal, args)...)
 rrule!!(::CoDual{typeof(throw)}, args::CoDual...) = throw(map(primal, args)...)
 
+@static if isdefined(Core, :throw_methoderror)
```
penelopeysm, line 1084:
```suggestion
# Only defined in v1.12+
@static if isdefined(Core, :throw_methoderror)
```

test/rules/misc.jl
```diff
@@ -1,4 +1,39 @@
...
+        _, (_, gx_tuple, gy_tuple) = Mooncake.value_and_gradient!!(c_tuple, f_tuple, x, y)
+        @test iszero(gx_tuple)
+        @test iszero(gy_tuple)
+
+        # Keyword arguments are rejected with a MethodError.
+        @test_throws MethodError Mooncake.stop_gradient(1.0; kw=1)
+        @static if isdefined(Core, :throw_methoderror)
+            f_kw(x) = sum(Mooncake.stop_gradient(x; kw=1))
+            @test_throws MethodError Mooncake.prepare_gradient_cache(f_kw, [1.0, 2.0])
+        end
```
penelopeysm: I think this test is now an anachronism. The rules for those are more generic than for `stop_gradient`, they handle all kinds of InexactError and MethodError. I think it would be better to test this in `test/rules/builtins.jl` instead, and to include a test for the InexactError case as well. In particular, you don't need to test this specifically with this particular function `Mooncake.stop_gradient` as that suggests that there's some special behaviour for this function. You could make up your own functions

```julia
using Mooncake

f(x, y) = 3; g(x) = f(x); Mooncake.prepare_gradient_cache(g, [1.0])

h(x) = Int(x[1]) + 0.5; Mooncake.prepare_gradient_cache(h, [1.5])
```

src/rules/builtins.jl
```diff
@@ -1078,6 +1078,22 @@ end
...
+    function frule!!(::Dual{typeof(Core.throw_methoderror)}, args::Dual...)
+        Core.throw_methoderror(map(primal, args)...)
+    end
+    function rrule!!(::CoDual{typeof(Core.throw_methoderror)}, args::CoDual...)
+        Core.throw_methoderror(map(primal, args)...)
+    end
```
penelopeysm, line 1093: Currently on 1.11, missing methods will cause a MooncakeRuleCompilationError to be thrown here:

https://github.com/chalk-lab/Mooncake.jl/blob/498fdee57202f15e98acd52acddff276c67ec91a/src/interpreter/ir_utils.jl#L265-L278

On 1.12, prior to this PR, missing methods throw MissingRuleForBuiltinException (which was different from 1.11). This now changes it to throw a MethodError (which is still different from 1.11). I'm okay with that, but it might be worth thinking about whether you want to unify the behaviour across 1.11 and 1.12 instead.

yebai:
Moved the `MethodError`/`InexactError` propagation tests from `test/rules/misc.jl` to `test/rules/builtins.jl` using generic functions (no reference to `stop_gradient`). The `MethodError` case is still guarded by `@static if isdefined(Core, :throw_methoderror)` since that builtin is 1.12+ only — leaving the 1.11 vs 1.12 behaviour difference as-is.

penelopeysm:
I don't see the new changes - did you push them?

yebai:
Here you go. Obviously, I forgot to push.

penelopeysm:
Thanks, looks good!

src/rules/builtins.jl
```diff
@@ -1034,7 +1034,12 @@ function rrule!!(
     end
 end
 
-@generated is_homogeneous_and_immutable(::P) where {P<:Tuple} = allequal(fieldtypes(P))
+# `@generated is_homogeneous_and_immutable(::P) where {P<:Tuple} = allequal(fieldtypes(P))`
+# Emitting a literal Bool is semantically equivalent here, but gives the compiler a simpler,
+# allocation-free result.
+@generated function is_homogeneous_and_immutable(::P) where {P<:Tuple}
+    return allequal(fieldtypes(P)) ? :(true) : :(false)
+end
```
yebai: The allocation behaviour differs between Mac and CI, so revert this. 

```suggestion
@generated is_homogeneous_and_immutable(::P) where {P<:Tuple} = allequal(fieldtypes(P))
```

src/interpreter/reverse_mode.jl
```diff
@@ -1957,7 +1957,10 @@ function rule_type(interp::MooncakeInterpreter{C}, sig_or_mi; debug_mode) where
     fwd_args_type = Tuple{map(fcodual_type, arg_types)...}
     fwd_return_type = forwards_ret_type(ir)
     Trdata_return = rdata_type(tangent_type(Treturn))
-    pb_args_type = Trdata_return === Union{} ? Union{} : Tuple{Trdata_return}
+    # `Trdata_return === Union{} ? Union{} : Tuple{Trdata_return}`
+    # For non-returning primals, Tuple{} means a zero-argument pullback; Union{} would
+    # instead mean there is no possible argument value.
+    pb_args_type = Trdata_return === Union{} ? Tuple{} : Tuple{Trdata_return}
```
yebai, line 1962: Likely quite subtle bug triggered by new test cases of builtins in this PR.

yebai:
@penelopeysm, can you help take a look at the remaining allocation issue? I have tried a few fixes, but none have worked. This is likely caused by the new tests directly targeting built-in functions. 

Alternatively, we could revert to testing at a higher level as before, which seems to sidestep this allocation issue.

penelopeysm:
> can you help take a look at the remaining allocation issue

Does it fail on main? Is it related to the `Union{}` change, if not what caused it to happen?

yebai:
This PR started as a new feature, but has evolved into a detailed interface and test design that is taking up too much time. 

@AstitvaAggarwal, can you help take over fixing the remaining issues?

AstitvaAggarwal:
yeah

yebai:
#1093 and #1094 fixed the allocation failures—they were caused by inaccurate allocation measurements from the `Base.@allocations` macro.

PR #1082, opened 2026-03-18
feat: native HVP interface and LazyFoRRule caching

yebai:
## Summary

- Adds `prepare_hvp_cache` / `value_and_hvp!!` — a native forward-over-reverse HVP interface. The outer forward-mode rule differentiates through gradient construction; `HVPCache` holds the compiled state.
- Introduces `LazyFoRRule{Trule,Tfwd,Trvs}` (formerly `FoRCache`): the frule for `build_derived_rrule` in forward-over-reverse mode. It lazily compiles the inner `DerivedRule` and its dual callables on the first call, then reuses them via `_copy` + tangent reconstruction on every subsequent call. The three type parameters are inferred at rule-construction time via a `@generated` helper (`__build_primitive_frule`) so there is no virtual dispatch on the hot path.
- `build_primitive_frule` is now called through `modify_fwd_ad_stmts!` and `build_frule` (instead of hardcoding `frule!!`), with `safe_for_literal` determining whether the rule is inlined or captured. This generalises the primitive frule dispatch consistently with the reverse-mode path.
- Adds `_copy` for `Tangent` (needed to deep-copy the cached rule on each HVP call).


src/rules/high_order_derivative_patches.jl
```diff
@@ -8,63 +8,117 @@
...
 
-    cache_key = ClosureCacheKey(interp.world, (sig_or_mi, debug_mode))
-    if haskey(cache.cache, cache_key)
-        rule, fwd_dual_callable, rvs_dual_callable = cache.cache[cache_key]
-        new_rule = _copy(rule)
+    # Cache hit: reuse compiled rule + dual callables. `debug_mode` is not re-checked
+    # because build_primitive_frule returns one LazyFoRRule per call site in the
+    # compiled IR, and every call through that site uses the same config (and therefore
+    # the same debug_mode) for the lifetime of the closure.
+    if isdefined(cache, :rule)
```
Technici4n: How safe is it to skip checking the signature? In principle this rule could be used for varying signatures.

src/rules/high_order_derivative_patches.jl
```diff
@@ -8,63 +8,117 @@
...
+    rvs_dual_callable::Trvs
+    LazyFoRRule{Trule,Tfwd,Trvs}() where {Trule,Tfwd,Trvs} = new()
 end
 
-struct FoRCache
-    cache::Dict{ClosureCacheKey,Any}
+@generated function __build_primitive_frule(
+    sig::Type{<:Tuple{typeof(build_derived_rrule),MooncakeInterpreter{C},SMI,S,Bool}}
+) where {C,SMI,S}
+    Trule = Core.Compiler.return_type(
```
Technici4n, line 55: I am not very familiar with this function but it sounds potentially brittle. Do we really need this? This function should only be called once during the preparation of the outer derivative.

src/interface.jl
```diff
@@ -744,3 +744,108 @@ function value_and_derivative!!(
...
+end
+
+"""
+    value_and_hvp!!(cache::HVPCache, v, x)
+    value_and_hvp!!(cache::HVPCache, vs::Tuple, x1, x2, ...)
+
+Given a cache prepared by [`prepare_hvp_cache`](@ref), compute the gradient of `f` at
+`x...` and the Hessian-vector product `H v`.
+
+**Single argument:** `v` is the tangent direction; returns `(∇f(x), H(x)v)`. For
```
Technici4n: It would be nice to return `f(x)` as well.

src/rules/high_order_derivative_patches.jl
yebai: This is already used in multiple places of the code base, eg, 

https://github.com/chalk-lab/Mooncake.jl/blob/498fdee57202f15e98acd52acddff276c67ec91a/src/interpreter/reverse_mode.jl#L920

yebai:
Thanks for the review!

- **Signature check (line 89):** Added a comment clarifying that `sig_or_mi` is also invariant per call site — each `LazyFoRRule` lives inside the closure for a fixed `grad_f`, so both the signature and config are constant for the lifetime of the closure.

- **`Core.Compiler.return_type` (line 34):** Needed for type stability — without it we fall back to `LazyFoRRule{Any,Any,Any}`, which causes virtual dispatch on `cache.rule`/`cache.fwd_dual_callable`/`cache.rvs_dual_callable` on every `value_and_hvp!!` call. The cost is paid once at outer frule compilation time, which is the right trade-off.

- **Return `f(x)` (line 817):** Done — `value_and_hvp!!` now returns `(f(x), gradient, hvp)`. Threaded through `grad_f` via `value_and_gradient!!` with no extra primal call.

- **`LazyFoRRule{Any,Any,Any}` caching bug (fdcb630):** Fixed the DI second-order test failure where multiple `LazyDerivedRule` instances (e.g. `collect` and `num_to_vec`) shared one call site via `build_rrule`'s `@nospecialize`, causing the single-slot `LazyFoRRule{Any,Any,Any}` cache to serve the wrong rule. Replaced with `DynamicFoRRule` — a `Dict`-keyed cache on `(sig, debug_mode)`. Also extracted `_for_rule_cached_dual` and `_compile_for_rule` as shared helpers to eliminate the duplication between `LazyFoRRule` and `DynamicFoRRule`.

PR #1085, opened 2026-03-19
fix(1.10): guard DebugRRule invalid CoDual specializations 

yebai:
This PR fixes a Julia 1.10 segfault in reverse-mode debug checking by guarding DebugRRule against invalid CoDual specialisations at compile time. Instead of letting Julia codegen reach the problematic OpaqueClosure path, we now fail early with a regular input-type error.

The concrete reproducer is a standalone reverse-mode debug example:

- build a debug rrule for `f = x -> 5x`
- call it as `rule(zero_fcodual(f), CoDual(5.0, 1.0))`

On Julia 1.10, the main segfaulted on that call. With this fix, it throws the expected input-type error instead.

This PR also adds that exact reproducer to the debug-mode tests and re-enables the previously disabled DispatchDoctor integration coverage, including debug_mode.jl and interface.jl. On this branch, the re-enabled DispatchDoctor suite passes on Julia 1.10.

This is a follow-up to the earlier work in PR #808. 

 What changed

  - Add a Julia 1.10-specific @generated guard for DebugRRule in src/debug_mode.jl
  - Reject invalid CoDual specialisations before codegen emits the crashing call path
  - Add the exact reverse-debug reproducer to test/debug_mode.jl
  - Re-enable the full DispatchDoctor integration coverage for this area

  Validation

  - On main, the standalone reverse-mode debug reproducer above segfaulted on Julia 1.10
  - On this branch, the same call throws the expected error instead
  - The re-enabled DispatchDoctor integration completed successfully on Julia 1.10-1.12

Discovered while working on #1082 

This PR is produced with help from Codex.

### DIFF
--- a/src/debug_mode.jl
+++ b/src/debug_mode.jl
@@ -44,7 +44,7 @@ Apply pre- and post-condition type checking. See [`DebugFRule`](@ref).
     # NOTE: @generated alone (without the type check) does NOT prevent the segfault -
     # returning an unconditional quote generates the same code as a plain function.
     # The early-return on mismatch is the critical part.
-    
+    #
     # See https://github.com/JuliaLang/julia/issues/61368
     @generated function (rule::DebugFRule{Trule})(x::Vararg{Dual,N}) where {Trule,N}
         # First, check tangent type consistency for all Dual inputs at compile time.


src/debug_mode.jl
```diff
@@ -20,15 +20,30 @@ end
...
 @static if VERSION < v"1.11-"
-    # On Julia 1.10, use @generated to check types at compile time, preventing the
-    # compiler from ever seeing rule.rule(x...) with mismatched types, which would
-    # cause a segfault (JuliaLang/julia#51016).
+    # On Julia 1.10, we encounter a segfault when an OpaqueClosure/MistyClosure is
+    # called with a specialised signature whose declared return type disagrees with
+    # what the IR actually returns (JuliaLang/julia#51016). Using @generated we
+    # check tangent types at compile time; for a bad specialisation we return
+    # :(error(...)) so the compiler never generates code for rule.rule(x...) with
+    # mismatched types. For well-typed calls the normal body is emitted.
```
sunxd3, line 42: add a ref to https://github.com/JuliaLang/julia/issues/61368 so we can remove this if/when backpoting happen

yebai: These are likely useful mechanisms for robustness for other cases, too. I think we should keep it even after https://github.com/JuliaLang/julia/issues/61368

In general, it is very challenging to navigate the Julia compiler landscape -- anything that can go wrong eventually does.

src/debug_mode.jl
```diff
@@ -20,15 +20,30 @@ end
...
+    #
+    # NOTE: @generated alone (without the type check) does NOT prevent the segfault -
+    # returning an unconditional quote generates the same code as a plain function.
+    # The early-return on mismatch is the critical part.
```
yebai, line 46:
```suggestion
    # The early-return on mismatch is the critical part.
    
    # See https://github.com/JuliaLang/julia/issues/61368
```

src/debug_mode.jl
```diff
@@ -217,11 +232,56 @@ _copy(x::P) where {P<:DebugRRule} = P(_copy(x.rule))
...
-@noinline function (rule::DebugRRule)(x::Vararg{CoDual,N}) where {N}
-    verify_fwds_inputs(rule.rule, x)
-    y, pb = rule.rule(x...)
-    verify_fwds_output(x, y)
-    return y::CoDual, DebugPullback(pb, primal(y), map(primal, x))
+@static if VERSION < v"1.11-"
+    # Julia 1.10 can segfault while codegen'ing an OpaqueClosure call for an invalid CoDual
+    # specialization before these runtime debug checks execute. Reject bad specializations at
+    # compile time so the compiler never generates `rule.rule(x...)` for them.
+    @generated function (rule::DebugRRule{Trule})(x::Vararg{CoDual,N}) where {Trule,N}
```
yebai, line 243:
```suggestion
    #
    # See https://github.com/JuliaLang/julia/issues/61368
    @generated function (rule::DebugRRule{Trule})(x::Vararg{CoDual,N}) where {Trule,N}
```

Issue #1084, opened 2026-03-18
Parameters overwritten on gpu


yolhan83:
Hello, just tryng out the new GPU support, wanted to point out an issue with friendly_tangents on gpu 
```julia
using DifferentiationInterface,CUDA
using Lux, LuxCUDA
using Random
import Mooncake
const device = cu
const rng = Random.default_rng(123)
function foo(ps,m,x,y)
    sum(abs2,m(x,ps).-y)/length(x)
end

model = Chain(
    Dense(1=>32,tanh),
    Dense(32=>32,tanh),
    Dense(32=>32,tanh),
    Dense(32=>10)
)
ps,st = Lux.setup(rng,model) 
ps = ps |> device
m = Lux.StatefulLuxLayer(model,nothing,device(st))

x = rand(Float32,1,200) |> device 
y = rand(Float32,10,200) |> device
foo(ps, m, x,y)

ps2 = deepcopy(ps)

g_moon = gradient(
    foo,
    AutoMooncake(;config = nothing),
    ps,
    Constant(m),
    Constant(x),
    Constant(y)
)

ps == ps2 # true

g_moon2 = gradient(
    foo,
    AutoMooncake(;config = Mooncake.Config(; friendly_tangents=true)),
    ps,
    Constant(m),
    Constant(x),
    Constant(y)
)

ps == ps2 # false
```
it seems like the parameters gets overwritten I also get 
```julia
g_moon2 == ps # true
```
and `g_moon2`  is actually wrong. 

ps : `g_moon` is good.

Also, little thing a little out of subject, but even making ps a ComponentArray doesn't help, I didn't find a way to use `gradient!` (and the same bug appears)

gdalle:
Hi, this may be an issue with the way either DI or Mooncake handles friendly tangents. Can you try to reduce the example to eliminate Lux?

yolhan83:
this reproduce : 
```julia
using DifferentiationInterface, CUDA
import Mooncake
p = (; a = cu([1.0,2.0]))
foo(p) = sum(abs2,p.a)
p_copy = deepcopy(p)
g_moon = gradient(foo,AutoMooncake(;config=nothing),p)
p == p_copy # true
g_moon2 = gradient(foo,AutoMooncake(;config=Mooncake.Config(;friendly_tangents=true)),p)
p == p_copy # false : Corupted
g_moon # good
g_moon2 # good (but may be lucky, was bad in the Lux case)
```
[...]

Issue #937, opened 2026-01-10
`friendly_tangents` can show misleading view for structured matrix wrappers


sunxd3:
@willtebbutt brought up this potential issue.

When using `friendly_tangents=true`, the `tangent_to_primal!!` conversion wraps the tangent's underlying matrix data directly into a `Symmetric` wrapper. Since the internal tangent data is not symmetric (only one triangle holds the gradient), the conversion produces misleading output where the gradient appears in both triangles.

```julia
using LinearAlgebra, Mooncake

f(S::Symmetric) = S[1, 2]
S = Symmetric([1.0 2.0; 2.0 4.0])
rule = Mooncake.build_rrule(f, S)
```

```julia
julia> _, grad_raw = Mooncake.value_and_pullback!!(rule, 1.0, f, S)
(2.0, (Mooncake.NoTangent(), Mooncake.Tangent{@NamedTuple{data::Matrix{Float64}, uplo::Mooncake.NoTangent}}((data = [0.0 1.0; 0.0 0.0], uplo = Mooncake.NoTangent()))))

julia> _, grad_friendly = Mooncake.value_and_pullback!!(rule, 1.0, f, S; friendly_tangents=true)
(2.0, (f, [0.0 1.0; 1.0 0.0]))

julia> grad_friendly[2]
2×2 Symmetric{Float64, Matrix{Float64}}:
 0.0  1.0
 1.0  0.0
```

## Problem

The raw tangent correctly stores `1.0` only at `[1,2]` (the stored triangle). When `tangent_to_primal!!` wraps this in `Symmetric`, the `[2,1]` entry is mirrored from `[1,2]`, making the gradient appear as `1.0` in both positions.

This is misleading: the user sees a symmetric gradient `[0 1; 1 0]` when the actual gradient information is only in the upper triangle `[0 1; 0 0]`.

## Round-trip behavior

The round-trip itself is correct because `primal_to_tangent!!` reads from the underlying storage, not the `Symmetric` view:

```julia
using LinearAlgebra, Mooncake

S = Symmetric([1.0 2.0; 2.0 4.0])
t_original = Mooncake.zero_tangent(S)
t_original.fields.data[1,2] = 1.0
```

```julia
julia> p = Mooncake.tangent_to_primal!!(deepcopy(S), t_original)
2×2 Symmetric{Float64, Matrix{Float64}}:
 0.0  1.0
 1.0  0.0

julia> p.data
2×2 Matrix{Float64}:
[...]

PR #1087, opened 2026-03-19
Mooncake copy_to_output functions support

### LINKED ISSUE #865: I don't trust `_copy_output` / `_copy_to_output` for all primal types
These probably need to be overwritten for custom types like `CuArray`. I think the current implementation would be problematic when differentiating through a function that returns a `CuArray`, but I am not 100% sure.

AstitvaAggarwal:
fix #865 

### Documentation Preview

Mooncake.jl documentation for PR #1087 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1087/

### Performance

Performance Ratio:
```
│                   sum_1000 │ 130.0 ns │     1.77 │        1.78 │     1.0 │        4.93 │   1.85 │
│                  _sum_1000 │  1.07 μs │     6.71 │        1.02 │  5570.0 │        34.1 │   1.08 │
│               sum_sin_1000 │  7.44 μs │     2.38 │        1.17 │    1.72 │        9.04 │   1.85 │
│              _sum_sin_1000 │  5.13 μs │     3.23 │        2.52 │   342.0 │        14.5 │   2.77 │
│                   kron_sum │ 215.0 μs │     12.1 │        3.43 │    7.32 │       471.0 │   17.6 │
│              kron_view_sum │ 261.0 μs │     11.9 │        4.83 │    17.5 │       455.0 │   15.4 │
│      naive_map_sin_cos_exp │  2.33 μs │     2.32 │        1.43 │ missing │        6.99 │   2.03 │
│            map_sin_cos_exp │  2.31 μs │     2.76 │        1.54 │    2.57 │         5.8 │   2.57 │
│      broadcast_sin_cos_exp │  2.47 μs │     2.52 │        1.45 │    5.29 │        1.36 │   1.95 │
│                 simple_mlp │ 357.0 μs │     4.73 │        2.83 │     1.6 │        8.46 │   3.26 │
│                     gp_lml │ 396.0 μs │     5.78 │         1.9 │    2.63 │     missing │   2.66 │
│ turing_broadcast_benchmark │  2.05 ms │     4.07 │        3.34 │ missing │        49.2 │   1.78 │
│         large_single_block │ 450.0 ns │      7.7 │        1.98 │  3840.0 │        29.2 │   2.16 │
```

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -418,20 +418,10 @@ function _copy_output(x::P) where {P}
     # Overload _copy_output to customise.
     nf == 0 && return x
 
-    try
-        if ismutable(x)
-            return _copy_output_mutable_cartesian(x, Val(nf))
-        else
-            return _copy_output_immutable_cartesian(x, Val(nf))
-        end
-    catch e
-        throw(
-            ArgumentError(
-                "Mooncake.jl does not currently have a method `_copy_output` for type $P. " *
-                "Please open an issue or define a custom `_copy_output` method for this type. " *
-                "Original error: $e",
-            ),
-        )
+    if ismutable(x)
+        _copy_output_mutable_cartesian(x, Val(nf))
+    else
+        _copy_output_immutable_cartesian(x, Val(nf))
     end
 end
 


ext/MooncakeCUDAExt/MooncakeCUDAExt.jl
```diff
@@ -86,6 +88,12 @@ include("ndual.jl")
 const CuFloatArray = CuArray{<:IEEEFloat}
 const CuComplexArray = CuArray{<:Complex{<:IEEEFloat}}
 const CuMaybeComplexArray = Union{CuFloatArray,CuComplexArray}
+Mooncake._copy_output(x::CuMaybeComplexArray) = copy(x)
```
yebai: Does this fix #1084? If so, I suggest we separate it into a smaller bugfix PR.

src/interface.jl
```diff
@@ -418,10 +418,20 @@ function _copy_output(x::P) where {P}
     # Overload _copy_output to customise.
     nf == 0 && return x
 
-    if ismutable(x)
-        _copy_output_mutable_cartesian(x, Val(nf))
-    else
-        _copy_output_immutable_cartesian(x, Val(nf))
+    try
```
yebai: The current approach appears somewhat brittle.

A more robust design would be to define a generic fallback method for `_copy_output` that raises an error by default. This method can then be selectively overloaded for a defined set of concrete types that are covered by Mooncake’s test suite.

For these supported cases, it is reasonable to dispatch on abstract types to minimise the number of required method implementations. While this introduces some risk—since not all subtypes of an abstract type may be explicitly tested—it represents a pragmatic and acceptable trade-off between coverage and maintainability.

src/test_utils.jl
```diff
@@ -1306,6 +1306,19 @@ function _test_tangent_interface(rng::AbstractRNG, p::P; interface_only=false) w
...
+    # Skip if p is not a valid output (e.g. contains Ptr, circular reference, or alias) else error.
+    try
+        Mooncake.__exclude_unsupported_output(p)
+        p_copy = Mooncake._copy_output(p)
+        @test has_equal_data(p, p_copy)
+        @test typeof(p_copy) == P
+        p_inplace = Mooncake._copy_to_output!!(p_copy, p)
+        @test has_equal_data(p, p_inplace)
+    catch e
+        e isa Mooncake.ValueAndPullbackReturnTypeError || rethrow(e)
```
yebai: What about other errors? Can we at least print a warning message if they are ignored?

test/tangents/tangents.jl
```diff
@@ -207,6 +207,83 @@ using DispatchDoctor: allow_unstable
         @test Mooncake.require_tangent_cache(P) == Val(expected_result)
     end
 
+    # Structs for _copy_output / _copy_to_output!! regression test cases
```
yebai: `test/interface.jl` is a more natural home for these tests.

src/interface.jl
AstitvaAggarwal: Im not sure i understand but the current generic fallback handles arbitrary structs by recursing into fields via `@generated` `_copy_output_immutable_cartesian`. if any field type is unsupported, it errors naturally via a `MethodError`.

yebai: The current generic fallback will fail silently for cases like https://github.com/chalk-lab/Mooncake.jl/pull/1101.  

What I mean is that friendly_tangents / `_copy_to_output!!` should be limited to a set of known types (including collections that consist of only known types, eg, Tuple, NamedTuple, Dict), and forbid structs that are not tested by Mooncake. This would help resolve #937, as we can handle it as a special case by overloading `_copy_output/_copy_to_output!!`. 

cc @penelopeysm @Technici4n

AstitvaAggarwal: how about calling `__exclude_unsupported_output` inside the copy function? we can lower the real call to an internal copy call, but the outer function weeds out unsupported objects that error out silently. This way we dont lose coverage, and any unsupported object will just need a new outer dispatch (like how I did for `CuArray`)

AstitvaAggarwal: for the issue 937, we mostly just needs a `Symmetric, Hermitian etc` dispatch for `tangent_to_primal_internal!!`

AstitvaAggarwal: added a PR for the above also. see #1102

src/test_utils.jl
```diff
@@ -1289,6 +1289,23 @@ function _test_tangent_interface(rng::AbstractRNG, p::P; interface_only=false) w
...
+        @test typeof(p_copy) == P
+        p_inplace = Mooncake._copy_to_output!!(p_copy, p)
+        @test has_equal_data(p, p_inplace)
+    catch e
+        if e isa Mooncake.ValueAndPullbackReturnTypeError
+            @warn "_copy_output/_copy_to_output!! test skipped for type $P: $e"
+        else
+            rethrow(e)
+        end
+    end
```
yebai, line 1308: This is a bit non-standard / not good enough. Can we add a utility `_is_unsupported_output(p)` to skip tests if not supported instead of a try-catch block? 


```julia
__is_unsupported_output(p; T=typeof(p)) = any(t -> t <: Nothing, Base.return_types(Mooncake.__exclude_unsupported_output, (T,)))
```

yebai:
This PR adds a few more tests to identify issues with `copy_output!!`. 

The motivation is to generalise tangent copying to handle arbitrary tangent types -- as noted in #1103, this is challenging because it requires supporting all Julia types, reminiscent of defining a new deepcopy method that works in all cases. 

Closing this, as it is unlikely to yield a robust solution even with significant effort.

PR #1093, opened 2026-03-21
refactor: simplify count_allocs using @generated

yebai:
## Summary

- Replaces the eval-loop of 22 explicit numbered methods in `count_allocs` with a single `@generated` function over `Vararg{Any,N}`.
- A companion `Tcount_allocs{F,Args}` struct with an `@noinline` callable replaces the old `@noinline` closure barrier (`@generated` bodies cannot contain closures).
- Removes the `@static VERSION >= v"1.12-"` guards at all call sites in `test_frule_performance` / `test_rrule_performance`.
- Replaces `@allocations` call sites in `test/rules/builtins.jl` and `test/rules/memory.jl` with `TestUtils.count_allocs`.

The motivation is that a plain `Vararg` method does not fully specialise on `Type{T}` arguments — they widen to `DataType` — causing spurious allocations to be reported. This was first noticed in #1079. A `@generated` function specialises fully on every argument type (including `Type{T}`), giving the correct zero-allocation count without needing explicit numbered overloads. Lumping `f` and its arguments into one `Vararg` (rather than a separate leading `f::F`) handles type constructors as the first argument without a special-case overload.

**MWE showing the problem with a plain `Vararg` method:**
```julia
vararg(f::F, x::Vararg{Any,N}) where {F,N} = nothing
vararg(rand, Xoshiro(1), Float64)
Base.specializations(@which vararg(rand, Xoshiro(1), Float64))
# => MethodInstance for vararg(::typeof(rand), ::Xoshiro, ::Type)
#                                                          ^^^^^^ widened — abstract dispatch

explicit(f::F, x1::X1, x2::X2) where {F,X1,X2} = nothing
explicit(rand, Xoshiro(1), Float64)
Base.specializations(@which explicit(rand, Xoshiro(1), Float64))
# => MethodInstance for explicit(::typeof(rand), ::Xoshiro, ::Type{Float64})
#                                                            ^^^^^^^^^^^^^^^ preserved
```

### DIFF
--- a/test/interface.jl
+++ b/test/interface.jl
@@ -1,5 +1,4 @@
 using Mooncake: TestUtils
-using Mooncake.TestUtils: has_equal_data
 using Mooncake:
     prepare_gradient_cache,
     prepare_pullback_cache,
@@ -158,8 +157,8 @@ end
 
             cache = Mooncake.prepare_pullback_cache(testf, x)
             v, pb = Mooncake.value_and_pullback!!(cache, x̄_unfriendly, testf, x)
-            @test has_equal_data(v, SimplePair(x.x1^2 + sin(x.x2), x.x1 * x.x2))
-            @test has_equal_data(
+            @test TestUtils.has_equal_data(v, SimplePair(x.x1^2 + sin(x.x2), x.x1 * x.x2))
+            @test TestUtils.has_equal_data(
                 pb[2],
                 Mooncake.Tangent((;
                     x1=2x.x1 * x̄.x1 + x.x2 * x̄.x2, x2=cos(x.x2) * x̄.x1 + x.x1 * x̄.x2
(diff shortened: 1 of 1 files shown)


src/test_utils.jl
```diff
@@ -1448,69 +1431,47 @@ function test_get_tangent_field_performance(t::Union{MutableTangent,Tangent})
...
-            # nonetheless
-            N > __MAX_ARGS_ALLOCS &&
-                @warn "using varargs method for `count_allocs` since there were $N arguments; this may lead to spurious allocations being reported if any arguments are `DataType`s"
-            stats = Base.gc_num()
-            @noinline clos = () -> f(x...)
-            clos()
-            diff = Base.GC_Diff(Base.gc_num(), stats)
-            return Base.gc_alloc_count(diff)
-        end
+# A plain Vararg method does not fully specialise on Type{T} arguments on Julia >= 1.12 —
```
penelopeysm: this is true pre 1.12 as well.

src/test_utils.jl
```diff
@@ -1448,69 +1431,47 @@ function test_get_tangent_field_performance(t::Union{MutableTangent,Tangent})
...
-    function count_allocs(f::F, x::Vararg{Any,N}) where {F,N}
-        test_hook(count_allocs, f, x...) do
-            @allocations f(x...)
-        end
+end
+# Needs a special case when `f` itself is a type constructor.
+@generated function count_allocs(::Type{F}, x::Vararg{Any,N}) where {F,N}
+    args = [:(x[$i]) for i in 1:N]
+    quote
+        test_hook(_MeasureAllocs(F, ($(args...),)), count_allocs, F, $(args...))
```
penelopeysm: since generated functions cause it to specialise (claude is right here), it seems like you could probably simplify this even further by lumping `f` and `x` into the same varargs, so just do

```julia
@generated function count_allocs(fx::Vararg{Any,N}) where {N}
   ...
end
```

essentially you are using the generated function to fix the lack of specialisation on `x`, but i think you could also use it to address the lack of specialisation on `f`

yebai:
@penelopeysm, excellent catches -- both fixed.

Issue #1097, opened 2026-03-22
build_rrule hangs on MTK-generated ODEProblem types (DynamicDerivedRule infinite type inference)


ChrisRackauckas-Claude:
## Summary

`Mooncake.build_rrule` succeeds for the top-level function, but execution hangs indefinitely during lazy sub-rule compilation (`DynamicDerivedRule`) when Mooncake encounters ModelingToolkit-generated ODEProblem types. The hang is in `abstract_call_gf_by_type` → `typeinf_edge` cycles in Mooncake's abstract interpretation.

This blocks using Mooncake as the outer AD for MTK-based models in SciMLSensitivity.jl (see SciML/SciMLSensitivity.jl#1358).

## MWE

```julia
# Mooncake v0.5.23, ModelingToolkit v11.17.0, OrdinaryDiffEq v6.108.0, Julia 1.11.9

using ModelingToolkit
using ModelingToolkit: t_nounits as t, D_nounits as D
using OrdinaryDiffEq
using SciMLStructures: Tunable, replace
using Mooncake

@parameters k
@variables X(t)
@named sys = System([D(X) ~ -k * X], t)
sys = mtkcompile(sys)

prob = ODEProblem(sys, [X => 1.0], (0.0, 1.0), [k => 0.5])

function loss(pval)
    new_p = replace(Tunable(), prob.p, pval)
    sol = solve(prob, Tsit5(); p = new_p, saveat = 0.1)
    return sum(reduce(hcat, sol.u))
end

println("Primal: ", loss([0.5]))  # works

# build_rrule succeeds (~15s):
rule = Mooncake.build_rrule(loss, [0.5])

# But execution hangs here during DynamicDerivedRule lazy compilation:
Mooncake.value_and_pullback!!(rule, 1.0, loss, [0.5])  # hangs
```

## Behavior

- **Without MTK** (plain `ODEProblem{true}(f!, u0, tspan, p)`): Works correctly, gradients match ForwardDiff.
- **With MTK** (even simple `D(X) ~ -k*X`): `build_rrule` succeeds in ~15s, but `value_and_pullback!!` hangs.
- **With MTK + DAE initialization** (algebraic constraints + guesses): Same hang, but the types are even more complex.

## Stack trace (from SIGTERM)

The hang is in deeply nested `abstract_call_gf_by_type` → `typeinf_edge` → `abstract_call_method` cycles:

```
[...]

PR #1115, opened 2026-03-28
prune recursion for faster rule building, usage

AstitvaAggarwal:
## Problem

`Mooncake.value_and_pullback!!` hangs indefinitely when differentiating functions that call SciML's `solve`:

```julia
using ModelingToolkit
using ModelingToolkit: t_nounits as t, D_nounits as D
using OrdinaryDiffEq
using SciMLStructures: Tunable, replace
using Mooncake

@parameters k
@variables X(t)
@named sys = System([D(X) ~ -k * X], t)
sys = mtkcompile(sys)
prob = ODEProblem(sys, [X => 1.0], (0.0, 1.0), [k => 0.5])

function loss(pval)
    new_p = replace(Tunable(), prob.p, pval)
    sol = solve(prob, Tsit5(); p = new_p, saveat = 0.1)
    return sum(reduce(hcat, sol.u))
end

println("Primal: ", loss([0.5]))
rule = Mooncake.build_rrule(loss, [0.5])               # completes quickly
Mooncake.value_and_pullback!!(rule, 1.0, loss, [0.5])  # hangs indefinitely
```

**Initial `build_rrule` completes quickly.** `prob` is non-const so Julia types it as `Any`, `solve` gets `DynamicDerivedRule`, and Mooncake never recurses into `solve`'s body at compile time.

**`value_and_pullback!!` hangs.** `DynamicDerivedRule` fires at runtime with concrete types, triggering a second `build_rrule` for the concrete `solve` signature. This is where the hang occurs.

---

## Root Cause

`build_rrule` has two phases:
- **Phase 1, Abstract interpretation via `MooncakeInterpreter`** (hang occurs here)
- **Phase 2, IR transformation**

### Ordering Bug in `abstract_call_gf_by_type`

The original code called `@invoke` **before** checking whether the call was a Mooncake primitive:

```julia
# ORIGINAL, @invoke fires first for every call
ret = @invoke CC.abstract_call_gf_by_type(interp::CC.AbstractInterpreter, ...)
# primitive check runs AFTER @invoke has already recursed into the body
any_prim = any_matches_primitive(applicable, C, M, interp.world)
```

### How the Hang Unfolds

Julia's IR lowering transforms `solve(prob, alg; kwargs...)`:

```
solve(prob, alg; kwargs...)
  -> Core.kwcall(nt, solve, prob, alg)    # Julia's lowering
  -> dispatches into var"#solve#28"       # solve's kw-body
  -> Core.kwcall(nt, solve_up, ...)       # registered Mooncake primitive, inside solve#28
```

`Core.kwcall(nt, solve_up, ...)` is a registered Mooncake primitive, so the primitive check would catch it immediately. But `@invoke` fires first, dispatching into `var"##solve_up#29"` before the check can fire.

Phase 1 obtains the IR of each function and walks its call statements. For each call statement, `abstract_call_gf_by_type` is fired, type information is inferred from the IR, and the process repeats recursively into the callee's IR. Since `MooncakeInterpreter` is repassed into `@invoke`, every call site at every level fires the override again, which fires `@invoke` again:

```
inference sees solve_up's body:
  -> call site 1: get_concrete_problem(...)                 # @invoke, recurse into body
      -> call site 1.1: get_updated_symbolic_problem(...)       # @invoke, recurse
          -> call site 1.1.1: ...                               # @invoke, recurse deeper
  -> call site 2: some_primitive(...)                       # @invoke, recurse into body
      -> ...                                                # primitive check never fires in time
```

For **non-primitives**, recursing into their bodies is necessary, as Mooncake needs to find inner primitives. For **primitives**, it is pure waste, since the primitive check would have caught them without seeing their internals. Since `@invoke` fires first, both are treated identically.

### Role of `@nospecialize` (Amplifier)

Inside `solve_up`'s body, `@nospecialize` is used extensively across packages like `ModelingToolkitBase` to reduce TTFX, permanently widening `ODEFunction`'s type parameters to abstract types. Two kinds of call sites result:

- **`Any(Any...)` calls,** where `find_method_matches` returns `FailedMethodMatch` immediately. O(1) per site, fast but numerous.
- **`concrete_f(Any...)` calls,** where Julia finds applicable methods and `@invoke` recurses into each body. These drive the counter growth and make N large.

### Complexity Analysis

Let:
- **N** = total unique `(method, specialization)` pairs in `solve_up`'s transitive call tree
- **D** = maximum depth of the call chain

**Time: O(N).** Each unique call site is visited once due to Julia's inference cache. N is enormous, touching most of the SciML stack and Julia stdlib. `@nospecialize`-induced `Any` types prevent branch pruning, increasing N further. 

**Space: O(D).** Each active inference frame is heap-allocated in Julia 1.12's stackless inference (`CC.Future`). D is unbounded, so heap usage grows without limit.

**Julia 1.12 stackless inference.** Prior to 1.12, the same recursion hit the stack size limit and raised a `StackOverflowError`. Julia 1.12's `CC.Future`-based stackless inference moved frames to the heap, removing the depth limit. The same O(N) traversal with O(D) heap now runs with no bound and no crash, resulting in a silent, large hang.

### Verification

```julia
const CALL_COUNTER = Ref(0)
# incremented at the top of abstract_call_gf_by_type
```

| Operation | Calls to `abstract_call_gf_by_type` | Outcome |
|-----------|--------------------------------------|---------|
| `build_rrule(loss, [0.5])` | 348 | Completes, `solve` gets `DynamicDerivedRule` (`prob::Any`) |
| Runtime `build_rrule` via `DynamicDerivedRule` inside `value_and_pullback!!` | 15,000+ before interrupt, still growing | Hangs, concrete `solve` sig triggers full O(N) traversal |

The 15,000+ calls are all within a single `build_rrule` invocation, one O(N) traversal of `solve_up`'s transitive call tree entirely within Phase 1.

---

## Fix

Check `any_matches_primitive` **before** any inference. If primitive, use `CC.NativeInterpreter` instead of `MooncakeInterpreter`:

```julia
# NEW, primitive check fires first (pure signature check, O(k), no recursion)
any_prim = any_matches_primitive(applicable, C, M, interp.world)
if any_prim
    native_interp = CC.NativeInterpreter(interp.world)
    ret = CC.abstract_call_gf_by_type(native_interp, f, arginfo, si, atype, sv, max_methods)
    # ... wrap with noinline_callmeta ...
end
# Not a primitive, safe to recurse with MooncakeInterpreter as before
@invoke CC.abstract_call_gf_by_type(interp::CC.AbstractInterpreter, ...)
```

- `any_matches_primitive` is a pure method table signature check, O(k), no body recursion
- `NativeInterpreter` does not repass `MooncakeInterpreter`, so recursion into the primitive's body uses Julia's standard inference with built-in termination mechanisms
- Full `CallMeta` (return type, exception type, effects) is still obtained from real inference, preserving downstream type propagation
- Non-primitive calls fall through to `@invoke` with `MooncakeInterpreter` as before, so inner primitives are still caught

---

## General Efficiency Improvement

This fix is not limited to the SciML hang case. Primitives are now detected first and inferred with `NativeInterpreter`, avoiding all unnecessary body recursion across any differentiated function. This makes `build_rrule` faster in general and preserves concrete return types for primitive calls, keeping downstream IR types concrete and avoiding `Mooncake.DynamicDerivedRule` cascades at rule-call time.

### Documentation Preview

Mooncake.jl documentation for PR #1115 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1115/

### Performance

Performance Ratio:
```
│                   sum_1000 │ 309.0 ns │      1.6 │        1.61 │   0.372 │         1.3 │   2.72 │
│                  _sum_1000 │  1.19 μs │     7.87 │        1.16 │  1330.0 │        23.6 │   1.44 │
│               sum_sin_1000 │  5.76 μs │     5.15 │        4.52 │    2.41 │        11.9 │   2.27 │
│              _sum_sin_1000 │  6.24 μs │     3.99 │        3.26 │   238.0 │        11.2 │   1.89 │
│                   kron_sum │ 366.0 μs │     8.09 │        2.68 │    15.9 │       222.0 │   12.3 │
│              kron_view_sum │ 385.0 μs │     8.52 │        3.99 │    11.5 │       243.0 │   8.84 │
│      naive_map_sin_cos_exp │  2.83 μs │     2.54 │        1.37 │ missing │        6.05 │   1.83 │
│            map_sin_cos_exp │  2.92 μs │      3.0 │         2.9 │    1.84 │        4.35 │   2.14 │
│      broadcast_sin_cos_exp │  3.03 μs │     2.72 │        2.29 │    2.98 │        1.09 │   1.91 │
│                 simple_mlp │ 441.0 μs │     4.25 │        2.71 │    1.64 │        7.76 │    3.0 │
│                     gp_lml │ 317.0 μs │     7.24 │        2.19 │    7.78 │     missing │   3.72 │
│ turing_broadcast_benchmark │  1.91 ms │     4.52 │        3.01 │ missing │        24.7 │    1.9 │
│         large_single_block │ 488.0 ns │     5.91 │        2.11 │  3620.0 │        22.7 │   1.85 │
```


AstitvaAggarwal:
@yebai This closes https://github.com/chalk-lab/Mooncake.jl/issues/1097 and the connected Mooncake issues.

src/interpreter/abstract_interpretation.jl
```diff
@@ -163,9 +153,18 @@ function Core.Compiler.abstract_call_gf_by_type(
...
         any_prim = any_matches_primitive(applicable, C, M, interp.world)
         if any_prim
+            # Use NativeInterpreter instead of MooncakeInterpreter to infer the primitive call.
+            # NativeInterpreter does not repass MooncakeInterpreter into inner calls, so
+            # recursion into the primitive's body is bounded and terminates. We still get the
+            # correct return type, exception type, and effects from full inference.
+            native_interp = CC.NativeInterpreter(interp.world)
+            ret = CC.abstract_call_gf_by_type(
+                native_interp, f, arginfo, si, atype, sv, max_methods
+            )
```
sunxd3: Two things worth calling out about this swap:

**Overlay table.** `MooncakeInterpreter` is wired to `OverlayMethodTable(interp.world, mooncake_method_table)` (see `CC.method_table` on L104). `NativeInterpreter` uses the default method table, so if any primitive's body dispatches to a `@mooncake_overlay`, inference will see the original method instead. This may not bite today, but it's a latent correctness issue — worth a comment explaining why it's safe for the primitive path to drop the overlay table, or an assertion that no current primitive body depends on overlays.

**Shared `sv`.** The `sv::AbsIntState` was created by `MooncakeInterpreter` — it holds interpreter-specific state (cache refs, edge tracking). Passing it to `NativeInterpreter` means two interpreters share the same inference frame. This works today because `CC.abstract_call_gf_by_type` mostly reads `sv` for type info rather than mutating interpreter-coupled state, but it's fragile across Julia versions. A short comment noting this assumption would help future maintainers.

src/interpreter/abstract_interpretation.jl
```diff
@@ -181,7 +180,17 @@ function Core.Compiler.abstract_call_gf_by_type(
...
+    # this Mooncake abstract_call_gf_by_type dispatch again, so any primitive encountered deeper in the call tree is also caught.
+    @invoke CC.abstract_call_gf_by_type(
+        interp::CC.AbstractInterpreter,
+        f::Any,
+        arginfo::CC.ArgInfo,
+        si::CC.StmtInfo,
+        atype::Any,
+        sv::CC.AbsIntState,
+        max_methods::Int,
+    )
```
sunxd3: Nit: "Not a primitives method's IR hence" reads awkwardly — maybe:

```suggestion
    # Not a primitive — recurse with MooncakeInterpreter so that any primitive
    # encountered deeper in the call tree is caught by the check above.
    @invoke CC.abstract_call_gf_by_type(
        interp::CC.AbstractInterpreter,
        f::Any,
        arginfo::CC.ArgInfo,
        si::CC.StmtInfo,
        atype::Any,
        sv::CC.AbsIntState,
        max_methods::Int,
    )
```

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.5.24"
+version = "0.5.25"
```
sunxd3, line 4: be careful about releasing

sunxd3 (changes requested):
I think the fix is good. The PR description makes sense, the code comments less so.

Could you improve the inline comments to describe the problem and add a ref to this PR? 

I think the comment needs to justify why NativeInterpreter is correct and safe. I think the comment in the last block is probably not necessary.

In general, thanks for looking into this, it must have been tricky to nail down.

src/interpreter/abstract_interpretation.jl
```diff
@@ -163,9 +154,19 @@ function Core.Compiler.abstract_call_gf_by_type(
...
     if !isa(matches, Core.Compiler.FailedMethodMatch)
         (; valid_worlds, applicable) = matches
-        # For all applicable method matches, we need to check if any of them could hit a primitive
+        # For applicable method matches in IR, we need to check if any of them is a primitive.
         any_prim = any_matches_primitive(applicable, C, M, interp.world)
         if any_prim
+            # Use NativeInterpreter instead of MooncakeInterpreter to infer primitive calls.
+            # This is efficient and safe because:
+            # 1. Primitive's Julia IR does not need to be searched further for primitives/inferred return type via MooncakeInterpreter.
+            # 2. NativeInterpreter's inference is sufficient to determine the return types.
```
sunxd3: `ret` is of type `CallMeta`, so "return types" is not entirely accurate

src/interpreter/abstract_interpretation.jl
```diff
@@ -163,9 +154,19 @@ function Core.Compiler.abstract_call_gf_by_type(
...
+            # 3. By not recursing with MooncakeInterpreter, we avoid unbounded recursion into primitive implementations.
```
sunxd3: I feel like the recursion was still bounded, just a lot. Or am I wrong?

src/interpreter/abstract_interpretation.jl
AstitvaAggarwal: return types mentioned here is a field of `CallMeta`, refering to the function's return type itself.

src/interpreter/abstract_interpretation.jl
AstitvaAggarwal: it was unbounded theoretically because it recurses first without checking the base condition. Now we check the condition first, so if there is a unbounded primal as long as it is declared a primitive (Mooncake cares only about this) Mooncake works fine independant of primal behaviour

yebai:
One thing to check: it appears to break documentation building, which may hint at some subtle behaviour changes. This may not mean this PR has issues, but it needs to be understood before merging.

sunxd3 (commented):
Could you give a short justification of why the ID change is benign (even I think they are). Also, I am not entirely satisfied with the code comments, maybe work with Claude to write these?

AstitvaAggarwal:
oh yeah, i was coming back to this. The doctest failure is caused by the PR's internal refactoring shifting all `BBCode` ID values by +12. IDs in `BBCode` are assigned from a global counter - every call to `ID()` increments it - so the absolute numeric values of IDs in any given `BBCode` depend on how many `ID()` calls have occurred before it is constructed. This fix adds method lookup earlier in `abstract_call_gf_by_type`, resulting in 12 additional ID() calls  before the doctest's `BBCode(ir)` is reached, pushing all the hardcoded expected values in the documentation.
e.g.  `(ID(96) → ID(108), ID(100) → ID(112), etc.)`, so we update those expected values to match.

src/interpreter/abstract_interpretation.jl
```diff
@@ -163,9 +154,21 @@ function Core.Compiler.abstract_call_gf_by_type(
...
         any_prim = any_matches_primitive(applicable, C, M, interp.world)
         if any_prim
+            # A primitive has a hand-written rrule!! - Mooncake never needs to inspect its body.
+            # Using NativeInterpreter here is safe because:
+            #   - return types (present in the returned CallMeta) are still inferred correctly by standard Julia inference
+            #   - inlining/const-folding is blocked below via noinline_callmeta, so the primitive
+            #     call is preserved in the caller's IR for Mooncake to dispatch its rrule!! at runtime
+            # Using MooncakeInterpreter would recurse into the primitive's entire call tree
+            # looking for further primitives - work that is pure waste, and unbounded for large
+            # packages like SciML. See PR #1115 for the full analysis.
```
yebai:
```suggestion
              # A primitive already has a hand-written `rrule!!`, so Mooncake does not need
              # to inspect its body when differentiating. The only thing we need here is the
              # ordinary `CallMeta` for the call site, especially the inferred return type.
              #
              # We therefore ask `NativeInterpreter` for the `CallMeta`. This avoids recursing
              # through the callee IR using Mooncake's primitive-search logic:
              # `MooncakeInterpreter` would walk nested calls in that body, check them for
              # primitives, and continue that search down the callee tree. That extra work is
              # unnecessary for a primitive with a hand-written rule.
              #
              # `noinline_callmeta` below then blocks inlining/const-folding so the primitive
              # call stays in the caller IR and Mooncake can dispatch its `rrule!!` at runtime.
              # See PR #1115 for more discussion.
```

sunxd3 (approved):

src/interpreter/abstract_interpretation.jl
```diff
@@ -163,9 +154,26 @@ function Core.Compiler.abstract_call_gf_by_type(
...
+            # We therefore ask `NativeInterpreter` for the `CallMeta`. This avoids recursing
+            # through the callee IR using Mooncake's primitive-search logic:
+            # `MooncakeInterpreter` would walk nested calls in that body, check them for
+            # primitives, and continue that search down the callee tree. That extra work is
+            # unnecessary for a primitive with a hand-written rule.
+            #
+            # `noinline_callmeta` below then blocks inlining/const-folding so the primitive
+            # call stays in the caller IR and Mooncake can dispatch its `rrule!!` at runtime.
+            # See PR #1115 for more discussion.
+            native_interp = CC.NativeInterpreter(interp.world)
```
yebai, line 173: @sunxd3 / @AstitvaAggarwal, can you take a look at why this regressed when overlayed methods were used? Please 

1. Create an MWE explaining why this PR caused regression on Mooncake's overlay
2. Open an issue for discussions

PR #1131, opened 2026-04-07
clear Mooncake OpaqueClosure Caches

AstitvaAggarwal:
This PR closes #1116 by adding `clear_mooncake_caches!(; gc::Bool=false)` as a public API function that clears all Julia-level caches `Mooncake` holds globally.               
                                                                                           
**Problem** :                                           

- When `Mooncake` differentiates a function for the first time, it compiles a reverse-mode rule and stores it in `GLOBAL_INTERPRETERS` across three caches : `oc_cache` (compiled `OpaqueClosures`), `code_cache` (`CodeInstance` objects), and `inf_cache` (`InferenceResult` objects). These caches never shrink on their own.         
                                                                                  
- In long-running `Mooncake` sessions we end up differentiating many distinct function signatures - these accumulate indefinitely. (The Discourse thread linked in #1116 reported the user hitting ~4.5 GB from Mooncake's caches with no way to reclaim it)                                                          
 
**Solution :**

- Our function empties `oc_cache`, `code_cache.dict`, and `inf_cache` in-place for both `ForwardMode` and `ReverseMode` interpreters. After calling it, `Mooncake` re-derives rules from scratch on the next use, and the `GC` can reclaim the now-unreferenced `DerivedRule`, `CodeInstance`, and `InferenceResult` objects.                
                                                                                           
- The `gc` keyword (default false) optionally runs `GC.gc(true)` immediately after clearing for users who want memory reclaimed right away rather than at the next `GC` cycle.                                                           
                                                                                          
**What is exactly freed :** 

Clearing the caches makes the Julia-level objects (`DerivedRule`, `CodeInstance`, `InferenceResult`) `GC` eligible. The JIT-compiled native machine code in main memory is allocated by LLVM outside the GC heap and cannot be reclaimed from user code and remains for the lifetime of the process (at least as of Julia v1.12), to clear this we must restart the REPL instance itself.

### Documentation Preview

Mooncake.jl documentation for PR #1131 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1131/

### Performance

Performance Ratio:
```
│                   sum_1000 │ 171.0 ns │     1.64 │        1.64 │   0.702 │        3.46 │   6.56 │
│                  _sum_1000 │ 952.0 ns │     7.08 │        1.01 │  4650.0 │        38.4 │   1.08 │
│               sum_sin_1000 │  7.18 μs │     3.55 │        1.44 │     1.6 │         9.8 │   1.78 │
│              _sum_sin_1000 │  6.02 μs │     3.84 │        2.02 │   259.0 │        11.6 │   2.19 │
│                   kron_sum │ 203.0 μs │     11.7 │        3.21 │    24.0 │       354.0 │   21.0 │
│              kron_view_sum │ 283.0 μs │     10.9 │        4.75 │    13.6 │       274.0 │    7.8 │
│      naive_map_sin_cos_exp │  2.64 μs │     2.66 │        1.36 │ missing │        6.04 │   1.91 │
│            map_sin_cos_exp │   2.6 μs │     3.07 │        1.45 │    1.95 │        5.13 │   2.44 │
│      broadcast_sin_cos_exp │   2.6 μs │     2.86 │        1.38 │     4.1 │        1.29 │   1.96 │
│                 simple_mlp │ 234.0 μs │     5.24 │        2.22 │    1.36 │        10.8 │   4.32 │
│                     gp_lml │ 206.0 μs │     12.1 │        3.59 │    12.4 │     missing │   8.14 │
│ turing_broadcast_benchmark │  2.17 ms │     4.61 │        2.84 │ missing │        25.0 │   1.85 │
│         large_single_block │ 401.0 ns │      5.4 │        1.97 │  4310.0 │        29.8 │   2.15 │
```


src/config.jl
```diff
@@ -5,6 +5,8 @@
         friendly_tangents::Bool=false,
         chunk_size::Union{Nothing,Int}=nothing,
         enable_nfwd::Bool=true,
+        empty_cache::Bool=false,
+        run_gc::Bool=false,
```
yebai: Let's drop `run_gc`, since one can always run `GC.gc(true)` themselves. 
```suggestion
```

test/interpreter/abstract_interpretation.jl
```diff
@@ -124,4 +124,53 @@ end
             @test grad[2] == [2.0, 0.0, 1.0]
         end
     end
+
+    @testset "Config(empty_cache=true)" begin
```
yebai, line 128: It looks like fairly excessive tests for this simple feature. Can you find the minimally necessary tests?

AstitvaAggarwal: addressed this and the above review

yebai (approved):

PR #1144, opened 2026-04-13
Fix ci failures

AstitvaAggarwal:
### Documentation Preview

Mooncake.jl documentation for PR #1144 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1144/

### Performance

_Pending. No performance comment found yet._


test/ext/dynamic_expressions/dynamic_expressions.jl
```diff
@@ -136,7 +136,7 @@ end
 end
 
 @testset "Use in DynamicExpressions + Optim optimization" begin
-    let
+    @test_broken let
```
yebai: Do you understand why this is broken? Let's try to fix it if it is a Mooncake issue.

AstitvaAggarwal: We are trying to use `Optim`, the transitive dependancy is : 
`Optim`->`LineSearches`->`NLSolversBase`->`DifferentiationInterface`. 
So since `DI` bounds `Mooncake`, this testset is broken for (`Mooncake >v0.5.25`)

gdalle: For future reference, the upper bound on Mooncake was introduced by @wsmoses in https://github.com/JuliaRegistries/General/pull/152662 because of the breaking Mooncake changes listed in https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/986, which had already caused issues in packages such as https://github.com/JuliaGeometry/Meshes.jl/pull/1324

PR #1153, opened 2026-04-20
Add cached value_and_jacobian!! interface.

### LINKED ISSUE #1145: Add `value_and_jacobian`
It is currently missing but helpful for Bijectors.jl, see, [here](https://github.com/TuringLang/Bijectors.jl/blob/e4d8585a2479de60708af2a0dea8ff929fb1df81/ext/BijectorsMooncakeExt.jl#L45-L66).

yebai:
Fix https://github.com/chalk-lab/Mooncake.jl/issues/1145

### Documentation Preview

Mooncake.jl documentation for PR #1153 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1153/

### Performance

Performance Ratio:
```
│              sum_1000 │ 180.0 ns │     1.56 │        1.62 │   0.667 │        3.23 │   6.23 │
│             _sum_1000 │ 962.0 ns │     6.84 │        1.03 │  1800.0 │        43.5 │   1.07 │
│          sum_sin_1000 │  7.19 μs │     3.52 │        1.35 │    1.48 │        11.2 │   1.78 │
│         _sum_sin_1000 │  6.02 μs │     3.61 │        1.89 │   248.0 │        13.9 │   2.17 │
│              kron_sum │ 214.0 μs │     12.9 │        3.27 │    16.6 │       348.0 │   18.8 │
│         kron_view_sum │ 295.0 μs │     10.8 │        5.05 │    25.5 │       278.0 │   7.07 │
│ naive_map_sin_cos_exp │  2.51 μs │     2.87 │         1.5 │ missing │        7.17 │   1.97 │
│       map_sin_cos_exp │  2.56 μs │     3.37 │        1.53 │    1.59 │        5.99 │   2.42 │
│ broadcast_sin_cos_exp │  2.49 μs │     3.02 │        1.52 │    4.44 │        1.38 │   2.01 │
│            simple_mlp │ 379.0 μs │     4.49 │        2.75 │    1.94 │        9.02 │   2.71 │
│                gp_lml │ 223.0 μs │     9.06 │        2.17 │    5.56 │     missing │   4.73 │
│    large_single_block │ 421.0 ns │     7.09 │        2.16 │  4560.0 │        40.0 │   2.05 │
```

### DIFF
--- a/docs/src/developer_documentation/ir_representation.md
+++ b/docs/src/developer_documentation/ir_representation.md
@@ -231,7 +231,7 @@ For example, consider
 julia> using Mooncake.BasicBlockCode: ID # to improve printing
 
 julia> bb_ir.blocks[3].insts[1]
-Compiler.NewInstruction(:(Base.add_int(ID(109), 1)), Int64, Compiler.NoCallInfo(), (0, 0, 0), 0x00002478)
+Compiler.NewInstruction(:(Base.add_int(ID(2), 1)), Int64, Compiler.NoCallInfo(), (0, 0, 0), 0x00002478)
 ```
 This is the first instruction of the third basic block.
 The first field is a call to `Base.add_int`, the second field is `Int64` (we promise that the other fields are just copies of the corresponding data from the `Core.Compiler.InstructionStream` in the original `IRCode` representation of this IR).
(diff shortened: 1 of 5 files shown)


yebai:
@penelopeysm, can you help take a look at this unblocks Bijectors switching to https://github.com/TuringLang/AbstractPPL.jl/pull/155. Many thanks!

penelopeysm:
I am somewhat confused: what does Bijectors have to do with AbstractPPL and why should Bijectors be using differentiation functionality that is in AbstractPPL? I assume that the point of this APPL stuff is to define ways to differentiate a probabilistic model. After all, that is what LogDensityFunction is responsible for. So, even though I still can't say that I agree with this direction, I can at least understand the rationale for making _DynamicPPL_ use it. But Bijectors doesn't define probabilistic models, so I don't really understand why the AbstractPPL PR is useful for Bijectors. If you are really dead set on removing DI from everything, then it seems to me that the solution in Bijectors is to just directly call `ForwardDiff.jacobian` etc.

yebai:
The dependency is already in place: Bijectors.jl depends on AbstractPPL.jl, likely for reasons related to density interfaces (see [`Project.toml`](https://github.com/TuringLang/Bijectors.jl/blob/ca29f081a1543da49cb78416184fe25daf986d77/Project.toml#L6)).

The pull request [AbstractPPL.jl PR #155](https://github.com/TuringLang/AbstractPPL.jl/pull/155) introduces the `ADProblems` interface. This interface is not specific to probabilistic models; it is designed to be broadly applicable to differentiable “problems” in general, with use cases like Bijectors considered from the outset. `ADProblems` could be in its own lightweight package principle, but AbstractPPL appears to be a good starting point for its initial implementation.

penelopeysm:
Yes, the dep is there because I added it in the VectorBijectors release (it depends on the optics defined in AbstractPPL). Okay, I see. In that case, then sure, I'm sure that this can be used there. I would probably recommend splitting it from AbstractPPL because if we have lots of AbstractPPLADBackendExt extensions precompilation can get quite annoying (and also AbstractPPL is a quite low-level dependency so it gets pulled into many different things).

PR #1170, opened 2026-05-07
Honor `mooncake_overlay` in primitive inference

### LINKED ISSUE #1169: `@mooncake_overlay` usage for or within primitives are silently ignored
> Note :  This is a regression introduced in #1115, the PR correctly uses `NativeInterpreter` for primitive call sites to prevent `MooncakeInterpreter` from recursing into primitive bodies. However `NativeInterpreter` is overlay-blind and it only sees the standard Julia method table - so when a primitive has a `@mooncake_overlay` that changes its return type, the inferred return type at the call site is wrong. 

> PR #1168 fixes this by replacing `NativeInterpreter` with `OverlayAwareNativeInterpreter`, which delegates everything to `NativeInterpreter` except method_table, which it redirects to `Mooncake`'s overlay table. This preserves the PR #1115 guarantee (no recursion into primitive bodies) while being overlay-aware.

When a primitive has a `@mooncake_overlay` that changes its return type, Mooncake infers the wrong return type for that call site. Downstream dispatch then compiles against the original type rather than the overlaid type, producing wrong gradients.
   
When `MooncakeInterpreter` encounters a primitive call site, it uses `NativeInterpreter` to get the `CallMeta` (return type, effects). `NativeInterpreter` is overlay-blind, i.e it only sees the standard Julia method table, not Mooncake's overlay table. So even though a `mooncake_overlay` exists and fires correctly at runtime, the inferred return type is from the original method body.                                                                                                                                                                          
                                                               
MWE :
```
  using Mooncake
  using Mooncake: CoDual, zero_fcodual, NoPullback, MinimalCtx, @is_primitive, @mooncake_overlay, NoRData
                                                                                                                                                                                        
  struct TypeA end
  struct TypeB end                                                                                                                                                                      
                                                               
  switch_type(x::TypeA) = x
  @mooncake_overlay switch_type(x::TypeA) = TypeB()
                                                                                                                                                                                        
  @is_primitive MinimalCtx Tuple{typeof(switch_type), TypeA}
  function Mooncake.rrule!!(f::CoDual{typeof(switch_type)}, x::CoDual{TypeA})                                                                                                           
      return zero_fcodual(TypeB()), NoPullback(f, x)           
  end

  use_b(::TypeA, x::Float64) = x        # gradient 1x
  use_b(::TypeB, x::Float64) = 2x       # gradient 2x
                                                                                                                                                                                        
  @is_primitive MinimalCtx Tuple{typeof(use_b), TypeA, Float64}
  function Mooncake.rrule!!(f::CoDual{typeof(use_b)}, ::CoDual{TypeA}, x::CoDual{Float64})                                                                                              
      pb(dy) = NoRData(), NoRData(), dy                                                                                                                                                 
      return zero_fcodual(Mooncake.primal(x)), pb
  end                                                                                                                                                                                   
                                                               
  @is_primitive MinimalCtx Tuple{typeof(use_b), TypeB, Float64}                                                                                                                         
  function Mooncake.rrule!!(f::CoDual{typeof(use_b)}, ::CoDual{TypeB}, x::CoDual{Float64})
      pb(dy) = NoRData(), NoRData(), 2dy                                                                                                                                                
      return zero_fcodual(2.0 * Mooncake.primal(x)), pb                                                                                                                                 
  end
                                                                                                                                                                                        
  function outer(x::Float64)                                   
      b = switch_type(TypeA())
      return use_b(b, x)
  end                                                                                                                                                                                   
   
  cache = Mooncake.prepare_gradient_cache(outer, 1.0)                                                                                                                                   
  val, (_, grad) = Mooncake.value_and_gradient!!(cache, outer, 1.0)

  # Expected: val=2.0, grad=2.0 (TypeB path via overlay)                                                                                                                                
  # Actual without fix: val=1.0, grad=1.0 (TypeA path, overlay ignored)
  @assert val == 2.0                                                                                                                                                                    
  @assert grad == 2.0                                          
```                                                                                                                                                          
   
This affects SciMLBase's Mooncake extension. `DiffEqBase.solve` calls `set_mooncakeoriginator_if_mooncake(ChainRulesOriginator())` to detect which AD system is active at runtime. The Mooncake extension registers an overlay that returns `MooncakeOriginator()` instead, this needed to make this context-dependent. Without correct inference, downstream `sensealg` dispatch compiles against `ChainRulesOriginator` instead of `MooncakeOriginator`.
                                                               
  References:
  - SciMLBase overlay: https://github.com/SciML/SciMLBase.jl/blob/b5e72aaa9304b451157a9a9849c2318f6af4d81e/ext/SciMLBaseMooncakeExt.jl#L35
  - DiffEqBase call site: https://github.com/SciML/DiffEqBase.jl/blob/e99aba76a9598255761240e4dc6490dc578f84f2/src/solve.jl#L590                                                        
  - Fix proposed in PR #1168

yebai:
`MooncakeInterpreter` previously routed every primitive call site through `NativeInterpreter` for `CallMeta`, but `NativeInterpreter` is overlay-blind: when a primitive has a `@mooncake_overlay` that changes its return type, the inferred type at the call site was wrong, and downstream dispatch compiled against the original type. Detect overlay matches via `Method.external_mt` and route them through `@invoke` instead, keeping the `NativeInterpreter` fast path for non-overlay primitives. 

Fixes #1169. Alternative to #1168. 

### Documentation Preview

Mooncake.jl documentation for PR #1170 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1170/

### Performance

Performance Ratio:
```
│              sum_1000 │ 190.0 ns │     1.53 │        1.53 │   0.737 │        3.43 │   6.64 │
│             _sum_1000 │  1.08 μs │     6.72 │        1.03 │  4700.0 │        35.5 │   1.07 │
│          sum_sin_1000 │  7.42 μs │     2.57 │        1.11 │    1.65 │        11.0 │   1.77 │
│         _sum_sin_1000 │  4.64 μs │     3.81 │        2.63 │   338.0 │        17.6 │   3.05 │
│              kron_sum │ 193.0 μs │     13.1 │        3.31 │    10.7 │       520.0 │   13.7 │
│         kron_view_sum │ 263.0 μs │     12.9 │        5.27 │    20.9 │       450.0 │    7.2 │
│ naive_map_sin_cos_exp │   2.2 μs │     2.81 │        1.55 │ missing │        8.48 │   2.17 │
│       map_sin_cos_exp │  2.15 μs │     3.35 │        1.66 │     1.6 │        7.34 │   2.73 │
│ broadcast_sin_cos_exp │  2.31 μs │     2.93 │        1.52 │    4.25 │        1.39 │   2.06 │
│            simple_mlp │ 344.0 μs │     5.04 │        2.91 │    1.61 │        9.67 │   3.21 │
│                gp_lml │ 360.0 μs │     6.17 │        1.79 │    2.76 │     missing │   3.18 │
│    large_single_block │ 471.0 ns │     5.94 │        1.96 │  4190.0 │        31.8 │   2.08 │
```


sunxd3:
On this issue, I slightly lean towards #1168 because indeed it's more general and the complexity diff w.r.t this PR is not too bad. A minor issue with the current code is the use of `external_mt` — not blocking, but less ideal in terms of interfacing with julia internals.

src/interpreter/abstract_interpretation.jl
```diff
@@ -216,6 +231,16 @@ function any_matches_primitive(applicable, C, M, world)
     false
 end
 
+function any_matches_overlay(applicable)
+    for app in applicable
+        method = VERSION < v"1.12-" ? app.method : app.match.method
+        if isdefined(method, :external_mt) && method.external_mt !== nothing
```
yebai: @sunxd3, this is not fundamental -- `any_matches_overlay` can be refactored in a way more friendly to Julia compiler internals.

sunxd3: maybe
```suggestion
        if isdefined(method, :external_mt) && method.external_mt === mooncake_method_table
```
to be a bit more precise?

src/interpreter/abstract_interpretation.jl
```diff
@@ -204,12 +219,19 @@ end
 
 function any_matches_primitive(applicable, C, M, world)
```
yebai, line 220: Note that there is already `any_matches_primitive` defined, which accesses Julia internals in similar ways.

sunxd3:
Thanks @yebai for the follow-up.

After some thought and investigation, I now lean toward this PR over #1168 for the scope of issue #1169. Use of `external_mt`, IMO, is fine and a small price to pay; the alternatives seem worse as far as I can tell.

Before merging, there is one particular case I want to think through. Maybe it is not that important, but I would like to be clear.

---

Comparing on four versions of Mooncake:

```text
pre-#1115                 4e2df06c0  c7c30f259^
main                      b79d67e1e
#1168                     ac5a35f72
#1170                     a0f11465e
```

on

```julia
using Mooncake

struct A end
struct B end

helper(::A) = A()
Mooncake.@mooncake_overlay helper(::A) = B()

primitive_wrapper(x::A) = helper(x)
Mooncake.@is_primitive Mooncake.DefaultCtx Mooncake.ReverseMode Tuple{
    typeof(primitive_wrapper), A
}

caller() = primitive_wrapper(A())
caller_arg(x::A) = primitive_wrapper(x)
```

```julia
interp = Mooncake.MooncakeInterpreter(Mooncake.DefaultCtx, Mooncake.ReverseMode)

for sig in (
    Tuple{typeof(primitive_wrapper), A},
    Tuple{typeof(caller)},
    Tuple{typeof(caller_arg), A},
)
    ir, rt = Base.code_ircode_by_type(sig; interp)[1]
    println(sig, " => ", rt)
end
```

Results:

| Version | `primitive_wrapper(::A)` | `caller()` | `caller_arg(::A)` |
|---|---:|---:|---:|
| pre-#1115 `4e2df06c0` | `B` | `B` | `B` |
| `#1168` `ac5a35f72` | `B` | `B` | `B` |
| main `b79d67e1e` | `B` | `A` | `A` |
| `#1170` `a0f11465e` | `B` | `A` | `A` |

---

This result indicates that #1115 introduced a behavior change, and #1168 reverted it. @AstitvaAggarwal and I had an offline chat and we'll dig deeper on this.

sunxd3:
@AstitvaAggarwal I added a developer doc on primitives and overlay, can you give it a read and let me know your thoughts?

AstitvaAggarwal:
looks good, dont think anything is incorrect.

sunxd3:
@yebai could you give this a look when you have some time? The plan is to merge this one and leave checking overlay output type consistency for the future.

yebai:
Let's discuss when we next meet.

PR #1172, opened 2026-05-12
rule reuse consistency checks

### LINKED ISSUE #681: Test suite extension: check that cached rules run correctly
In the DI test suite, each operator is tested like this:

```julia
(; f, x, grad, x0) = scenario
@test gradient(f, backend, x) == grad  # without prep
prep = prepare_gradient(f, backend, x0)  # prepare on different input
@test gradient(f, prep, backend, x) == grad  # first test with prep
@test gradient(f, prep, backend, x) == grad  # second test with prep
```
Preparing on a different input is essential to avoid data leakage, it has allowed me to uncover some nasty bugs (https://github.com/JuliaDiff/FiniteDiff.jl/issues/185). But here what I had in mind is this second prepared test, to make sure that the preparation result is not invalidated or destroyed by a first differentiation call. One example might be a pullback closure which is invalidated after one reverse pass in the case of mutated arguments (https://github.com/JuliaDiff/DifferentiationInterface.jl/discussions/678).

_Originally posted by @gdalle in https://github.com/chalk-lab/Mooncake.jl/pull/389#discussion_r2252001181_

AstitvaAggarwal:
closes #681.

When a rule is compiled via `build_frule`/`build_rrule`, the resulting closure is intended to be reusable: compile once, call many times. However, there was no test enforcing this. A rule whose closure captures mutable state that gets corrupted on a first call would pass all existing tests but silently return wrong results on any subsequent use, which is exactly the  class of bug described in #681.

This PR adds a "Reuse" testset inside `Mooncake.TestUtils.test_rule` that runs two complete forward+backward cycles with the same compiled rule and asserts the results are identical. Crucially, "Reuse" runs before "Interface (1)" and "Correctness"  testsets so it exercises the true first and second calls on a fresh compiled rule. If a rule corrupts its internal state on the first call, the second call will produce a different primal output or different cotangents and "Reuse" will fail.

The check is implemented via two dedicated functions `test_frule_reuse` and `test_rrule_reuse`. Both functions construct two independent sets of inputs from deepcopies and run the same rule twice. `test_frule_reuse` compares primal outputs and output/input tangents. `test_rrule_reuse` compares primal outputs, `rdata` from the pullback, and `fdata` accumulated in the input buffer. No finite differences are involved, so the check is fast and has no step-size or domain-constraint sensitivity.

`test/test_utils.jl` adds two negative tests using real `@is_primitive` + `frule!!`/`rrule!!` registrations with a deliberately broken rule (a counter is incremented during each call, causing a different primal on the second call) to demonstrate that the reuse check catches this class of bug.

### Documentation Preview

Mooncake.jl documentation for PR #1172 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1172/

### Performance

Performance Ratio:
```
│              sum_1000 │ 210.0 ns │     1.29 │        1.48 │   0.619 │         3.2 │   6.06 │
│             _sum_1000 │   1.1 μs │     6.04 │        1.02 │  3520.0 │        35.2 │   1.06 │
│          sum_sin_1000 │  7.43 μs │     2.56 │        1.12 │    1.67 │        11.2 │   1.71 │
│         _sum_sin_1000 │  4.58 μs │     3.85 │        2.66 │   402.0 │        18.0 │   3.09 │
│              kron_sum │ 197.0 μs │     12.9 │        3.26 │    7.02 │       488.0 │   23.8 │
│         kron_view_sum │ 273.0 μs │     12.6 │        5.26 │    28.1 │       441.0 │   13.8 │
│ naive_map_sin_cos_exp │  2.33 μs │     2.78 │        1.51 │ missing │        8.08 │   2.06 │
│       map_sin_cos_exp │  2.34 μs │     3.23 │        1.53 │    1.48 │        6.88 │   2.51 │
│ broadcast_sin_cos_exp │  2.26 μs │     3.05 │        1.57 │    4.34 │        1.41 │    2.1 │
│            simple_mlp │ 346.0 μs │     4.98 │        2.99 │    1.95 │        11.3 │   3.21 │
│                gp_lml │ 164.0 μs │     12.9 │        2.69 │    5.23 │     missing │   5.88 │
│    large_single_block │ 471.0 ns │     6.55 │        1.96 │  3910.0 │        36.7 │   2.06 │
```

### DIFF
--- a/src/test_utils.jl
+++ b/src/test_utils.jl
@@ -558,6 +558,47 @@ function test_frule_correctness(
     @test any(isapprox_results)
 end
 
+function test_frule_reuse(x_ẋ...; frule)
+    @nospecialize x_ẋ
+    x_ẋ_a = map(_deepcopy, x_ẋ)
+    x_ẋ_b = map(_deepcopy, x_ẋ)
+    y_ẏ_a = frule(x_ẋ_a...)
+    y_ẏ_b = frule(x_ẋ_b...)
+    @test has_equal_data(primal(y_ẏ_a), primal(y_ẏ_b))
+    @test has_equal_data(tangent(y_ẏ_a), tangent(y_ẏ_b))
+    @test all(map(has_equal_data, map(tangent, x_ẋ_a), map(tangent, x_ẋ_b)))
+end
+
+function test_rrule_reuse(rng::AbstractRNG, x_x̄...; rrule)
+    @nospecialize rng x_x̄
+    function make_inputs(x_x̄)
+        x_x̄ = map(_deepcopy, x_x̄)
+        x = map(primal, x_x̄)
+        x̄_zero = map(zero_tangent, x)
+        x̄_fwds = map(Mooncake.fdata, x̄_zero)
+        x_x̄_rule = map(
+            (xi, fi) -> fcodual_type(_typeof(xi))(_deepcopy(xi), fi), x, x̄_fwds
+        )
+        return x_x̄_rule, x̄_zero
+    end
+    inputs_a, x̄_zero_a = make_inputs(x_x̄)
+    inputs_b, x̄_zero_b = make_inputs(x_x̄)
+
+    # First forward+backward pass.
+    y_ȳ_a, pb_a!! = rrule(inputs_a...)
+    ŷ = randn_tangent(rng, primal(y_ȳ_a))
+    x̄_rvs_a = pb_a!!(Mooncake.rdata(ŷ))
+
+    # Second forward+backward pass with the same rule.
+    y_ȳ_b, pb_b!! = rrule(inputs_b...)
+    x̄_rvs_b = pb_b!!(Mooncake.rdata(_deepcopy(ŷ)))
+    @test has_equal_data(primal(y_ȳ_a), primal(y_ȳ_b))
+    @test has_equal_data(x̄_rvs_a, x̄_rvs_b)
+    @test all(
+        map(has_equal_data, map(Mooncake.fdata, x̄_zero_a), map(Mooncake.fdata, x̄_zero_b))
+    )
+end
+
 # Assumes that the interface has been tested, and we can simply check for numerical issues.
 function test_rrule_correctness(
     rng::AbstractRNG,
@@ -1034,9 +1075,6 @@ function test_rule(
                 test_rvs && test_rrule_interface(x_x̄...; rrule)
             end
 
-            # Snapshot of RNG so "Reuse" testset uses identical probe tangents to "Correctness" testset.
-            rng_snapshot = deepcopy(rng)
-
             # Test that answers are numerically correct / consistent.
             @testset "Correctness" begin
                 if test_fwd && !interface_only
@@ -1049,23 +1087,13 @@ function test_rule(
                 end
             end
 
-            # Verify rules are not invalidated by a first differentiation call.
+            # Verify rules give identical results on a second call.
             @testset "Reuse" begin
                 if test_fwd && !interface_only
-                    test_frule_correctness(
-                        rng_snapshot, x_ẋ...; frule, unsafe_perturb, atol, rtol
-                    )
+                    test_frule_reuse(x_ẋ...; frule)
                 end
                 if test_rvs && !interface_only
-                    test_rrule_correctness(
-                        rng_snapshot,
-                        x_x̄...;
-                        rrule,
-                        unsafe_perturb,
-                        output_tangent,
-                        atol,
-                        rtol,
-                    )
+                    test_rrule_reuse(rng, x_x̄...; rrule)
                 end
             end
 
--- a/src/test_utils.jl
+++ b/src/test_utils.jl
@@ -589,14 +589,18 @@ function test_rrule_reuse(rng::AbstractRNG, x_x̄...; rrule)
     y_primal_a = _deepcopy(primal(y_ȳ_a))
     # Set output cotangent via fdata (vectors communicate cotangents through fdata, not rdata).
     ŷ_delta = randn_tangent(rng, primal(y_ȳ_a))
-    ȳ_a = increment!!(set_to_zero!!(zero_tangent(primal(y_ȳ_a), tangent(y_ȳ_a))), ŷ_delta)
+    ȳ_a = increment!!(
+        set_to_zero!!(zero_tangent(primal(y_ȳ_a), tangent(y_ȳ_a))), ŷ_delta
+    )
     x̄_rvs_a = pb_a!!(Mooncake.rdata(ȳ_a))
 
     # Second forward pass with the same rule.
     y_ȳ_b, pb_b!! = rrule(inputs_b...)
     y_primal_b = _deepcopy(primal(y_ȳ_b))
     # Same output cotangent as run A.
-    ȳ_b = increment!!(set_to_zero!!(zero_tangent(primal(y_ȳ_b), tangent(y_ȳ_b))), _deepcopy(ŷ_delta))
+    ȳ_b = increment!!(
+        set_to_zero!!(zero_tangent(primal(y_ȳ_b), tangent(y_ȳ_b))), _deepcopy(ŷ_delta)
+    )
     x̄_rvs_b = pb_b!!(Mooncake.rdata(ȳ_b))
 
     @test has_equal_data(y_primal_a, y_primal_b)
--- a/src/test_utils.jl
+++ b/src/test_utils.jl
@@ -1034,6 +1034,9 @@ function test_rule(
                 test_rvs && test_rrule_interface(x_x̄...; rrule)
             end
 
+            # Snapshot of RNG so "Reuse" testset uses identical probe tangents to "Correctness" testset.
+            rng_snapshot = deepcopy(rng)
+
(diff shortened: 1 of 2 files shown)


src/test_utils.jl
```diff
@@ -1034,6 +1034,9 @@ function test_rule(
                 test_rvs && test_rrule_interface(x_x̄...; rrule)
             end
 
+            # Snapshot of RNG so "Reuse" testset uses identical probe tangents to "Correctness" testset.
+            rng_snapshot = deepcopy(rng)
```
yebai: It would be better if we could eliminate the need for sharing random seeds.

src/test_utils.jl
```diff
@@ -1046,6 +1101,16 @@ function test_rule(
...
+            # Verify rules give identical results on a second call.
+            @testset "Reuse" begin
+                if test_fwd && !interface_only
+                    test_frule_reuse(x_ẋ...; frule)
+                end
+                if test_rvs && !interface_only
+                    test_rrule_reuse(rng, x_x̄...; rrule)
+                end
+            end
+
```
sunxd3: it looks like to me that the `Correctness` test group is the first differentiation call. so `Reuse` would be testing the 2nd and 3rd call, which is doesn't seems to be what we want? (as far as I can tell, we want to be comparing 1st and 2nd calls)

outside of this, I am a little unsure about the effectiveness of this check: is there a MWE that would pass the current (main branch) test but would be caught with the added test from this PR?

AstitvaAggarwal: We care only that two consecutive calls give the same result (if a rule corrupts its state, it'll do so on every call, not just the second one). If ordering does matter i.e a rule that fails specifically on call 2 but recovers afterwards would already be caught by Correctness since it verifies each call against finite differences.

About 2,  Im not sure I understand but if there was anything wrong with our current testset it would have been caught ? Also this is a preventive check to guard against preexisting/future rules that accidentally corrupt their internal state across calls.

sunxd3: I am reflecting on what Will wrote in https://github.com/chalk-lab/Mooncake.jl/issues/681, it seems to me that it's a possible scenario that the first call corrupts the state, so the first call is correct, but all the following calls will be wrong, but consistently wrong.

my thoughts were: current rules are already good, but to make sure the added checks are actually doing what it's designed to do, at least we need to show a negative example and the check we can catch it (it doesn't need to be in the tests even)

AstitvaAggarwal: I think ill run the reuse checks first (before interface, correctness), the correctness checks can catch the error but its not what it was designed for anyways.

sunxd3:
👍 I'll do another round by the end of today

src/test_utils.jl
```diff
@@ -1028,6 +1081,20 @@ function test_rule(
     redirector = print_results ? ((f, x) -> f()) : redirect_stdout
     ts = redirector(devnull) do
         @testset "$(typeof(x))" begin
+            rng_reuse = deepcopy(rng)
```
sunxd3: for reuse, maybe we don't need to control the rng state? it looks like to me the rng is only used to generate `ŷ_delta = randn_tangent(rng, primal(y_ȳ_a))` and `ŷ_delta` is used in the two calls?

sunxd3 (commented):
the content looks good, could you try refactoring a little bit to reduce the code duplications if possible?

src/test_utils.jl
AstitvaAggarwal: for testing it would be the convention to use randomnly generated cotangents. are you suggesting we use a fixed/hardcoded `ŷ_delta` for all test calls to the `reuse` testset?

sunxd3: No, I am proposing something narrower. I am suggesting that maybe the deepcopy of the rng is not necessary, or for reuse test, a fixed rng like `Xoshiro(123)` might be good enough. Does this make sense at all?

src/test_utils.jl
```diff
@@ -558,6 +558,60 @@ function test_frule_correctness(
...
+        x = map(primal, x_x̄)
+        x̄_zero = map(zero_tangent, x)
+        x̄_fwds = map(Mooncake.fdata, x̄_zero)
+        x_x̄_rule = map(
+            (xi, fi) -> fcodual_type(_typeof(xi))(_deepcopy(xi), fi), x, x̄_fwds
+        )
+        return x_x̄_rule, x̄_zero
+    end
+    inputs_a, x̄_zero_a = make_inputs(x_x̄)
+    inputs_b, x̄_zero_b = make_inputs(x_x̄)
```
sunxd3: this is a style nit: I don't think we need to define a function for this?

src/test_utils.jl
```diff
@@ -558,6 +558,62 @@ function test_frule_correctness(
...
+        set_to_zero!!(zero_tangent(primal(y_ȳ_b), tangent(y_ȳ_b))), _deepcopy(ŷ_delta)
+    )
+    x̄_rvs_b = pb_b!!(Mooncake.rdata(ȳ_b))
+
+    @test has_equal_data(y_primal_a, y_primal_b)
+    @test has_equal_data(x̄_rvs_a, x̄_rvs_b)
+    @test all(
+        map(has_equal_data, map(Mooncake.fdata, x̄_zero_a), map(Mooncake.fdata, x̄_zero_b))
+    )
+end
```
```suggestion
function test_rrule_reuse(rng::AbstractRNG, x_x̄...; rrule)
    @nospecialize rng x_x̄
    x = map(primal, x_x̄)
    x̄_zero_a = map(zero_tangent, x)
    inputs_a = map(
        (x, x̄_f) -> fcodual_type(_typeof(x))(_deepcopy(x), x̄_f),
        x,
        map(Mooncake.fdata, x̄_zero_a),
    )
    x̄_zero_b = map(zero_tangent, x)
    inputs_b = map(
        (x, x̄_f) -> fcodual_type(_typeof(x))(_deepcopy(x), x̄_f),
        x,
        map(Mooncake.fdata, x̄_zero_b),
    )

    # Skip the deepcopy when tangent(y_ȳ) is NoFData: such primals (e.g. Float64, or
    # Module-containing types like Core.TypeName) can't be mutated by the pullback
    y_ȳ_a, pb_a!! = rrule(inputs_a...)
    y_primal_a = tangent(y_ȳ_a) isa NoFData ? primal(y_ȳ_a) : _deepcopy(primal(y_ȳ_a))
    ȳ_delta = randn_tangent(rng, primal(y_ȳ_a))
    ȳ_a = increment!!(
        set_to_zero!!(zero_tangent(primal(y_ȳ_a), tangent(y_ȳ_a))), _deepcopy(ȳ_delta)
    )
    x̄_rvs_a = pb_a!!(Mooncake.rdata(ȳ_a))

    y_ȳ_b, pb_b!! = rrule(inputs_b...)
    y_primal_b = tangent(y_ȳ_b) isa NoFData ? primal(y_ȳ_b) : _deepcopy(primal(y_ȳ_b))
    ȳ_b = increment!!(
        set_to_zero!!(zero_tangent(primal(y_ȳ_b), tangent(y_ȳ_b))), _deepcopy(ȳ_delta)
    )
    x̄_rvs_b = pb_b!!(Mooncake.rdata(ȳ_b))

    @test has_equal_data(y_primal_a, y_primal_b)
    @test has_equal_data(x̄_rvs_a, x̄_rvs_b)
    @test all(
        map(has_equal_data, map(Mooncake.fdata, x̄_zero_a), map(Mooncake.fdata, x̄_zero_b))
    )
end
```

I think this is a bit more consistent with style convention

src/test_utils.jl
```diff
@@ -1028,6 +1084,17 @@ function test_rule(
...
     ts = redirector(devnull) do
         @testset "$(typeof(x))" begin
+            # Verify rules give identical results on a second call,
+            # i.e. the rule does not corrupt its internal state across calls.
+            @testset "Reuse" begin
+                if test_fwd && !interface_only
+                    test_frule_reuse(x_ẋ...; frule)
+                end
+                if test_rvs && !interface_only
+                    test_rrule_reuse(x_x̄...; rrule)
```
```suggestion
                    test_rrule_reuse(rng, x_x̄...; rrule)
```

test/test_utils.jl
```diff
@@ -147,4 +174,30 @@
...
+    end
+    @testset "test_rrule_reuse catches state-corrupting Mooncake rules" begin
+        _rvs_counter[] = 0
+        broken_rrule = build_rrule(
+            get_interpreter(ReverseMode), Tuple{typeof(_broken_rvs_f),Float64}
+        )
+        inner_ts = Test.DefaultTestSet("inner"; verbose=false)
+        Test.push_testset(inner_ts)
+        TestUtils.test_rrule_reuse(
+            zero_codual(_broken_rvs_f), zero_codual(1.5); rrule=broken_rrule
```
sunxd3:
```suggestion
            Xoshiro(123), zero_codual(_broken_rvs_f), zero_codual(1.5); rrule=broken_rrule
```

sunxd3 (commented):
we are getting there! some nits I think worth considering

sunxd3:
~what is the nature of https://github.com/chalk-lab/Mooncake.jl/actions/runs/26028904572/job/76509403381?pr=1172?~

~why consuming rng for testing reuse leads to these failures?~
edited: use `Xoshiro(123)` instead of `rng`

PR #1173, opened 2026-05-13
max_norm_perturbation kwarg

### LINKED ISSUE #765: restrict finite differences in test_rule
For functions that are differentiable on a domain ``D``, but undefined outside that domain, testing with finite differences can result in errors.

It would be useful to extend the API of `TestUtils.test_rule` to allow for this.

But I am unsure what form the generic interface should take, I am opening this to get ideas and then I would happy to make a PR.

AstitvaAggarwal:
closes #765 

This PR adds a `max_fd_step` keyword argument to test_rule, `test_frule_correctness`, and `test_rrule_correctness`. When provided, only finite-difference step sizes `ε ≤ max_fd_step` are used. This is useful for functions defined on a restricted domain (e.g. `log`, `sqrt`, `cholesky`) where large perturbations would step outside it.

Behaviour change: tangent normalisation in `test_frule_correctness`

`test_frule_correctness` now normalises the tangent direction before running finite differences:
`ẋ = map(normalize_tangent ∘ tangent, x_ẋ)`

Previously the raw tangent from x_ẋ was used directly. With normalisation, ẋ has unit norm, so `max_fd_step` directly controls the step size in input space (rather than a step scaled by the tangent magnitude). The correctness check is unaffected, the `FD` estimate and the `AD` result are both linear in `ẋ`, so scaling `ẋ` scales both by the same factor. `test_rrule_correctness` already normalised its tangent direction; this makes `test_frule_correctness` consistent with it.

### Documentation Preview

Mooncake.jl documentation for PR #1173 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1173/

### Performance

Performance Ratio:
```
│              sum_1000 │ 170.0 ns │     1.59 │        1.76 │   0.706 │        3.42 │    6.6 │
│             _sum_1000 │ 972.0 ns │     7.16 │        1.04 │  1990.0 │        40.0 │   1.06 │
│          sum_sin_1000 │  7.25 μs │      3.5 │        1.36 │    1.46 │        10.9 │   1.74 │
│         _sum_sin_1000 │  6.47 μs │      3.5 │        1.77 │   228.0 │        12.4 │   2.05 │
│              kron_sum │ 219.0 μs │     12.7 │         3.6 │    13.4 │       347.0 │   19.7 │
│         kron_view_sum │ 300.0 μs │     11.1 │        5.26 │    24.4 │       313.0 │   11.8 │
│ naive_map_sin_cos_exp │  2.32 μs │     3.08 │        1.54 │ missing │        7.65 │   2.14 │
│       map_sin_cos_exp │  2.29 μs │     3.59 │        1.69 │    1.55 │        6.71 │   2.72 │
│ broadcast_sin_cos_exp │  2.32 μs │     3.19 │        1.59 │    3.81 │        1.45 │   2.12 │
│            simple_mlp │ 352.0 μs │     5.26 │        2.68 │    2.21 │        9.67 │   2.95 │
│                gp_lml │ 172.0 μs │     11.3 │        2.54 │    4.84 │     missing │   5.31 │
│    large_single_block │ 421.0 ns │     6.09 │         1.9 │  4910.0 │        33.3 │   2.05 │
```

### DIFF
--- a/src/test_utils.jl
+++ b/src/test_utils.jl
@@ -475,6 +475,16 @@ function _diff(p::P, q::P) where {P}
     return increment!!(_scale(-1.0, t2), t1)
 end
 
+function _check_ε_list(ε_list, max_norm_perturbation)
+    isempty(ε_list) && throw(
+        ArgumentError(
+            "max_norm_perturbation=$max_norm_perturbation filters out all finite-difference " *
+            "step sizes; the smallest available is 1e-8, below which floating-point rounding " *
+            "errors dominate the estimate.",
+        ),
+    )
+end
+
 # Assumes that the interface has been tested, and we can simply check for numerical issues.
 function test_frule_correctness(
     rng::AbstractRNG,
@@ -491,7 +501,7 @@ function test_frule_correctness(
 
     # Run original function on deep-copies of inputs.
     x = map(primal, x_ẋ)
-    ẋ = map(x -> normalize_tangent(tangent(x)), x_ẋ)
+    ẋ = map(normalize_tangent ∘ tangent, x_ẋ)
     x_primal = _deepcopy(x)
     y_primal = x_primal[1](x_primal[2:end]...)
 
@@ -502,6 +512,7 @@ function test_frule_correctness(
         ε -> isnothing(max_norm_perturbation) || ε <= max_norm_perturbation,
         [1e-2, 1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8],
     )
+    _check_ε_list(ε_list, max_norm_perturbation)
     fd_results = Vector{Any}(undef, length(ε_list))
     for (n, ε) in enumerate(ε_list)
         x′_l = _add_to_primal(x, _scale(ε, ẋ), unsafe_perturb)
@@ -600,6 +611,7 @@ function test_rrule_correctness(
         ε -> isnothing(max_norm_perturbation) || ε <= max_norm_perturbation,
         [1e-2, 1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8],
     )
+    _check_ε_list(ε_list, max_norm_perturbation)
     fd_results = Vector{Any}(undef, length(ε_list))
     for (n, ε) in enumerate(ε_list)
         x′_l = _add_to_primal(x, _scale(ε, ẋ), unsafe_perturb)
@@ -921,7 +933,10 @@ __get_primals(xs) = map(x -> x isa Union{Dual,CoDual} ? primal(x) : x, xs)
         print_results=true,
         output_tangent=nothing,
         atol=1e-3,
-        rtol=1e-3
+        rtol=1e-3,
+        frule=nothing,
+        rrule=nothing,
+        max_norm_perturbation::Union{Nothing,Float64}=nothing,
     )
 
 Run standardised tests on the `rule` for `x`.
--- a/src/test_utils.jl
+++ b/src/test_utils.jl
@@ -491,7 +491,7 @@ function test_frule_correctness(
 
     # Run original function on deep-copies of inputs.
     x = map(primal, x_ẋ)
-    ẋ = map(normalize_tangent ∘ tangent, x_ẋ)
+    ẋ = map(x -> normalize_tangent(tangent(x)), x_ẋ)
     x_primal = _deepcopy(x)
     y_primal = x_primal[1](x_primal[2:end]...)
 
--- a/src/test_utils.jl
+++ b/src/test_utils.jl
@@ -483,7 +483,7 @@ function test_frule_correctness(
     unsafe_perturb::Bool,
     rtol=1e-3,
     atol=1e-3,
-    max_norm_perturbation::Union{Nothing,Float64}=nothing,
+    max_fd_step::Union{Nothing,Float64}=nothing,
 )
     @nospecialize rng x_ẋ
 
@@ -498,17 +498,13 @@ function test_frule_correctness(
     # Use finite differences to estimate Frechet derivative. Compute the estimate at a range
     # of different step sizes. We'll just require that one of them ends up being close to
     # what AD gives.
-    !isnothing(max_norm_perturbation) &&
-        max_norm_perturbation < 1e-8 &&
-        throw(
-            ArgumentError(
-                "max_norm_perturbation=$max_norm_perturbation < 1e-8; the smallest available step size is 1e-8.",
-            ),
+    ε_list = [1e-2, 1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8]
+    if max_fd_step !== nothing
+        ε_list = filter(≤(max_fd_step), ε_list)
+        length(ε_list) ≥ 2 || throw(
+            ArgumentError("max_fd_step=$max_fd_step is too small (leaves <2 FD steps)")
         )
-    ε_list = filter(
-        ε -> isnothing(max_norm_perturbation) || ε <= max_norm_perturbation,
-        [1e-2, 1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8],
-    )
+    end
     fd_results = Vector{Any}(undef, length(ε_list))
     for (n, ε) in enumerate(ε_list)
         x′_l = _add_to_primal(x, _scale(ε, ẋ), unsafe_perturb)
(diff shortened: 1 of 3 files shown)


sunxd3:
thanks, I'll take a look today

src/test_utils.jl
```diff
@@ -586,7 +596,10 @@ function test_rrule_correctness(
 
     # Use finite differences to estimate vjps. Compute the estimate at a range of different
     # step sizes. We'll just require that one of them ends up being close to what AD gives.
-    ε_list = [1e-2, 1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8]
+    ε_list = filter(
+        ε -> isnothing(max_norm_perturbation) || ε <= max_norm_perturbation,
+        [1e-2, 1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8],
+    )
```
sunxd3: worth add a check here for the filtered `ε_list`: if too few left or empty

src/test_utils.jl
```diff
@@ -477,22 +477,31 @@ end
...
+    @nospecialize rng x_ẋ
 
-    x_ẋ = map(_deepcopy, x_ẋ) # defensive copy
+    x_ẋ = map(_deepcopy, x_ẋ) # defensive copy
 
     # Run original function on deep-copies of inputs.
-    x = map(primal, x_ẋ)
-    ẋ = map(tangent, x_ẋ)
+    x = map(primal, x_ẋ)
+    ẋ = map(x -> normalize_tangent(tangent(x)), x_ẋ)
```
sunxd3: a nit on style, maybe do something like `normalize_tangent ∘ tangent` instead of the anonymous function?

src/test_utils.jl
sunxd3, line 940: update this too? it was outdated before this PR but worth fixing with it

test/test_utils.jl
```diff
@@ -147,4 +147,17 @@
...
+        # With a bound of 1e-3, all ε values keep x+ε*ẋ > 0 and the rule test passes.
+        ts = TestUtils.test_rule(
+            rng,
+            log,
+            0.005;
+            is_primitive=false,
+            print_results=false,
+            max_norm_perturbation=1.0e-3,
+        )
+        @test !ts.anynonpass
```
sunxd3: why do we need this?

test/test_utils.jl
```diff
@@ -147,4 +147,17 @@
         @test TestUtils.count_allocs(Mooncake.fdata_type, Tuple{Float64}) == 0
         @test TestUtils.count_allocs(Mooncake.fdata_type, Tuple{Vector{Float64}}) == 0
     end
+    @testset "max_norm_perturbation kwarg for testing rules" begin
+        rng = Xoshiro(1)
```
sunxd3: this feels inconsistent coding style

test/rules/low_level_maths.jl
```diff
@@ -133,4 +133,11 @@
...
         @test !is_primitive(C, M, Tuple{typeof(\),T,T}, world)
     end
+
+    @testset "near-boundary domain-restricted functions" begin
+        test_rule(StableRNG(123), log, 0.005; is_primitive=true, max_norm_perturbation=1e-3)
+        test_rule(
+            StableRNG(123), sqrt, 0.005; is_primitive=true, max_norm_perturbation=1e-3
+        )
+    end
 end
```
sunxd3, line 140: does this overlap with the one in test_utils.jl?

src/test_utils.jl
```diff
@@ -477,22 +477,31 @@ end
...
     y_primal = x_primal[1](x_primal[2:end]...)
 
     # Use finite differences to estimate Frechet derivative. Compute the estimate at a range
     # of different step sizes. We'll just require that one of them ends up being close to
     # what AD gives.
-    ε_list = [1e-2, 1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8]
+    ε_list = filter(
+        ε -> isnothing(max_norm_perturbation) || ε <= max_norm_perturbation,
+        [1e-2, 1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8],
+    )
```
sunxd3: check here also

sunxd3 (commented):
thanks for the effort

test/test_utils.jl
AstitvaAggarwal: will fix this, its actually a field in `Test.DefaultTestSet` i was using it as `@test !ts.anynonpass` to assert that all internal tests passed.

AstitvaAggarwal:
thanks for the reviews @sunxd3! I have tried addressing all of them.

src/test_utils.jl
```diff
@@ -475,24 +475,44 @@ function _diff(p::P, q::P) where {P}
...
 
+function _check_ε_list(ε_list, max_norm_perturbation)
+    isempty(ε_list) && throw(
+        ArgumentError(
+            "max_norm_perturbation=$max_norm_perturbation filters out all finite-difference " *
+            "step sizes; the smallest available is 1e-8, below which floating-point rounding " *
+            "errors dominate the estimate.",
+        ),
+    )
+end
```
sunxd3: this is a style nit: we can just check the value of `max_norm_perturbation` directly before 
```
ε_list = filter(
        ε -> isnothing(max_norm_perturbation) || ε <= max_norm_perturbation,
        [1e-2, 1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8],
    )
```
and error if too small, try to keep it a oneliner?

src/test_utils.jl
```diff
@@ -477,22 +477,38 @@ end
...
 # Assumes that the interface has been tested, and we can simply check for numerical issues.
 function test_frule_correctness(
-    rng::AbstractRNG, x_ẋ...; frule, unsafe_perturb::Bool, rtol=1e-3, atol=1e-3
+    rng::AbstractRNG,
+    x_ẋ...;
+    frule,
+    unsafe_perturb::Bool,
+    rtol=1e-3,
+    atol=1e-3,
+    max_norm_perturbation::Union{Nothing,Float64}=nothing,
```
sunxd3: rename this to something like `max_fd_step`?

src/test_utils.jl
```diff
@@ -477,22 +477,38 @@ end
...
+        max_norm_perturbation < 1e-8 &&
+        throw(
+            ArgumentError(
+                "max_norm_perturbation=$max_norm_perturbation < 1e-8; the smallest available step size is 1e-8.",
+            ),
+        )
+    ε_list = filter(
+        ε -> isnothing(max_norm_perturbation) || ε <= max_norm_perturbation,
+        [1e-2, 1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8],
+    )
```
```suggestion
    ε_list = [1e-2, 1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8]
    if max_fd_step !== nothing
        ε_list = filter(≤(max_fd_step), ε_list)
        length(ε_list) ≥ 2 || throw(ArgumentError(
            "max_fd_step=$max_fd_step is too small (leaves <2 FD steps)"))
    end
```

src/test_utils.jl
```diff
@@ -908,7 +935,10 @@ __get_primals(xs) = map(x -> x isa Union{Dual,CoDual} ? primal(x) : x, xs)
         print_results=true,
         output_tangent=nothing,
         atol=1e-3,
-        rtol=1e-3
+        rtol=1e-3,
+        frule=nothing,
+        rrule=nothing,
```
sunxd3, line 938: update the doc argument list for these?

test/test_utils.jl
```diff
@@ -147,4 +147,14 @@
         @test TestUtils.count_allocs(Mooncake.fdata_type, Tuple{Float64}) == 0
         @test TestUtils.count_allocs(Mooncake.fdata_type, Tuple{Vector{Float64}}) == 0
     end
+    @testset "max_norm_perturbation kwarg for testing rules" begin
+        TestUtils.test_rule(
+            StableRNG(123),
+            x -> log(x),
```
sunxd3, line 156: I sense that there is a better way to use an anonymous function here? or at least worth some comments

sunxd3 (commented):
for `test_frule_correctness` now `x` is normalized, which I think is still correct, but a behavior change, could you write about this change in the PR description?

Issue #847, opened 2025-11-11
Things to check in 1.12


Technici4n:
Some comments from the 1.12 update PR #714, to be checked at some point.

- [ ] Resizing `bb_code.debuginfo.codelocs` to `3n` in `bbcode.jl` seems like a quick fix for incorrect handling of line numbers elsewhere.
- [ ] One of the `@zero_derivative` calls for logging in `avoiding_non_differentiable_code.jl` might be unnecessary.
- [ ] The `integration testing for invalid global ref errors` test is currently disabled for 1.12, is this the right thing to do?
- [ ] Why was `has_equal_data_internal` added for `Method`, `CodeInstance` and `MethodInstance`?
- [ ] `resolve_unbound_globalrefs` needs to be checked again.

yebai:
> [!WARNING]
>  The comments below are by Copilot. It may not be entirely accurate, but it is still useful, as it is more detailed than the above. 

## Analysis of the 5 open items (from PR #714 review comments + source)

All 5 items remain in the current codebase. Here's what I found for each:

---

### 1. `codelocs` resizing to `3n` in `bbcode.jl` (lines 668–677)

In Julia 1.12+, `debuginfo.codelocs` changed format from one integer per instruction to three integers per instruction (flattened), so a valid vector has length `3n` for `n` instructions.

The current code handles both `> 3n` (truncate) and `< 3n` (pad with zeros) cases. Technici4n questioned *why* the vector ever has the wrong size — if Mooncake constructs it properly via `BBCode`, it should always be exactly `3n`. serenity4 never answered whether there is a genuine upstream case where this occurs.

**Suspicion:** This is a defensive band-aid. The root cause is likely in how `BBCode` is constructed from `IRCode` (or vice versa) — Technici4n raised `new_inst_vec` as a potential culprit too. Worth auditing the `IRCode → BBCode → IRCode` round-trip to see if `codelocs` is ever mis-sized for a legitimate reason.

---

### 2. Possibly unnecessary `@zero_derivative` for logging (`avoiding_non_differentiable_code.jl`, lines 63–74 vs 94+)

There are multiple `@zero_derivative` rules for `CoreLogging` dispatch chains. Technici4n asked whether the `Core._call_latest` + `handle_message` rule (lines 63–74) was made redundant by the earlier `current_logger_for_env` / `shouldlog` rules. mhauru replied he tried both independently and they weren't individually sufficient, but couldn't explain why.

[...]

PR #1180, opened 2026-05-28
remove redundant `resolve_unbound_globalrefs`

AstitvaAggarwal:
**closes part 5 of #847** 

A constant binding is a name tied to a value that is expected never to change (eg: function names like `sin` or a package-defined constant). Julia tracks a global world age that increments every time a new function or constant is defined. Every piece of compiled code carries a `valid_worlds` range which is the span of world ages for which that compiled code is considered correct. When Mooncake calls `lookup_ir` to extract the compiled code for a function `f`, it gets back a copy of that code stamped with a wide range like `[w_compile, ∞)`.

Before Julia `v1.12`, constants could never be redefined, so this wide stamp was always safe. Julia `v1.12` introduced partitioned bindings -> constants can now be redefined at a new world age. This made Julia's internal `verify_ir` checker stricter: it now scans whether every name referenced inside the code was actually a stable constant across the entire claimed `valid_worlds` range. A name defined at world `80` inside code stamped `[50, ∞)` fails this check because it didn't exist at worlds `50–79`.

**Details about the PR fix :**
  
PR #714 introduced `resolve_unbound_globalrefs` as a workaround. Before `verify_ir` ran, it scanned the code and replaced any name that would fail the check with the actual value at that moment (baking the value directly into the code so `verify_ir` saw a concrete value instead of a name reference). It was already marked `# XXX: this is probably unsound` because if the name was later rebound to a new value, the compiled rule would silently keep using the old baked-in value and produce wrong gradients.
  
The same PR also introduced `set_valid_world!(ir, interp.world)`, called at every entry point into Mooncake's build rule pipeline, always before `normalise!` is called. This narrows the stamp from `[w_compile, ∞)` to the single point `[w_now, w_now]` (the world age the Mooncake's AD compiler sees during rule build time). 

After this narrowing, `verify_ir` only checks if the const binding names are a stable constant in that single world age. Now if you notice, by the time `resolve_unbound_globalrefs` ran, `verify_ir` already passed cleanly without any patching. The function `resolve_unbound_globalrefs` was redundant and Ive removed it. Now the behaviour is the same as before: if `f` or anything it depends on changes after the rule is built, the cached rule will silently use the old compiled version. This is expected as well because `build_rrule` is an explicit compilation step and the user is responsible for rebuilding the rule if `f` changes.
  

### Documentation Preview

Mooncake.jl documentation for PR #1180 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1180/

### Performance

Performance Ratio:
```
│              sum_1000 │ 180.0 ns │     1.62 │        1.62 │   0.722 │        4.01 │   7.01 │
│             _sum_1000 │   1.1 μs │     5.94 │        1.02 │  4610.0 │        42.7 │   1.06 │
│          sum_sin_1000 │  7.42 μs │     2.61 │        1.11 │    1.64 │        11.3 │   1.72 │
│         _sum_sin_1000 │  4.58 μs │     3.93 │        2.68 │   396.0 │        19.6 │   3.11 │
│              kron_sum │ 195.0 μs │     13.0 │        3.29 │    13.3 │       491.0 │   22.7 │
│         kron_view_sum │ 261.0 μs │     12.8 │        5.33 │    28.9 │       455.0 │   13.8 │
│ naive_map_sin_cos_exp │  2.22 μs │     2.83 │        1.54 │ missing │        8.51 │   2.17 │
│       map_sin_cos_exp │  2.22 μs │     3.22 │        1.63 │    1.53 │        7.22 │   2.64 │
│ broadcast_sin_cos_exp │  2.31 μs │     2.93 │        1.55 │    4.64 │        1.39 │   2.09 │
│            simple_mlp │ 344.0 μs │     5.03 │        2.93 │    1.98 │        9.43 │   3.21 │
│                gp_lml │ 166.0 μs │     12.0 │        2.61 │    4.79 │     missing │   5.44 │
│    large_single_block │ 451.0 ns │     6.23 │        2.04 │  4170.0 │        41.3 │   2.18 │
```

### DIFF
--- a/src/developer_tools.jl
+++ b/src/developer_tools.jl
@@ -30,6 +30,8 @@ true
 function primal_ir(interp::MooncakeInterpreter, sig::Type{<:Tuple}; normalize=true)::IRCode
     ir, _ = lookup_ir(interp, sig)
     @static if VERSION > v"1.12-"
+        # Pins valid_worlds to a single world so verify_ir's GlobalRef const-binding
+        # check passes for all currently-defined bindings without patching the IR.
         ir = set_valid_world!(ir, interp.world)
     end
     normalize || return ir
--- a/src/developer_tools.jl
+++ b/src/developer_tools.jl
@@ -30,8 +30,7 @@ true
 function primal_ir(interp::MooncakeInterpreter, sig::Type{<:Tuple}; normalize=true)::IRCode
     ir, _ = lookup_ir(interp, sig)
     @static if VERSION > v"1.12-"
-        # Pins valid_worlds to a single world so verify_ir's GlobalRef const-binding
-        # check passes for all currently-defined bindings without patching the IR.
+        # Pin to one world so verify_ir's GlobalRef check passes; see `set_valid_world!`.
         ir = set_valid_world!(ir, interp.world)
     end
     normalize || return ir
--- a/src/interpreter/forward_mode.jl
+++ b/src/interpreter/forward_mode.jl
@@ -187,6 +187,8 @@ function generate_dual_ir(
     # Grab code associated to the primal.
     primal_ir, _ = lookup_ir(interp, sig_or_mi)
     @static if VERSION > v"1.12-"
+        # Pins valid_worlds to a single world so verify_ir's GlobalRef const-binding
+        # check passes for all currently-defined bindings without patching the IR.
         primal_ir = set_valid_world!(primal_ir, interp.world)
     end
     nargs = length(primal_ir.argtypes)
--- a/src/interpreter/forward_mode.jl
+++ b/src/interpreter/forward_mode.jl
@@ -187,8 +187,7 @@ function generate_dual_ir(
     # Grab code associated to the primal.
     primal_ir, _ = lookup_ir(interp, sig_or_mi)
     @static if VERSION > v"1.12-"
-        # Pins valid_worlds to a single world so verify_ir's GlobalRef const-binding
-        # check passes for all currently-defined bindings without patching the IR.
+        # Pin to one world so verify_ir's GlobalRef check passes; see `set_valid_world!`.
         primal_ir = set_valid_world!(primal_ir, interp.world)
     end
     nargs = length(primal_ir.argtypes)
--- a/src/interpreter/ir_normalisation.jl
+++ b/src/interpreter/ir_normalisation.jl
@@ -233,7 +233,6 @@ expressions in the same way as a primitive `:call` expression, i.e. via an `rrul
 """
 splatnew_to_call(x) = Meta.isexpr(x, :splatnew) ? Expr(:call, _splat_new_, x.args...) : x
 
-
 """
     intrinsic_to_function(inst)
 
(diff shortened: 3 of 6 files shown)


sunxd3:
thanks for this. the removal looks sound after a first pass, I'll do a second pass more carefully later

sunxd3:
Could you give some a concrete example to explain why `resolve_unbound_globalrefs` is redundant? (for instance, go through an example and show that the function makes no impact)

AstitvaAggarwal:
@sunxd3 The function `resolve_unbound_globalrefs` is a no-op once `set_valid_world!` has run:

```julia
using Mooncake
import Core.Compiler as CC

function resolve_unbound_globalrefs(ir, @nospecialize ex)
    isa(ex, Expr) || return ex
    args = nothing
    for i in eachindex(ex.args)
        arg = ex.args[i]
        if isa(arg, GlobalRef) && arg.mod !== Core && arg.mod !== Base
            (valid_worlds, alldef) = CC.scan_leaf_partitions(
                nothing, arg,
                CC.WorldWithRange(CC.min_world(ir.valid_worlds), ir.valid_worlds),
            ) do _, _, bpart
                CC.is_defined_const_binding(CC.binding_kind(bpart))
            end
            if !alldef ||
                CC.max_world(valid_worlds) < CC.max_world(ir.valid_worlds) ||
                CC.min_world(valid_worlds) > CC.min_world(ir.valid_worlds)
                if isdefined(arg.mod, arg.name)
                    arg = getglobal(arg.mod, arg.name)
                end
            end
            if args === nothing
                args = ex.args[1:(i - 1)]
            end
        end
        args === nothing && continue
        push!(args, arg)
    end
    args === nothing && return ex
    resolved = Expr(ex.head)
    resolved.args = args
    return resolved
end

w_now = Base.get_world_counter()
ir_wide   = (valid_worlds = CC.WorldRange(UInt(1), typemax(UInt)),)
ir_narrow = (valid_worlds = CC.WorldRange(w_now, w_now),)

expr = Expr(:call, GlobalRef(Mooncake, :zero_tangent), Core.Argument(2))
result_wide   = resolve_unbound_globalrefs(ir_wide, expr)
result_narrow = resolve_unbound_globalrefs(ir_narrow, expr)

println("Wide   [1, typemax(UInt)]: first arg is GlobalRef? ", result_wide.args[1] isa GlobalRef)
# false because GlobalRef is replaced with concrete value, unsound bake-in fired

println("Narrow [w_now, w_now]:     first arg is GlobalRef? ", result_narrow.args[1] isa GlobalRef)
# true  because there is no replacement, function was a no-op
```
With a wide valid_worlds, the function fires and replaces `GlobalRef(Mooncake, :zero_tangent)` with the concrete function object. With `[w_now, w_now]` (`set_valid_world!` always  produces this before `normalise!` runs), it does nothing. Since `set_valid_world!` and `resolve_unbound_globalrefs` were introduced in the same commit and `set_valid_world!` is always called first, the function could never fire.

sunxd3 (approved):

Issue #1209, opened 2026-06-14
Forward-over-reverse: LazyFRule slot-type goes stale across a mid-pass world-age advance -> convert MethodError in _build_rule!


ChrisRackauckas-Claude:
## Summary

Forward-over-reverse (`prepare_hvp_cache` / `value_and_hvp!!`, `prepare_hessian_cache` / `value_gradient_and_hessian!!`) can throw a `convert` `MethodError` from `_build_rule!` when a `LazyFRule`'s predicted slot type goes stale across a world-age advance that happens *during* the forward pass.

```
MethodError: Cannot `convert` an object of type
  Mooncake.DerivedFRule{Tuple{DiffEqBase.var"##solve_call#30", ...},
    MistyClosure{OpaqueClosure{..., Mooncake.Dual{ODESolution{...}}}}, ...}
to an object of type
  ... MistyClosure{OpaqueClosure{..., Mooncake.Dual}}
```
from `setproperty!` → `_build_rule!` at `src/interpreter/forward_mode.jl`.

The reverse-mode gradient of the same function is computed correctly; only the forward-over-reverse path fails.

## Mechanism

`LazyFRule`'s type parameter `Trule` is predicted by `frule_type(interp, mi)` using the `ForwardMode` interpreter that exists when the *enclosing* rule is derived (at `prepare_*`). The rule is built lazily on first call in `_build_rule!`, which calls `get_interpreter(ForwardMode)` again:

```julia
@noinline function _build_rule!(rule::LazyFRule{sig,Trule}, args) where {sig,Trule}
    interp = get_interpreter(ForwardMode)
    rule.rule = build_frule(interp, rule.mi; debug_mode=rule.debug_mode)   # <-- convert error here
    return __call_rule(rule.rule, args)
end
```

`get_interpreter` mints a fresh interpreter whenever `Base.get_world_counter()` has advanced. If methods are defined *between* prediction and build, re-inference of the same `mi` can produce a **tighter** return type (here abstract `Dual` → concrete `Dual{ODESolution{...}}`), so the built `DerivedFRule` no longer matches the predicted `Trule`, and the `rule.rule = ...` setfield fails to `convert`.

[...]

PR #1213, opened 2026-06-15
Fix Lazy/Dynamic rules rebuilding at the current world instead of the prediction world (#1218)

### LINKED ISSUE #1218: Lazy/Dynamic rules rebuild at the current world instead of the prediction world (breaks after a mid-pass world advance)
### Summary

`LazyFRule` / `DynamicFRule` / `LazyDerivedRule` / `DynamicDerivedRule` predict their rule type in the **enclosing rule's derivation world**, but build the rule lazily on first call using `get_interpreter(ForwardMode/ReverseMode)` at the **current** world. If a method is defined between prediction and build — a mid-pass world advance — re-inference of the same `MethodInstance` can re-tighten the callee's inferred return type, so the built rule no longer matches the predicted `Trule`:

- for the Lazy rules, `rule.rule = built` throws a `convert` `MethodError` (the slot is typed `Trule`);
- for the Dynamic rules, the rebuilt rule silently reflects the post-advance world rather than the world the enclosing rule was compiled against.

### Mechanism

```julia
@noinline function _build_rule!(rule::LazyFRule{sig,Trule}, args) where {sig,Trule}
    interp = get_interpreter(ForwardMode)            # <-- current world, not prediction world
    rule.rule = build_frule(interp, rule.mi; debug_mode=rule.debug_mode)
    return __call_rule(rule.rule, args)
end
```

`get_interpreter` mints a fresh interpreter whenever `Base.get_world_counter()` has advanced, so the lazy build can run in a strictly newer world than the one used to predict `Trule`.

### Reproduction (minimal)

Defining a more specific method between rule construction and first call advances the world and tightens the callee's return type:

```julia
inner(x) = Float32(x) * 2.0f0
@noinline callee(x) = inner(x)
f(x) = callee(x)                       # static :invoke -> LazyFRule

rule = Mooncake.build_frule(f, 1.5)
@eval inner(x::Float64) = x * 2.0      # world advance; tightens callee Float32 -> Float64
Base.invokelatest(rule, Mooncake.zero_dual(f), Mooncake.Dual(1.5, 1.0))
# pre-fix: convert MethodError in _build_rule!
```

### Fix

Pin each rebuild to the world the rule type was predicted at (stored per rule) via
`get_interpreter(mode, world)` + `skip_world_age_check=true`, for all four Lazy/Dynamic rule
types in forward and reverse mode.

### Relationship to #1209

This is the **world-advance trigger** that surfaced while investigating #1209. It is a real,
self-contained bug with the reproduction above. It is **distinct** from #1209's headline MWE
(forward-over-reverse HVP through a SciML `InterpolatingAdjoint(ReverseDiffVJP())` solve): that
case involves **no world advance** — there, `frule_type` predicts an abstract return type via
inference type-complexity widening on the un-normalised IR while `build_frule` produces a
concrete one, and even reconciling the types yields a silently-wrong HVP. #1209 remains open.

yebai:
Fixes #1218.

`LazyFRule` / `DynamicFRule` / `LazyDerivedRule` / `DynamicDerivedRule` predict their rule type
in the enclosing rule's derivation world but rebuilt it lazily in the *current* world. A mid-pass
world advance can re-tighten the callee's inferred return type, so the rebuilt rule no longer
matches the predicted `Trule` and `setfield!` fails to `convert` (Lazy), or the Dynamic rebuild
silently reflects the post-advance world. Pin the rebuild to the prediction world (stored per
rule) via `get_interpreter(mode, world)` + `skip_world_age_check`, for all four rule types in
forward and reverse mode. Includes a `@zero_derivative` for the new two-arg `get_interpreter`.

### Documentation Preview

Mooncake.jl documentation for PR #1213 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1213/

### Performance

Performance Ratio:
```
│              sum_1000 │ 181.0 ns │     1.55 │        1.61 │   0.718 │        3.49 │   6.97 │
│             _sum_1000 │  1.09 μs │      6.2 │        1.04 │  1560.0 │        35.3 │   1.07 │
│          sum_sin_1000 │  7.43 μs │      2.6 │        1.11 │    1.65 │        11.1 │   1.71 │
│         _sum_sin_1000 │  4.69 μs │     3.74 │         2.6 │   324.0 │        17.3 │   3.04 │
│              kron_sum │ 198.0 μs │     13.1 │        3.28 │    10.7 │       503.0 │   14.6 │
│         kron_view_sum │ 261.0 μs │     12.8 │        5.32 │    24.1 │       435.0 │   10.4 │
│ naive_map_sin_cos_exp │  2.21 μs │     2.83 │        1.53 │ missing │        8.73 │   2.16 │
│       map_sin_cos_exp │  2.23 μs │     3.35 │        1.63 │    1.52 │        7.33 │   2.68 │
│ broadcast_sin_cos_exp │  2.22 μs │      3.1 │        1.57 │    4.52 │        1.46 │   2.17 │
│            simple_mlp │ 340.0 μs │     5.23 │        2.94 │    1.66 │        10.7 │   3.25 │
│                gp_lml │ 369.0 μs │     5.99 │        1.72 │    2.88 │     missing │   3.22 │
│    large_single_block │ 471.0 ns │     5.57 │        1.96 │  4040.0 │        32.2 │   2.08 │
```

### DIFF
--- a/src/rules/high_order_derivative_patches.jl
+++ b/src/rules/high_order_derivative_patches.jl
@@ -1,6 +1,6 @@
 @zero_derivative MinimalCtx Tuple{typeof(get_interpreter),Type{<:Mode}}
 @zero_derivative MinimalCtx Tuple{
-    typeof(build_rrule_checks),MooncakeInterpreter,Any,Bool,Bool
+    typeof(build_rrule_checks),MooncakeInterpreter,Any,Bool,Bool,Bool
 }
 @zero_derivative MinimalCtx Tuple{typeof(is_primitive),Type,Type{<:Mode},Type,UInt}
 
--- a/src/rules/high_order_derivative_patches.jl
+++ b/src/rules/high_order_derivative_patches.jl
@@ -1,4 +1,5 @@
 @zero_derivative MinimalCtx Tuple{typeof(get_interpreter),Type{<:Mode}}
+@zero_derivative MinimalCtx Tuple{typeof(get_interpreter),Type{<:Mode},UInt}
 @zero_derivative MinimalCtx Tuple{
     typeof(build_rrule_checks),MooncakeInterpreter,Any,Bool,Bool,Bool
 }
--- a/src/interpreter/abstract_interpretation.jl
+++ b/src/interpreter/abstract_interpretation.jl
@@ -336,18 +336,24 @@ const GLOBAL_INTERPRETERS = Dict(
 )
 
 """
-    get_interpreter(mode::Type{<:Mode})
+    get_interpreter(mode::Type{<:Mode}; world::UInt=Base.get_world_counter())
 
-Returns a `MooncakeInterpreter` appropriate for the current world age. Will use a cached
-interpreter if one already exists for the current world age, otherwise creates a new one.
+Returns a `MooncakeInterpreter` for `mode` at `world`, defaulting to the current world age and
+served from a process-wide cache. A non-current `world` (e.g. when a Lazy/Dynamic rule rebuilds
+at its stored prediction world) yields a fresh, uncached interpreter, leaving the shared
+current-world cache untouched.
 
 This should be prefered over constructing a `MooncakeInterpreter` directly.
 """
-function get_interpreter(mode::Type{<:Mode})
-    if GLOBAL_INTERPRETERS[mode].world != Base.get_world_counter()
-        GLOBAL_INTERPRETERS[mode] = MooncakeInterpreter(DefaultCtx, mode)
+function get_interpreter(mode::Type{<:Mode}; world::UInt=Base.get_world_counter())
+    return if world != Base.get_world_counter()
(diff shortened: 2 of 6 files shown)


src/rules/high_order_derivative_patches.jl
```diff
@@ -1,6 +1,6 @@
 @zero_derivative MinimalCtx Tuple{typeof(get_interpreter),Type{<:Mode}}
```
sunxd3, line 1: maybe add a rule for the new `get_interpreter`?

sunxd3 (approved):
Thanks, Hong. By my estimation, the fix is correct.

I think it's worthwhile to create a separate issue than #1209 for clarity.

src/rules/high_order_derivative_patches.jl
yebai: Thanks, excellent catch!

PR #1216, opened 2026-06-17
fix `debuginfo` copy & `line` padding for final `IRCode`.

AstitvaAggarwal:
**PR Issue description :**

Fixes part 1 of #847 .

In Julia `1.12+`, every `IRCode` has a `debuginfo.codelocs` array storing exactly `3` `Int32` values per instruction (file index, line number, column) - `3n` total for `n` instructions. When Mooncake converts `IRCode` to `BBCode`, it copies `debuginfo`. During AD transforms, instructions are added or removed from the `BBCode` blocks but `debuginfo.codelocs` is never updated. When converting back `BBCode` to `IRCode`, the sizes disagree, and depending on where instructions were added or removed, the debug info can be misaligned or wrong for added, removed, and already-present instructions.

The current code handles this with a defensive resize where it pads with zeros if too short, truncate if too long. This is a workaround that hides the root cause rather than fixing it.

**Note :**
1. Currently internal helper `new_inst_vec` also reads `stmts.line` incorrectly in Julia `1.12+`. `stmts.line` is a flat `Vector{Int32}` of length `3n`, map was called on it - treating it as length n. So each NewInstruction got a single Int32 instead of the proper `Tuple{Int32,Int32,Int32}`. This meant all `inst.line` fields in `BBCode` had wrong values also. Also fixed by indexing `x.line[3i-2]`, `x.line[3i-1]`, `x.line[3i]` for each instruction `i` access in `new_inst_vec`. 

2. PR also updates `ir_representation.md` doctests to reflect the new correct behaviour :  `NewInstruction` line fields now show proper `Tuple{Int32,Int32,Int32}` values, and `IR` display line numbers for inserted instructions now reflect their actual source location from the `linetable` (doctests run on latest Julia version only)

**MWE :** 

```  
  using Mooncake
  import Mooncake: BBCode, new_inst, ID
  import Core.Compiler as CC

  # for any function
  ir = Base.code_ircode(sin, Tuple{Float64})[1][1]
  bb = BBCode(ir)

  # Simulating what the AD transform does: add an instruction
  push!(bb.blocks[1].inst_ids, ID())
  push!(bb.blocks[1].insts, new_inst(nothing))

  new_ir = CC.IRCode(bb)
  n = length(new_ir.stmts) 

  # Before the fix: debuginfo.codelocs has wrong size (stmts.line was resized but
  # debuginfo.codelocs was copied from the unmodified BBCode debuginfo)

  # After the fix:
  @assert length(new_ir.stmts.line) == 3n
  @assert length(new_ir.debuginfo.codelocs) == 3n

```

### Documentation Preview

Mooncake.jl documentation for PR #1216 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1216/

### Performance

Performance Ratio:
```
│              sum_1000 │ 110.0 ns │     1.82 │        1.91 │   0.909 │        4.65 │    9.1 │
│             _sum_1000 │ 851.0 ns │     6.12 │        1.01 │  3950.0 │        40.4 │   1.13 │
│          sum_sin_1000 │  5.78 μs │     2.45 │        1.16 │    1.71 │        11.4 │   1.76 │
│         _sum_sin_1000 │  3.59 μs │     3.77 │        2.74 │   402.0 │        18.2 │   3.15 │
│              kron_sum │ 151.0 μs │     12.9 │        3.51 │    8.23 │       529.0 │   24.9 │
│         kron_view_sum │ 167.0 μs │     16.3 │        6.31 │    35.5 │       600.0 │   18.8 │
│ naive_map_sin_cos_exp │  1.75 μs │     2.75 │         1.5 │ missing │        8.94 │   2.19 │
│       map_sin_cos_exp │   1.7 μs │      3.3 │        1.65 │    1.69 │        7.62 │   2.77 │
│ broadcast_sin_cos_exp │  1.82 μs │     2.85 │        1.47 │    4.63 │        1.47 │   2.09 │
│            simple_mlp │ 240.0 μs │     5.35 │        2.94 │    2.35 │        10.9 │    3.5 │
│                gp_lml │ 146.0 μs │     11.1 │        2.82 │     4.6 │     missing │   5.79 │
│    large_single_block │ 370.0 ns │     5.44 │        1.92 │  4210.0 │        37.8 │   2.03 │
```

### DIFF
--- a/src/interpreter/bbcode.jl
+++ b/src/interpreter/bbcode.jl
@@ -571,7 +571,22 @@ Convert an `Compiler.InstructionStream` into a list of `Compiler.NewInstruction`
 """
 function new_inst_vec(x::CC.InstructionStream)
     stmt = @static VERSION < v"1.11.0-rc4" ? x.inst : x.stmt
-    return map((v...,) -> NewInstruction(v...), stmt, x.type, x.info, x.line, x.flag)
+    @static if VERSION > v"1.12-"
+        # In Julia 1.12+, x.line is a flat Vector{Int32} of length 3n (3 codeloc values per
+        # instruction). Extract the proper Tuple{Int32,Int32,Int32} for each instruction.
+        n = length(stmt)
+        return [
+            NewInstruction(
+                stmt[i],
+                x.type[i],
+                x.info[i],
+                (x.line[3i - 2], x.line[3i - 1], x.line[3i]),
+                x.flag[i],
+            ) for i in 1:n
+        ]
+    else
+        return map((v...,) -> NewInstruction(v...), stmt, x.type, x.info, x.line, x.flag)
+    end
 end
 
 # Maps from positional names (SSAValues for nodes, Integers for basic blocks) to IDs.
--- a/src/interpreter/bbcode.jl
+++ b/src/interpreter/bbcode.jl
@@ -666,15 +666,17 @@ function CC.IRCode(bb_code::BBCode)
     cfg = control_flow_graph(bb_code)
     insts = _lines_to_blocks(insts, cfg)
     @static if VERSION > v"1.12-"
-        lines = CC.copy(bb_code.debuginfo.codelocs)
-        n = length(insts)
-        if length(lines) > 3n
-            resize!(lines, 3n)
-        elseif length(lines) < 3n
-            for _ in (length(lines) + 1):3n
-                push!(lines, 0)
-            end
-        end
+        # Reconstruct codelocs from each instruction's own line field (a Tuple{Int32,Int32,Int32}
+        # in Julia 1.12+). This is now always 3n entries by construction, regardless of how
+        # many instructions were added or removed during BBCode transforms. The old approach of
+        # copying debuginfo.codelocs and resizing was a defensive workaround: instructions added
+        # or removed during transforms don't update codelocs, causing a size mismatch and
+        # misalignment between debug info and line numbers, potentially affecting added, removed,
+        # and already-present instructions depending on where the changes occurred.
+        lines = Int32[v for inst in insts for v in inst.line]
+        debuginfo = CC.copy(bb_code.debuginfo)
+        resize!(debuginfo.codelocs, length(lines))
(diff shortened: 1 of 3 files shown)


src/interpreter/bbcode.jl
```diff
@@ -666,15 +682,17 @@ function CC.IRCode(bb_code::BBCode)
...
+        # in Julia 1.12+). This is now always 3n entries by construction, regardless of how
+        # many instructions were added or removed during BBCode transforms. The old approach of
+        # copying debuginfo.codelocs and resizing was a defensive workaround: instructions added
+        # or removed during transforms don't update codelocs, causing a size mismatch and
+        # misalignment between debug info and line numbers, potentially affecting added, removed,
+        # and already-present instructions depending on where the changes occurred.
+        lines = Int32[v for inst in insts for v in inst.line]
+        debuginfo = CC.copy(bb_code.debuginfo)
+        resize!(debuginfo.codelocs, length(lines))
+        copyto!(debuginfo.codelocs, lines)
```
sunxd3:
```suggestion
        debuginfo.codelocs = lines
```
would this work? Could you verify that semantics of CC,copy on debuginfo is the same as before so that we are not mutating things unintentionally?

test/interpreter/bbcode.jl
```diff
@@ -64,6 +64,25 @@ end
...
+            # Insert an instruction into the first block
+            push!(bb.blocks[1].inst_ids, ID())
+            push!(bb.blocks[1].insts, new_inst(nothing))
+            new_ir = CC.IRCode(bb)
+            n = length(new_ir.stmts)
+            @test length(new_ir.stmts.line) == 3n
+            @test length(new_ir.debuginfo.codelocs) == 3n
+            @test new_ir.stmts.line == new_ir.debuginfo.codelocs
+        end
+    end
```
sunxd3, line 104: these feel a bit tautological, maybe use some existing tests?

sunxd3 (changes requested):

sunxd3 (approved):

PR #1223, opened 2026-06-22
Stop is_primitive specialising per call-signature

### LINKED ISSUE #1222: is_primitive over-specialises per call-signature
`is_primitive` is called once per call-node signature during the AD transform, so its specialisation count grows with the number of distinct signatures in the typed primal IR. On larger / deeply-composed models this becomes a noticeable chunk of rule-build time (roughly half of Mooncake's AD-specific inference on the model that prompted this).

The cause is small: both methods carry `@nospecialize sig`, but bind `sig` to a `where` parameter, which forces specialisation regardless (the compiler must solve for `Tsig`), so the `@nospecialize` is effectively a no-op:

```julia
# src/interpreter/contexts.jl
function is_primitive(
    ctx::Type{MinimalCtx}, mode::Type{<:Mode}, sig::Type{Tsig}, world::UInt
) where {Tsig<:Tuple}
    @nospecialize sig
    ...
end
```

## Self-contained reproduction

Only depends on Mooncake. We synthesise a primal whose IR has `K` distinct call-signatures by chaining `K` distinctly-typed ops:

```julia
import Mooncake
using Mooncake: prepare_gradient_cache

struct Op{N} end
@inline (::Op{N})(x) where {N} = muladd(1.0, x, N)

@generated function chain(x, ::Val{K}) where {K}     # body has K distinct Op{n}() call nodes
    ex = :x
    for n in 1:K
        ex = :(Op{$n}()($ex))
    end
    return ex
end
struct ChainF{K} end
(::ChainF{K})(x) where {K} = chain(x, Val(K))

nspec(f) = sum(m -> (s = m.specializations; s isa Core.MethodInstance ? 1 : count(!isnothing, s)), methods(f))

for K in (16, 32, 64, 128)
    prepare_gradient_cache(ChainF{K}(), 1.0)
    println("K=$K  is_primitive_spec=$(nspec(Mooncake.is_primitive))")
end
```

On current `main` the specialisation count grows with `K` (≈594 → 824); a genuinely `@nospecialize`d method would stay roughly constant. The same effect shows up on real models — e.g. `to_submodel` chains of `K` DynamicPPL submodels — where it scaled to a few thousand specialisations and tens of seconds of build time.

## Suggested fix

Despecialising naively turns out to need care, because of two interacting constraints:

1. **Free type variables.** Simply relaxing the slot to `::Type{<:Tuple}` / `::Type` makes `is_primitive` (and `build_primitive_rrule`/`build_primitive_frule`, which are `::Type{<:Tuple}`) stop matching signatures that contain a free type variable, e.g. the `jl_type_unionall` foreigncall sig `Tuple{…, TypeVar, Type{AbstractArray{a,1}}}`. Covered by `typevar_tester`.
2. **Nested AD.** `is_primitive` is already a forward-mode `@zero_derivative` primitive, so forward-over-reverse (HVP/Hessian) doesn't differentiate through its `try`/`catch`. But that declaration types its `sig` argument as `Type`; once `sig` is despecialised to statically `Any`, the declaration stops matching and forward mode descends into the body (`:the_exception` is unsupported in forward mode). Covered by the nested-AD tests in `interface.jl`.

Handling both takes three small coordinated changes:

- `is_primitive` (both methods): drop the `where {Tsig<:Tuple}` static parameter → untyped `@nospecialize sig`.
- `build_primitive_frule` / `build_primitive_rrule` fallbacks: `::Type{<:Tuple}` → `@nospecialize(sig::Type)` (constraint 1).
- the `is_primitive` `@zero_derivative` declaration's `sig` argument: `Type` → `Any` (constraint 2).

With these, `is_primitive` specialisations collapse to a small constant, warmed-marginal rule-build cost drops ~30% (cold ~−23%), and gradients / forward-over-reverse HVP & Hessian (type-stability and allocations) are unchanged. A regression test guarding the specialisation count is added in `test/interpreter/contexts.jl`.

## Scope

After `is_primitive` there's no comparable single offender — the rest of the per-signature work (`tuple_map`, `lazy_zero_rdata`, `DerivedRule`, `tangent_type`, …) is essential (runtime-hot or genuinely type-specific). Further build-time reductions would be structural: hand-written primitive rules at numeric-kernel boundaries, and reducing distinct-type proliferation in the primal.

yebai:
`is_primitive` is called once per call-node signature during the AD transform, so specialising it makes compile time scale with the number of distinct signatures in the typed primal IR. On deeply composed DynamicPPL models, this dominates: its specialisation count grows into the thousands.

The `where {Tsig<:Tuple}` static parameter defeated the existing `@nospecialize sig`. Drop it (untyped `@nospecialize sig`) so the method is genuinely despecialised. Two supporting changes keep behaviour identical:

- Broaden the `build_primitive_frule`/`build_primitive_rrule` fallbacks from `::Type{<:Tuple}` to `@nospecialize(sig::Type)`, so signatures containing a free type variable (e.g. the `jl_type_unionall` foreigncall sig) still match once `is_primitive` no longer normalises them via the static parameter.
- Broaden the `is_primitive` zero-derivative declaration's `sig` argument from `Type` to `Any`. `is_primitive` is a forward-mode primitive so that nested forward-over-reverse AD does not differentiate through its `try`/`catch`; with a despecialised (statically `Any`) `sig` argument, the previous `Type` bound no longer matched, which would otherwise re-expose the `catch` to the forward-mode transform.

Specialisations collapse to a small constant (no growth with model size); warmed-marginal rule-build cost drops ~30% (from 5x to 3.5x primal compile time). Reverse/forward dispatch, gradients, and forward-over-reverse HVP/Hessian type-stability and allocations are unchanged.

Fix https://github.com/chalk-lab/Mooncake.jl/issues/1222

### Documentation Preview

Mooncake.jl documentation for PR #1223 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1223/

### Performance

Performance Ratio:
```
│              sum_1000 │ 190.0 ns │     1.43 │        1.37 │   0.632 │        3.16 │   5.91 │
│             _sum_1000 │ 962.0 ns │     6.76 │        1.03 │  3610.0 │        37.0 │   1.07 │
│          sum_sin_1000 │   7.4 μs │     3.78 │        1.35 │    1.44 │        10.7 │   1.74 │
│         _sum_sin_1000 │  6.17 μs │     3.66 │        1.87 │   249.0 │        12.8 │   2.11 │
│              kron_sum │ 223.0 μs │     12.4 │        3.16 │    22.0 │       306.0 │   18.5 │
│         kron_view_sum │ 296.0 μs │     11.7 │        5.09 │    24.4 │       311.0 │   12.5 │
│ naive_map_sin_cos_exp │  2.35 μs │     3.03 │        1.52 │ missing │        7.52 │   2.14 │
│       map_sin_cos_exp │  2.29 μs │     3.61 │        1.66 │    1.52 │        6.69 │    2.7 │
│ broadcast_sin_cos_exp │  2.45 μs │     3.04 │        1.58 │    3.87 │        1.38 │   2.05 │
│            simple_mlp │ 379.0 μs │     4.62 │        2.62 │    2.04 │        8.55 │   2.73 │
│                gp_lml │ 394.0 μs │     5.74 │        1.75 │    2.89 │     missing │   3.03 │
│    large_single_block │ 421.0 ns │     5.24 │        1.93 │  4630.0 │        32.7 │   2.05 │
```

### DIFF
--- a/src/Mooncake.jl
+++ b/src/Mooncake.jl
@@ -109,7 +109,7 @@ callable `struct` with type parameters which are the result of this computation.
 context, the motivation for using this function is the same as that of using staged
 programming (e.g. via `@generated` functions) more generally.
 """
-build_primitive_frule(::Type{<:Tuple}) = frule!!
+build_primitive_frule(@nospecialize(sig::Type)) = frule!!
 
 """
     rrule!!(f::CoDual, x::CoDual...)
@@ -166,7 +166,7 @@ callable `struct` with type parameters which are the result of this computation.
 context, the motivation for using this function is the same as that of using staged
 programming (e.g. via `@generated` functions) more generally.
 """
-build_primitive_rrule(::Type{<:Tuple}) = rrule!!
+build_primitive_rrule(@nospecialize(sig::Type)) = rrule!!
 
 #! format: off
 @stable default_mode = "disable" default_union_limit = 2 begin
--- a/src/interpreter/contexts.jl
+++ b/src/interpreter/contexts.jl
@@ -154,9 +154,7 @@ Observe that `is_primitive` returns `false` for the world age prior to declaring
 primitive, but `true` afterwards. For more information on Julia's world age mechanism, see
 https://docs.julialang.org/en/v1/manual/methods/#Redefining-Methods .
 """
-function is_primitive(
-    ctx::Type{MinimalCtx}, mode::Type{<:Mode}, sig::Type{Tsig}, world::UInt
-) where {Tsig<:Tuple}
+function is_primitive(ctx::Type{MinimalCtx}, mode::Type{<:Mode}, sig, world::UInt)
     @nospecialize sig
     try
         Base.invoke_in_world(world, _is_primitive, ctx, mode, sig)::Bool
@@ -170,9 +168,7 @@ function is_primitive(
     end
 end
 
-function is_primitive(
-    ctx::Type{DefaultCtx}, mode::Type{<:Mode}, sig::Type{Tsig}, world::UInt
-) where {Tsig<:Tuple}
+function is_primitive(ctx::Type{DefaultCtx}, mode::Type{<:Mode}, sig, world::UInt)
     @nospecialize sig
     # This function returns `true` if the method is a primitive in either 
     # `DefaultCtx` _or_ `MinimalCtx`.
--- a/src/rules/high_order_derivative_patches.jl
+++ b/src/rules/high_order_derivative_patches.jl
@@ -3,7 +3,7 @@
 @zero_derivative MinimalCtx Tuple{
     typeof(build_rrule_checks),MooncakeInterpreter,Any,Bool,Bool,Bool
 }
-@zero_derivative MinimalCtx Tuple{typeof(is_primitive),Type,Type{<:Mode},Type,UInt}
+@zero_derivative MinimalCtx Tuple{typeof(is_primitive),Type,Type{<:Mode},Any,UInt}
 
 @is_primitive MinimalCtx Tuple{
     typeof(build_derived_rrule),MooncakeInterpreter{C},Any,Any,Bool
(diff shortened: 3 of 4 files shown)


sunxd3 (approved):
LGTM, thanks!

PR #1225, opened 2026-06-23
GPU rules for `vcat`, `cat`, `hcat`, `permutedims`.

AstitvaAggarwal:
### Documentation Preview

Mooncake.jl documentation for PR #1225 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1225/

### Performance

Performance Ratio:
```
│                   sum_1000 │ 180.0 ns │     1.62 │        1.72 │   0.667 │        3.28 │   6.23 │
│                  _sum_1000 │ 971.0 ns │      6.8 │        1.01 │  1630.0 │        41.6 │   1.06 │
│               sum_sin_1000 │  7.23 μs │     3.51 │        1.33 │    1.46 │        10.9 │   1.72 │
│              _sum_sin_1000 │  6.58 μs │     3.46 │        1.72 │   211.0 │        12.2 │   1.98 │
│                   kron_sum │ 219.0 μs │     12.8 │        3.26 │    8.21 │       363.0 │   18.6 │
│              kron_view_sum │ 301.0 μs │     10.7 │        4.92 │    23.7 │       308.0 │   11.7 │
│      naive_map_sin_cos_exp │  2.39 μs │     2.99 │        1.49 │ missing │         7.5 │   2.08 │
│            map_sin_cos_exp │  2.28 μs │      3.7 │        1.68 │    1.56 │         6.7 │   2.69 │
│      broadcast_sin_cos_exp │  2.34 μs │     3.15 │        1.67 │    4.19 │        1.46 │   2.13 │
│                 simple_mlp │ 375.0 μs │     4.51 │        2.52 │     1.4 │        8.24 │   2.77 │
│                     gp_lml │ 389.0 μs │     5.61 │        1.66 │    4.69 │     missing │   3.25 │
│ turing_broadcast_benchmark │  1.66 ms │     6.94 │        3.58 │ missing │        43.5 │   2.36 │
│         large_single_block │ 501.0 ns │     7.24 │        1.72 │  3930.0 │        26.7 │   1.72 │
```


test/integration_testing/flux/flux.jl
```diff
@@ -150,11 +150,11 @@ const TEST_MODELS = [
         "Maxout(Dense(5 => 4, tanh), 3)",
     ),
     (
-        _gpu_disabled,
+        _gpu_enabled,
         SkipConnection(Dense(2 => 2), vcat),
         randn(Float32, 2, 3),
         "SkipConnection(Dense(2 => 2), vcat)",
-    ),  # vcat tangent mixes CPU/GPU: Illegal conversion of CUDA.DeviceMemory to Ptr{Float32}
```
yebai, line 157: The rules for `vcat/hcat/cat` on CuArrays are left unimplemented on purpose — if the underlying issue has been fixed, let's investigate what changed.

AstitvaAggarwal: So I tested further by disabling the rules, the three functions being diffed fail as the following places:
- `vcat` -> `Base._typed_vcat` -> `CUDAKernels.##_#6` -> `kernel_compile` -> `cufunction` (contains `Base.@lock (try/finally`)
- `hcat` -> `Base._typed_hcat` -> `CUDAKernels.##_#6` -> `kernel_compile` -> `cufunction`
- `cat` -> `Base._copy_or_fill!` -> `CUDAKernels.##_#6` -> `kernel_compile` -> `cufunction`

The root cause is `Base.@lock` inside `cufunction` (used for thread-safe kernel caching), which Mooncake can't handle in reverse mode due to the `try/finally` block. I tried adding a `@zero_derivative` rule for `cufunction` (to skip past the `@lock`), and then `prepare_cuda_state` (skip `cuCtxGetCurrent` `ccall` inside). After both, `CUDAKernels.##_#6` itself then hits `cuOccupancyMaxPotentialBlockSize` (raw ccall). (this continues)

The explicit rules at the `vcat/hcat/cat` level might be the right intervention point. But in case Im seeing something incorrectly, happy to be corrected to the right approach ahead.

AstitvaAggarwal: for clarity, I'm of the understanding that the underlying issue was handling the relevant `CUDA` `ccall`s along these paths. Is it preferred to mark all these calls as non differentiable/with their respective rules?

ext/MooncakeCUDAExt/MooncakeCUDAExt.jl
```diff
@@ -1379,27 +1379,163 @@ function rrule!!(::CoDual{typeof(reduce)}, op::CoDual, x::CoDual{<:CuArray})
...
+
+for _fn in (:vcat, :hcat)
+    @eval @is_primitive(MinimalCtx, Tuple{typeof($_fn),CuMaybeComplexArray,AbstractArray})
+    @eval function frule!!(
+        ::Dual{typeof($_fn)}, ::Dual{<:CuMaybeComplexArray}, ::Dual{<:AbstractArray}
+    )
+        _throw_mixed_cat_error($_fn)
+    end
+    @eval function rrule!!(
+        ::CoDual{typeof($_fn)}, ::CoDual{<:CuMaybeComplexArray}, ::CoDual{<:AbstractArray}
```
sunxd3: this only cover the two argument case

ext/MooncakeCUDAExt/MooncakeCUDAExt.jl
```diff
@@ -1379,27 +1379,163 @@ function rrule!!(::CoDual{typeof(reduce)}, op::CoDual, x::CoDual{<:CuArray})
...
+    primals = map(primal, args)
+    fdatas = map(tangent, args)
+    out = cat(primals...; pkw...)
+    dy_out = zero(out)
+    pb!!(::NoRData) = (
+        _cu_concat_pb!(fdatas, dy_out, dim);
+        (NoRData(), NoRData(), NoRData(), map(_ -> NoRData(), args)...)
+    )
+    return CoDual(out, dy_out), pb!!
+end
```
sunxd3, line 1527: `cat`'s `dim` argument could take `Tuple` and `Val`, this assumes dims to be a single integer?

ext/MooncakeCUDAExt/MooncakeCUDAExt.jl
```diff
@@ -1379,27 +1379,163 @@ function rrule!!(::CoDual{typeof(reduce)}, op::CoDual, x::CoDual{<:CuArray})
...
+@inline function _cu_concat_pb!(fdatas, dy_out, dim)
+    offset = 0
+    for i in eachindex(fdatas)
+        n = size(fdatas[i], dim)
+        fdatas[i] .+= selectdim(dy_out, dim, (offset + 1):(offset + n))
+        offset += n
     end
 end
 
+@is_primitive(MinimalCtx, Tuple{typeof(vcat),Vararg{CuMaybeComplexArray}})
```
sunxd3: old error message functions would match `Numbers`, consider add them back for better error message?

ext/MooncakeCUDAExt/MooncakeCUDAExt.jl
```diff
@@ -1452,54 +1464,136 @@ function rrule!!(
...
+# Pure-GPU calls never reach here — Vararg{CuMaybeComplexArray} is more specific and wins.
+@inline _throw_if_any_cu(fn, ::CuMaybeComplexArray, _...) = _throw_mixed_cat_error(fn)
+@inline _throw_if_any_cu(fn, ::AbstractArray, rest...) = _throw_if_any_cu(fn, rest...)
+@inline _throw_if_any_cu(fn) = nothing
+
+# Vararg{AbstractArray} rules for vcat, hcat, and cat. These are less specific than the
+# Vararg{CuMaybeComplexArray} rules above, so pure-GPU calls still dispatch to those rules.
+# For mixed GPU/CPU calls _throw_if_any_cu detects the CuArray at any position and throws.
+# For pure-CPU calls the rules compute correct pullbacks using the same _cu_concat_pb! helper.
+@is_primitive(MinimalCtx, Tuple{typeof(vcat),Vararg{AbstractArray}})
```
sunxd3: I am not sure `AbstractArray` is a good idea: could you verify `view`, adjoints etc. not breaking on CPU with the extension loaded?

AstitvaAggarwal: The PR is still WIP, will request a review when ready

AstitvaAggarwal: changed the PR to draft as well for this, I think this needs more thought. sorry for the delay

ext/MooncakeCUDAExt/MooncakeCUDAExt.jl
AstitvaAggarwal: fixed

ext/MooncakeCUDAExt/MooncakeCUDAExt.jl
AstitvaAggarwal: fixed

sunxd3 (approved):
Great work, we are really close.

Two smaller items left worth fixing:

## 1. `CuMaybeWrappedArray` declares `ComplexF16` wrappers the rule can't handle

`ext/MooncakeCUDAExt/MooncakeCUDAExt.jl:127-131`

The Adjoint/Transpose arms use `CuFloatOrComplex`, which includes `Complex{Float16}`:

```julia
Adjoint{<:CuFloatOrComplex,<:CuMaybeComplexArray},
Transpose{<:CuFloatOrComplex,<:CuMaybeComplexArray},
```

But these rules canonicalise through `arrayify`, whose Adjoint/Transpose methods
(`src/rules/blas.jl`) are bounded `Union{IEEEFloat,BlasFloat}`, and `ComplexF16` is not in
that set. So `Adjoint{ComplexF16}`/`Transpose{ComplexF16}` — and bare `CuArray{ComplexF16}`
(`CuMaybeComplexArray` on L128) — are declared primitives but then fail inside the rule
(`MethodError` from `arrayify`, or a kernel exception), instead of falling through to the
interpreter like other unsupported cases.

Fix: bound the complex arms to `ComplexF32/ComplexF64` so the primitive declaration matches
what the rule can actually canonicalise, mirroring the `BlasFloat` bound already used on the
`SubArray` arm (L131).

Also fix the comment at L119-126, which is inaccurate on two points:
- The example `vcat(view(cu_f16_vec, 1:4), y)` — a 1-D contiguous `view` of a CuArray
  returns a `CuArray`, not a `SubArray`, so it is unaffected by the `SubArray` bound and
  works. Use a strided/2-D view (e.g. `view(cu_f16_mat, 1:2, :)`) to illustrate the point.
- "excludes Float16/Complex{Float16}, unlike its Adjoint/Transpose/Symmetric siblings" —
  the siblings do handle `Float16` but not `Complex{Float16}` (same `Union{IEEEFloat,
  BlasFloat}` bound). Tightening the complex arms above makes this consistent.

## 2. The `f_view` test does not exercise what it claims

`test/ext/cuda/cuda.jl:1325-1327`

```julia
# Float16 SubArrays are excluded from CuMaybeWrappedArray (Finding 3);
# confirm this now falls through cleanly instead of erroring.
f_view(x, y) = sum(vcat(view(x, 1:2), y))
```

`view(x16, 1:2)` on a 1-D contiguous CuArray returns a `CuArray`, not a `SubArray`, so this
hits the normal pure-GPU `vcat` rule and duplicates the "Bare Float16" test just above it
(L1314). It never touches the SubArray-exclusion path the comment describes.

Fix: to test the exclusion, use a strided view that stays a `SubArray`
(e.g. `view(cu_f16_mat, 1:2, :)`) and assert it errors (`@test_throws`); otherwise remove
the test as redundant.

sunxd3 (approved):
LG, thanks!

sunxd3:
Don't release yet until we look into the CI failures
c.f. https://github.com/TuringLang/Bijectors.jl/pull/482

PR #1229, opened 2026-06-30
drop deprecated `NNlib` API

### LINKED ISSUE #1230: Fix NNLib integration test errors
see : https://github.com/chalk-lab/Mooncake.jl/actions/runs/28446894455/job/84354551549

AstitvaAggarwal:
The four `softmax`/`logsoftmax` pullbacks called `∇softmax_data`/`∇logsoftmax_data`, which are deprecated in NNlib 0.9.37 (https://github.com/FluxML/NNlib.jl/pull/718). This causes 25 JET test failures because the deprecation warning triggers runtime dispatch via `@invokelatest` and `StackTraces.lookup`.

 This PR's fix : version-split the pullbacks using `@static` if `pkgversion(NNlib) >= v"0.9.37"` :

- `NNlib ≥ 0.9.37` : call `NNlib.∇softmax!`/`∇logsoftmax!` directly, which dispatches to `cudnnSoftmaxBackward` on GPU, `1.5`–`3.4×` faster than broadcasting for `dims=1` (benchmarks in https://github.com/FluxML/NNlib.jl/pull/718). JET-clean since these are not deprecated.

- `NNlib < 0.9.37` : inline the gradient formulas directly. Calling `∇softmax_data` always takes its fast `else`-branch in `Mooncake` (`within_gradient` is always false), but still allocates a full N-element intermediate before returning. The inlined formula uses `mapreduce` to fuse the multiply into the reduction kernel without materialising `dy.*y`, writing directly into `dx` with only a small sum allocation.
 See: https://github.com/JuliaGPU/GPUArrays.jl/blob/110608541b4bd47948ed03b9fd9aa64e46e780cb/src/host/mapreduce.jl#L58-L71

Also fixes : #1230 

---

Xianda's edit:

Astitva's fix is sound, but I opted for approach with least changes needed: older code are kept as it (they won't be tested any more as the CI will be using the most up-to-date version of NNlib). Functions like `∇logsoftmax_data` were not public functions, albeit being [recommended by NNlib error message](https://github.com/FluxML/NNlib.jl/blob/4ed25976bb6c67a4f2d04630e26fd1f37fbde6eb/src/deprecations.jl#L10-L34). They are replaced by public functions that do the same thing. 

Compat with NNlib pre 0.9.37 is a tradeoff between cleanliness and backward compat. We opted for backward compat in this incident as it seems the Julia ecosystem might depend on older versions of NNlib. We should bump the compat and remove the method check once the ecosystem move on in the future.

### Documentation Preview

Mooncake.jl documentation for PR #1229 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1229/

### Performance

Performance Ratio:
```
│              sum_1000 │ 220.0 ns │     1.32 │        1.32 │   0.545 │        2.64 │    5.1 │
│             _sum_1000 │ 972.0 ns │     6.85 │        1.05 │  2910.0 │        38.4 │   1.06 │
│          sum_sin_1000 │  7.19 μs │     3.47 │        1.34 │    1.49 │        11.0 │   1.76 │
│         _sum_sin_1000 │   6.1 μs │     3.58 │        1.84 │   245.0 │        13.3 │   2.15 │
│              kron_sum │ 281.0 μs │     10.9 │        3.25 │    11.8 │       308.0 │   15.9 │
│         kron_view_sum │ 323.0 μs │     11.0 │        4.93 │    22.6 │       322.0 │   11.8 │
│ naive_map_sin_cos_exp │   2.5 μs │     2.99 │        1.46 │ missing │         7.1 │   1.99 │
│       map_sin_cos_exp │  2.34 μs │     3.54 │        1.59 │    1.55 │        6.58 │   2.68 │
│ broadcast_sin_cos_exp │  2.47 μs │     3.12 │        1.58 │    4.21 │        1.38 │   2.01 │
│            simple_mlp │ 394.0 μs │     4.95 │        2.75 │    1.55 │        8.39 │   2.68 │
│                gp_lml │ 184.0 μs │     12.8 │         3.9 │    10.0 │     missing │   7.54 │
│    large_single_block │ 421.0 ns │      5.6 │        1.93 │  4400.0 │        32.9 │   2.05 │
```


sunxd3:
I wouldn't be offended if we just bump the compat version of NNLib to 0.9.37, is there a reason we must maintain the backward compat with NNlib?

AstitvaAggarwal:
Just my opinion but I wouldnt want `Mooncake` to be limited by `NNlib`'s compatibility with other packages. `0.9.37` is pretty recent as well, and any `Mooncake` version should work regardless of `NNlib`s version . Additionally, the old depreciated calls were wasting some allocations as well.

sunxd3 (approved):

sunxd3:
@AstitvaAggarwal thanks!

ext/MooncakeNNlibExt.jl
```diff
@@ -123,7 +115,15 @@ function Mooncake.rrule!!(
     res = zero_fcodual(y)
     function logsoftmax_pb!!(::NoRData)
         _, dx = arrayify(x)
-        dx .+= ∇logsoftmax_data(tangent(res), y; dims=1)
+        dy = tangent(res)
+        # TODO: Drop the fallback once Mooncake only supports NNlib >= 0.9.37.
+        # Until then, use the public softmax backpass API when available and delegate
+        # NNlib < 0.9.37 to the legacy `_data` helpers rather than owning the formulas here.
```
sunxd3:
```suggestion
        # NNlib < 0.9.37 to the legacy `_data` helpers.
        # See https://github.com/chalk-lab/Mooncake.jl/pull/1229 for more context.
```

ext/MooncakeNNlibExt.jl
```diff
@@ -123,7 +115,15 @@ function Mooncake.rrule!!(
     res = zero_fcodual(y)
     function logsoftmax_pb!!(::NoRData)
         _, dx = arrayify(x)
-        dx .+= ∇logsoftmax_data(tangent(res), y; dims=1)
+        dy = tangent(res)
+        # TODO: Drop the fallback once Mooncake only supports NNlib >= 0.9.37.
```
sunxd3:
```suggestion
        # TODO: Drop the fallback once NNlib >= 0.9.37 is more widely supported.
```

Issue #1149, opened 2026-04-18
Crash with sparse matrix when friendly_tangents=true in Mooncake 0.5.25 (works in 0.5.24)


Parvm1102:
The sparse matrix inputs fail when using `friendly_tangents = true` with Mooncake 0.5.25 and above.
However it works on Mooncake 0.5.24 (julia 1.12.5).

MWE:
```julia
using Mooncake, SparseArrays

A = sparse([1,2], [1,2], [1,2], 2, 2)
f(A) = sum(Float64, A.nzval)

println("forward = ", f(A))

cache = Mooncake.prepare_gradient_cache(
    f,
    A;
    config = Mooncake.Config(friendly_tangents=true),
)
out = Mooncake.value_and_gradient!!(cache, f, A)
println(out)
```
It produces this error and stacktrace:
```bash
ERROR: MethodError: no method matching zero(::Mooncake.FriendlyTangentCache{Mooncake.AsRaw, Nothing})
The function `zero` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  zero(::Type{Union{}}, Any...)
   @ Base number.jl:315
  zero(::Type{Dates.DateTime})
   @ Dates ~/.julia/juliaup/julia-1.12.5+0.x64.linux.gnu/share/julia/stdlib/v1.12/Dates/src/types.jl:458
  zero(::Type{Dates.Date})
   @ Dates ~/.julia/juliaup/julia-1.12.5+0.x64.linux.gnu/share/julia/stdlib/v1.12/Dates/src/types.jl:459
  ...

Stacktrace:
  [1] iszero(x::Mooncake.FriendlyTangentCache{Mooncake.AsRaw, Nothing})
    @ Base ./number.jl:42
  [2] _iszero
    @ ~/.julia/juliaup/julia-1.12.5+0.x64.linux.gnu/share/julia/stdlib/v1.12/SparseArrays/src/SparseArrays.jl:43 [inlined]
  [3] _noshapecheck_map
    @ ~/.julia/juliaup/julia-1.12.5+0.x64.linux.gnu/share/julia/stdlib/v1.12/SparseArrays/src/higherorderfns.jl:181 [inlined]
  [4] map
    @ ~/.julia/juliaup/julia-1.12.5+0.x64.linux.gnu/share/julia/stdlib/v1.12/SparseArrays/src/higherorderfns.jl:1187 [inlined]
  [5] macro expansion
[...]

PR #1249, opened 2026-07-10
Widen `friendly_tangents` support for `Adjoint`/`Transpose`, `Symmetric`/`Hermitian`/`SymTridiagonal`

### LINKED ISSUE #1250: `friendly_tangents` has no reconstruction support for `Adjoint`/`Transpose`
`friendly_tangent_cache`/`tangent_to_friendly_internal!!` currently only have dedicated support for `Symmetric`, `Hermitian`, and `SymTridiagonal`. Adjoint and Transpose fall through to the raw internal Tangent representation.
Unlike `Symmetric`/`Hermitian`, `Adjoint`/`Transpose` aren't lossy because every entry maps to exactly one parent entry relabeled, so there's no structural reason they can't get the same treatment.

MWE:

```
  using Mooncake, LinearAlgebra

  A = randn(3, 4)
  f(x) = sum(abs2, x)

  cache = Mooncake.prepare_gradient_cache(
      f, Adjoint(A); config=Mooncake.Config(; friendly_tangents=true)
  )
  y, (_, grad) = Mooncake.value_and_gradient!!(cache, f, Adjoint(A))

  grad
  # Mooncake.Tangent{@NamedTuple{parent::Matrix{Float64}}}(...)
  # expected: a plain 4×3 Matrix{Float64}, matching size(Adjoint(A))
```

Even with `friendly_tangents=true` set, `grad` comes back as a raw Tangent, not a plain array and so comparing it directly against another backend's gradient (e.g. `isapprox(grad, zygote_grad)`) fails with a `MethodError`, and any downstream code expecting an array-like result breaks. ([Lux's Mooncake Tests](https://github.com/LuxDL/Lux.jl/pull/1703)).

AstitvaAggarwal:
Adds `friendly_tangent_cache`/`tangent_to_friendly_internal!!` support for `Adjoint`/`Transpose`, and brings `Symmetric`/`Hermitian`/`SymTridiagonal` onto the same mechanism so all five agree on how they present a dense
gradient, whether the type appears on its own or nested inside another one of these wrappers.

- `Adjoint`/`Transpose` are exact relabellings of their parent (just moved, and conjugated for `Adjoint`), so this is safe over a plain array, however deeply nested. This also holds for a `SubArray` parent with no repeated indices, and for a `Diagonal` parent.

- `Symmetric`/`Hermitian`/`SymTridiagonal` mirror their one stored gradient value into both dense positions, matching the standard matrix calculus convention for gradients of a symmetric/Hermitian matrix (the symmetrised `G + Gᵀ`, or `G + Gᴴ` for `Hermitian`). This now holds whether the type is the top-level differentiated value or sits nested inside `Transpose`/`Adjoint`.

- Two cases stay `AsRaw`: a `SubArray` with repeated indices (structurally the same as
  the symmetric-matrix mirroring above, but with no established convention to lean on, so
  this stays conservative instead of picking one), and a triangular wrapper's implicit
  diagonal (a primal-shape constant, not a real tangent value).

- An array type `arrayify` doesn't recognise at all now logs a `@debug` hint pointing at what's missing, instead of looking identical to a deliberately excluded structural type.

Handles both matrix and vector parents, and allocates buffers via `similar(x.parent, ...)`/`similar(x, ...)` so they stay on the primal's device.
Tests split out into their own file, `test/tangents/friendly_tangent.jl`.

closes #1250

### Documentation Preview

Mooncake.jl documentation for PR #1249 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1249/

### Performance

Performance Ratio:
```
│                   sum_1000 │ 170.0 ns │     1.77 │        1.77 │   0.706 │        3.48 │    6.6 │
│                  _sum_1000 │ 982.0 ns │     6.69 │        1.01 │  4260.0 │        49.3 │   1.05 │
│               sum_sin_1000 │  7.22 μs │     3.42 │        1.36 │    1.48 │        13.4 │   1.77 │
│              _sum_sin_1000 │  6.24 μs │     3.53 │        1.85 │   226.0 │        13.4 │   2.09 │
│                   kron_sum │ 220.0 μs │     13.1 │        3.59 │    19.9 │       321.0 │   19.7 │
│              kron_view_sum │ 308.0 μs │     10.8 │        4.99 │    24.2 │       302.0 │   9.66 │
│      naive_map_sin_cos_exp │  2.35 μs │     3.08 │        1.57 │ missing │        8.42 │   2.12 │
│            map_sin_cos_exp │  2.34 μs │     3.45 │        1.67 │    1.45 │        6.62 │   2.63 │
│      broadcast_sin_cos_exp │  2.41 μs │     3.24 │        1.56 │    3.79 │         1.4 │   2.04 │
│                 simple_mlp │ 349.0 μs │     5.58 │        2.92 │    2.22 │        9.77 │   2.84 │
│                     gp_lml │ 401.0 μs │     5.84 │        1.75 │    2.62 │     missing │    3.1 │
│ turing_broadcast_benchmark │  1.67 ms │      6.9 │        3.54 │ missing │        38.3 │   2.39 │
│         large_single_block │ 421.0 ns │     7.04 │        1.95 │  4740.0 │        42.7 │   2.05 │
```


sunxd3:
@AstitvaAggarwal thanks! The semantics look right to me (dense output shaped like the wrapper, `conj` for `Adjoint` only), tests are solid. Two things:

The cache methods dispatch on eltype only, but the conversion assumes `val(tangent.fields.parent)` is an array — only true for `Array`/`CuArray` parents. Something like `adjoint(view(A, :, 1:2))` or `Transpose(Diagonal(v))` works on `main` with `friendly_tangents=true` (raw path), but throws at the first gradient call with this PR:

```
MethodError: no method matching permutedims!(::Matrix{Float64}, ::Mooncake.Tangent{@NamedTuple{parent::…, indices::…}}, (2, 1))
```

Could you add a guard so these fall back to `AsRaw`, similar to the #1149 fix? e.g.

```julia
tangent_type(typeof(parent(x))) <: AbstractArray ||
    return FriendlyTangentCache{AsRaw}(nothing)
```

plus a regression test (e.g. `adjoint(view(randn(4, 3), :, 1:2))` returns `AsRaw` and converts without error).

Also, have a look at `arrayify` in `src/rules/blas.jl` — `arrayify(x::Adjoint, dx)` already gives `adjoint(parent_dx)`, which transposes, conjugates (`Adjoint` only, not `Transpose`), and handles vector parents (the `(1, N)` shape) in one go. It's also how the rules canonicalise these tangents, so we don't keep a parallel copy of the logic here. The four conversion methods then collapse into one, and the two cache methods into one via `LinearAlgebra.AdjOrTrans`:

```julia
@unstable function Mooncake.tangent_to_friendly_internal!!(
    tangent_as_friendly::AbstractMatrix{T}, x::LinearAlgebra.AdjOrTrans{T}, tangent
) where {T<:Union{IEEEFloat,BlasFloat}}
    _, dx = arrayify(x, tangent)
    return tangent_as_friendly .= dx
end
```

(`.=` rather than `permutedims!`/`copyto!` so wrapped GPU arrays go through broadcast.) I checked this gives identical results on the cases your tests cover, and widening to view/`Diagonal` parents later is just loosening the guard, since `arrayify` already recurses through those.

Minor: could you also mention `Adjoint`/`Transpose` in the `friendly_tangent_cache` docstring table and the reference-implementation note in `docs/src/utilities/defining_rules.md`?

AstitvaAggarwal:
@sunxd3 Thanks for the reviews! 

I have skipped the `AsRaw guard`. With `arrayify` doing the reconstruction, the cases flagged (`view`, `Diagonal`) just work correctly now instead of needing a fallback (it also fixes the nested-wrapper crash e.g. handling `Adjoint(Transpose(x))`). If I add the guard on top it would undo this. For genuinely unsupported types, `arrayify` throws its own clear error instead of crashing, so that concern's still covered. Ive also added more tests, doc updates.

yebai:
One high-level comment here is to add a new friendly_tangent test file and keep all related tests there.

sunxd3:
Thanks for the commit. I still think the guard on `tangent_type` is important. Here's a concrete case, run on the current branch: 

```julia
julia> using Mooncake, LinearAlgebra

julia> x = Transpose(UnitUpperTriangular([1.0 2.0 3.0; 4.0 5.0 6.0; 7.0 8.0 9.0]))
3×3 transpose(::UnitUpperTriangular{Float64, Matrix{Float64}}) with eltype Float64:
 1.0   ⋅    ⋅ 
 2.0  1.0   ⋅ 
 3.0  6.0  1.0

julia> rule = Mooncake.build_rrule(sum, x);

julia> _, grads = Mooncake.value_and_gradient!!(rule, sum, x; friendly_tangents=true);

julia> grads[2]
3×3 Matrix{Float64}:
 1.0  0.0  0.0
 1.0  1.0  0.0
 1.0  1.0  1.0
```

The diagonal should be `0.0`. The `1.0`s are the unit values of the tangent's `UnitUpperTriangular` re-wrap leaking through `arrayify` — not derivatives.

```julia
julia> t = Mooncake.zero_tangent(x.parent);  # tangent of the UnitUpperTriangular: all zeros

julia> _, dx = Mooncake.arrayify(x.parent, t);  # the presentation the conversion broadcasts from

julia> dx
3×3 UnitUpperTriangular{Float64, Matrix{Float64}}:
 1.0  0.0  0.0
  ⋅   1.0  0.0
  ⋅    ⋅   1.0
```

If I understand correctly, reading a reverse cotangent back through `arrayify` is only roundtrip safe when the parent's tangent is itself a plain array.

If this make sense, here are some diff you could use as reference:

<details><summary>Details</summary>
<p>

```diff
diff --git a/docs/src/utilities/defining_rules.md b/docs/src/utilities/defining_rules.md
index b65d46e11..b3445d683 100644
--- a/docs/src/utilities/defining_rules.md
+++ b/docs/src/utilities/defining_rules.md
@@ -137,12 +137,9 @@ The existing overloads for `LinearAlgebra.Symmetric`, `LinearAlgebra.Hermitian`,
 `LinearAlgebra.SymTridiagonal` in `src/rules/linear_algebra.jl` follow exactly this
 pattern and serve as reference implementations.
 
-`Symmetric`/`Hermitian` only store one triangle of the full matrix, so their tangent has no
-slot for the un-stored half. `LinearAlgebra.Adjoint`/`Transpose` have no such gap: every
-entry maps to exactly one parent entry, just transposed (and, for `Adjoint`, conjugated).
-Their overloads in the same file therefore take a different approach: they
-reconstruct the friendly gradient via `arrayify` (`src/rules/blas.jl`), the same
-canonicalisation utility the differentiation rules themselves use, rather than accessing
-`tangent.fields` directly. This lets the conversion handle any parent type `arrayify`
-already recurses through (plain arrays, views, `Diagonal`, other `Adjoint`/`Transpose`
-wrappers, ...) instead of assuming the parent's tangent is always a plain array.
+`LinearAlgebra.Adjoint`/`Transpose` opt into a plain array representation only when the
+parent's canonical tangent is itself an array. In that case every stored coordinate maps to
+exactly one logical coordinate, just transposed (and, for `Adjoint`, conjugated), so the
+conversion is isometric. Structural parents such as views, `Symmetric`, and triangular
+wrappers remain `AsRaw`: although `arrayify` can present their tangents as logical arrays,
+that presentation may drop or duplicate stored coordinates or inject implicit values.
diff --git a/src/rules/linear_algebra.jl b/src/rules/linear_algebra.jl
index 376c28bd6..ac5011836 100644
--- a/src/rules/linear_algebra.jl
+++ b/src/rules/linear_algebra.jl
@@ -56,13 +56,18 @@ end
     return tangent_as_friendly
 end
 
-# Symmetric/Hermitian/SymTridiagonal only store one triangle (or the diagonals) of the
-# full matrix, so their tangent has no slot for the un-stored half. Adjoint/Transpose have
-# no such gap: every logical entry maps to exactly one parent entry, just transposed. The
-# friendly tangent is a plain matrix shaped like the wrapper itself, reconstructed via
-# `arrayify` (src/rules/blas.jl, the same canonicalisation the differentiation rules use),
-# so parent types it already recurses through (SubArray, Diagonal, nested Adjoint/Transpose,
-# ...) are handled automatically rather than reimplemented here.
+# Adjoint/Transpose are isometric relabellings when the parent's canonical tangent is
+# already an array: every parent entry maps to exactly one logical entry, just transposed
+# (and conjugated for Adjoint). The friendly tangent is therefore a plain matrix shaped
+# like the wrapper itself.
+#
+# `arrayify` also supports structural parents such as SubArray, Symmetric, and triangular
+# wrappers, but those presentations may drop or duplicate stored coordinates or inject
+# implicit values. Reading a reverse cotangent through such a presentation is not generally
+# valid. Conservatively keep those cases AsRaw; they need representation-specific proofs.
+# The tangent_type gate below is a proxy for "arrayify has a leaf method": true for every
+# array-valued tangent type today (Array, CuArray, Memory), but an extension adding one
+# without an arrayify method would error in the conversion rather than fall back to AsRaw.
 #
 # The buffer uses similar(x.parent, ...) instead of a hardcoded Matrix, so it stays on the
 # same device as the primal. A CPU buffer can't be written to from a GPU tangent.
@@ -80,6 +85,8 @@ end
 function Mooncake.friendly_tangent_cache(
     x::LinearAlgebra.AdjOrTrans{T}
 ) where {T<:Union{IEEEFloat,BlasFloat}}
+    tangent_type(typeof(x.parent)) <: AbstractArray ||
+        return FriendlyTangentCache{AsRaw}(nothing)
     FriendlyTangentCache{AsCustomised}(similar(x.parent, T, size(x)))
 end
 
diff --git a/src/tangents/tangents.jl b/src/tangents/tangents.jl
index 821a12e70..794467e9a 100644
--- a/src/tangents/tangents.jl
+++ b/src/tangents/tangents.jl
@@ -1386,7 +1386,7 @@ caches instead.
 | Mutable struct with fields and standard `MutableTangent` | `FriendlyTangentCache{AsMutableFields}` — per-field `NamedTuple` at runtime *(non-composite, internal mode)* |
 | `AbstractDict` | `FriendlyTangentCache{AsPrimal}` *(non-composite)* |
 | `LinearAlgebra.Symmetric` / `Hermitian` / `SymTridiagonal` | `FriendlyTangentCache{AsCustomised}` *(non-composite)* |
-| `LinearAlgebra.Adjoint` / `Transpose` (of an `IEEEFloat`/`BlasFloat`-eltype array) | `FriendlyTangentCache{AsCustomised}` *(non-composite)* |
+| `LinearAlgebra.Adjoint` / `Transpose` (when the parent's canonical tangent is an array) | `FriendlyTangentCache{AsCustomised}` *(non-composite)* |
 | Everything else (Julia primitive types, float arrays, non-differentiable arrays, custom-tangent types, zero-field types) | `FriendlyTangentCache{AsRaw}` *(non-composite)* |
 
 Override to opt a type into a non-composite mode with a custom buffer:
diff --git a/test/ext/cuda/cuda.jl b/test/ext/cuda/cuda.jl
index 2c4d1de5c..6e1c8d16b 100644
--- a/test/ext/cuda/cuda.jl
+++ b/test/ext/cuda/cuda.jl
@@ -1160,6 +1160,30 @@ const _MooncakeCUDAExt = Base.get_extension(Mooncake, :MooncakeCUDAExt)
             @test grads[4] ≈ sum(mask)
         end
 
+        @testset "friendly tangents for Adjoint/Transpose of a CuArray" begin
+            # The AdjOrTrans friendly path (src/rules/linear_algebra.jl) allocates its
+            # buffer via similar(x.parent, ...) and converts with an arrayify-fed
+            # broadcast, so the friendly gradient must come back as a CuMatrix in
+            # wrapper shape without any CPU round-trip.
+            x = adjoint(CuArray(randn(rng, 2, 3)))
+            cache = prepare_gradient_cache(
+                sum, x; config=Mooncake.Config(; friendly_tangents=true)
+            )
+            val, grads = value_and_gradient!!(cache, sum, x)
+            @test val ≈ sum(Array(parent(x)))
+            @test grads[2] isa CuMatrix{Float64}
+            @test size(grads[2]) == (3, 2)
+            @test Array(grads[2]) == ones(3, 2)
+
+            # Conjugation for complex Adjoint happens in the on-device broadcast.
+            xc = adjoint(CuArray(randn(rng, ComplexF64, 2, 3)))
+            txp = CuArray(randn(rng, ComplexF64, 2, 3))
+            tx = Mooncake.build_tangent(typeof(xc), txp)
+            fx = Mooncake.tangent_to_friendly!!(xc, tx)
+            @test fx isa CuMatrix{ComplexF64}
+            @test Array(fx) == Array(txp)'
+        end
+
         # Verify that unsupported GPU operations throw user-friendly ArgumentErrors rather
         # than silent wrong answers or opaque internal crashes.  Each case exercises an
         # explicit catch-all rule that blocks an unimplemented differentiation path.
diff --git a/test/interface.jl b/test/interface.jl
index 348d2c7f9..4cd334987 100644
--- a/test/interface.jl
+++ b/test/interface.jl
@@ -266,6 +266,24 @@ end
             @test dx_sym2[2].m === dx_sym[2].m
             @test dx_sym2[2].m == grads_sym[2].m
 
+            # Regression for #1250: wrappers around a raw array have an isometric logical
+            # presentation, so their friendly gradients are plain arrays in wrapper shape.
+            A = reshape(collect(1.0:12.0), 3, 4)
+            f_wrapper = x -> sum(abs2, x)
+            for W in (LinearAlgebra.Adjoint, LinearAlgebra.Transpose)
+                x = W(A)
+                cache_wrapper = Mooncake.prepare_gradient_cache(
+                    f_wrapper, x; config=Mooncake.Config(; friendly_tangents=true)
+                )
+                value, (_, grad) = Mooncake.value_and_gradient!!(
+                    cache_wrapper, f_wrapper, x
+                )
+                @test value == sum(abs2, x)
+                @test grad isa Matrix{Float64}
+                @test size(grad) == size(x) == (4, 3)
+                @test grad == 2 .* Matrix(x)
+            end
+
             # Vector of structs: friendly gradient returns a Vector of the same struct type
             # (MWE 3 from temp/friendly_tangent_mwes.jl).
             f_vec = (v::Vector{ScalarBox}) -> sum(b.x^2 for b in v)
diff --git a/test/tangents/tangents.jl b/test/tangents/tangents.jl
index 1188195aa..66be0b30d 100644
--- a/test/tangents/tangents.jl
+++ b/test/tangents/tangents.jl
@@ -316,8 +316,8 @@ using DispatchDoctor: allow_unstable
         @test Mooncake.tangent_to_friendly!!(A, tx) == permutedims(tx_parent)
     end
 
-    # Adjoint/Transpose of a Vector give a (1, N) row shape. The parent's tangent is then
-    # a Vector too, so this exercises the copyto! path and not permutedims!.
+    # Adjoint/Transpose of a Vector give a (1, N) row shape; the parent's tangent is then
+    # a Vector too, so the broadcast maps a 1-d tangent into a 2-d buffer.
     @testset "Adjoint of a Vector" begin
         A = LinearAlgebra.Adjoint(randn(3))
         tx_parent = randn(3)
@@ -363,72 +363,22 @@ using DispatchDoctor: allow_unstable
         @test Mooncake.tangent_to_friendly!!(A, tx) == permutedims(tx_parent)
     end
 
-    # Regression: the reconstruction goes through `arrayify` (src/rules/blas.jl), which
-    # recurses through parent types it already has a method for, rather than assuming the
-    # parent's own tangent is a plain array. Before this, a non-Array parent (SubArray,
-    # Diagonal, or another Adjoint/Transpose) crashed inside permutedims!/copyto! with a
-    # MethodError, since the parent's tangent was itself a nested Tangent, not an array.
-    @testset "Adjoint of a SubArray (view)" begin
-        P = view(randn(4, 3), :, 1:2)
-        A = LinearAlgebra.Adjoint(P)
-        tx_parent_data = randn(4, 3)
-        tx_P = Mooncake.build_tangent(
-            typeof(P), tx_parent_data, NoTangent(), NoTangent(), NoTangent()
-        )
-        tx = Mooncake.build_tangent(typeof(A), tx_P)
-        cache = Mooncake.friendly_tangent_cache(A)
-        @test cache isa Mooncake.FriendlyTangentCache{Mooncake.AsCustomised}
-        dest = cache.buffer
-        @test size(dest) == size(A)
-        @test Mooncake.tangent_to_friendly_internal!!(dest, A, tx) === dest
-        expected = (tx_parent_data[P.indices...])'
-        @test dest == expected
-        @test Mooncake.tangent_to_friendly!!(A, tx) == expected
-    end
-
-    @testset "Transpose of a Diagonal (raw constructor)" begin
-        # transpose(::Diagonal)/adjoint(::Diagonal) simplify to the Diagonal itself, so the
-        # raw type constructor is needed to actually get a Transpose{...,<:Diagonal} wrapper.
-        v = randn(3)
-        D = LinearAlgebra.Diagonal(v)
-        A = LinearAlgebra.Transpose(D)
-        tx_diag_data = randn(3)
-        tx_D = Mooncake.build_tangent(typeof(D), tx_diag_data)
-        tx = Mooncake.build_tangent(typeof(A), tx_D)
-        cache = Mooncake.friendly_tangent_cache(A)
-        @test cache isa Mooncake.FriendlyTangentCache{Mooncake.AsCustomised}
-        dest = cache.buffer
-        @test Mooncake.tangent_to_friendly_internal!!(dest, A, tx) === dest
-        @test dest == LinearAlgebra.Diagonal(tx_diag_data)
-        @test Mooncake.tangent_to_friendly!!(A, tx) == LinearAlgebra.Diagonal(tx_diag_data)
-    end
-
-    @testset "nested Adjoint(Transpose(...))" begin
-        B = randn(3, 3)
-        T = LinearAlgebra.Transpose(B)
-        A = LinearAlgebra.Adjoint(T)
-        tx_B = randn(3, 3)
-        tx_T = Mooncake.build_tangent(typeof(T), tx_B)
-        tx = Mooncake.build_tangent(typeof(A), tx_T)
-        dest = Mooncake.friendly_tangent_cache(A).buffer
-        @test Mooncake.tangent_to_friendly_internal!!(dest, A, tx) === dest
-        # Adjoint(Transpose(B)) == B for real B, so the tangent map is the identity too.
-        @test dest == tx_B
-        @test Mooncake.tangent_to_friendly!!(A, tx) == tx_B
-    end
-
-    @testset "nested Transpose(Adjoint(...)) (complex)" begin
-        B = randn(ComplexF64, 3, 3)
-        Ad = LinearAlgebra.Adjoint(B)
-        A = LinearAlgebra.Transpose(Ad)
-        tx_B = randn(ComplexF64, 3, 3)
-        tx_Ad = Mooncake.build_tangent(typeof(Ad), tx_B)
-        tx = Mooncake.build_tangent(typeof(A), tx_Ad)
-        dest = Mooncake.friendly_tangent_cache(A).buffer
-        @test Mooncake.tangent_to_friendly_internal!!(dest, A, tx) === dest
-        # Transpose(Adjoint(B))[i,j] == conj(B[i,j]), so the tangent map conjugates too.
-        @test dest == conj.(tx_B)
-        @test Mooncake.tangent_to_friendly!!(A, tx) == conj.(tx_B)
+    # Isometry check: the friendly presentation must preserve the tangent inner product,
+    # so gradients read through it pair correctly with perturbations. Note this cannot
+    # catch a missing conj for Adjoint (the error cancels between fx and fy inside the
+    # real inner product); the complex value testsets above pin conjugation down.
+    @testset "Adjoint/Transpose pairing" begin
+        for T in (Float64, ComplexF64),
+            W in (LinearAlgebra.Adjoint, LinearAlgebra.Transpose)
+
+            A = W(randn(T, 3, 2))
+            tx = Mooncake.build_tangent(typeof(A), randn(T, 3, 2))
+            ty = Mooncake.build_tangent(typeof(A), randn(T, 3, 2))
+            fx = Mooncake.tangent_to_friendly!!(A, tx)
+            fy = Mooncake.tangent_to_friendly!!(A, ty)
+            friendly_dot = sum(real.(conj.(fx) .* fy))
+            @test Mooncake._dot(tx, ty) ≈ friendly_dot
+        end
     end
 
     @testset "Adjoint of degenerate-length Vectors" begin
@@ -443,42 +393,36 @@ using DispatchDoctor: allow_unstable
         end
     end
 
-    @testset "Transpose of a Symmetric (raw constructor)" begin
-        # Composes with Symmetric, an already-supported wrapper that only stores half its
-        # matrix (unlike SubArray/Diagonal, which store everything). Symmetric always
-        # presents a fully mirrored dense matrix when read, so transpose(Symmetric(...)) is
-        # a no-op on values even though the raw storage itself isn't symmetric.
-        S = LinearAlgebra.Symmetric(randn(3, 3))
-        A = LinearAlgebra.Transpose(S)
-        tx_S_data = randn(3, 3)
-        tx_S = Mooncake.build_tangent(typeof(S), tx_S_data, NoTangent())
-        tx = Mooncake.build_tangent(typeof(A), tx_S)
-        cache = Mooncake.friendly_tangent_cache(A)
-        @test cache isa Mooncake.FriendlyTangentCache{Mooncake.AsCustomised}
-        dest = cache.buffer
-        @test Mooncake.tangent_to_friendly_internal!!(dest, A, tx) === dest
-        expected = Matrix(LinearAlgebra.Symmetric(tx_S_data, Symbol(S.uplo)))
-        @test dest == expected
-        @test Mooncake.tangent_to_friendly!!(A, tx) == expected
-    end
-
-    @testset "Transpose of an arrayify-unsupported parent throws clearly" begin
-        # No AsRaw fallback guard is used here: a parent type with no arrayify method
-        # throws arrayify's own clear error instead, rather than silently degrading or
-        # crashing with a confusing MethodError.
+    @testset "structural parents fall back to AsRaw" begin
         struct MyWeirdArrayForFriendlyTangentTest{T} <: AbstractMatrix{T}
             data::Matrix{T}
         end
         Base.size(x::MyWeirdArrayForFriendlyTangentTest) = size(x.data)
-        w = MyWeirdArrayForFriendlyTangentTest(randn(3, 3))
-        A = LinearAlgebra.Transpose(w)
-        tx_w = Mooncake.Tangent((; data=randn(3, 3)))
-        tx = Mooncake.build_tangent(typeof(A), tx_w)
-        dest = Mooncake.friendly_tangent_cache(A).buffer
-        err = @test_throws ErrorException Mooncake.tangent_to_friendly_internal!!(
-            dest, A, tx
+
+        B = randn(3, 3)
+        unique_view = view(B, [1, 2], :)
+        repeated_view = view(B, [1, 1], :)
+        @test typeof(unique_view) === typeof(repeated_view)
+
+        parents = (
+            "view with unique indices" => unique_view,
+            "view with repeated indices" => repeated_view,
+            "Diagonal" => LinearAlgebra.Diagonal(randn(3)),
+            "Symmetric" => LinearAlgebra.Symmetric(randn(3, 3)),
+            "unit triangular" => LinearAlgebra.UnitUpperTriangular(randn(3, 3)),
+            "nested wrapper" => LinearAlgebra.Adjoint(randn(3, 3)),
+            "unsupported custom array" => MyWeirdArrayForFriendlyTangentTest(randn(3, 3)),
         )
-        @test occursin("A new method of `Mooncake.arrayify` is needed", err.value.msg)
+
+        @testset "$label / $W" for (label, parent) in parents,
+            W in (LinearAlgebra.Adjoint, LinearAlgebra.Transpose)
+
+            A = W(parent)
+            cache = Mooncake.friendly_tangent_cache(A)
+            @test cache isa Mooncake.FriendlyTangentCache{Mooncake.AsRaw}
+            tx = Mooncake.zero_tangent(A)
+            @test Mooncake.tangent_to_friendly!!(cache, A, tx, Mooncake.NoCache()) === tx
+        end
     end
 
     @testset "tangent_to_friendly!! routing" begin

```

</p>
</details>

AstitvaAggarwal:
@sunxd3 @yebai pushed a round covering both comments.

@sunxd3 - applied the guard as given, so structural parents like `Transpose(UnitUpperTriangular(...))` fall back to `AsRaw` instead of leaking implicit values. While checking how far it needed to go, loosened it back up for a few cases that are actually fine: nested `Adjoint`/`Transpose` over a plain array (any depth), a `SubArray` with no repeated indices, and `Diagonal` (repeated-index views and triangular wrappers still correctly stay `AsRaw`).

Bigger one: `Symmetric`/`Hermitian`/`SymTridiagonal` now mirror their stored value into both dense positions (the usual `G + Gᵀ`/`G + Gᴴ` convention), consistently whether nested or standalone, by routing both through `arrayify` - which already did this mirroring internally for `Symmetric`, just never got exposed via `friendly_tangent_cache` (also added the missing `arrayify` methods for `Hermitian`/`SymTridiagonal`). This is the part I'd like another review on, it's more of a judgment call from my side than the rest.

Unrecognised array types now log a debug hint instead of looking identical to a deliberately excluded structural type.
Also worth flagging on the repeated-index `SubArray` case: it's not actually a hard impossibility, it's the same shape of tradeoff as the `Symmetric` mirroring (one accumulated value shown at more than one position). There's just no established convention for "gradient of a repeated-index view" to lean on the way there is for symmetric matrices, so I kept it `AsRaw` rather than invent one. Made a small judgment call on this.


@yebai - moved the tests into `test/tangents/friendly_tangent.jl` as asked. Left the two `test/interface.jl` testsets where they are, they're real end-to-end tests via `build_rrule`, different in kind from the presentation-layer tests in the new file - happy to move them too if you'd rather.

PR #1262, opened 2026-07-29
Friendly gradients for `Adjoint` and `Transpose`

### LINKED ISSUE #1250: `friendly_tangents` has no reconstruction support for `Adjoint`/`Transpose`

yebai:
Fixes #1250. These wrappers only relabel entries, so `AsPrimal` can keep the wrapper around
the parent's gradient instead of returning a raw `Tangent` — no conversion hook, and nested
parents work. Non-differentiable parent eltypes stay `AsRaw` (#1149). 

### Documentation Preview

Mooncake.jl documentation for PR #1262 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1262/

### Performance

Performance Ratio:
```
│                   sum_1000 │ 180.0 ns │     1.72 │        1.67 │   0.667 │        3.23 │   6.23 │
│                  _sum_1000 │ 962.0 ns │     6.76 │        1.03 │  3260.0 │        37.2 │   1.07 │
│               sum_sin_1000 │  7.49 μs │     3.39 │        1.32 │    1.45 │        10.5 │   1.69 │
│              _sum_sin_1000 │  6.81 μs │     3.35 │        1.68 │   204.0 │        11.7 │   1.94 │
│                   kron_sum │ 206.0 μs │     13.4 │        3.25 │    24.0 │       366.0 │   20.2 │
│              kron_view_sum │ 292.0 μs │     11.4 │         5.0 │    25.1 │       314.0 │   12.2 │
│      naive_map_sin_cos_exp │  2.46 μs │     2.95 │        1.49 │ missing │        7.42 │   2.01 │
│            map_sin_cos_exp │  2.34 μs │      3.5 │        1.65 │    1.47 │        6.48 │   2.62 │
│      broadcast_sin_cos_exp │  2.52 μs │     3.06 │        1.43 │    3.74 │        1.34 │   1.94 │
│                 simple_mlp │ 342.0 μs │      4.9 │        2.77 │    1.54 │        9.08 │   2.98 │
│                     gp_lml │ 387.0 μs │     5.56 │        1.65 │    4.51 │     missing │   3.19 │
│ turing_broadcast_benchmark │  1.66 ms │     6.87 │        3.59 │ missing │        38.1 │   2.38 │
│         large_single_block │ 421.0 ns │      7.3 │        2.02 │  4640.0 │        32.1 │   2.05 │
```

### DIFF
--- a/docs/src/utilities/defining_rules.md
+++ b/docs/src/utilities/defining_rules.md
@@ -93,6 +93,7 @@ When `friendly_tangents=true` is passed to `value_and_gradient!!` or `prepare_gr
 - **`AbstractArray` with `IEEEFloat` (or complex) eltype**: plain array tangent, unchanged.
 - **Callables with no captured differentiable state**: `NoTangent()`, unchanged.
 - **`AbstractDict`**: a dict of the same type as the primal, with the same keys and gradient values.
+- **`LinearAlgebra.Adjoint` / `Transpose`** over a differentiable parent eltype: the same wrapper around the parent's gradient — an `AbstractMatrix` of `size(x)`, not a dense `Matrix`.
 - **Everything else** (primitive types, zero-field types, mutable structs with custom tangent types): raw Mooncake tangent, unchanged unless customised as described below.
 
 For example, with `friendly_tangents=false` (default), an immutable struct `Foo` with fields `a::Float64` and `b::Vector{Float64}` returns a `Mooncake.Tangent` wrapping `(a = da, b = db)`, and a mutable struct `Bar` with the same fields returns a `Mooncake.MutableTangent` wrapping `(a = da, b = db)`. With `friendly_tangents=true` both unwrap to the plain `NamedTuple` `(a = da, b = db)` where `da::Float64` and `db::Vector{Float64}`.
--- a/src/rules/linear_algebra.jl
+++ b/src/rules/linear_algebra.jl
@@ -18,6 +18,23 @@
 # tangent accumulation), and the non-stored triangle is zero-initialised. copyto! copies
 # the full data matrix (including complex entries) to dest, which is a plain Matrix{T}.
 
+# Adjoint and Transpose lose nothing — each entry is one parent entry, relabelled — so AsPrimal
+# can rebuild the wrapper around the parent's gradient, giving an `AbstractMatrix` of `size(x)`
+# rather than a raw `Tangent`. Reconstruction recurses, so `Adjoint(Symmetric(A))` works too.
+#
+# The conjugation in `Adjoint(dparent)` is right: Mooncake cotangents are real gradients
+# (∂L/∂re + i ∂L/∂im) and x[i, j] == conj(parent[j, i]), so d/dx[i, j] == conj(dparent[j, i]).
+#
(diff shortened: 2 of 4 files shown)


sunxd3 (commented):
this looks good

a corner case discussed in https://github.com/chalk-lab/Mooncake.jl/pull/1249 is when the `parent` has a structured tangent:

```julia
using Mooncake
using LinearAlgebra

B = [
    1.0 2.0
    3.0 4.0
    5.0 6.0
]

x = transpose(view(B, [1, 1], :))

rule = Mooncake.build_rrule(sum, x)
_, grads = Mooncake.value_and_gradient!!(
    rule, sum, x; friendly_tangents=true
)

dx = grads[2]

Matrix(dx)         # [2 2; 2 2]
parent(parent(dx)) # [2 2; 0 0; 0 0]
```

The gradient stored for B is `[2 2; 0 0; 0 0]`. `AsPrimal` reapplies the repeated-index view, so array op (e.g. `Matrix`) on `dx` display the values twice. `AsRaw` might make more sense for something like this, although I consider it a corner case.

yebai:
It's quite tricky to have a general solution for friendly tangents, as it is subjective and has various edge / special cases. In the longer term, we might consider removing the friendly tangent option from Mooncake, and providing tools like `as_primal` (or some other name) so users can do the conversion on their side 


```julia
_, grads = Mooncake.value_and_gradient!!(rule, sum, x)

g = Mooncake.as_primal(grads)
```
cc @ChrisRackauckas as this seems to be popping up in the SciML-verse several times, eg, `ComponentArrays`.

ChrisRackauckas:
Yes indeed, this seems to be a bit tricky right now and I'm not sure of a great way to handle all of the special array types showing up.

yebai:
I suggest we merge this PR as the edge case in https://github.com/chalk-lab/Mooncake.jl/pull/1262#pullrequestreview-4815728579 is related but not new, and rethink the friendly tangent in a separate effort.

PR #1264, opened 2026-07-29
Better rules for nnlib dropout and scatter

yebai:
### Documentation Preview

Mooncake.jl documentation for PR #1264 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1264/

### Performance

Performance Ratio:
```
│                   sum_1000 │ 190.0 ns │     1.58 │        1.58 │   0.684 │        3.32 │   6.64 │
│                  _sum_1000 │  1.07 μs │     6.16 │        1.05 │  3520.0 │        41.4 │   1.08 │
│               sum_sin_1000 │  7.42 μs │     2.49 │        1.13 │    1.65 │        11.0 │   1.74 │
│              _sum_sin_1000 │  4.76 μs │     3.73 │        2.59 │   363.0 │        17.5 │   3.02 │
│                   kron_sum │ 200.0 μs │     12.7 │        3.27 │     7.3 │       550.0 │   18.4 │
│              kron_view_sum │ 264.0 μs │     13.0 │        5.34 │    30.0 │       533.0 │   14.1 │
│      naive_map_sin_cos_exp │  2.33 μs │     2.81 │        1.48 │ missing │        7.94 │   2.06 │
│            map_sin_cos_exp │  2.23 μs │     3.39 │        1.62 │    1.57 │        7.13 │   2.65 │
│      broadcast_sin_cos_exp │  2.45 μs │      2.8 │        1.49 │    4.05 │        1.32 │   1.95 │
│                 simple_mlp │ 344.0 μs │     5.21 │        2.96 │    1.65 │        10.0 │   3.21 │
│                     gp_lml │ 165.0 μs │     13.5 │        3.91 │    12.1 │     missing │   7.86 │
│ turing_broadcast_benchmark │  1.62 ms │     7.88 │        4.11 │ missing │        54.2 │   2.21 │
│         large_single_block │ 481.0 ns │     6.08 │        2.19 │  4170.0 │        31.7 │   2.04 │
```


sunxd3:
Two issue found by agents:

## 1. Float32 tie counts saturate past 2^24, giving an invalid subgradient

The tie count in `_scatter_extremum_grads!` accumulates by repeated `+ 1.0f0` into a `Float32` array ([`ext/MooncakeNNlibExt.jl#L320-L322`](https://github.com/chalk-lab/Mooncake.jl/blob/e8bd3e341a1fc8fb84958c193d82daaf8f558126/ext/MooncakeNNlibExt.jl#L320-L322)): once a destination's count reaches 2^24, `total + 1` rounds back and every further increment is dropped, so each tied share is inflated and the shares no longer sum to 1.

```julia
using Mooncake, NNlib

n = 2^24 + 2^20                      # 17_825_792 sources, all tied at one destination
src, idx = ones(Float32, n), ones(Int, n)

f(x) = sum(NNlib.scatter(max, x, idx))
cache = Mooncake.prepare_gradient_cache(f, src)
val, grad = Mooncake.value_and_gradient!!(cache, f, src)

grad[2][1]              # 5.9604645e-8 (= 2^-24); expected 1/n = 5.6098489e-8
sum(Float64.(grad[2]))  # 1.0625; expected 1.0
```

The count saturated at 16,777,216 instead of reaching 17,825,792. At `n = 2^24` exactly the sum is still `1.0`, so the cliff is precisely Float32's integer-representability limit. The comment on L318-319 dismisses 2^24 as "beyond any real tie group", but the failing input is a ~68 MB array of bitwise-equal values, which is reachable in practice (e.g. exact ReLU zeros or init-filled tensors scattered onto few destinations). Since the Float16 path widens to Float32 (L320), it hits the same cliff.

## 2. The new view increment silently accepts a wrong-shaped cotangent

`increment_and_get_rdata!(f::AbstractArray{P}, ::NoRData, t::AbstractArray{P})` does `f .+= t` with no shape check ([`src/tools_for_rules.jl#L485`](https://github.com/chalk-lab/Mooncake.jl/blob/e8bd3e341a1fc8fb84958c193d82daaf8f558126/src/tools_for_rules.jl#L485)). A malformed ChainRules rrule returning a `(1×n)` cotangent for an `(m×n)` input fails loudly on a plain `Array` (via `increment!!`'s size assertion) but silently broadcasts into every row of a view's fdata:

```julia
using Mooncake, ChainRulesCore
using Mooncake: zero_fcodual
const CRC = ChainRulesCore

bad_sum(x) = sum(x)
function CRC.rrule(::typeof(bad_sum), x::AbstractMatrix{<:Real})
    bad_sum_pb(dy) = CRC.NoTangent(), fill(dy, 1, size(x, 2))  # wrong shape: (1, n), not (m, n)
    return sum(x), bad_sum_pb
end
Mooncake.@from_rrule Mooncake.DefaultCtx Tuple{typeof(bad_sum),AbstractMatrix{<:Base.IEEEFloat}}

# View input: no error, silently wrong gradient.
co = zero_fcodual(view(ones(2, 3), 1:2, 1:3))
out, pb = Mooncake.rrule!!(zero_fcodual(bad_sum), co)
pb(1.0)
Mooncake.tangent(co).data.parent   # [1.0 1.0 1.0; 1.0 1.0 1.0] — the (1×3) cotangent broadcast into both rows

# Plain Array input: the same malformed rule fails loudly.
co2 = zero_fcodual(ones(2, 3))
out2, pb2 = Mooncake.rrule!!(zero_fcodual(bad_sum), co2)
pb2(1.0)                           # AssertionError
```

yebai:
Thanks @sunxd3 -- good catches! Hopefully these are fixed now. 

As a general note, numerical edge cases fixed in this PR are very rare, and I am not sure they matter in practise as deep learning often deals with lower precision than here.

sunxd3:
make sense — thanks for working on this!

yebai:
@sunxd3 I reviewed this carefully and removed some spurious changes -- they were added to work around a limitation in ChainRules support (see 98aa00f). It turns out supporting abstract arrays robustly in ChainRules support may be hard. In these cases, writing native Mooncake rules is preferred.

PR #856, opened 2025-11-17
Support Julia 1.13

yebai:
Prepare Mooncake for Julia 1.13 while retaining compatibility with older supported versions.

| Upstream change | Mooncake change |
| --- | --- |
| [Julia #56201](https://github.com/JuliaLang/julia/pull/56201) fixes the compiler bug behind #319. | Skip `patch_for_319.jl` on 1.13 and use `NativeInterpreter()` for inference. Keep the workaround on older Julia; supplied interpreters still control optimization only. |
| [Julia #59165](https://github.com/JuliaLang/julia/pull/59165) represents foreigncall names as `Expr(:tuple, name[, library])`, including singleton tuples for bare names. | Accept the new representation alongside older `Core.tuple` expressions, preserving existing tuple handling. |
| [Julia #58768](https://github.com/JuliaLang/julia/pull/58768) adds indexed `memoryrefnew(::Memory, ::Int[, ::Bool])`. | Widen the existing forward/reverse rules from `MemoryRef` to `Union{Memory,MemoryRef}`. |

Require AllocCheck 0.2.6 for its non-allocating classification of `get_pgcstack_fallback`. Add non-blocking 1.13 CI lanes with prerelease selection limited to those lanes, clarify the prerelease support policy, and bump Mooncake to 0.5.54.

### Documentation Preview

Mooncake.jl documentation for PR #856 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR856/

### Performance

Performance Ratio:
Ratio of time to compute gradient and time to compute function.
Warning: results are very approximate! See [here](https://github.com/chalk-lab/Mooncake.jl/tree/main/bench#inter-framework-benchmarking) for more context.
```
┌────────────────────────────┬──────────┬──────────┬─────────────┬─────────┬─────────────┬────────┐
│                      Label │   Primal │ Mooncake │ MooncakeFwd │  Zygote │ ReverseDiff │ Enzyme │
│                     String │   String │   String │      String │  String │      String │ String │
├────────────────────────────┼──────────┼──────────┼─────────────┼─────────┼─────────────┼────────┤
│                   sum_1000 │ 180.0 ns │     1.67 │        1.56 │    1.61 │        7.07 │   7.02 │
│                  _sum_1000 │ 962.0 ns │      6.8 │        1.03 │  1390.0 │        39.4 │   1.25 │
│               sum_sin_1000 │  6.92 μs │     3.53 │        1.41 │    2.87 │        12.6 │   2.43 │
[...]

PR #1272, opened 2026-08-11
Support Julia 1.14 Compiler API changes

IanButterworth:
> **Note:** this patch was written by Claude Code, driven by me. I've verified it
> precompiles and produces correct gradients on nightly and 1.12, but I have not
> run the full test suite locally — I'm relying on CI for that. Please review
> accordingly.

Mooncake currently fails to precompile on Julia nightly (`1.14.0-DEV`), starting with:

```
ERROR: LoadError: UndefVarError: `WorldView` not defined in `Compiler`
@ src/interpreter/abstract_interpretation.jl:90
```

Three separate changes on Julia master are involved:

- **JuliaLang/julia#59413** removed `Compiler.WorldView` — caches now carry their own
  `WorldRange` — and added `Compiler.OverlayCodeCache`, the replacement wrapper for
  interpreters (like this one) that keep a global code cache alongside the local
  inference cache. The new spelling follows Julia's own `@newinterp` macro: return
  `OverlayCodeCache(globalcache, inf_cache)` from `code_cache`, and define
  `get`/`getindex`/`haskey`/`setindex!` on the cache type directly rather than on
  `WorldView{Cache}`.
- **JuliaLang/julia#60795** replaced the `Vector{InferenceResult}` inference cache with
  `Compiler.InferenceCache`. This changes the type of `MooncakeInterpreter.inf_cache`, and
  means `empty!` no longer applies to it (hence `empty_inference_cache!`, which keeps
  `empty_mooncake_caches!` working).
- **JuliaLang/julia#61714** dropped the redundant `world` argument from
  `IRInterpretationState` (it's now always taken from `get_inference_world(interp)`, which
  is what was being passed), and parametrised `IRInterpretationState.callstack` on the
  interpreter type, so it is no longer a `Vector{AbsIntState}`.

The `WorldView{BugPatchInterpreter}` methods in `patch_for_319.jl` are gated rather than
migrated: `CC.code_cache(::BugPatchInterpreter)` delegates to the wrapped
`NativeInterpreter`, so a `WorldView{BugPatchInterpreter}` is never constructed and those
methods are unreachable on any version.

Every change is version-gated (`isdefined` checks where a binding moved, `VERSION` checks
where a signature changed), so 1.10–1.13 take exactly the paths they did before.

### Verification

On nightly (`1.14.0-DEV.2894`), before: fails to precompile. After:

```
julia> gradient(x -> sum(abs2, x), AutoMooncake(), [1.0, 2.0, 3.0])
([2.0, 4.0, 6.0],)
```

Also re-checked on 1.12.6 (same result), and confirmed the `Config(empty_cache=true)` path
still works on nightly. The `Mooncake/Gradient-Of-Square-Norm` case from
[Julia-TTFX-Snippets](https://github.com/IanButterworth/Julia-TTFX-Snippets), which is how
I hit this, now runs on nightly instead of erroring during precompile.

I did not run the full test suite locally — CI coverage on this PR is the real check,
particularly for anything exercising the interpreter cache on 1.10/1.11.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -22,6 +22,7 @@ jobs:
     name: ${{ matrix.test_group }}-${{ matrix.version }}-${{ matrix.arch }}
     runs-on: ubuntu-latest
     if: github.event_name != 'schedule'
+    continue-on-error: ${{ matrix.version == '1.13' }}
     strategy:
       fail-fast: false
       matrix:
@@ -55,6 +56,7 @@ jobs:
         version:
           - 'lts'
           - '1.12'
+          - '1.13'
         arch:
           - x64
         include:
@@ -70,7 +72,7 @@ jobs:
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
-          include-all-prereleases: false
+          include-all-prereleases: ${{ matrix.version == '1.13' }}
       - uses: julia-actions/cache@v3
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
@@ -86,6 +88,7 @@ jobs:
     name: ${{matrix.test_group.test_type}}-${{ matrix.test_group.label }}-${{ matrix.version }}-${{ matrix.arch }}
     runs-on: ubuntu-latest
     if: github.event_name != 'schedule'
+    continue-on-error: ${{ matrix.version == '1.13' }}
     strategy:
       fail-fast: false
       matrix:
@@ -115,6 +118,7 @@ jobs:
         version:
           - 'lts'
           - '1.12'
+          - '1.13'
         arch:
           - x64
         include:
@@ -127,7 +131,7 @@ jobs:
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
-          include-all-prereleases: false
+          include-all-prereleases: ${{ matrix.version == '1.13' }}
       - uses: julia-actions/cache@v3
       - uses: julia-actions/julia-buildpkg@v1
       - run: |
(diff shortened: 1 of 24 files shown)


sunxd3:
@IanButterworth Thanks for starting this. Because 1.14 is still a moving target, we'll probably slow on working towards it until a rc version is up. But this is looking really promising, thanks!

IanButterworth:
Yeah makes sense. I did this so we can benchmark it on nightly. It doesn't need to be merged.

IanButterworth:
We're setting up a [julia CI job to run the TTFX-snippets](https://github.com/JuliaCI/julia-buildkite/pull/626), which includes this. Locally I've been using this branch, but I don't really want to set up CI to use custom branches. Would it be possible to get 1.14 working in a release? i.e. this PR, or something better if this isn't quite right?

Otherwise we can't track how julia performs with mooncake during the development work

yebai:
Sure, this can likely be merged together with https://github.com/chalk-lab/Mooncake.jl/pull/856

@sunxd3 can you help with that?

sunxd3:
sounds good, let's get https://github.com/chalk-lab/Mooncake.jl/pull/856/ merged first. the CIs are passing, should be close

sunxd3:
@IanButterworth I pushed some changes to enabled couple of CI on nightly. Good new is that first order tests in `basic` are mostly passing (not just compile), but second order has problems.

I am not super comfortable with merging the changes to main yet because how messy the PR looks right now.

I'll try to keep this branch up to date with main. Do you think it's okay to run the CI jobs you mentioned on this branch for a short while?

IanButterworth:
Great. I would rather not get into managing custom branches for the TTFX test set, so it can just pick Mooncacke up when nightly works again. Thanks for considering it

sunxd3:
Great, thanks for starting this anyways.

I'll take over this PR if you don't mind. I'll let you know if thing look more right.

I can start another PR on the changes you started if you rather not keeping a fork too.

IanButterworth:
Feel free to take over

PR #1277, opened 2026-08-22
Preserve primitive rules in forward-over-reverse

### LINKED ISSUE #1246: Forward-over-reverse through triangular solves reaches unsupported memmove after copy inlining

### LINKED ISSUE #1276: no-inline policy does not protect generated `rrule!!` calls
Primitive identity is lost when reverse mode rewrites a primal call into a rule call. The forward-over-reverse optimiser therefore sees:

```julia
rrule!!(CoDual(original_function, ...), ...)
```

rather than the original primitive signature.

The no-inline policy checks the actual callee signature in `ForwardMode`. Although the original function may be a forward primitive, the corresponding `rrule!!` signature is not:

```text
Core.memoryrefnew primitive in ForwardMode: true
corresponding rrule!! call primitive in ForwardMode: false
```

Consequently, generated rule calls receive no `NoInlineCallInfo`, and their implementations may be inlined, exposing internal operations or foreign calls.

## Example

```julia
using Mooncake

f(x) = begin
    y = copy(x[1:(end - 1)])
    push!(y, x[end])
    sum(y)
end

x = [1.0, 2.0, 3.0]
cache = Mooncake.prepare_hvp_cache(f, x)
Mooncake.value_and_hvp!!(cache, f, ones(3), x)
```

For this example's generated reverse-forward pass, 101 `rrule!!` call sites were found after inference; none had `NoInlineCallInfo`, and 51 were removed by inlining. These include calls associated with vector growth and memory allocation.

Related: https://github.com/chalk-lab/Mooncake.jl/issues/1246, https://github.com/SciML/SciMLSensitivity.jl/issues/1608

cc @ChrisRackauckas

yebai:
Delay inlining until after forward transformation, fixing HVPs and removing LAPACK workaround while retaining tests.

Fix https://github.com/chalk-lab/Mooncake.jl/issues/1276; Fix #1246 properly. 

### Documentation Preview

Mooncake.jl documentation for PR #1277 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1277/

### Performance

Performance Ratio:
```
│                   sum_1000 │ 150.0 ns │     1.53 │        1.61 │   0.667 │        3.27 │   6.61 │
│                  _sum_1000 │ 842.0 ns │     6.15 │        1.03 │  4300.0 │        42.0 │   1.13 │
│               sum_sin_1000 │  5.79 μs │     2.51 │        1.18 │     1.8 │        10.7 │   1.73 │
│              _sum_sin_1000 │  3.59 μs │     3.97 │        2.66 │   492.0 │        17.7 │   3.11 │
│                   kron_sum │ 186.0 μs │     11.6 │        3.57 │    7.42 │       481.0 │   22.3 │
│              kron_view_sum │ 193.0 μs │     14.4 │        6.07 │    32.3 │       579.0 │   16.8 │
│      naive_map_sin_cos_exp │  1.74 μs │     3.05 │        1.53 │ missing │        8.41 │   2.17 │
│            map_sin_cos_exp │  1.68 μs │     3.69 │        1.66 │    1.68 │        7.42 │   2.85 │
│      broadcast_sin_cos_exp │  1.75 μs │      3.3 │        1.65 │    5.02 │        1.52 │   2.25 │
│                 simple_mlp │ 281.0 μs │     5.32 │        2.84 │     1.9 │        13.5 │   3.12 │
│                     gp_lml │ 153.0 μs │     11.7 │        2.79 │    5.08 │     missing │   5.71 │
│ turing_broadcast_benchmark │  1.32 ms │     7.47 │        3.82 │ missing │        55.5 │   2.25 │
│         large_single_block │ 380.0 ns │     10.8 │        2.22 │  3660.0 │        34.5 │   2.03 │
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Mooncake"
 uuid = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 authors = ["Will Tebbutt, Hong Ge, and contributors"]
-version = "0.5.46"
+version = "0.5.47"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/src/interpreter/ir_utils.jl
+++ b/src/interpreter/ir_utils.jl
@@ -49,16 +49,22 @@ is used to represent all instructions in the working-IR layer.
 const InstVector = Vector{NewInstruction}
 
 """
-    new_inst(stmt, type=Any, flag=CC.IR_FLAG_REFINED)::NewInstruction
+    new_inst(stmt, type=Any, flag=CC.IR_FLAG_REFINED; noinline=false)::NewInstruction
 
 Create a `NewInstruction` with fields:
 - `stmt` = `stmt`
 - `type` = `type`
 - `info` = `CC.NoCallInfo()`
 - `line` = `Int32(1)`
-- `flag` = `flag`
+- `flag` = `flag`, with `CC.IR_FLAG_NOINLINE` added when `noinline=true`
 """
-function new_inst(@nospecialize(stmt), @nospecialize(type)=Any, flag=CC.IR_FLAG_REFINED)
+function new_inst(
+    @nospecialize(stmt),
+    @nospecialize(type)=Any,
+    flag=CC.IR_FLAG_REFINED;
+    noinline::Bool=false,
+)
+    noinline && (flag |= CC.IR_FLAG_NOINLINE)
     return NewInstruction(stmt, type, CC.NoCallInfo(), Int32(1), flag)
 end
 
--- a/src/interpreter/reverse_mode.jl
+++ b/src/interpreter/reverse_mode.jl
@@ -917,6 +917,8 @@ codegen which produces the forwards- and reverse-passes.
     ssa rather than each argument.
 - `debug_mode`: if `true`, run in "debug mode" -- wraps all rule calls in `DebugRRule`. This
     is applied recursively, so that debug mode is also switched on in derived rules.
+- `noinline_primitive_rules`: preserve primitive rule-call boundaries when the generated
+    reverse rule will subsequently be differentiated by forward mode.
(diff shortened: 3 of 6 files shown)


sunxd3 (approved):
Thanks, looks good!

PR #1283, opened 2026-08-28
interpreter: Reset constant fdata between rule calls

### LINKED ISSUE #1282: Captured const globals retain cotangents across rule reuse
On Julia 1.10.12, 1.11.9, and 1.12.7, reusing a derived rule can produce incorrect gradients when a const global contains differentiable mutable data.

```julia
using Mooncake, LinearAlgebra

const B = [1.0, 2.0]

function f(p, flag)
    A = LowerTriangular([p[1] 0.0; p[2] p[1]] + I)
    return sum(abs2, A \ (flag ? p : B))
end

function grad(rule, p)
    dp = zeros(length(p))
    y, pb = rule(
        Mooncake.zero_fcodual(f),
        Mooncake.CoDual(p, dp),
        Mooncake.zero_fcodual(false),
    )
    pb(one(Mooncake.primal(y)))
    return copy(dp)
end

a = [0.3, 0.5]
b = [1.5, 2.0]
rule = Mooncake.build_rrule(f, a, false)

grad(rule, a) # correct
grad(rule, b) # [-0.38349, -0.45947]
```

Finite differences at `b` give approximately `[-0.18944, -0.15360]`. Each call supplies fresh zero fdata for the explicit arguments, so the retained state is internal to the rule.

Julia places `GlobalRef(B)` directly in the `PhiNode`. `make_ad_stmts!(::IDPhiNode)` passes it through `inc_or_const`, which stores `uninit_fcodual(B)` in shared opaque-closure data. Its mutable fdata consequently has rule lifetime rather than invocation lifetime, and pullbacks accumulate stale cotangents into later calls.

The problem applies generally to non-literal constants with mutable fdata. On current `main`, `const_codual_stmt` also captures such `CoDual`s without resetting them, so non-phi const-global uses can fail similarly.

A class-level fix should retain the constant primal while resetting or materialising its derivative scratch once per rule invocation. Phi operands must reference an SSA value that dominates the phi because the reset call cannot be embedded directly in a phi operand.

yebai:
Captured constant `CoDual`s retained mutable fdata across rule calls, corrupting later gradients for read-only globals. Reset their fdata when the forward closure extracts shared data.

Fixes #1282.

### Documentation Preview

Mooncake.jl documentation for PR #1283 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1283/

### Performance

Performance Ratio:
```
│                   sum_1000 │ 140.0 ns │     1.51 │        1.64 │    2.07 │        11.4 │   8.44 │
│                  _sum_1000 │ 852.0 ns │     6.03 │        1.02 │  1420.0 │        37.5 │   1.27 │
│               sum_sin_1000 │  5.79 μs │      2.6 │        1.21 │    2.02 │        11.4 │   2.09 │
│              _sum_sin_1000 │   3.6 μs │     3.87 │        2.63 │   332.0 │        18.6 │   3.42 │
│                   kron_sum │ 158.0 μs │     4.03 │         3.6 │    8.26 │       605.0 │   17.5 │
│              kron_view_sum │ 206.0 μs │     3.66 │        5.55 │    16.0 │       568.0 │   9.58 │
│      naive_map_sin_cos_exp │  1.78 μs │     2.86 │        1.47 │ missing │        9.12 │   2.39 │
│            map_sin_cos_exp │  1.73 μs │     3.46 │        1.61 │    2.17 │        8.09 │    3.0 │
│      broadcast_sin_cos_exp │  1.87 μs │     3.05 │        1.49 │    7.85 │        1.81 │   2.32 │
│                 simple_mlp │ 254.0 μs │     5.18 │        2.86 │    1.71 │        12.9 │   3.42 │
│                     gp_lml │ 194.0 μs │      6.9 │        2.39 │    5.53 │     missing │   5.39 │
│ turing_broadcast_benchmark │  1.32 ms │     7.37 │        3.82 │ missing │        50.7 │   2.27 │
│         large_single_block │ 370.0 ns │     11.8 │        2.11 │  7530.0 │        72.2 │   3.76 │
```

### DIFF
--- a/test/interpreter/reverse_mode.jl
+++ b/test/interpreter/reverse_mode.jl
@@ -41,7 +41,7 @@ const STALE_RVS_FNS = Function[stale_rvs_mid]
 stale_rvs_dyn(x) = (STALE_RVS_FNS[1])(x)
 
 @testset "s2s_reverse_mode_ad" begin
-    @testset "const global fdata is reset between rule calls" begin
+    @testset "const global fdata is reset between rule calls (#1282)" begin
         f = S2SGlobals.const_vector_phi
         rule = build_rrule(f, [0.3, 0.5], false)
 
--- a/test/interpreter/reverse_mode.jl
+++ b/test/interpreter/reverse_mode.jl
@@ -5,6 +5,12 @@ non_const_global = 5.0
 const const_float = 5.0
 const const_int = 5
 const const_bool = true
+const const_vector = [1.0, 2.0]
+
+function const_vector_phi(p, flag)
+    A = LowerTriangular([p[1] 0.0; p[2] p[1]] + I)
+    return sum(abs2, A \ (flag ? p : const_vector))
+end
 
 # used for regression test for issue 184
 struct A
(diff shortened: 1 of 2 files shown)


test/interpreter/reverse_mode.jl
```diff
@@ -35,6 +41,21 @@ const STALE_RVS_FNS = Function[stale_rvs_mid]
 stale_rvs_dyn(x) = (STALE_RVS_FNS[1])(x)
 
 @testset "s2s_reverse_mode_ad" begin
+    @testset "const global fdata is reset between rule calls" begin
```
yebai:
```suggestion
    @testset "const global fdata is reset between rule calls (#1282)" begin
```

PR #1298, opened 2026-09-04
docs: clarify mutable aliasing limitations

### LINKED ISSUE #1295: `getfield` breaks derivative-storage aliasing for `NoTangent` parents
A `NoTangent` parent can contain a differentiable mutable field that aliases another active argument. Reading that field creates fresh derivative storage instead of reusing the argument’s seeded storage, violating the aliasing invariant and silently dropping contributions.

```julia
using Mooncake

struct Box
    x::Vector{Float64}
end

Mooncake.tangent_type(::Type{Box}) = Mooncake.NoTangent

function f(box, x)
    box.x[1] *= 2
    return x[1]
end

x = [3.0]
box = Box(x)
cache = Mooncake.prepare_gradient_cache(f, box, x)

Mooncake.value_and_gradient!!(cache, f, box, x)
```

Current result:

```julia
(6.0, (NoTangent(), NoTangent(), [1.0]))
```

The gradient for `x` should be `[2.0]`.

The generic `getfield`/`lgetfield` rules infer from the parent’s `tangent_type` that no derivative storage exists, then construct fresh storage for the extracted field via `uninit_dual` or `uninit_fdata`. The field primal still aliases `x`, but its derivative storage does not. Mutation through `box.x` therefore updates a detached buffer.

This reproduces in reverse mode, forward mode with `enable_nfwd=false`, and debug mode.

The motivating case is `DataType`:

```julia
tangent_type(DataType) === NoTangent
tangent_type(Core.SimpleVector) === Vector{Any}
```

A `DataType` is opaque, but its `parameters` field is a `Core.SimpleVector`. If opacity is transitive, projecting `T.parameters` from a `NoTangent` `DataType` must produce `NoTangent`; consulting the field’s standalone `tangent_type` incorrectly reintroduces differentiability across the opaque boundary.

Three properties conflict:

1. `tangent_type(P)` uniquely determines every valid tangent for `P`.
2. `NoTangent` is transitive opacity: every value projected from an opaque parent is also `NoTangent`.
3. Active and opaque aliases of the same mutable object may coexist and mutate safely.

This requires provenance-sensitive semantics. `tangent_type(P)` would describe the tangent representation when a `P` is active, while `NoTangent` could represent an inactive occurrence of any primal type. Consumers such as `_svec_ref` must then accept and propagate `NoTangent` rather than assume `Vector{Any}` solely from the primal type.

Mutable aliases require a separate validity rule. If the same mutable primal is reachable through active and opaque references, mutation cannot preserve both derivative states without hidden shared shadow storage. Mooncake should either retain such storage independently of the exposed tangent or reject mixed-activity mutation explicitly.

The earlier aliasing example therefore does not establish that storage should be reused. It instead exposes the need to define opacity, canonical tangent types, and mutable aliasing consistently.

For this issue, forbidding mixed active/opaque mutable aliases is sufficient to prevent silently wrong derivatives:

- Detect when mutable storage is reachable through both an active and an opaque path.
- Reject the program before mutation can occur.
- Change the regression test to expect a clear error, not gradient `[2.0]`.

This restores soundness without redesigning tangent representation or making `NoTangent` universally valid for every primal type. Literal transitive `NoTangent` semantics for immutable projections such as `DataType.parameters` remain a separate representation question. The required outcome here is that unsupported opaque/active mutable aliasing fails locally rather than returning `[1.0]`.

yebai:
Document that mutations can produce incorrect derivatives when mutable storage is shared between an active argument and an inactive path, including a `NoTangent` parent or global. Add minimal examples for both cases and clarify that read-only global access remains safe.

Fixes #1295.

### Documentation Preview

Mooncake.jl documentation for PR #1298 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1298/

### Performance

Performance Ratio:
```
│                   sum_1000 │ 201.0 ns │     1.49 │        1.55 │    1.35 │        5.23 │   6.78 │
│                  _sum_1000 │  1.08 μs │     6.06 │        1.05 │  1050.0 │        40.4 │   1.18 │
│               sum_sin_1000 │  7.43 μs │     2.51 │        1.12 │    2.69 │        11.8 │   2.05 │
│              _sum_sin_1000 │  4.62 μs │     3.83 │        2.67 │   254.0 │        18.4 │   3.21 │
│                   kron_sum │ 191.0 μs │     3.33 │        3.36 │    6.76 │       509.0 │   12.4 │
│              kron_view_sum │ 263.0 μs │      3.0 │        5.28 │    15.1 │       456.0 │   6.78 │
│      naive_map_sin_cos_exp │  2.24 μs │     2.75 │         1.5 │ missing │        9.08 │   2.34 │
│            map_sin_cos_exp │  2.24 μs │     3.33 │        1.62 │    2.05 │        7.77 │   2.94 │
│      broadcast_sin_cos_exp │  2.34 μs │     2.92 │        1.52 │    5.06 │         1.8 │    2.3 │
│                 simple_mlp │ 346.0 μs │     4.85 │        2.88 │    1.69 │        9.57 │   3.53 │
│                     gp_lml │ 364.0 μs │     4.21 │        1.75 │     3.9 │     missing │   3.41 │
│ turing_broadcast_benchmark │  1.62 ms │     7.61 │        4.05 │ missing │        43.9 │   2.21 │
│         large_single_block │ 471.0 ns │     5.27 │        1.96 │  7060.0 │        38.7 │   2.29 │
```

### DIFF
--- a/docs/src/known_limitations.md
+++ b/docs/src/known_limitations.md
@@ -74,6 +74,49 @@ Observe that while it has correctly computed the identity function, the gradient
 
 The takeaway: do not attempt to differentiate functions which modify global state. Reading globals is fine; mutating globals is not.
 
+## Mutable aliases across inactive paths
+
+Mooncake may silently return incorrect derivatives when mutable storage is reachable through a differentiable argument and an inactive path. Inactive paths include fields of `NoTangent` parents and globals, both treated as constants. Paths can receive separate derivative storage, so mutation changes the shared primal without updating every derivative. Reverse and `frule!!`-based forward modes are affected. Reading globals is safe only while aliased storage remains unmodified. See [issue #1295](https://github.com/chalk-lab/Mooncake.jl/issues/1295).
+
+Here `box.x` and `x` refer to the same vector, but `box` is inactive:
+
+```jldoctest opaque-alias
+julia> struct OpaqueBox
+           x::Vector{Float64}
+       end
+
+julia> Mooncake.tangent_type(::Type{OpaqueBox}) = NoTangent;
+
+julia> function mutate_through_box(box, x)
+           box.x[1] *= 2
+           return x[1]
+       end;
+
+julia> x = [3.0]; box = OpaqueBox(x);
+
+julia> cache = Mooncake.prepare_gradient_cache(mutate_through_box, box, x);
+
+julia> Mooncake.value_and_gradient!!(cache, mutate_through_box, box, x)
+(6.0, (NoTangent(), NoTangent(), [1.0]))
+```
+
+A global alias has the same problem even when its binding is constant:
+
+```jldoctest global-alias
+julia> const GLOBAL_ALIAS = [3.0];
+
+julia> function mutate_through_global(x)
+           GLOBAL_ALIAS[1] *= 2
+           return x[1]
+       end;
+
+julia> cache = Mooncake.prepare_gradient_cache(mutate_through_global, GLOBAL_ALIAS);
+
+julia> Mooncake.value_and_gradient!!(cache, mutate_through_global, GLOBAL_ALIAS)
+(6.0, (NoTangent(), [1.0]))
+```
+
+Both examples report `[1.0]`; the correct derivative is `[2.0]` because the mutation doubles the aliased input. Until mixed active/inactive aliases are rejected or supported, do not mutate them.
 
 ## Passing Differentiable Data as a Type
 
--- a/docs/src/known_limitations.md
+++ b/docs/src/known_limitations.md
@@ -74,49 +74,52 @@ Observe that while it has correctly computed the identity function, the gradient
 
 The takeaway: do not attempt to differentiate functions which modify global state. Reading globals is fine; mutating globals is not.
 
-## Mutable aliases across inactive paths
+## Mutable aliases involving `NoTangent` parents or globals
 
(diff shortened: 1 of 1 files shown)


docs/src/known_limitations.md
```diff
@@ -74,6 +74,49 @@ Observe that while it has correctly computed the identity function, the gradient
...
+Mooncake may silently return incorrect derivatives when mutable storage is reachable through a differentiable argument and an inactive path. Inactive paths include fields of `NoTangent` parents and globals, both treated as constants. Paths can receive separate derivative storage, so mutation changes the shared primal without updating every derivative. Reverse and `frule!!`-based forward modes are affected. Reading globals is safe only while aliased storage remains unmodified. See [issue #1295](https://github.com/chalk-lab/Mooncake.jl/issues/1295).
+
+Here `box.x` and `x` refer to the same vector, but `box` is inactive:
+
+```jldoctest opaque-alias
+julia> struct OpaqueBox
+           x::Vector{Float64}
+       end
+
+julia> Mooncake.tangent_type(::Type{OpaqueBox}) = NoTangent;
```
sunxd3: this example fails even without the `NoTangent` overwrite, because preparation builds per argument tangent separately

sunxd3 (commented):
reads sensible!

yebai:
This should be fixed now. 

The cross-argument aliasing issue is fixable and will be fixed in #1215

docs/src/known_limitations.md
yebai: Good catch.

PR #1301, opened 2026-09-05
Document that explicit invoke of a primitive bypasses its rule

### LINKED ISSUE #1300: Explicit invoke can inline through a primitive and bypass its custom rrule!!
## Summary

An explicit `invoke(f, Tuple{Float64}, x)` of a Mooncake primitive bypasses its custom rule, whereas an ordinary call uses it. Found while reviewing #856; reproduced on Julia 1.11.9, 1.12.6, 1.12.7 and 1.13.0-rc4. It is not specific to any Julia version.

## MWE

```julia
using Mooncake

@inline leaf(x) = sin(x)
Mooncake.@is_primitive Mooncake.DefaultCtx Mooncake.ReverseMode Tuple{typeof(leaf),Float64}

function Mooncake.rrule!!(::Mooncake.CoDual{typeof(leaf)}, x::Mooncake.CoDual)
    # Deliberately distinguish the custom derivative from differentiation of sin.
    return Mooncake.zero_fcodual(sin(x.x)), dy -> (Mooncake.NoRData(), 7dy)
end

ordinary(x) = leaf(x)
explicit(x) = invoke(leaf, Tuple{Float64}, x)

for f in (ordinary, explicit)
    cache = Mooncake.prepare_gradient_cache(f, 0.3)
    println(f, " => ", Mooncake.value_and_gradient!!(cache, f, 0.3))
end
```

```text
ordinary => (0.29552020666133955, (Mooncake.NoTangent(), 7.0))
explicit => (0.29552020666133955, (Mooncake.NoTangent(), 0.955336489125606))
```

The explicit-invoke gradient is `cos(0.3)`: the body of `leaf` was differentiated instead of the rule being used.

## Cause

Mooncake matches rules by argument types and assumes a call reaches the method that ordinary dispatch selects. `invoke` selects a method directly, and the Julia compiler resolves it in `abstract_invoke` during inference and inlining, before Mooncake's AD transform runs, so Mooncake never sees a call to the primitive.

## When it happens

All three must hold:

1. Code that Mooncake differentiates through contains `invoke(f, types, args...)` with a compile-time constant `types`. `@invoke` and macros that expand to `invoke` count; `Base.Math.@horner` is one, `@evalpoly` is not.
2. `Tuple{typeof(f), typeof(args)...}` matches an `@is_primitive` signature for the AD mode.
3. The `invoke` is not inside a method that Mooncake already replaced with a rule. The most common real-world use, a specialised method forwarding to the generic method of the same function with the same arguments, can therefore never trigger it.

## What happens

- **Invoked method inlined** (small methods, `@inline`, constant propagation): the body is differentiated and the rule is skipped. Silently wrong only if the rule's derivative differs from the body's, as in the MWE.
- **Invoked method not inlined** (`@noinline`, large method): the surviving call still matches the rule by argument types, so the rule is substituted even though it was written for the method dispatch would select. If `invoke` selected a different method, value and derivative are both wrong:

  ```julia
  @noinline f(x::Real) = sin(x)
  f(x::Float64) = cos(x)
  Mooncake.@is_primitive Mooncake.DefaultCtx Tuple{typeof(f),Float64}   # rule for the Float64 method
  g(x) = invoke(f, Tuple{Real}, x)                                      # primal is sin(x)
  # main, 1.12.6: value_and_gradient!! gives (cos(0.3), -sin(0.3)); truth is (sin(0.3), cos(0.3))
  ```

Neither case raises an error.

## Not affected

- `invoke` of a function without a rule: differentiated through the invoked method, correct.
- `invoke` with a non-constant `types`, and `invokelatest`: already unsupported, fail at run time.

## Ecosystem status

No site in Base, the stdlibs (Julia 1.12.6, 112 `invoke` sites) or 585 surveyed packages (549 sites) meets all three conditions. The one latent case is `Base.Math.@horner`, which expands to `invoke(evalpoly, Tuple{Any,Tuple}, x, coeffs)` and is used in StatsFuns, Distributions, LogExpFunctions and SpecialFunctions. It is harmless while `evalpoly` has no Mooncake rule; adding one would trigger the first outcome at every such site.

## Resolution

Documented as a known limitation in #1301. Intercepting the call would need an `abstract_invoke` hook, since nothing survives to the AD transform otherwise, and that is not worth overriding compiler behaviour for a shape no real code hits. Revisit if a rule for `evalpoly` or another macro-invoked function is ever added.

sunxd3:
Documents #1300 as a known limitation. closes #1300 

### Documentation Preview

Mooncake.jl documentation for PR #1301 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1301/

### Performance

Performance Ratio:
```
│                   sum_1000 │ 171.0 ns │      1.7 │        1.64 │    1.53 │        7.91 │    7.2 │
│                  _sum_1000 │ 981.0 ns │     6.69 │        1.02 │  1110.0 │        37.8 │    1.2 │
│               sum_sin_1000 │  7.05 μs │     3.59 │        1.38 │     2.8 │        13.7 │   2.22 │
│              _sum_sin_1000 │  5.89 μs │     3.83 │        1.96 │   202.0 │        14.6 │   2.49 │
│                   kron_sum │ 242.0 μs │     3.63 │        2.98 │    5.45 │       310.0 │   9.03 │
│              kron_view_sum │ 294.0 μs │     3.47 │        5.02 │    12.9 │       317.0 │    6.6 │
│      naive_map_sin_cos_exp │  2.37 μs │     3.02 │        1.51 │ missing │        9.61 │   2.52 │
│            map_sin_cos_exp │  2.39 μs │     3.43 │        1.62 │    2.05 │        6.97 │   3.19 │
│      broadcast_sin_cos_exp │   4.9 μs │     1.59 │       0.767 │     2.8 │       0.927 │   1.24 │
│                 simple_mlp │ 356.0 μs │     4.69 │        2.64 │    1.69 │        9.31 │   3.13 │
│                     gp_lml │ 393.0 μs │     4.19 │        1.73 │    4.36 │     missing │   3.37 │
│ turing_broadcast_benchmark │  1.62 ms │     7.13 │         3.6 │ missing │        40.1 │   2.47 │
│         large_single_block │ 421.0 ns │     5.71 │        2.02 │  7280.0 │        37.6 │    2.4 │
```


docs/src/known_limitations.md
```diff
@@ -205,6 +205,12 @@ true
...
+
+This typically arises when `invoke` is used to reach a generic fallback while a rule exists for the specialised method, or through macros that expand to `invoke`, such as `Base.Math.@horner`. Do not call a function that has a rule through `invoke`: call it directly, or wrap the `invoke` in a function and write a rule for that wrapper. `invoke` of a function without a rule is differentiated normally; `invoke` whose `types` argument is not a compile-time constant, and `invokelatest`, are not supported.
```
yebai:
````suggestion
## Explicit `invoke` of a primitive

An explicit `invoke` encountered during differentiation bypasses Mooncake's protection against inlining primitives.
Inlining can silently discard the custom rule; surviving calls can use the rule for ordinary dispatch, changing the value and derivative ([#1300](https://github.com/chalk-lab/Mooncake.jl/issues/1300)).

```julia
f(x::Real) = sin(x)
f(x::Float64) = cos(x)
g(x::Float64) = invoke(f, Tuple{Real}, x)
```

With a rule for `f(::Float64)`, `g` may incorrectly use that rule although its primal computes `sin(x)`.

This can occur when differentiated code explicitly invokes a generic fallback despite a rule for the specialised call.
It can also arise indirectly: [`Base.Math.@horner`](https://github.com/JuliaLang/julia/blob/v1.12.7/base/math.jl#L176-L179) expands to `invoke(evalpoly, Tuple{Any, Tuple}, ...)`.
That becomes problematic if a matching `evalpoly` rule is added; the macro alone does not trigger the limitation.
Without a matching rule, Mooncake differentiates the invoked method normally.

Simply replacing `invoke` with ordinary dispatch can change the primal's behaviour.
Wrap the `invoke` in a function with its own rule.
Calls inside an existing primitive's body are unaffected.
````

sunxd3: thanks for reviewing!

PR #1304, opened 2026-09-08
Run existing DiffTests under GC pressure to expose Julia 1.13 failure

sunxd3:
## Purpose
Capture the failing “before” state for #1303 using the existing DiffTests integration suite, not a synthetic regression.

The original DiffTests.neural_step test already ran on PR #856 and passed. After merging, the same Julia 1.13 lane failed its two forward-rule reuse assertions. The synthetic reproducer initially added here also passed in CI, so it has been removed.

The net change is two lines of CI configuration: run the existing diff_tests lane with --heap-size-hint=32M on all three Julia versions. This asks Julia to collect more aggressively; it is not a hard memory limit. No production code, test inputs, or assertions are changed. Other CI lanes are unchanged.

## Local verification
Using the existing full lane:
```sh
LABEL=diff_tests TEST_TYPE=integration_testing julia --heap-size-hint=32M --eval 'include("test/run_extra.jl")'
```

- Julia 1.10.12: 2,846/2,846 pass
- Julia 1.12.7: 2,846/2,846 pass
- Julia 1.13.0-rc4: 2,844 pass, 2 fail, both in DiffTests.neural_step forward-rule Reuse

On Julia 1.13 the first/second calls disagreed: primal 0.5 versus 0.9990183901683176, tangent 3.0e-323 versus 0.0014630377487576307. The isolated existing neural_step case also failed both reuse assertions under the same heap hint. An unmodified full-lane run without the hint failed one of those assertions locally, confirming the timing-sensitive nature of the original failure.

YAML parsing, shell-routing checks for affected and unaffected lanes, and git diff --check pass.

## CI interpretation
This is an intentionally failing before-state investigation, not a production fix. Local results do not guarantee the same failure under CI; the new CI run must establish that.

The existing Julia 1.13 continue-on-error setting is unchanged. Inspect individual DiffTests job logs rather than relying only on overall workflow status.

### Documentation Preview

Mooncake.jl documentation for PR #1304 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1304/

### Performance

Performance Ratio:
Ratio of time to compute gradient and time to compute function.
Warning: results are very approximate! See [here](https://github.com/chalk-lab/Mooncake.jl/tree/main/bench#inter-framework-benchmarking) for more context.
```
┌────────────────────────────┬──────────┬──────────┬─────────────┬─────────┬─────────────┬────────┐
[...]

PR #1305, opened 2026-09-09
Fix forward-mode GC preservation

### LINKED ISSUE #1303: Forward-mode AD drops GC.@preserve, allowing live buffers to be collected
## Summary

Mooncake's forward-mode transformation drops the lifetime guarantee of `GC.@preserve`. Temporary primal and tangent buffers can consequently be collected while generated code still holds raw pointers to their data.

This explains the Julia 1.13 `DiffTests.neural_step` failures after #856, but the underlying preservation defect is reproducible on **both Julia 1.12.7 and 1.13.0-rc4**. It is not a 1.13-only validity requirement.

## Minimal reproducer

This affects `main` after #856 was merged, at commit [9caca9f](https://github.com/chalk-lab/Mooncake.jl/commit/9caca9f37e56495a1170ac543097434471cac894).

```julia
using Mooncake

const observed = Ref(WeakRef(nothing))

Base.@noinline function observe_memory(x::Memory{Float64})
    observed[] = WeakRef(x)
    return nothing
end

Base.@noinline function is_still_alive()
    GC.gc(true)
    return observed[].value !== nothing
end

Mooncake.@zero_derivative Mooncake.DefaultCtx Tuple{typeof(observe_memory),Memory{Float64}}
Mooncake.@zero_derivative Mooncake.DefaultCtx Tuple{typeof(is_still_alive)}

function preserve_probe(x)
    y = fill(x, 5)
    observe_memory(y.ref.mem)
    return GC.@preserve y is_still_alive()
end

fwd = Mooncake.build_frule(preserve_probe, 1.25)
println("Julia ", VERSION)
println("primal: ", preserve_probe(1.25))
for i in 1:3
    result = fwd(Mooncake.zero_dual(preserve_probe), Mooncake.Dual(1.25, 1.0))
    println("forward ", i, ": ", Mooncake.primal(result))
end
```

On both tested Julia versions:

```text
primal: true
forward 1: false
forward 2: false
forward 3: false
```

Expected: all results are `true`. The backing memory must remain alive inside the preserve region. The observer holds only a weak reference; the zero-derivative rules avoid differentiating the diagnostic machinery. This reproducer requires no BLAS, unsafe memory reads, or heap-size hints.

## Cause

[`lift_gc_preservation`](https://github.com/chalk-lab/Mooncake.jl/blob/9caca9f37e56495a1170ac543097434471cac894/src/interpreter/ir_normalisation.jl#L406-L410) replaces `:gc_preserve_begin` with a call to `gc_preserve` and replaces `:gc_preserve_end` with `nothing`.

The reverse rule captures the preserved values until its pullback runs, but the [forward rule](https://github.com/chalk-lab/Mooncake.jl/blob/9caca9f37e56495a1170ac543097434471cac894/src/interpreter/ir_normalisation.jl#L354-L360) is a no-op:

```julia
frule!!(::Dual{typeof(gc_preserve)}, ::Vararg{Dual,N}) where {N} =
    zero_dual(nothing)
```

Thus forward AD erases the region. BLAS rules' `GC.@preserve args` does not compensate when those trailing arguments contain raw pointers rather than the managed owners.

## Evidence from the failing integration test

CI failed twice in `test_frule_reuse` for `DiffTests.neural_step`: [first failure](https://github.com/chalk-lab/Mooncake.jl/actions/runs/34254752848/job/102157613466), [rerun](https://github.com/chalk-lab/Mooncake.jl/actions/runs/34254752848/job/102177487485).

| Result | First call | Second call (correct) |
|---|---:|---:|
| Primal | 0.7338560145069507 | 0.9990183901683176 |
| Tangent | −0.23346424198932497 | 0.001463037748757631 |

A local native GDB trace, without modifying Mooncake rules, captured the following sequence in one failing execution:

1. The five correct weights were copied into buffer `0x7fffc0242da0`.
2. A hardware watchpoint caught the same buffer being overwritten in compiler `BitSet` code: `_growend0!` → `find_ssavalue_uses!` → `InferenceState`.
3. The first `cblas_ddot64_` then received that exact pointer containing corrupted data. The other dot-product operand was correct.
4. The second evaluation received a different buffer with the correct weights.

That execution returned a first primal of 0.5 instead of 0.9990183901683176 and also failed the tangent comparison. Another failing run's object-header dump matched Julia's swept pool-cell free-list layout. Disabling GC prevented the failure in the tested stress case.

### Why the existing case exposes this on 1.13

The generated 1.12 LLVM retains temporary primal/tangent array objects in GC slots through intervening lazy-rule calls, loading their data pointers immediately before BLAS. The 1.13 LLVM eliminates those wrappers and carries raw data pointers across intervening allocation/compilation without retaining their managed owners. Cold lazy-rule compilation provides the collection/reuse window.

The exact upstream optimization has not been bisected. The deterministic reproducer establishes Mooncake's preservation violation independently of that version difference.

## Proposed fix

Retain native preserve-begin/end regions through **forward-mode** normalization and transformation, mapping their operands to the corresponding `Dual` values so both primal and tangent storage remain rooted. Leave reverse mode's existing treatment intact.

A process-local prototype of this approach passed:

- The reproducer above on Julia 1.12.7 and 1.13.0-rc4, including repeated calls.
- The original `neural_step` case under `--heap-size-hint=32M`: **35/35** checks.
- The complete DiffTests integration group on Julia 1.13: **2,846 passed, 0 failed**.

This is an experimentally validated fix direction, not a landed patch or complete corner-case coverage. The production fix should include regression coverage for preservation of primal and tangent storage across collection.

The pointer-root recovery pass previously considered in #856 is narrower: it repairs particular foreign-call patterns, but cannot fix the general contract demonstrated above, which contains no foreign call. Warming up the rule or weakening the reuse check would hide the failure rather than fix it.

sunxd3:
Fixes #1303. Replaces the closed before-state demo #1304.

Forward AD was lowering native GC-preserve scopes to a no-op, allowing temporary primal and tangent storage to be collected while BLAS still held their raw pointers. Retain the native scope and preserve the corresponding `Dual` operands. Reverse-mode behavior is unchanged.

The regression forces GC at a pointer-use boundary and checks both owners through weak references, without dereferencing stale pointers. It also checks prepared-rule reuse and collection after return. Before the fix, both in-scope checks fail on Julia 1.13; with the fix, all five assertions pass on 1.10.12, 1.11.6, 1.12.7, and 1.13.0-rc4. Mode-specific IR checks and a type-stability/zero-allocation dot probe also pass on all four versions.

The PR runs the existing DiffTests group with `--heap-size-hint=32M` on all CI versions. The unchanged before-state job failed on 1.13 and passed on 1.12/LTS: [CI run](https://github.com/chalk-lab/Mooncake.jl/actions/runs/34285907374). The causal experiments and reproduction directions are in [the issue](https://github.com/chalk-lab/Mooncake.jl/issues/1303#issuecomment-5593566237).

### Documentation Preview

Mooncake.jl documentation for PR #1305 is available at:
https://chalk-lab.github.io/Mooncake.jl/previews/PR1305/

### Performance

Performance Ratio:
```
│                   sum_1000 │ 347.0 ns │     1.44 │        1.37 │    1.45 │         5.2 │   7.29 │
│                  _sum_1000 │  1.21 μs │     10.9 │         1.2 │   957.0 │        60.8 │   2.43 │
│               sum_sin_1000 │  5.72 μs │     4.07 │        3.21 │    4.95 │        20.2 │   3.22 │
│              _sum_sin_1000 │  5.93 μs │     5.21 │        1.79 │   197.0 │        20.2 │   4.59 │
│                   kron_sum │ 372.0 μs │     4.71 │        5.98 │    11.2 │       254.0 │   11.1 │
│              kron_view_sum │ 392.0 μs │     3.18 │        4.18 │    12.5 │       277.0 │   11.0 │
│      naive_map_sin_cos_exp │  2.96 μs │     2.55 │        1.42 │ missing │        16.7 │   2.07 │
│            map_sin_cos_exp │  2.71 μs │      5.4 │        2.33 │    2.35 │        15.7 │   3.86 │
│      broadcast_sin_cos_exp │  3.01 μs │     4.12 │        2.91 │    6.32 │        1.95 │   2.53 │
│                 simple_mlp │ 447.0 μs │     4.25 │        2.69 │    1.69 │        8.88 │   2.99 │
│                     gp_lml │ 259.0 μs │     5.75 │        2.49 │    6.22 │     missing │   4.06 │
│ turing_broadcast_benchmark │  2.21 ms │     5.13 │        2.54 │ missing │        29.5 │   1.61 │
│         large_single_block │ 490.0 ns │     7.37 │         2.1 │  6150.0 │        83.0 │   2.37 │
```

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -137,6 +137,8 @@ jobs:
       - run: |
           if [ ${{ matrix.test_group.test_type }} == 'ext' ]; then
             julia --code-coverage=user --eval 'include("test/run_extra.jl")'
+          elif [ "$LABEL" == 'diff_tests' ]; then
+            julia --heap-size-hint=32M --eval 'include("test/run_extra.jl")'
           else
             julia --eval 'include("test/run_extra.jl")'
           fi
--- a/docs/src/developer_documentation/forwards_mode_design.md
+++ b/docs/src/developer_documentation/forwards_mode_design.md
@@ -215,6 +215,10 @@ The purpose of converting `Expr(:foreigncall...)`, `Expr(:new, ...)` and `Expr(:
 
 The purpose of translating `Expr(:call, ::IntrinsicFunction, ...)` is to do with type stability -- see the docstring for the [Mooncake.IntrinsicsWrappers](@ref) module for more info.
 
+Native `gc_preserve_begin` / `gc_preserve_end` scopes are retained in forward mode.
+Their operands are mapped to `Dual` values, keeping both primal and tangent storage
+alive throughout the differentiated pointer-use interval. Preserving raw pointers alone
+does not keep their owning Julia objects alive.
 
 #### Statement Transformation
 
--- a/docs/src/developer_documentation/forwards_mode_design.md
+++ b/docs/src/developer_documentation/forwards_mode_design.md
@@ -215,10 +215,10 @@ The purpose of converting `Expr(:foreigncall...)`, `Expr(:new, ...)` and `Expr(:
 
 The purpose of translating `Expr(:call, ::IntrinsicFunction, ...)` is to do with type stability -- see the docstring for the [Mooncake.IntrinsicsWrappers](@ref) module for more info.
 
-Native `gc_preserve_begin` / `gc_preserve_end` scopes are retained in forward mode.
-Their operands are mapped to `Dual` values, keeping both primal and tangent storage
-alive throughout the differentiated pointer-use interval. Preserving raw pointers alone
-does not keep their owning Julia objects alive.
+Native `gc_preserve_begin` / `gc_preserve_end` scopes are retained in forwards-mode AD.
+The preserved owners are mapped to `Dual` values, keeping both primal and tangent storage alive throughout the scope.
+The end expression still consumes the native begin token, not a `Dual`.
+Preserving raw pointers alone does not keep their owning Julia objects alive.
 
 #### Statement Transformation
 
--- a/src/developer_tools.jl
+++ b/src/developer_tools.jl
@@ -35,7 +35,9 @@ function primal_ir(interp::MooncakeInterpreter, sig::Type{<:Tuple}; normalize=tr
     end
     normalize || return ir
     _, spnames = is_vararg_and_sparam_names(sig)
-    return normalise!(ir, spnames)
+    return normalise!(
+        ir, spnames; preserve_gc=interp isa MooncakeInterpreter{<:Any,ForwardMode}
+    )
 end
 
 """
--- a/src/interpreter/forward_mode.jl
+++ b/src/interpreter/forward_mode.jl
@@ -407,8 +407,11 @@ function modify_fwd_ad_stmts!(
     stmt::Expr, dual_ir::IRCode, ssa::SSAValue, captures::Vector{Any}, info::DualInfo
 )
     if isexpr(stmt, :gc_preserve_begin) || isexpr(stmt, :gc_preserve_end)
-        # The begin operands refer to Duals, keeping primal and tangent owners alive.
-        # The end operand remains the native begin token, not a Dual.
+        # Forward AD inserts a captures argument before the original arguments.
+        # Without shifting its argument references, gc_preserve_begin would preserve
+        # the wrong values. inc_args shifts them by one to refer to the intended Duals.
+        # gc_preserve_end refers to the result of its matching begin, not an argument
+        # position, so inc_args leaves that reference unchanged. See PR #1305.
         replace_call!(dual_ir, ssa, inc_args(stmt))
     elseif isexpr(stmt, :invoke) || isexpr(stmt, :call)
         raw_args = isexpr(stmt, :invoke) ? stmt.args[2:end] : stmt.args
--- a/src/interpreter/forward_mode.jl
+++ b/src/interpreter/forward_mode.jl
@@ -407,7 +407,8 @@ function modify_fwd_ad_stmts!(
     stmt::Expr, dual_ir::IRCode, ssa::SSAValue, captures::Vector{Any}, info::DualInfo
 )
     if isexpr(stmt, :gc_preserve_begin) || isexpr(stmt, :gc_preserve_end)
-        # Preserve the Dual operands, keeping both primal and tangent owners alive.
+        # The begin operands refer to Duals, keeping primal and tangent owners alive.
+        # The end operand remains the native begin token, not a Dual.
         replace_call!(dual_ir, ssa, inc_args(stmt))
     elseif isexpr(stmt, :invoke) || isexpr(stmt, :call)
         raw_args = isexpr(stmt, :invoke) ? stmt.args[2:end] : stmt.args
--- a/src/interpreter/forward_mode.jl
+++ b/src/interpreter/forward_mode.jl
@@ -212,7 +212,7 @@ function generate_dual_ir(
 
     # Normalise the IR.
     isva, spnames = is_vararg_and_sparam_names(sig_or_mi)
-    primal_ir = normalise!(primal_ir, spnames)
+    primal_ir = normalise!(primal_ir, spnames; preserve_gc=true)
 
     # Keep a copy of the primal IR with the insertions
     dual_ir = CC.copy(primal_ir)
@@ -406,7 +406,10 @@ __get_primal(x::Dual) = primal(x)
 function modify_fwd_ad_stmts!(
     stmt::Expr, dual_ir::IRCode, ssa::SSAValue, captures::Vector{Any}, info::DualInfo
 )
-    if isexpr(stmt, :invoke) || isexpr(stmt, :call)
+    if isexpr(stmt, :gc_preserve_begin) || isexpr(stmt, :gc_preserve_end)
+        # Preserve the Dual operands, keeping both primal and tangent owners alive.
+        replace_call!(dual_ir, ssa, inc_args(stmt))
+    elseif isexpr(stmt, :invoke) || isexpr(stmt, :call)
         raw_args = isexpr(stmt, :invoke) ? stmt.args[2:end] : stmt.args
         sig_types = map(raw_args) do x
             t = CC.widenconst(get_forward_primal_type(info.primal_ir, x))
--- a/src/interpreter/ir_normalisation.jl
+++ b/src/interpreter/ir_normalisation.jl
@@ -12,7 +12,7 @@ unchanged, but makes AD more straightforward. In particular, replace
 7. `gc_preserve_begin` / `gc_preserve_end` exprs so that memory release is delayed.
 
 With `preserve_gc=true`, leave native GC preservation scopes intact for forward AD,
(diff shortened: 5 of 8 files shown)


sunxd3:
### Before and after: actual forward-mode IR

These are captured Julia 1.13.0-rc4 IR dumps for:

```julia
f(x) = GC.@preserve x unsafe_load(pointer(x))
```

The input is `Vector{Float64}`. The before comparison used the affected transform methods from [the reviewed PR base](https://github.com/chalk-lab/Mooncake.jl/commit/9caca9f37e56495a1170ac543097434471cac894), restored in a separate Julia process. The after dump uses this PR. The instruction listings below are actual output, not pseudocode.

#### Original primal IR

```julia
1 1 ─ %1 = $(Expr(:gc_preserve_begin, Core.Argument(2)))
  │   %2 =   builtin Base.getfield(_2, :ref)::MemoryRef{Float64}
  │   %3 =   builtin Base.getfield(%2, :ptr_or_offset)::Ptr{Nothing}
  │   %4 = intrinsic Base.bitcast(Ptr{Float64}, %3)::Ptr{Float64}
  │   %5 = intrinsic Base.pointerref(%4, 1, 1)::Float64
  │        $(Expr(:gc_preserve_end, :(%1)))
  └──      return %5
```

`_2` (`Argument(2)`) is the array `x`. The begin expression keeps that owner alive. `%1` names the result of the begin expression. The end refers to `%1` to identify which begin it matches; it does not refer to a function argument.

#### Before: normalization loses the native scope

```julia
27 1 ─        dynamic (Mooncake.gc_preserve)(_2)::Any
   │   %2 =   dynamic (Mooncake.lgetfield)(_2, Val{:ref}())::MemoryRef{Float64}
   │   %3 =   dynamic (Mooncake.lgetfield)(%2, Val{:ptr_or_offset}())::Ptr{Nothing}
   │   %4 =   dynamic (Mooncake.IntrinsicsWrappers.bitcast)(Ptr{Float64}, %3)::Ptr{Float64}
   │   %5 =   dynamic (Mooncake.IntrinsicsWrappers.pointerref)(%4, 1, 1)::Float64
   │        nothing::Core.Const(nothing)
   └──      return %5
```

Normalization replaces the begin with an ordinary helper call and the end with `nothing`. Forward AD differentiates that helper call through its no-op rule:

```julia
frule!!(::Dual{typeof(gc_preserve)}, ::Vararg{Dual,N}) where {N} =
    zero_dual(nothing)
```

Optimization eliminates the unused no-op. The resulting forward IR has both pointer loads but **no preservation scope**:

```julia
27 1 ─ %1  =   builtin Base.getfield(_3, :primal)::Vector{Float64}
   │   %2  =   builtin Mooncake.getfield(%1, :ref, :not_atomic)::MemoryRef{Float64}
   │   %3  =   builtin Base.getfield(_3, :tangent)::Vector{Float64}
   │   %4  =   builtin Base.getfield(%3, :ref)::MemoryRef{Float64}
   │   %5  =   builtin Mooncake.getfield(%2, :ptr_or_offset, :not_atomic)::Ptr{Nothing}
   │   %6  =   builtin Base.getfield(%4, :ptr_or_offset)::Ptr{Nothing}
   │   %7  = intrinsic Mooncake.bitcast(Ptr{Mooncake.NoTangent}, %6)::Ptr{Mooncake.NoTangent}
   │   %8  = intrinsic (Core.Intrinsics.bitcast)(Ptr{Float64}, %5)::Ptr{Float64}
   │   %9  = intrinsic (Core.Intrinsics.bitcast)(Ptr{Float64}, %7)::Ptr{Float64}
   │   %10 = intrinsic (Core.Intrinsics.pointerref)(%8, 1, 1)::Float64
   │   %11 = intrinsic (Core.Intrinsics.pointerref)(%9, 1, 1)::Float64
   │   %12 = %new(Mooncake.Dual{Float64, Float64}, %10, %11)::Mooncake.Dual{Float64, Float64}
   └──       return %12
```

A raw pointer does not keep its owning Julia object alive. Once the pointers have been extracted, the explicit guarantee that the owners survive through the loads is gone. This example demonstrates the lost scope; it need not itself trigger GC.

#### After: both pointer loads remain inside the native scope

```julia
1 1 ─ %1  = $(Expr(:gc_preserve_begin, Core.Argument(3)))
  │   %2  =   builtin Base.getfield(_3, :primal)::Vector{Float64}
  │   %3  =   builtin Mooncake.getfield(%2, :ref, :not_atomic)::MemoryRef{Float64}
  │   %4  =   builtin Base.getfield(_3, :tangent)::Vector{Float64}
  │   %5  =   builtin Base.getfield(%4, :ref)::MemoryRef{Float64}
  │   %6  =   builtin Mooncake.getfield(%3, :ptr_or_offset, :not_atomic)::Ptr{Nothing}
  │   %7  =   builtin Base.getfield(%5, :ptr_or_offset)::Ptr{Nothing}
  │   %8  = intrinsic Mooncake.bitcast(Ptr{Mooncake.NoTangent}, %7)::Ptr{Mooncake.NoTangent}
  │   %9  = intrinsic (Core.Intrinsics.bitcast)(Ptr{Float64}, %6)::Ptr{Float64}
  │   %10 = intrinsic (Core.Intrinsics.bitcast)(Ptr{Float64}, %8)::Ptr{Float64}
  │   %11 = intrinsic (Core.Intrinsics.pointerref)(%9, 1, 1)::Float64
  │   %12 = intrinsic (Core.Intrinsics.pointerref)(%10, 1, 1)::Float64
  │   %13 = %new(Mooncake.Dual{Float64, Float64}, %11, %12)::Mooncake.Dual{Float64, Float64}
  │         $(Expr(:gc_preserve_end, :(%1)))
  └──       return %13
```

#### Why this branch calls `inc_args`

Forward AD inserts a captures argument before the original arguments. It holds values the AD code needs, such as differentiation rules. The original arguments also become `Dual`s. For this example, the argument positions change as follows:

| Position | Original function | Forward AD code |
| --- | --- | --- |
| 1 | `f` | captures |
| 2 | `x` | Dual for `f` |
| 3 | — | Dual for `x` |

`gc_preserve_begin` must follow that shift. Leaving `Argument(2)` unchanged would preserve the Dual for **the function**, rather than the Dual for **the array**. `inc_args` changes the reference to `Argument(3)`. In the after IR, the begin refers to `_3`, and the following `getfield(_3, :primal)` and `getfield(_3, :tangent)` instructions show the two arrays it keeps alive. This branch does not construct a `Dual`; it adjusts the reference to the argument that already contains one.

`gc_preserve_end` needs no such adjustment. It refers to `%1`, the result of its matching begin, not an argument position. `inc_args` only shifts argument references, so it leaves `%1` unchanged. References to other computed values likewise stay unchanged; their defining instructions are transformed by AD. Neither preservation expression is dispatched through `frule!!`.

The lifetime regression independently checks both owners under forced GC, collection after return, and prepared-rule reuse without dereferencing stale pointers. In the stressed DiffTests before/after CI comparison, Julia 1.13 changes from 2,845 passes and one failure to 2,846/2,846; LTS and 1.12 remain at 2,846/2,846. Reverse-mode's existing pullback-based preservation is unchanged.

PR #1331, opened 2026-09-18
Add forward-mode rule for BLAS axpy!

AstitvaAggarwal:
`axpy!` had no `frule!!`, so forward-mode differentiation traced into its `ccall` and hit a missing intrinsic-wrapper error (reverse mode gets away without a dedicated rule via a generic fallback, which is why plain gradients through `axpy!` already worked fine).

This showed up computing native HVP (forward-over-reverse) for `GaussAdjoint` in SciML/SciMLSensitivity.jl#1648: `GaussAdjoint`'s reverse pass calls `axpy!` directly, and an HVP forward-differentiates that whole reverse pass, hitting the missing rule.

Adds `frule!!`/`rrule!!` for `axpy!`, mirroring the existing `scal!` rule right above it (same `@is_primitive` shape, same Wirtinger-conjugate convention for the scalar's cotangent), plus a test case in the existing `blas`/`blas_Float64`/`blas_Float32`/`blas_ComplexF64`/`blas_ComplexF32` groups.

Full `blas_Float64` group passes (190346 tests). Also checked by hand against analytical forward/reverse/HVP values on an independent test function, and against the real `GaussAdjoint` HVP case (matches ForwardDiff to ~5e-6).


src/rules/blas.jl
```diff
@@ -453,6 +453,81 @@ function rrule!!(
...
 
+# Without this, differentiating `axpy!` traces into its ccall, for which there is no
+# `frule!!` -- reverse mode has a generic ccall fallback that gets away without one, but
+# forward mode does not. Surfaced via SciMLSensitivity.jl#1648 (GaussAdjoint's HVP forward-
+# differentiates its own reverse pass, which calls `axpy!` directly).
+@is_primitive(
+    MinimalCtx,
+    Tuple{
+        typeof(BLAS.axpy!),Integer,P,X,Integer,X,Integer
+    } where {P<:BlasFloat,X<:Union{Ptr{P},AbstractArray{P}}}
```
yebai, line 464: The registration couples the concrete types of `X` and `Y`, although Julia's `BLAS.axpy!` method allows them to differ. A valid `Vector` plus `SubArray` call therefore misses this primitive and reaches the unsupported foreign-call rule.

```suggestion
        typeof(BLAS.axpy!),Integer,P,X,Integer,Y,Integer
    } where {
        P<:BlasFloat,
        X<:Union{Ptr{P},AbstractArray{P}},
        Y<:Union{Ptr{P},AbstractArray{P}},
    }
```

src/rules/blas.jl
```diff
@@ -453,6 +453,81 @@ function rrule!!(
...
+    n = primal(_n)
+    incx = primal(_incx)
+    incy = primal(_incy)
+    a, da = extract(a_da)
+    X, dX = arrayify(X_dX)
+    Y, dY = arrayify(Y_dY)
+
+    # Fréchet derivative of Y_new = a*X + Y: dY_new = da*X + a*dX + dY.
+    BLAS.axpy!(n, da, X, incx, dY, incy)
+    BLAS.axpy!(n, a, dX, incx, dY, incy)
```
yebai, line 486: When `X === Y`, `dX` and `dY` share storage. The first `axpy!` changes the value read by the second, so the derivative is wrong. Copy the logical source tangent only for this aliasing case.

```suggestion
    if X === Y
        _, dX_source = viewify(n, X_dX, incx)
        dX_source = copy(dX_source)
        BLAS.axpy!(n, da, X, incx, dY, incy)
        BLAS.axpy!(n, a, dX_source, 1, dY, incy)
    else
        BLAS.axpy!(n, da, X, incx, dY, incy)
        BLAS.axpy!(n, a, dX, incx, dY, incy)
    end
```

src/rules/blas.jl
```diff
@@ -453,6 +453,81 @@ function rrule!!(
     return X_dX, scal_adjoint
 end
 
+# Without this, differentiating `axpy!` traces into its ccall, for which there is no
+# `frule!!` -- reverse mode has a generic ccall fallback that gets away without one, but
+# forward mode does not. Surfaced via SciMLSensitivity.jl#1648 (GaussAdjoint's HVP forward-
+# differentiates its own reverse pass, which calls `axpy!` directly).
```
yebai, line 459: The current comment incorrectly suggests that reverse mode has a usable foreign-call fallback. The fallback throws when no rule exists, so explain why both rules are defined here.

```suggestion
# `BLAS.axpy!` ends at a BLAS foreign call with no Mooncake derivative rule. This wrapper
# supplies both modes; forward-over-reverse paths invoke `axpy!` directly.
```

yebai (approved):

